PR #1837, opened 2014-02-16
[mustache.php] Fixes is_dir not correctly handling 'file://' in filename

### LINKED ISSUE #1835: [mustache.php] realpath does not resolve to dir
Made a test case for an error in mustache.php tests.

``` php
<?php
class Dir_Loader{
  private $baseDir;

  public function __construct($baseDir){
    $this->baseDir = $baseDir;

    if (strpos($this->baseDir, '://') === -1) {
      $this->baseDir = realpath($this->baseDir);
    }

    if (!is_dir($this->baseDir)) {
      echo 'FilesystemLoader baseDir must be a directory: ' . $baseDir;
    }
  }
}

function test(){
  $baseDir = realpath(dirname(__FILE__).'/../../somedir/someinnerdir');

  $loader = new Dir_Loader('file://' . $baseDir);
}
test();
```

Hard to upload as a test because of dir locations, but running test in folders will show broken behaviour 

```
../somedir/someinnerdir$ 
```

k-fish:
Should probably be merged with  #1732 , but they deal with different fixes, so I decided to stick with the one feature per pull paradigm.

The 'file://' stripping I found in https://github.com/facebook/hhvm/blob/master/hphp/runtime/base/file-stream-wrapper.cpp#L47 , not sure if file-stream-wrapper usage would be more appropriate in this context.

Fixes #1835

### DIFF
--- a/hphp/test/slow/ext_file/is_dir.php
+++ b/hphp/test/slow/ext_file/is_dir.php
@@ -1,0 +1,6 @@
+<?php
+function test(){
+  $baseDir = 'file://' . realpath(dirname(__FILE__));
+  var_dump(is_dir($baseDir));
+}
+test();
--- a/hphp/test/slow/ext_file/is_dir.php.expect
+++ b/hphp/test/slow/ext_file/is_dir.php.expect
@@ -1,0 +1,1 @@
+bool(true)


iamhabbeboy:
check your file path, that solve my problem

PR #3577, opened 2014-08-24
Make access log %h behave more like Apache

joelastpass:
The access log format appears to be Apache like -- but doesn't behave like Apache when hostname lookups are off, it logs a blank.   It should log IP address in this case just like Apache, this accomplishes that.


joelastpass:
This is still a useful contribution -- can you reopen this?

jwatzman:
It's probably easiest at this point to open a new pull request, if you don't mind.

jwatzman:
(But more than happy to take a look at the new PR!)

PR #3930, opened 2014-10-08
Implemented variables_order&request_order directive

reeze:
This also makes cli mode consist with PHP if S&E is not
in variables_order $_SERVER/$_ENV will not available in cli.

before the diff HHVM's variables_order=SEGPC and request_order=GPC
after the diff thoes directive are default one:

 - variables_order=EGPCS
 - request_order=  and it will be the same as variables_order

@see <http://php.net/manual/en/ini.core.php#ini.variables-order>


reeze:
Start a new PR, the previous could not been updated to phabricator

JoelMarcey:
Trying to resync something. Open and close.

PR #4443, opened 2014-12-13
Update exception classes, fixes #4442

### LINKED ISSUE #4442: Minor interop issue with ReflectionParameter
Demo: http://3v4l.org/5ZCMb

ReflectionParameter's constructor throws an `Exception` instead of `ReflectionException`, which actually broke our test suite as we were catching the latter specifically :)

Seldaek:
Fix for #4442

### DIFF
--- a/hphp/test/slow/reflection_classes/reflection_param_exception.php
+++ b/hphp/test/slow/reflection_classes/reflection_param_exception.php
@@ -1,0 +1,39 @@
+<?php
+
+class A {
+  public static function bar($baz) {}
+}
+
+// Bad function.
+try {
+  new ReflectionParameter(1, 'baz');
+} catch (ReflectionException $e) {
+  echo 'caught1';
+}
+
+// Bad param name.
+try {
+  new ReflectionParameter(['A', 'bar'], 'not-baz');
+} catch (ReflectionException $e) {
+  echo 'caught2';
+}
+
+// Out of bound param.
+new ReflectionParameter(['A', 'bar'], 0);
+try {
+  new ReflectionParameter(['A', 'bar'], 1);
+} catch (ReflectionException $e) {
+  echo 'caught3';
+}
+try {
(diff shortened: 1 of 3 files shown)


JoelMarcey:
@Seldaek Hi. Are you planning on addressing the test failures in your pull request? Thanks.

Seldaek:
I'm not against it but TBH I have zero experience with HHVM and I can't even figure out why one of the build failed.

hphp/test/slow/reflection_classes/reflection_param_exception.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+class A {
+  public static function bar($baz) {}
+}
+
+// Bad function.
+try {
+  new ReflectionParameter('A::foo', 'baz');
```
staabm: `A::foo` needs to be e.g. a number to trigger the 1st newly added exception

hphp/test/slow/reflection_classes/reflection_param_exception.php
```diff
@@ -0,0 +1,19 @@
...
+} catch (ReflectionException $e) {
+  echo 'caught';
+}
+
+// Bad param name.
+try {
+  new ReflectionParameter('A::bar', 'not-baz');
```
staabm: `not-baz` needs to be a non-string & non-int to trigger the 2nd newly added exception

staabm:
@Seldaek added some feedback what you need to adjust to make your changes pass the provided impl.

JoelMarcey:
@Seldaek It looks like @staabm might have helped you fix your issues here. Can you send an update to your pull request so we can re-run the tests and see what happens? Thanks.

Seldaek:
So after looking at it I found other mistakes and I now updated the PR with more tests and a few more fixes to the code to make sure everything passes. Travis failed and I have no idea why, but maybe you can check it out again.

PR #4459, opened 2014-12-15
Change the iso format in ext_datetime.cpp

chizhengwang:
<?php
$date = new DateTime('@0');
var_dump($date);

The this output of this code should be

object(DateTime)#1 (3) {
  ["date"]=>
  string(19) "1970-01-01 00:00:00"
  ["timezone_type"]=>
  int(1)
  ["timezone"]=>
  string(6) "+00:00"
}

instead of
 
object(DateTime)#1 (3) {
  ["date"]=>
  string(19) "1970-01-01 00:00:00.000000"
  ["timezone_type"]=>
  int(1)
  ["timezone"]=>
  string(6) "+00:00"
}.


sgolemon:
Simon makes a good case for why this PR should not be accepted.

https://reviews.facebook.net/D30303#25

PR #4545, opened 2014-12-30
Add a system profiler and hook some actions of interest to New Relic

rrh:
This PR re-does an earlier PR which was abandoned in the summer 2014:
  https://github.com/facebook/hhvm/pull/3082
At the time, PR 3082 was abandoned in favor of local development using
a subclass of the recently developed DebugHookHandler. However, using
DebugHookHandler proved to be too intrusive, heavy weight and expensive,
and I find that this mechanism is sufficient, albeit it is yet another
mechanism.

Don't have ProfilerFactory::setExternalProfiler delete a pre-existing
external profiler, as that is not the sole ownership point.

Don't assume that the External profiler wants to ignore builtin functions.

### DIFF
--- a/hphp/runtime/ext/ext_system_profiler.cpp
+++ b/hphp/runtime/ext/ext_system_profiler.cpp
@@ -19,6 +19,6 @@
 
 namespace HPHP {
 
-SystemProfiler *g_system_profiler = 0;
+SystemProfiler *g_system_profiler = nullptr;
 
 }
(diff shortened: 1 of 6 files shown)


hphp/runtime/ext/ext_system_profiler.cpp
```diff
@@ -0,0 +1,24 @@
...
+   | obtain it through the world-wide-web, please send a note to          |
+   | license@php.net so we can mail you a copy immediately.               |
+   +----------------------------------------------------------------------+
+*/
+
+#include "hphp/runtime/ext/ext_system_profiler.h"
+
+namespace HPHP {
+
+SystemProfiler *g_system_profiler = 0;
```
sgolemon: nit; use nullptr, not 0

rrh: fixed

PR #4551, opened 2014-12-31
refactor FastLZ cmake finder

jrobeson:
Changes of note:

 * actually find the library in cmake (instead of specifying manually)
 * rename from libfastlz to fastlz to match the library name
 * don't add the third-party includes if we already specified the includes dir


glensc:
please consider adding to 3.3-LTS branch as well. thanks!

paulbiss:
cc @jwatzman

paulbiss:
Also, in general if you want something backported you should open a PR with the cherry-pick against that branch so we can discuss it there.

glensc:
I will make PR. thanks

PR #4559, opened 2015-01-02
Fix typo in hphp/test/run regarding UTF8/BMP.

rrh:
This is a pure php5 script, and as such presumably isn't subject
to standard static analysis like hack scripts are.


sgolemon:
Oopsie... My bad. :)

Issue #4136, opened 2014-11-04
11 tests failed using HHVM 3.3.1 on CentOS 6.5


wjywbs:
I compiled HHVM 3.3.1 on CentOS 6.5 and ran test/all. 11 tests out of 11825 tests failed.
```
test/slow/ext_icu/uspoof.php
test/slow/ext_zlib/qlzfail.php
test/zend/good/ext/intl/tests/calendar_getDayOfWeekType_basic.php
test/zend/good/ext/mbstring/tests/bug54494.php
test/zend/good/ext/pcre/tests/bug47662.php
test/zend/good/ext/standard/tests/strings/bug51059.php
test/zend/good/ext/zlib/tests/gzreadgzwriteplain.php
test/zend/good/ext/zlib/tests/gzwrite_variation1.php
```
I think these are false positives:
```
test/slow/ext_gd/imagewebp.php
test/slow/ext_xdebug/remote/feature.php
test/zend/good/ext/date/tests/bug67253.php
```
```
cat test/slow/ext_icu/uspoof.php.diff
030+ bool(false)
031+ Failed: 1
032+ Got: 
033+ array(2) {
034+   [0]=>
035+   array(4) {
036+     ["file"]=>
037+     string(61) "/hhvm/hphp/test/slow/ext_icu/uspoof.php"
038+     ["line"]=>
039+     int(120)
040+     ["function"]=>
041+     string(2) "VS"
042+     ["args"]=>
043+     array(2) {
044+       [0]=>
045+       bool(false)
046+       [1]=>
047+       bool(true)
048+     }
049+   }
050+   [1]=>
051+   array(4) {
052+     ["file"]=>
053+     string(61) "/hhvm/hphp/test/slow/ext_icu/uspoof.php"
054+     ["line"]=>
055+     int(186)
056+     ["function"]=>
057+     string(27) "test_SpoofChecker_setchecks"
058+     ["args"]=>
059+     array(0) {
060+     }
061+   }
062+ }
030- bool(true)
......
092+ bool(true)
```
```
test/slow/ext_zlib/qlzfail.php: HHVM 3.3.0 didn't fail this one.
Fatal error: Call to undefined function qlzcompress() in /hhvm/hphp/test/slow/ext_zlib/qlzfail.php on line 6
```
```
test/zend/good/ext/intl/tests/calendar_getDayOfWeekType_basic.php
001+ int(1)
001- int(3)
```
```
[...]

PR #4573, opened 2015-01-05
Call gzclearerr() in gzseek().

wjywbs:
This clears gzeof so data can be read. This only occurs when the user read
a gz file to the end, seek and read again with buffer size >= 8192.

Part of #4136.
Fixed test/zend/good/ext/zlib/tests/gzreadgzwriteplain.php.


JoelMarcey:
@wjywbs Can you sign the CLA so we can import this and review internally? Thanks.

wjywbs:
The corporate CLA was signed yesterday, but I'm not sure how long it will take effect.

wjywbs:
@JoelMarcey Can you please check the status of the CLA? Thanks.

JoelMarcey:
@wjywbs Thanks. Checking internally now.

Issue #4577, opened 2015-01-06
Crash when throwing an exception from a destructor with track_errors on


jmarrama:
```php
<?php
ini_set('track_errors', 1);
class Foo {
        public function __destruct() {
                throw new Exception("FOO");
        }
}

$f = new Foo();
```

Results in a segfault. Fix coming shortly

jmarrama:
Issue is fixed, closing

PR #4578, opened 2015-01-06
Fix segfault with exceptions in object dtors and track_errors

jmarrama:
This fixes the issue tracked in #4577 where throwing an exception
from an object destructor when the 'track_errors' ini setting is
on causes HHVM to segfault, due to accessing the frame pointer after
it has been destroyed.


jmarrama:
This also fixes the same problem in HHVM 3.3.1

PR #4585, opened 2015-01-06
Add compile time option to disable thrift extension

adsr:
The original aim was to make this a runtime option, but there doesn't seem to be a clean way to do it. Each extension constructor adds the instance to s_registered_extensions, which is later iterated and queried in-the-raw in several places.


JoelMarcey:
@adsr Not sure if saw these comments https://github.com/facebook/hhvm/commit/128614b888f323f0ffd59f20c505db0fc01191ba but this had to be reverted.

Let me know if you want to submit another pull request with the same change or if you want me to do it manually internally, giving you credit in the diff of course. 

However, I would be curious if your next pull request has the same path problem as before. Normally all files are prefixed with `hphp` when changes are made to the `runtime` directory. So that was weird.

Sorry about the confusion.

adsr:
Strange. Not sure what happened there but I opened another pull request and it appears to have the correct path now.

PR #4593, opened 2015-01-08
Implemented directives: variables_order&request_order

### LINKED ISSUE #1806: HHVM doesn't use request_order from php.ini
in my php.ini I have following:
request_order = "GP" which mean I do not want have Cookies in my $\_REQUEST. But with hhvm I still have Cookies data in $\_REQUEST variable.

reeze:
This also makes cli mode consist with PHP if S/E is not
in variables_order $_SERVER/$_ENV will not available in cli.

before the diff HHVM's default is variables_order=SEGPC and
request_order=GPC after the diff thoes directives becames:

  - variables_order=EGPCS
  - request_order=  and it will be the same as variables_order

@see <http://php.net/manual/en/ini.core.php#ini.variables-order>

closes #1806


reeze:
Finally :)
Thank you @JoelMarcey for you help!

PR #4601, opened 2015-01-08
Add compile time option to disable thrift extension

adsr:
@JoelMarcey This one appears to have the correct path. Not sure what happened in the previous pull request.


JoelMarcey:
@adsr Hi. Have you gotten this to build correctly? See my comment at 

https://reviews.facebook.net/D31179

jrobeson:
can you possibly add the ability to disable thrift to hhvm-third-party also? that way thrift won't be compiled if not needed.

fredemmott:
Thrift is needed, even if the extension is not available - we use its' concurrency classes in core parts of the runtime.

JoelMarcey:
Closing -- see comments in pull request.

Issue #4250, opened 2014-11-14
Investigate create_function() memory leaks


fredemmott:
cc @huzhiguang, @reeze, @lbadger, @ibrezac
mentioned in #2851
Memory leak #4 on http://lamp.baidu.com/2014/11/03/hhvm-in-baidu/

wjywbs:
Another leak is to make a static string of "aaa..." as the argument of var_dump.
https://github.com/facebook/hhvm/blob/c5a485f66c07cad8945b4bd6ca3186e1b8c89e40/hphp/compiler/analysis/emitter.cpp#L4508

corientdev:
Guys any solution on this i am still having the issue hhvm stops every 4-5 days.

Wondering if i can edit something in wordpress to stop this..? Because i don't have any plans to run my wordpress set up on php-fpm so i can modify the wordpress script to make it fully compatible with HHV. Please someone help its frustrating..

gferon:
It's actually quite easy, as stated in another ticket, you need to replace `create_function` calls by anonymous functions.

Example: in `wp-includes/pomo/po.php`, you need to replace
```php
$is_final = create_function('$context', 'return $context == "msgstr" || $context == "msgstr_plural";');
```

by
```php
$is_final = function($context) {
	return $context == "msgstr" || $context == "msgstr_plural";
};
```

There is another call to `create_function` in this file, one in `wp-includes/pomo/translations.php`. There might also be more calls in whatever Wordpress plugin you use.

corientdev:
Any clue why hhvm does not support this `create_function`? i mean it seems pretty do able and will enhance the compatibility of hhvm with php.

Curious to know..

huzhiguang:
@corientdev ,hhvm support create_function,but this feature will affect hhvm performance.

corientdev:
let it affect thats not the concern for me but why it causes memory leaks and ultimately stops the daomon?

paulbiss:
8d0bac44832df988cba9f247beaec5e7041ef6ed

wjywbs:
@paulbiss I think the static strings created during `compile_string()` are still not freed, right? In Fred's https://github.com/facebook/hhvm/issues/4250#issuecomment-63901992, a static string of "aaa..." will be   created as the argument of var_dump. This leaks much faster than the bytecode.

(Note that my PR #4607 wasn't accepted. https://reviews.facebook.net/D31263)

paulbiss:
[...]

PR #4607, opened 2015-01-09
Store static strings created by create_function() on a thread local map.

wjywbs:
The memory of these static strings and compiled units are freed when the
script finishes. Part of #4250.


jwatzman:
Closing due to inactivity. https://github.com/facebook/hhvm/wiki/Human-Timeouts

PR #4641, opened 2015-01-13
Set type call with double as argument

### LINKED ISSUE #4624: settype($val, 'double') not works (PHP5 incompatiblity)
`settype($val, 'double')` leaves `$val` unconverted, see http://3v4l.org/9bfAV.

`double` is in fact deprecated variant, but `gettype(1.0)` still returns `'double'`.

Stelian:
closes #4624


Majkl578:
Can you add `closes #4624` to the summary? It will auto-close the issue once this PR lands.

PR #4643, opened 2015-01-14
use strict should fatal

### LINKED ISSUE #4556: No use strict easter egg
<?php
    use strict;

From PHP 5.4 onwards, this produces the following error:

    Fatal error: You seem to be trying to use a different language... in /in/WY1Vs on line 3

Yet HHVM produces no error. Perhaps it should have something like "Fatal error: Why not try Hack instead?" ;)

Also, shhh. This is supposed to be seeekrit.

Majkl578:
closes #4556

### DIFF
--- a/hphp/compiler/parser/parser.cpp
+++ b/hphp/compiler/parser/parser.cpp
@@ -2423,6 +2423,11 @@ void Parser::onNamespaceEnd() {
 }
 
 void Parser::onUse(const std::string &ns, const std::string &as) {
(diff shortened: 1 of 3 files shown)


jwatzman:
Closing due to inactivity. https://github.com/facebook/hhvm/wiki/Human-Timeouts

dg:
@Majkl578 what about 

```php
<?php
namespace x;
use strict;
```

PR #4652, opened 2015-01-15
Experimental - No collection

psmarshall:
Don't actually merge this in, this is just for comments.

The aim of this diff is to disable collection of reference counted objects so that a tracing garbage collector can be built and tested against the reference counting. In many places if an object were to be released, the reference count would not actually be decremented to 0, so this has been changed everywhere I could find it.


fredemmott:
Closing this as it's not meant for merge - it'll stay around on phabricator for discussion.

PR #4666, opened 2015-01-17
Fix Memcache::connect with unix socket

yaslama:
Without this fix, the connection fails with the following errors
Warning: Name or service not known in hphp/test/slow/ext_memcache/unix-socket.php
  on line 4
Warning: Memcache: Can't connect to unix:///run/memcached/memcached.sock:0 in hphp/test/slow/ext_memcache/unix-socket.php on line 4


clementhk:
LGTM, I might fix it in `isServerReachable()` to check if the socket file exists so it doesn't pass into memcache if it is not valid, but this works.

DmitrySoshnikov:
Yeah, as a quick fix it's OK, though ideally we should be checking for the valid socket file as well.

PR #4680, opened 2015-01-21
fix year in NEWS

sobit:


jwatzman:
Thanks for the PR! This was fixed by https://github.com/facebook/hhvm/pull/4686 which has been merged (though we're having a bit of trouble with our sync script so it could be a few hours before that makes it back out to GitHub).

PR #4707, opened 2015-01-26
made the url parsing faster and updated license headers

MattEagle:
Line 38 (return "") was never getting executed because int strip would always be either 0, 7, or 8. Therefore the last return statement could only be url. So I fixed that as well as eliminating the need for int strip by inlining the strchr calls.

It's a small change, but it has a minor (very minor) performance improvement.

### DIFF
--- a/hphp/runtime/base/url.cpp
+++ b/hphp/runtime/base/url.cpp
@@ -2,7 +2,7 @@
    +----------------------------------------------------------------------+
    | HipHop for PHP                                                       |
    +----------------------------------------------------------------------+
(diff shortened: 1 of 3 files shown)


jwatzman:
Did you intend to close this PR? Looks like you deleted the source repo for it. All of our tooling requires the original github PR to remain open until the change is merged, even though the review goes on inside Phabricator.

MattEagle:
Yeah. I still have the source on my machine, but I figured I'm better off fixing documented issues than trying to make micro-optimizations to things that work fine the way they are.

jwatzman:
Cool, we can always revive this if you want to later.

Issue #4533, opened 2014-12-26
Deadlock in jemalloc when a signal occurs in malloc and hhvm bt_handler() tries to malloc again


wjywbs:
I added a fopen() in Logger::log() if the output file is closed. (I'm not sure if this is related to this issue.) This doesn't happen very often, but all threads deadlock in jemalloc.

```
#0  0x00007fcc372ea264 in __lll_lock_wait () from /lib64/libpthread.so.0
#1  0x00007fcc372e5508 in _L_lock_854 () from /lib64/libpthread.so.0
#2  0x00007fcc372e53d7 in pthread_mutex_lock () from /lib64/libpthread.so.0
#3  0x00007fcc3a8488dd in je_malloc_mutex_lock (mutex=0x7fcc32800bf8)
    at include/jemalloc/internal/mutex.h:77
#4  je_arena_tcache_fill_small (arena=0x7fcc328000c0, 
    tbin=tbin@entry=0x7fcc10228248, binind=binind@entry=17, 
    prof_accumbytes=prof_accumbytes@entry=0) at src/arena.c:1475
#5  0x00007fcc3a86f8cf in je_tcache_alloc_small_hard (
    tcache=tcache@entry=0x7fcc10228000, tbin=tbin@entry=0x7fcc10228248, 
    binind=binind@entry=17) at src/tcache.c:72
#6  0x00007fcc3a83e5c5 in je_tcache_alloc_small (zero=false, size=640, 
    tcache=0x7fcc10228000) at include/jemalloc/internal/tcache.h:303
#7  je_arena_malloc (try_tcache=true, zero=false, size=<optimized out>, 
    arena=0x0) at include/jemalloc/internal/arena.h:957
#8  je_imalloct (arena=0x0, try_tcache=true, size=<optimized out>)
    at include/jemalloc/internal/jemalloc_internal.h:771
#9  je_imalloc (size=<optimized out>)
    at include/jemalloc/internal/jemalloc_internal.h:780
#10 malloc (size=<optimized out>) at src/jemalloc.c:929
#11 0x00007fcc365cbecb in __fopen_internal () from /lib64/libc.so.6
#12 0x0000000001a1130b in real_fopen (
    filename=filename@entry=0x7fff4455f773 "hhvm", modes=<optimized out>)
    at bfdio.c:120
#13 0x000000000199a708 in bfd_fopen (filename=0x7fff4455f773 "hhvm", 
[...]

Issue #2408, opened 2014-04-10
c_DOMDocument::sweep causes segfault in Joomla! JForm Test


chalet16:
Code (extracted from joomlacms repository)

```
<?php
function load($data)
{
    $element     = new SimpleXMLElement($data);
    $fields = $element->xpath('descendant-or-self::field');
    foreach ($fields as $field) {
        $loadeddom = dom_import_simplexml($field);
        $current   = simplexml_load_string('<books></books>');
        $olddom    = dom_import_simplexml($current);
        $addeddom  = $olddom->ownerDocument->importNode($loadeddom);
        $olddom->parentNode->replaceChild($addeddom, $olddom);
   }
}

$loadMergeDocument = '
   <fields>
      <field name="published" type="list">
         <option value="1">JYES</option>
         <option value="0">JNO</option>
      </field>
      <field name="abstract" label="Abstract" />
      <fields label="A general group">
         <field name="access" />
         <field name="ordering" />
      </fields>
      <fields name="params">
         <field name="show_abstract" type="radio">
            <option value="1">JYes</option>
            <option value="0">JNo</option>
         </field>
      </fields>
      <fieldset>
         <field name="language" type="text" />
      </fieldset>
   </fields>
';

load($loadMergeDocument);

```

PR #4712, opened 2015-01-26
Output simple stacktrace if jemalloc is found in the stacktrace to avoid deadlock.

wjywbs:
jemalloc is most likely crashed and acquired its lock if it's found in the
stacktrace. Be conservative and avoid malloc again and deadlock.

Part of #4533 and #2408.


reeze:
Why this was closed?
We encountered the deadlock problem before, do you have another solution or ? :-)

fredemmott:
Based on the conversation on phabricator, it looks like an alternative approach is going to be taken.

reeze:
:+1:

PR #4736, opened 2015-01-29
Add support for thread-local setlocale() and localeconv().

keyurdg:
This also fixes an icompatibility (with Zend) where printf doesn't respect the
locale.


JoelMarcey:
Hi @keyurdg, we are going to have to revert this pull request because of some segfaults happening on Ubuntu 14.04. Not sure what in this diff is causing it, but it is the commit that started the problem.

See #4954

So I am re-opening this.

Out of curiosity, what system did you build and test this PR on?

cc: @jasone, @sgolemon, @rrh

keyurdg:
Hi @JoelMarcey I tested it on Cent 6 and 7 and we are running this in Prod on Cent 7.

Do you have a core dump for the backtrace? I'll also try this on an Ubuntu 14.04 VM tomorrow.

JoelMarcey:
All I have is what is in https://github.com/facebook/hhvm/issues/4954

@rrh -- have any more debug info that you can provide?

hphp/test/slow/ext_string/setlocale-multiple-calls.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+test(LC_TIME, 'fr_FR');
```
glensc: maybe use utf-8 locales. looks better in github and in terminal :)

`fr_FR.UTF-8`

sgolemon:
FYI Looks like all the side-issues with this diff have been worked out.  Should be landing it today after all the test runs have completed.  Thanks for all your patience!

sgolemon:
3df8f4c239984df4d9b53c2b451f34218025f2f0 should take care of it.

keyurdg:
Thanks much for shepherding this through @sgolemon!

PR #4759, opened 2015-02-02
Prefer webscalesqlclient if available

fredemmott:
This lets us do neat stuff once https://reviews.facebook.net/D17031 lands


fredemmott:
going to do this differently.

PR #4783, opened 2015-02-05
Fix segfault in mysql_escape_string without connection

### LINKED ISSUE #4777: HHVM 3.5 segfaults if calling mysql_real_escape_string() without a connection open
```PHP
<?php
var_dump(mysql_real_escape_string('1234'));
```

yay segfault.

SiebelsTim:
Fixes #4777
We might have hit some more segfaults with this

Test Plan:
I don't have a php installation with mysql_* still available, so I think the error message doesn't match php5. But do we worry enough for a deprecated function?


DmitrySoshnikov:
> We might have hit some more segfaults with this

If the segfault was caused by [accessing](https://github.com/facebook/hhvm/blob/master/hphp/runtime/ext/mysql/mysql_common.cpp#L188) `mysql` property, then yes, it fixes segfaults in other similar cases where the default connection is called.

> so I think the error message doesn't match 

I don't see a test with the warning message (when you don't have `@`), but FWIW, PHP shows:
```
PHP Warning:  mysql_real_escape_string(): Access denied for user ''@'localhost' (using password: NO) in /home/dmitrys/mysql.php on line 4
PHP Warning:  mysql_real_escape_string(): A link to the server could not be established in /home/dmitrys/mysql.php on line 4
```
I think it's fine not to exactly match this, as long as it's deprecated.

Issue #4818, opened 2015-02-10
Empty response body on fatal errors


Yogu:
For reasonable error pages, we register a shutdown callback that will output a nice page when it detects that the script has not terminated gracefully. This works fine with the normal php engine, but in hhvm, it only works for non-fatal errors. For fatal errors, like calling an undefined function, the response body is not sent.

The following script:

```
<?php
echo "before\n";
register_shutdown_function(function() {
        header('Test: Header');
        echo "shutdown\n";
});
call_undefined();
```

sends the header, but suppresses the whole response body:

```
< HTTP/1.1 500 hphp_invoke
< Server: nginx/1.6.2
< Date: Tue, 10 Feb 2015 23:42:16 GMT
< Content-Type: text/html; charset=utf-8
< Transfer-Encoding: chunked
< Connection: keep-alive
< X-Powered-By: HHVM/3.5.0
< Test: Header
< 
```

I'm using the hhvm package on ubuntu 14.04 via fastcgi with nginx and the defaults for `server.ini` and `php.ini`.

See also [this stackoverflow question](https://stackoverflow.com/questions/23086903/hhvm-fastcgi-nginx-how-to-make-it-display-fatal-errors-in-the-browser) which has 12 upvotes but no satisfying answer.

iansltx:
There are a couple config changes that you need to make, and then you can use set_error_handler() to catch things and print out a nice error page. If your code fatals before your handler is set then you've got issues there, but it works a lot of the time...

hhvm.error_handling.call_user_handler_on_fatals	= 1

The above should be in your php.ini.

There are a few other useful flags that you can set this way, see here:

https://github.com/facebook/hhvm/blob/4d9b5b19afc0581856c79298f0672d7b55df9339/hphp/runtime/base/runtime-option.cpp#L840

As a heads-up, whereas the Zend runtime won't call the error handler on fatals, HHVM will, with an error code of 16777217. So you'll handle full-on fatals a bit different than you would under Zend, but that way is actually more consistent.

Thing is, the config options have defaults that don't patch Zend, so that needs fixing. I'll create an issue for that now...like I should've a few weeks ago when I figured this stuff out.

If you've got any more questions about setting this up, I'm on #hhvm on Freenode, and I've run into (and largely solved) this exact issue in my code...would love to help.

paulbiss:
@iansltx has the right idea. Not leaking errors to the browser is something we do intentionally, though you're more than welcome to setup an error handler that does this for you.

[...]

PR #4820, opened 2015-02-11
Match PHP5 error handling defaults unless configured otherwise

### LINKED ISSUE #4819: Error handling config defaults don't match Zend
...resulting in issues like #4818, which I've run into as well.

Probably just need to change here:

https://github.com/facebook/hhvm/blob/4d9b5b19afc0581856c79298f0672d7b55df9339/hphp/runtime/base/runtime-option.cpp#L841

for that default, and here:

https://github.com/facebook/hhvm/blob/4d9b5b19afc0581856c79298f0672d7b55df9339/hphp/runtime/base/runtime-option.cpp#L843

for throwing errors rather than exceptions on undef functions.

The latter of the two is convenient as-is, but unexpected when you're coming from PHP 5.6.

iansltx:
Resolves #4819 and probably #4818. Split into two commits in case we want to leave bad method call config default the way it is now.

### DIFF
--- a/hphp/runtime/base/runtime-option.cpp
+++ b/hphp/runtime/base/runtime-option.cpp
@@ -840,7 +840,7 @@ void RuntimeOption::Load(IniSetting::Map& ini, Hdf& config,
     Config::Bind(CallUserHandlerOnFatals, ini,
                  error["CallUserHandlerOnFatals"], true);
     Config::Bind(ThrowExceptionOnBadMethodCall, ini,
-                 error["ThrowExceptionOnBadMethodCall"], true);
+                 error["ThrowExceptionOnBadMethodCall"], false);
     Config::Bind(LogNativeStackOnOOM, ini,
                  error["LogNativeStackOnOOM"], false);
     Config::Bind(MaxLoopCount, ini, error["MaxLoopCount"], 0);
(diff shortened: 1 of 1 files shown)


iansltx:
May need to tweak error handling further so that E_ALL being thrown to the error handler won't result in a blank 200 rather than a blank 500. Maybe excluding error code 16777217 (HHVM-specific fatal delivered to logger method) from E_ALL will fix behavior.

Majkl578:
@iansltx: That error-specific code for fatals concerns me too as it breaks one of our tests in nette/tracy, but it should be done in a separate PR. Also see my comment in Phabricator.

iansltx:
Thanks for linking to that rev. I think that one issue here is that, when the error handler flag is turned off, the normal way of grabbing fatals, via register_shutdown_function(), doesn't worth either. So displaying a pretty error page on-fatal, which is doable in PHP5, is impossible under HHVM.

May be best to drop this PR entirely, as well as #4819, and instead address register_shutdown_function() handling so that fatals hit that callable. Which would be a more compatible, albeit more difficult, fix.

Majkl578:
I think catching fatals in shutdown functions works now: http://3v4l.org/NPI6v

iansltx:
I'll check into this further. I wasn't able to catch the fatal in my testing prior; I'd instead get a hphp_invoke 500. Similar to what's happening in #4818.

iansltx:
Closing this as switching configs has unintended consequences on the fatal handler side, and the bad method call exception should probably have a test rather than just swapping a config value.

PR #4840, opened 2015-02-14
Memset with AVX2 support

sathvikl:
When configured with -DENABLE_AVX2=ON, ENABLE_AVX2 symbol will be set. 
This version of memset has support for AVX2 instructions and even for cases less than 64 bytes has a smaller ratio of branches/instruction sequence. 
I will submit another version which will speed up for ivy bridge, where AVX2 instruction set is absent.


denji:
@sathvikl Bug when enabling `-DENABLE_SSP=ON` ref: facebook/hhvm#5015

Issue #4863, opened 2015-02-19
Closures in construct = memory leak


terah:
```php

<?php


class CallbackBucket
{
    public $callbacks = [];

    public function __construct()
    {
        $this->callbacks[] = function() {  return "Hello"; };
        $this->callbacks[] = function() {  return "Hello"; };
        $this->callbacks[] = function() {  return "Hello"; };
    }
}

function echoHumanFileSize()
{
    $bytes  = memory_get_usage();
    $sz     = 'BKMGTP';
    $factor = floor(( strlen($bytes) - 1 ) / 3);
    $mem    = sprintf("%.2f", $bytes / pow(1024, $factor)) . $sz[(int)$factor];
    echo("Memory Usage: {$mem}" . PHP_EOL);
}

echo "MEMORY IS FINE HERE\n";
for ( $i = 0; $i < 10000000 ; $i++ )
{
    $object = new CallbackBucket();
    // Have to unset the callback to recover memory
    unset($object->callbacks);
    if ( $i % 1000000 === 0 )
    {
        echoHumanFileSize();
    }
}

echo "MEMORY LEAKS HERE\n";
for ( $i = 0; $i < 10000000 ; $i++ )
{
    $object = new CallbackBucket();
    if ( $i % 1000000 === 0 )
    {
        echoHumanFileSize();
    }
}

//bash-4.2# hhvm --version
//HipHop VM 3.6.0-dev (rel)
//Compiler: heads/master-0-ga9e2a34d62dd95d72eb31b2e73ce475d73d87e2f
//Repo schema: e05b63d7e028ccfb71cf09a8df4753e70080d510

//bash-4.2# hhvm tests/test_hhvm_mem_usage2.php
//MEMORY IS FINE HERE
//Memory Usage: 2.00M
//Memory Usage: 2.00M
//Memory Usage: 2.00M
//Memory Usage: 2.00M
//Memory Usage: 2.00M
//Memory Usage: 2.00M
//Memory Usage: 2.00M
//Memory Usage: 2.00M
//Memory Usage: 2.00M
//Memory Usage: 2.00M
//MEMORY LEAKS HERE
//Memory Usage: 2.00M
//Memory Usage: 246.00M
//Memory Usage: 490.00M
//Memory Usage: 734.00M
//Memory Usage: 0.96G
//Memory Usage: 1.19G
//Memory Usage: 1.43G
//Memory Usage: 1.67G
//Memory Usage: 1.91G
[...]

Issue #4846, opened 2015-02-16
extract() + namespaces = memory leak


terah:
``` php
<?php

// The namespace is part of the bug
// remove it and the bug disappears
namespace HHVMTest\BuggyStaticClass;

class BuggyExtractClass
{
    public static function bigIteration()
    {
        for ( $i = 0; $i < 1000000 ; $i++ )
        {
            static::modifyRow(['value' => 1, 'value2' => 'something else', 'value3' => 2.3]);
            static::modifyRow(['value' => 1, 'value2' => 'something else', 'value3' => 2.3]);
            static::modifyRow(['value' => 1, 'value2' => 'something else', 'value3' => 2.3]);
            if ( $i % 100000 === 0 )
            {
                echoHumanFileSize();
            }
        }
        return true;
    }

    protected static function modifyRow($args)
    {
        // Memory is lost here
        extract($args);
    }
}

BuggyExtractClass::bigIteration();

function echoHumanFileSize()
{
    $bytes  = memory_get_usage();
    $sz     = 'BKMGTP';
    $factor = floor(( strlen($bytes) - 1 ) / 3);
    $mem    = sprintf("%.2f", $bytes / pow(1024, $factor)) . $sz[(int)$factor];
    echo("Memory Usage: {$mem}" . PHP_EOL);
}

//bash-4.2# hhvm --version
//HipHop VM 3.6.0-dev (rel)
//Compiler: heads/master-0-ga9e2a34d62dd95d72eb31b2e73ce475d73d87e2f
//Repo schema: e05b63d7e028ccfb71cf09a8df4753e70080d510

// Output
//bash-4.2# hhvm test_hhvm_mem_usage.php
//Memory Usage: 2.00M
//Memory Usage: 84.00M
//Memory Usage: 166.00M
//Memory Usage: 248.00M
//Memory Usage: 330.00M
//Memory Usage: 414.00M
//Memory Usage: 496.00M
//Memory Usage: 578.00M
//Memory Usage: 660.00M
//Memory Usage: 742.00M
```

wjywbs:
The difference without and with namespace in hhas is:

```
   .method [protected static] modifyRow($args) {
-    FPushFuncD 1 "__SystemLib\\extract"
+    FPushFuncU 1 "HHVMTest\\BuggyStaticClass\\extract" "extract"
```

[...]

PR #4882, opened 2015-02-23
Tests for reported memory leaks

terah:
Tests for:
https://github.com/facebook/hhvm/issues/4863
https://github.com/facebook/hhvm/issues/4846


sgolemon:
Per comment on https://reviews.facebook.net/D33885 I'm closing this PR without mergoing, @edwinsmith has the repro case he can use in his GC work which will address the underlying cause.

PR #4903, opened 2015-02-26
fixed a type for building extensions 

Degola:
while including hphp/runtime/base/base-includes.h in extensions source the building failed with
error: ‘HHVM_VERISON_BRANCH’ was not declared in this scope
 HHVM_GET_MODULE(geoip);

The corresponding typo is in hphp/runtime/ext/extension.h and is fixed with this pull request.


sgolemon:
Thanks for the PR, but @rrh sent the same thing in and that's been merged already.

Issue #1871, opened 2014-02-19
Generator behavior doesn't match PHP5


asm89:
Apart from syntax (https://github.com/facebook/hhvm/issues/1627) and instanceof (https://github.com/facebook/hhvm/issues/1787) issues, generators in hhvm and zend behave differently.

The following gist contains three examples that show different behavior/output between hhvm and zend:
https://gist.github.com/asm89/9102248

JeroenDeDauw:
+1 I'm also running into this

noony:
https://3v4l.org/CjaLS
https://3v4l.org/95Q6f
https://3v4l.org/Clno2

Seems ok in 3.11. Can we close ?

fredemmott:
Yep, likely fixed at the same time as the PHP7 changes were implemented.

PR #4915, opened 2015-02-27
Implement auto-priming for generators

nikic:
This implements auto-priming for generators to match the PHP 5 behavior. It's a partial fix of #1871, there are a number of other incompatibilities. Async generators don't use auto-priming.

The implementation uses the approach described in https://github.com/facebook/hhvm/issues/1871#issuecomment-53123517. This is my first PR to HHVM, so I have no idea what I'm doing ;)


paulbiss:

nikic:
@paulbiss Done :)

fredemmott:
We review on phabricator- this change is here: https://reviews.facebook.net/D34161

nikic:
Any updates here?

jano:
Sorry, I don't have good news. The usage of generators in our codebase seems to be significantly higher than I originally expected and I can't justify spending significant amount of time on something outside of scope of my work priorities :(. I will let folks from our HHVM OSS team comment on what is the plan here.

nikic:
Push

JoelMarcey:
OK, so this diff needs non-trivial internal changes that at one point was on our plate to do, but is no longer. 

I think what I am going to do here is close this pull request to get it out of our queue. Then, I am going to file an internal task and assign it to proper people to figure out what, if anything, we can do to move forward with this.

Then, if something is done, we can push out the code of this PR, maintaining authorship credit, etc.

nikic:
Awwwwww

Maybe we can couple this to the PHP 7 runtime switch? If I got it right it's going to allow per-dir/per-file granularity, so if we change this only for PHP 7 mode this might allow gradually porting things?

JoelMarcey:
@nikic Let me see what we can do. Maybe that is a possibility. Good idea. The diff as it stands seems a little dense for a runtime switch, but maybe that is ok. This is all on our end now. I can bring this up at our internal meeting today. We might have to wait until @jano gets back too.

fredemmott:
I think @paulbiss was of the opinion that this was too invasive for a runtime switch

paulbiss:
Hmm, the original conversation was around doing an `#ifdef FACEBOOK`. Looking at this again, I still think switching on a runtime setting would be intrusive but given how long this has sat waiting for a resolution on our end I'm coming around to the idea of doing it that way. My main concern there would be having test coverage for both modes.

@jano @sgolemon @jwatzman: What are your thoughts?

jwatzman:
Yeah, IMO if we can do anything to get this in, ifdef or runtime option or whatever, we should do it. This comes up fairly often as an annoying incompatibility, and is worth a little ugliness to get right.

As far as test coverage, I don't see that as being much of an issue -- I assume we have fairly adequate coverage of the existing mode, so instead of changing the tests as this PR does right now, we can just add the changed tests as new ones, have two copies of a few of them for each mode, doesn't seem too bad.

paulbiss:
Okay, unless anyone strongly objects I think gating on a runtime option (and default to the correct PHP behavior) and testing both modes with a .opts files as necessary is a good enough solution. The plumbing here is going to be a little gross but it should be doable. I'm really sorry we dropped the ball here, I'd be happy to help get this over the line.

It might be worth discussing adding some logging when we detect incompatible behavior at runtime so we can start to clean things up internally.

PR #4917, opened 2015-02-27
Fix #4916 by using is_a() to check if the classname provided is valid

### LINKED ISSUE #4916: Default argument throws exception
```php
<?php

$spl = new SPLFileInfo("README.md");
$spl->setFileClass();
$spl->setInfoClass();
$o = $spl->openFile();
$f = $spl->getFileInfo();
var_dump(get_class($o), get_class($f));

?>
```
Fatal error: Uncaught exception 'UnexpectedValueException' with message 'SplFileInfo::setFileClass() expects parameter 1 to be a class name derived from SplFileObject, 'SplFileObject' given' in :

The same happens when "SplFileObject" is passed (and "SplFileInfo" to setInfoClass).

running the above code in php returns:
```
string(13) "SplFileObject"
string(11) "SplFileInfo"
```

bjori:
This allows the user to "reset" previously passed in values by passing
in the default value again.
Furthermore, it fixes the exception that is thrown when no arguments
are passed.


jmikola:
@bjori: I'll follow up with a regression test PR against your branch, so we can get that into here as well.

Additionally, I updated the PHP documentation (likely responsible for HHVM's implementation difference) in http://svn.php.net/viewvc?view=revision&revision=336017

bjori:
The TEST_RUN_MODE="-m jit quick" seems to have failed due to unrelated error (g++ killed 40% through the build)
I don't seem to have permission to restart the build.

paulbiss:

PR #4935, opened 2015-03-02
Fix a typo in a macro name related to HHVM versions and DSOs.

rrh:
[[Was this code ever tested?]]


sgolemon:
Hrmm, this should have been caught by our dso_test build.  Bothers me that it didn't catch it. :(

PR #4939, opened 2015-03-03
Extend ZEND_GET_MODULE to present getModuleBuildInfo.

rrh:
Function getModuleBuildInfo must now be callable at dlopen time
per recent changes to API/DSO revision level checking.

With this change zend extensions can be DSOs.


sgolemon:
Thanks, Robert!

PR #4941, opened 2015-03-03
Adding empty back-end for PPC64 and arch specific code

lbianc:
This change adds an empty back-end for PPC64 architecture. Currently HHVM works on PPC64 with JIT disabled.


JoelMarcey:
There was agreement that we can close this pull request, but @lbianc would like to have an IRC or Skype discussion with @sgolemon about this.

PR #4943, opened 2015-03-03
PDOMySqlStatement::nextRowset() now returning true on empty result

### LINKED ISSUE #4942: PDOMySqlStatement::nextRowset() returns false on queries with no result
Hello guys,

I was having problems with `nextRowset()` in PDO (along with MySQL) while using multiple DML queries. As it seems from PHP 5 (althought docs are not very informative), `nextRowset()` should suceed (i.e. return true) even when there is no result. Now HHVM returns false.

So, I believe it makes sense to return true (mainly because of the ability to use `rowCount()`).

chmeldax:
Fixes #4942

According to MySQL docs:

Anyway, to put some light on my solution with m_result:
```
mysql_store_result() returns a null pointer if the statement did not return a result set (for example, if it was an INSERT statement).

mysql_store_result() also returns a null pointer if reading of the result set failed. You can check whether an error occurred by checking whether mysql_error() returns a nonempty string, mysql_errno() returns nonzero, or mysql_field_count() returns zero.
```

Also, I've tried to fix bug with shadowing public property `row_count` by local variable.

### DIFF
--- a/hphp/runtime/ext/pdo_mysql.cpp
+++ b/hphp/runtime/ext/pdo_mysql.cpp
@@ -1284,22 +1284,29 @@ bool PDOMySqlStatement::nextRowset() {
     return false;
   }
 
-  my_ulonglong row_count;
+  my_ulonglong affected_count;
   if (!m_conn->buffered()) {
     m_result = mysql_use_result(m_server);
-    row_count = 0;
+    affected_count = 0;
   } else {
     m_result = mysql_store_result(m_server);
-    if ((my_ulonglong)-1 == (row_count = mysql_affected_rows(m_server))) {
+    if ((my_ulonglong)-1 == (affected_count = mysql_affected_rows(m_server))) {
       handleError(__FILE__, __LINE__);
       return false;
     }
   }
-
+  row_count = affected_count;
+  
   if (!m_result) {
-    return false;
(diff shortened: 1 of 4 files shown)


hphp/runtime/ext/pdo_mysql.cpp
```diff
@@ -1284,22 +1284,29 @@ bool PDOMySqlStatement::nextRowset() {
...
   } else {
     m_result = mysql_store_result(m_server);
-    if ((my_ulonglong)-1 == (row_count = mysql_affected_rows(m_server))) {
+    if ((my_ulonglong)-1 == (affected_count = mysql_affected_rows(m_server))) {
       handleError(__FILE__, __LINE__);
       return false;
     }
   }
-
+  row_count = affected_count;
```
sgolemon: Where is row_count being declared?

chmeldax: I believe it's here: https://github.com/facebook/hhvm/blob/master/hphp/runtime/ext/pdo_driver.h#L649 :)

sgolemon: Ugh... smh for bad naming conventions (not your fault, but ugh)

PR #4956, opened 2015-03-05
cmake changes for using LLVM library

sathvikl:
Using find_package directive provided by cmake is more robust.


fredemmott:
Will re-open when updated (see phabricator)

Issue #5002, opened 2015-03-11
HHVM 3.6 binary packages have incorrect mysql socket path


fredemmott:
- Need rebuilding with mysql-server installed
- Given we're now building a mysqlclient, we should probably error if cmake can't figure out where the socket is

See discussion in #4987

r3wt:
damnit. damnit. damnit! GRRR

!(http://puu.sh/gxWmi/dbd6dd7cba.png)

PDowney:
For those early upgraders that ended up killing their wordpress-based websites (like me), you can temporarily get around this issue by switching your server from using unix sockets. Just return hhvm.conf and server.ini to their defaults, and then edit wordpress' wpconfig.php file to use 127.0.0.1 instead of localhost.

r3wt:
@PDowney what about someone running an ORM like Redbeans? do we have options

PDowney:
I have no experience with Redbean whatsoever, but i imagine you'd just change it to point to 127.0.0.1 instead of localhost in the same way.

So if the text below is how it's set up, just replace localhost I guess, and then change hhvm.conf and server.ini back to defaults. 
R::setup('mysql:host=localhost;dbname=mydatabase',
        'user','password'); //mysql

frijsdijk:
Even if I set hhvm.mysql.socket=/var/run/mysqld/mysqld.sock in /etc/hhvm/php.ini, which shows up in a hhvm-phpinfo, it still seems to look for /tmp/mysql.sock in stead.

Btw, this works as well as a quick work-around (but dirty):

ln -s /var/run/mysqld/mysqld.sock /tmp/mysql.sock

fredemmott:
@frijsdijk : probably #5003

fredemmott:
This will be fixed in 3.6.1 - #5003 won't be, but is less of an issue once this is.

fredemmott:
Fixed in 3.6.1

PR #5005, opened 2015-03-11
[master] Raise a cmake error if MySQL socket path can't be identified

fredemmott:
Tested by:
 1. deleted CMakeCache.txt, re-ran cmake with mysql-server installed, worked
 2. deleted CMakeCache.txt, removed mysql-server. Re-ran cmake, failed
 3. deleted CMakeCache.txt, specified -DMYSQL_UNIX_SOCK_ADDR=/var/run/mysqld/mysqld.sock, worked

Tested at runtime too for 1 and 3:

```
hphpd> mysql_connect('localhost')
mysql_connect('localhost')

Warning: mysql_connect(): Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)
```

refs facebook/hhvm#5002


fredemmott:
Cherry pick version: #5006

PR #5032, opened 2015-03-17
Travis modernize build-in

denji:
* Try to use gcc-4.9 soon gcc-5
* `SCRIPT_DIR` correct work for Darwin
* `CPUS` more correct definition cores
* Stable version `libevent-2.0.22-stable`, `curl-7.41.0` and `jemalloc-3.6.0`
* clang added but not enabled, the problem is in the patches clang compiler
* silent `apt-get` removed unnecessary items from the logfile
* check `more /proc/meminfo` and `vmstat`
* enable `ccache`, disable `cotire`

> **Travis CI Support**
apt-fast: We are speeding up apt-get with a proxy we run on the VM hosts.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,19 @@
 language: cpp
 
-# This doesn't work - travis defaults to plain gcc if unknown
-# http://github.com/travis-ci/travis-ci/issues/979
-# http://llvm.org/apt/
+os:
+  - linux
+#  - osx
+#osx_image: xcode61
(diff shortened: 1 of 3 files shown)


jrobeson:
https://github.com/facebook/hhvm/commit/658ca01c71251ea1b4b8667d402915269906e86e

JoelMarcey:
We aren't supporting the Travis builds any longer.

PR #5042, opened 2015-03-18
Check for empty docroot before looking for trailing slash.

adsr:
We ran into this case during warmup requests on a debug build. I didn't look into what the docroot is supposed to be for warmup requests, or why we only ran into this now, but this prevents a segfault in one case on our end.


jwatzman:
Closing per https://github.com/facebook/hhvm/wiki/Human-Timeouts -- if you're still intending to work on this, let us know, and we can reopen when you update.

PR #5073, opened 2015-03-24
Fix fread on an SSL socket potentially waiting for data from socket when OpenSSL already has data

### LINKED ISSUE #4315: fread may not return all the data for TLS/OpenSSL sockets
I'm trying to isolate a repro for this, but I wanted to write down what I know before I quit for the weekend (and also in case an HHVM dev is able to easily spot the issue with the info I already have).

We use persistent sockets to send data back and forth from our PHP code to other backend servers. In regular PHP, this works fine regardless of whether or not we use TLS. With HHVM, this only works reliably for non-SSL/TLS sockets. When using a TLS socket and HHVM, depending on the data sent by the other server, the fread call may not return the full response and hang polling for the remainder.

The response from our server that HHVM is reading has a header that contains the number of bytes that was sent and then the sent data. So we fread the header, check for integrity, get the number of bytes to read, and then fread until we have read that many bytes. When using HHVM on a persistent TLS socket, some responses are never completely read. We'll see something like the following in our error log:
```Request returned no data after the header and then 7 more reads (expected 3899 more bytes of 61243). ```

If we retry and the other server resends the exact same response, it gets stuck in the same place. It doesn't matter if we read one byte at a time or try to read the entire thing in one fread call. fread will return some data and then no more.

I was wondering if maybe SSLSocket needs to override waitForData and check SSL_pending, but I have not yet checked to see if that fixes the issue.

The thread's stack will be something like this while waiting for the remainder of the data:
```
0x00007fb3a30d1963 in poll () from /lib/x86_64-linux-gnu/libc.so.6
#0  0x00007fb3a30d1963 in poll () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00000000008fe2d9 in HPHP::Socket::waitForData() ()
#2  0x00000000009aa389 in HPHP::SSLSocket::readImpl(char*, long) ()
#3  0x00000000009bc478 in HPHP::File::read(long) ()
#4  0x000000000112011b in HPHP::f_fread(HPHP::Resource const&, long) ()
#5  0x0000000003b383eb in ?? ()
#6  0x0000000002139bf3 in enterTCHelper ()
#7  0x0000000002139c69 in HPHP::jit::x64::BackEnd::enterTCHelper(unsigned char*, HPHP::jit::TReqInfo&) ()
#8  0x0000000002139675 in HPHP::jit::MCGenerator::enterTC(unsigned char*, void*) ()
#9  0x00000000021388c9 in HPHP::ExecutionContext::invokeFunc(HPHP::TypedValue*, HPHP::Func const*, HPHP::Variant const&, HPHP::ObjectData*, HPHP::Class*, HPHP::VarEnv*, HPHP::StringData*, HPHP::ExecutionContext::InvokeFlags) ()
#10 0x0000000002196647 in HPHP::ExecutionContext::invokeUnit(HPHP::TypedValue*, HPHP::Unit const*) ()
#11 0x00000000009b9797 in HPHP::include_impl_invoke(HPHP::String const&, bool, char const*) ()
#12 0x0000000000991850 in HPHP::hphp_invoke(HPHP::ExecutionContext*, std::string const&, bool, HPHP::Array const&, HPHP::VRefParamValue const&, std::string const&, std::string const&, bool&, std::string&, bool, bool, bool) ()
#13 0x0000000000b0b765 in HPHP::HttpRequestHandler::executePHPRequest(HPHP::Transport*, HPHP::RequestURI&, HPHP::SourceRootInfo&, bool) ()
#14 0x0000000000b0cdfa in HPHP::HttpRequestHandler::handleRequest(HPHP::Transport*) ()
```

apaloma:
Resolves #4315. OpenSSL may have data ready without the socket necessarily having data ready to read.

### DIFF
--- a/hphp/runtime/base/ssl-socket.cpp
+++ b/hphp/runtime/base/ssl-socket.cpp
@@ -349,7 +349,7 @@ SmartPtr<SSLSocket> SSLSocket::Create(
 }
 
 bool SSLSocket::waitForData() {
-  if (m_ssl_active && SSL_pending(m_handle)) {
+  if (m_data->m_ssl_active && SSL_pending(m_data->m_handle)) {
     return true;
   }
 
--- a/hphp/runtime/base/ssl-socket.cpp
+++ b/hphp/runtime/base/ssl-socket.cpp
@@ -348,13 +348,21 @@ SmartPtr<SSLSocket> SSLSocket::Create(
   return sock;
 }
 
+bool SSLSocket::waitForData() {
+  if (m_ssl_active && SSL_pending(m_handle)) {
+    return true;
+  }
+
+  return Socket::waitForData();
+}
+
 int64_t SSLSocket::readImpl(char *buffer, int64_t length) {
   int64_t nr_bytes = 0;
   if (m_data->m_ssl_active) {
     bool retry = true;
     do {
       if (m_data->m_is_blocked) {
-        Socket::waitForData();
+        waitForData();
         if (timedOut()) {
           break;
         }
--- a/hphp/runtime/base/ssl-socket.h
+++ b/hphp/runtime/base/ssl-socket.h
@@ -63,6 +63,7 @@ struct SSLSocket : Socket {
   // overriding ResourceData
   const String& o_getClassNameHook() const override { return classnameof(); }
 
+  bool waitForData();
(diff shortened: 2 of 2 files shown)


apaloma:
Sorry about the build issue. It was a patch from an older build.

PR #5074, opened 2015-03-25
The Extension API check does not work.

keyurdg:
The Extension class constructor implementation is provided in the cpp
file. So when the extension's static object s_##name_extension is constructed,
the constructor called is in the HHVM binary and hence the m_hhvmAPIVersion
private member variable is initialized to HHVM_API_VERSION defined in the
binary itself, not what it was when the extension was compiled.
The patch moves the Extension constructor to the header file to fix the issue.

### DIFF
--- a/hphp/runtime/ext/extension.cpp
+++ b/hphp/runtime/ext/extension.cpp
@@ -108,10 +108,7 @@ static const char* dlerror() {
 
 ///////////////////////////////////////////////////////////////////////////////
 
-Extension::Extension(litstr name, const char *version /* = "" */)
-    : m_hhvmAPIVersion(HHVM_API_VERSION)
-    , m_name(name)
-    , m_version(version ? version : "") {
+void Extension::registerExtension(litstr name) {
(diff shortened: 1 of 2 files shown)


keyurdg:
Ah never mind, I see this was fixed by commit 1b2addb77d

keyurdg:
I'm reopening as I see Sara's commit is not in [3.6 LTS](https://github.com/facebook/hhvm/blob/HHVM-3.6/hphp/runtime/ext/extension.cpp#L112). I wonder if my patch would qualify as a bug fix until the changes from Master come out in an LTS. 

We at Etsy had a situation where we were running 3.4 and upgraded to 3.5 while forgetting to rebuild the extensions (all in Dev thankfully) and it caused weird memory corruption bugs. It took us a while to figure out the cause.

r3wt:
@keyurdg ive had the same problem with redis and libgmp when upgrade from 3.3 to 3.5. glad to see someone has solved it

keyurdg:
Reopened as PR 5145

PR #5097, opened 2015-03-31
Add some configurability to hh_format

Aatch:
Adds the ability to set the values of a few important settings in the
formatter. These are currently pulled from the `.hhconfig` file. Each
option is prefixed with "format." to avoid potential conflicts with
other configuration variables.

The options all default to the values/behaviour that were already
present.

There are currently 4 options:

* Indent offset ("indent_offset") - Set the number of spaces used for
  indentation. Default 2.
* Line width ("line_width") - The maximum number of characters in a
  line. Default 80.
* Indent case ("indent_case") - Indent cases in switch statements.
  Default off.
* Align cascaded calls ("align_cascaded_calls") - Align cascaded method
  calls on the '->' or '?->' like this:

  ```php
  $foo->bar()
      ->baz()
      ->done();
  ```

  Defaults to off.

### DIFF
--- a/hphp/hack/src/format/formatConfig.ml
+++ b/hphp/hack/src/format/formatConfig.ml
@@ -1,0 +1,61 @@
+(**
+ * Copyright (c) 2014, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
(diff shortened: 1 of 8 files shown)


fredemmott:
Will re-open when updated. See comments on phabricator.

PR #5107, opened 2015-04-01
Fix TRIM CHARS

gfleury:
Change HHVM TRIM CHARS for  compatibility with PHP5


fredemmott:
Closing for now (https://github.com/facebook/hhvm/wiki/Human-Timeouts) - will re-open when test failures on reviews.facebook.net are addressed.

PR #5122, opened 2015-04-03
Improve reflection compatibilty

simonwelsh:
Corrects some of the incompatibilites in ext/reflection.

Major changes are:
* Error output when passing incorrect number of arguments
* NPH for ReflectionMethod
* __debugInfo methods to hide internals

Test Plan: Existing tests including previously bad ones


simonwelsh:
Ugh, `Test Plan` is still case-sensitive.

simonwelsh:
Notice a couple of things when doing this:
* References to objects stop being references when ref-count hits 1
* PHP has a different return value for property/method modifiers if the property/method has the same name as a private property/method in a parent class

fredemmott:
Closing for failing tests for > ~ 1 week. Will re-open when updated.

PR #5126, opened 2015-04-04
Move some constants out of IDL

simonwelsh:
More killing of IDL

Test Plan: existing tests, especially good Zend ones


fredemmott:
Consistently failing tests for > 1 week - see Phabricator. WIll re-open when updated.

PR #5138, opened 2015-04-07
imporve hhvm report level performance

huzhiguang:
HHVM new version create msg varible before compare report level,this way can reduce hhvm more performance.
Found that the problem is my colleague @wwbmmm and I have fixed this problem.

### DIFF
--- a/hphp/runtime/base/runtime-error.cpp
+++ b/hphp/runtime/base/runtime-error.cpp
@@ -24,7 +24,29 @@
 
 namespace HPHP {
 ///////////////////////////////////////////////////////////////////////////////
+static int64_t g_warning_counter = 0;
+static int64_t g_notice_counter = 0;
 
+bool checkReportLevel(ErrorMode mode) {
+  int errnum = static_cast<int>(mode);
+  if (!g_context->errorNeedsHandling(errnum, true,
+                              ExecutionContext::ErrorThrowMode::Never)) {
+    return true;
+  } else if (RuntimeOption::WarningFrequency <= 0 ||
+        (g_warning_counter++) % RuntimeOption::WarningFrequency != 0) {
+      return true;
+  } else if (RuntimeOption::NoticeFrequency <= 0 ||
+             (g_notice_counter++) % RuntimeOption::NoticeFrequency != 0) {
+    return true;
+  }
+  return false;
+}
+
+#define CHECK_REPORT_LEVEL(mode) \
+   if (checkReportLevel(mode)) { \
+      return ; \
+   } 
+   
 /*
  * Careful in these functions: they can be called when tl_regState is
  * REGSTATE_DIRTY.  ExecutionContext::handleError is dirty-reg safe,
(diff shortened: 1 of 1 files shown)


hphp/runtime/base/runtime-error.cpp
```diff
@@ -24,7 +24,29 @@
 
 namespace HPHP {
 ///////////////////////////////////////////////////////////////////////////////
+static int64_t g_warning_counter = 0;
+static int64_t g_notice_counter = 0;
+
+bool checkReportLevel(ErrorMode mode) {
```
reeze: How about adding a `static` modifier？  or can we make it inline function?

hphp/runtime/base/runtime-error.cpp
```diff
@@ -24,7 +24,29 @@
...
+  int errnum = static_cast<int>(mode);
+  if (!g_context->errorNeedsHandling(errnum, true,
```
reeze: Since we checked need handling here, can we remove   https://github.com/huzhiguang/hiphop-php/blob/patch-13/hphp/runtime/base/runtime-error.cpp#L326 the checking  here? they are duplicated.

hphp/runtime/base/runtime-error.cpp
```diff
@@ -24,7 +24,29 @@
...
+  } else if (RuntimeOption::NoticeFrequency <= 0 ||
+             (g_notice_counter++) % RuntimeOption::NoticeFrequency != 0) {
+    return true;
+  }
+  return false;
+}
 
+#define CHECK_REPORT_LEVEL(mode) \
+   if (checkReportLevel(mode)) { \
+      return ; \
```
reeze: no space here

hphp/runtime/base/runtime-error.cpp
```diff
@@ -62,6 +85,7 @@ void raise_recoverable_error(const std::string &msg) {
 }
 
 void raise_recoverable_error_without_first_frame(const std::string &msg) {
+  CHECK_REPORT_LEVEL(ErrorMode::RECOVERABLE_ERROR)
```
reeze: I think we should add a semicolon here.

hphp/runtime/base/runtime-error.cpp
```diff
@@ -93,6 +117,7 @@ void raise_disallowed_dynamic_call(const std::string& msg) {
 }
 
 void raise_recoverable_error(const char *fmt, ...) {
+  CHECK_REPORT_LEVEL(ErrorMode::RECOVERABLE_ERROR)
```
reeze: So does here and the others.

JoelMarcey:
Due to our week timeout policy (https://github.com/facebook/hhvm/wiki/Human-Timeouts) I am closing this pull request. If changes are made as requested on https://reviews.facebook.net/D36597 (at this point it needs at least a rebase), then please feel free to reopen the request and we will continue to iterate on the diff.

Thanks!

PR #5144, opened 2015-04-08
Fix ZendSorting performance.

keyurdg:
Details: The old implementation use zend-collator which does a lot of useless
conversions to UTF-16 back to UTF-8 and comparisons using ICU, all of which
gets really expensive.
The new implementation ports the existing Zend functionality to HHVM and
relies on regular string/numeric comparison functions where possible. The
performance is an order of magnitude faster.


solhuebner:
Thank you for your effort!

Any news on this?

PR #5145, opened 2015-04-08
The Extension API check does not work

keyurdg:
The Extension class constructor implementation is provided in the cpp file. So when the extension's static object s_##name_extension is constructed,
the constructor called is in the HHVM binary and hence the m_hhvmAPIVersion
private member variable is initialized to HHVM_API_VERSION defined in the
binary itself, not what it was when the extension was compiled.
The patch moves the Extension constructor to the header file to fix the issue.

The proper fix is coming in commit 1b2addb77d21b43447098c7b0cf1cfed31feebb9,
this is a placeholder until then. I've added a dummy getModuleBuildInfo()
method to aid in the transition

This reopens PR #5074 properly rebased to HHVM-3.6

### DIFF
--- a/hphp/runtime/ext/extension.cpp
+++ b/hphp/runtime/ext/extension.cpp
@@ -108,10 +108,7 @@ static const char* dlerror() {
 
 ///////////////////////////////////////////////////////////////////////////////
 
-Extension::Extension(litstr name, const char *version /* = "" */)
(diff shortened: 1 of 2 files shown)


fredemmott:
Closing as I don't think this is suitable for 3.6; see phabricator.

PR #5149, opened 2015-04-08
Hhvm 3.6

AndroidTox:


paulbiss:
I don't think this is what you intended to do.

Issue #4444, opened 2014-12-13
Prioritize MacOS build/binary support


Ocramius:
As it stands, `brew install hhvm` provided by https://github.com/mcuadros/homebrew-hhvm is broken.
Same goes for building HHVM from source on MacOS.

While I don't love MacOS too much, I'm currently using it, as most of the people in my github network (according to google analytics).

It would probably be a good idea to provide either binaries or a compilable version, and to prioritize the issue.

I know for sure that builds on MacOS stopped working because of @travis-ci workers being too slow at picking up jobs for this platform, so I'm pinging them as well.

christopherowen:
#4636

reeze:
:+1:

lovek323:
:+1

lovek323:
:+1:

lima01:
+1

ptica:
+1

cweagans:
:+1:

craigcarnell:
:+1:

Ocramius:
Folks, upvoting does not really help: make a contribution instead.

lovek323:
Yeah, but it allows me to get notifications of updates to the issue. Though, I'm starting to see the flaw in my thinking. :smile:

steveh:
You can subscribe to notifications instead on the right nav bar -->

lovek323:
And now I feel stupid :disappointed:

santagada:
I feel a little impotent because there is a huge todo list of many patches ready to make it work and none of them are checked, so I think this is waiting for something that I have zero control of. I saw that 3.6 is supposed to be released soon and with support for osx. I wish I knew what is missing.

denji:
@santagada [`proxygen`](https://github.com/facebook/proxygen#installing) appeared in HHVM-3.6 and does not support OSX.

santagada:
Will it be built using clang or gcc? If it is going to use clang than I think all of those patches for folly, hhvm-thirdparty and proxygen need to be applied.

Knowing wich compiler the main devs want to support and what exactly needs to be done (porting proxygen to osx) then someone can help (my c++ foo is very weak or I would give it a try).

simonwelsh:
* It's likely to support gcc first, with clang support coming later
* 3.6.0 won't support OS X (the branch has already been cut). [This](http://hhvm.com/blog/8405/coming-soon-in-hhvm) blog post just says it's coming later in 2015.

wjywbs:
[...]

PR #5173, opened 2015-04-14
Cache time zone validity and add lock for calling timelib on Mac.

wjywbs:
date_default_timezone_set() is called once for every WordPress request in
stock WordPress 4.1.1. Cache the time zone validity for better performance.

A lock is needed when calling timelib_timezone_id_is_valid() because
setlocale() on Mac is not thread safe. When the first argument to setlocale
is LC_CTYPE, loadlocale() will call __setrunelocale() to modify the global
locale object and double free can happen and crash.

http://www.opensource.apple.com/source/Libc/Libc-997.90.3/locale/FreeBSD/setrunelocale.c

Part of #4444.

### DIFF
--- a/hphp/runtime/base/program-functions.cpp
+++ b/hphp/runtime/base/program-functions.cpp
@@ -1762,6 +1762,10 @@ void hphp_process_init() {
   pthread_attr_destroy(&attr);
 
   Process::InitProcessStatics();
+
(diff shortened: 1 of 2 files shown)


fredemmott:
See comments on reviews.facebook.net re test failures. Will re-open when updated.

PR #5188, opened 2015-04-19
add JsonSerializable interface on DateTimeImmutable class for json_encode

### LINKED ISSUE #5137: DateTimeImmutable not json_encoded by HHVM
http://3v4l.org/ragDN
`DateTime` is json_encoded but not `DateTimeImmutable`

```php
//works
echo json_encode(new DateTime('2015-05-03'));
//does not work returns empty brackets
echo json_encode(new DateTimeImmutable('2015-05-03')); //{}
```

noony:
fix #5137


fredemmott:
Sorry, going to revert this - the right fix is to modify ObjectData::toArray()

PR #5191, opened 2015-04-20
Fix building gd extension with libvpx 1.4.0

mtorromeo:
In libvpx 1.4.0 the compatibility layer for some of the older constants have been removed [1].
The use of this constants was deprecated since 2010 in favor of the newer VPX_ prefixed names.
The same issue as been reported in the php bugtracker [2].

Thanks!

[1] https://gerrit.chromium.org/gerrit/#/c/70164/3/vpx/vpx_image.h,unified
[2] https://bugs.php.net/patch-display.php?bug_id=69479&patch=php-gd-newvpx.diff&revision=latest


yan12125:
Excellent work :+1: With this patch, the build process runs fluently on Arch Linux and all quick tests and slow/ext passed. I hope this PR can be merged as soon as possible.

PR #5222, opened 2015-04-29
socket timeout support float

thbourlove:
float value is supported by php5 (http://php.net/manual/en/function.stream-socket-client.php)


thbourlove:
Any news on this pull request? All I need is to bind a float value as timeout of socket client.

paulbiss:
We handle all pull request communication using phabricator, please see the link, it looks like this PR is waiting on changes from you. Thanks!

https://reviews.facebook.net/D37821

thbourlove:
@paulbiss thanks.

PR #5249, opened 2015-05-05
Update Joomla to 3.4.1

photodude:


photodude:
@fredemmott does a new PR for this need to be opened or can this one be reopened?

fredemmott:
Please send a new PR; try also deleteing framework_downloads before re-running. If this works correctly, we would expect to see a change to one of the .expect files too, not just frameworks.yaml.

photodude:
@fredemmott I'll have to see about setting up a local environment and updating in the same fashion as  adding a new framework for a new PR. This PR was created by just modifying frameworks.yaml via github.com

fredemmott:
Ah, OK, that explains why it doesn't work. Yep, this'll need testing before sending. The yaml file also tries to disable tests that don't exist in this version.

photodude:
Good to know, I'll have to review the tests list for additions and removals too.

fredemmott:
Please actually run it.
- Most additions and removals will be fine
- Removals are only a problem if they are referenced in the yaml file for some reason - and the script will tell you about these. Manual review shouldn't be neccessary
- You need to run it to generate the expect file anyway.

photodude:
Will, do. 

Are there guidlines on what tests are necessary to be included in the yaml file, or is all of that handled by the run?

fredemmott:
They're only added if they're problematic for some reason - eg if take an hour to run, or are unreliable (assuming that calling time() twice in a row gives the same result is a common one).

photodude:
So in general practice, would it be best to remove all of the tests for the framework from the yaml file and then add them back if they are "problematic for some reason"?

fredemmott:
Leave them as-is. Remove them if there's a reason to - eg they don't exist any more, or the test has been modified.

It can take many runs to find out which tests fail intermittently, so giving it a clean slate each time it's updated isn't generally a good idea.

fredemmott:
oh, you'll want to run with --record to update the .expect file, instead of merely compare against it.

PR #5340, opened 2015-05-12
Fix protected output buffer flushing

tstarling:
* Implement OutputDisabled and Flushable output buffer flags and fix RPC early f
lush
* Fix HEAD handling, don't require buffering
* Migrate HTTP/1.0 handling to new OutputDisabled flag
* Allow obFlush() of top-level output buffer

### DIFF
--- a/hphp/runtime/base/execution-context.cpp
+++ b/hphp/runtime/base/execution-context.cpp
@@ -200,6 +200,7 @@ static void safe_stdout(const  void  *ptr,  size_t  size) {
 }
 
 void ExecutionContext::writeStdout(const char *s, int len) {
+  fflush(stdout);
   if (m_stdout == nullptr) {
     if (s_stdout_color) {
       safe_stdout(s_stdout_color, strlen(s_stdout_color));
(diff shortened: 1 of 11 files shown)


fredemmott:
Getting this out of the HHVM queue for now, pending fixes in proxygen.

paulbiss:
Not actually blocked on proxygen per @afrind

JoelMarcey:
So, we need to figure out what is going on with our internal FB extension test failures before this can be considered to land. I have filed an internal task describing the problem and copying the right people so if there is movement, then everyone will know.

I am going to close this PR for the meantime until if and when we can figure out what is going on. If we do, we can land this appropriately.

tstarling:
OK. In case anyone's forgotten, this PR is preparatory work towards the goal of allowing large streaming output, larger than available memory, in both the Proxygen and FastCGI backends. I have a patch already written which implements said streaming on top of this preparatory work.

JoelMarcey:
@tstarling Thanks Tim. I will make sure that the task itself doesn't fall of our radar for sure.

JoelMarcey:
T8186806

PR #5354, opened 2015-05-15
One-bit refcount analysis

psmarshall:
Don't actually merge me in, this is just an instrumented build to show the behaviour of a one-bit-refcount / multiple reference bit (MRB). @edwinsmith


JoelMarcey:
I am cleaning out our PR queue. The actual phabricator diff will remain for further discussion https://reviews.facebook.net/D38511 if you would like to chime on on this.

@edwinsmith @bnell

PR #5366, opened 2015-05-18
Folly/Singleton.h is no longer experimental

samveen:
Singleton.h has been pulled out of experimental and moved into the main folly tree as per facebook/folly@d58181e677965b609be89b49ed7fbc4e78bf70b5


paulbiss:
@samveen see my comment on your other PR. This is going to break the build.

PR #5368, opened 2015-05-18
Folly/Singleton.h is no longer experimental

samveen:
Singleton.h has been pulled out of experimental and moved into the main folly tree as per facebook/folly@d58181e677965b609be89b49ed7fbc4e78bf70b5 (same as #5366, for HHVM-3.6)

### DIFF
--- a/hphp/hhvm/process-init.cpp
+++ b/hphp/hhvm/process-init.cpp
@@ -32,7 +32,7 @@
 #include "hphp/system/systemlib.h"
 #include "hphp/util/logger.h"
 
-#include <folly/experimental/Singleton.h>
+#include <folly/Singleton.h>
 
 #include <libgen.h> // For dirname(3).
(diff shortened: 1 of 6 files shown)


paulbiss:
@samveen I'm not sure I understand why we need this. The 3.7 branch is still pegged to a version of folly that has singleton in the experimental directory, this would definitely break the 3.7 build.

samveen:
I think this was caused by a screwup in my `git archive` while trying to build an RPM:

1. `git pull --recursive`
2. `git checkout HHVM-3.7`
3. `git archive` and then recursively add the submodules into the same archive. 

git archive managed to archive the `HHVM-3.7` branch, but took the `master` for the submodules.

SiebelsTim:
I think `git pull --recursive` is not ideal.

Try `git submodule update --init --recursive` next time :)

PR #5386, opened 2015-05-22
Fix Multipart/form-data remote dos Vulnerability

huzhiguang:
Bug analyze refer:
   https://bugs.php.net/bug.php?id=69364&edit=3

PHP fix way also using smart_str_appends buf_value is 128 byte allocation problems(O(M*N)/128),HHVM fix way buf_value using std::string,std::string malloc len (const size_type __len = __old_size + max(__old_size, (size_t) __n) + 1;) when std::string full size.

So hhvm fix way is safe.

### DIFF
--- a/hphp/runtime/server/upload.cpp
+++ b/hphp/runtime/server/upload.cpp
@@ -424,8 +424,6 @@ static int find_boundary(multipart_buffer *self, char *boundary) {
 static int multipart_buffer_headers(multipart_buffer *self,
                                     header_list &header) {
   char *line;
-  std::string key;
-  std::string buf_value;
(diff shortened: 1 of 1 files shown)


reeze:
:+1:

huzhiguang:
@fredemmott ,I can delete this PR,but security fix send  to where?

jwatzman:
#hhvm on Freenode IRC is a good place to start. In any event, I'm pulling your patch into the proper process now, so you don't need to do anything else, thanks!

fredemmott:
https://www.facebook.com/whitehat is the correct place

PR #5397, opened 2015-05-25
Enhance checking for -latomic

yan12125:
This is an enhancement of #5218. With GCC 5-20150519 and CXXFLAGS="-O2", the original line ```s.is_lock_free()``` is optimized out, so the checking result is incorrect. I managed to fix it by printing out the value.


JoelMarcey:
We haven't heard anything re: this pull request in two weeks. Per our human timeout policy (https://github.com/facebook/hhvm/wiki/Human-Timeouts), I am closing this out. If you have further changes as requested in the diff, please feel free to re-open.

Issue #4993, opened 2015-03-10
HHVM 3.6.0 ini_get not working


sandeepone:
I'm unable to get ```ini_get('post_max_size')``` return empty string, which worked in 3.5.0.

~~~
echo ini_get('post_max_size');  // output ""
echo ini_get('max_post_size');  // output ""

echo ini_get('hhvm.server.max_post_size');  //output 104857600
~~~

Note: i'm not sure which one to use in HHVM so tried both, in Zend php it is 'post_max_size'
Thank you

jeroenvermeulen-eu:
This bug is probably what breaks image management in te backend of Magento on HHVM 3.6.0.
This is a serious problem.

fcorriga:
even with ini_set('post_max_size', '25000');
ini_get('post_max_size') returns empty string

fcorriga:
the problem occurs with upload_max_size too

sandeepone:
I don't think we can set the post_max_size with ini_set, IMO it can only be set from ini file. Yes same issue with upload_max_size. A qucik ugly fix is use this ```ini_get('hhvm.server.max_post_size')```

fcorriga:
ini_get('hhvm.server.max_post_size') returs empty string too

fcorriga:
well, actually it returns a boolean value (false)

fredemmott:
cc @JoelMarcey

rebeld:
FWIW, I'm only seeing this behaviour when vending through a webserver (nginx via fastcgi in my case). If I call ini_get through hhvn via the command line, it returns the values I would expect for 'upload_max_filesize' and 'post_max_size'.

fcorriga:
exactly.

BradStephens:
For post_max_size, use ini_get('hhvm.server.max_post_size').
For upload_max_filesize, use ini_get('hhvm.server.upload.upload_max_file_size').

Here's code for Magento users. app/code/core/Mage/Adminhtml/Block/Media/Uploader.php

```
public function getPostMaxSize()
{
    $post_max_size = ini_get('post_max_size');
    return $post_max_size ? $post_max_size : ini_get('hhvm.server.max_post_size');
}

public function getUploadMaxSize()
{
    $upload_max_filesize = ini_get('upload_max_filesize');
    return $upload_max_filesize ? $upload_max_filesize : ini_get('hhvm.server.upload.upload_max_file_size');
}
```

fcorriga:
Does not work for me.

[...]

PR #5400, opened 2015-05-26
Expose per directory ini settings in each request

rrh:
This refactor makes the 5 PHP_INI_PERDIR ini settings visible to
each request. This commit appears to fix the bug reported in
https://github.com/facebook/hhvm/issues/4993
and first introduced in:
https://github.com/facebook/hhvm/commit/4bfeda9418331e6fea0de66a106b9ba5d04728a9

There are now tests that run in CLI mode and in FCGI mode that expose
the values of the 5 PHP_INI_PERDIR settings. These new tests failed
without this commit, and now pass with the commit.

The strategy used here was similar to that taken for the SerializationSizeLimit
https://github.com/facebook/hhvm/commit/a0080d5f7e0949daee275024e91b5d49de5a39c5
We take working values come preferentially from the thread/request local data if it has been set, and then from the default values scraped from HDF. A variety of strategies are used here to determine if the request data has been set or not; most solutions rely in there being an easily determined default value.

This commit uses a variety of strategies for implementing the necessary thread/request data. In some cases static __thread is used; in others the RequestInjectionData has to be used for non-plain-old-data.

New tests have been added that use the -v (HDF) and -d (ini) initializations. In addition, as a one off developer test, the binding type of PHP_INI_PERDIR was temporarily changed to PHP_INI_USER and the expected behavior (such as allowing ini_set to have side effects) was observed.

This commit exposes some nits involving the reporting of MiB vs Bytes in what seems to be identical context. In addition the corresponding HDF weirdness of using MiB vs PHP using B is documented.


lhorace:
@rrh Does your repo contain that patch for #4993 ? Okay for testing?

JoelMarcey:
Since this is not going to be worked on any longer externally, I am going to close this right now and create an internal task to come up with a way to support per-dir, most likely taking some aspect of what was done here into the architecture.

Issue #4071, opened 2014-10-27
Open-source idx implementation or remove from hhi


fredemmott:
eg https://github.com/facebook/hhvm/commit/1c0fbc72ca6e3aac1bc4600a2df31e80849568e9

That change is only safe because idx() happens to be unused now in that subtree - otherwise, that change would have broken the test for anyone other than Facebook.

Should also check for other FBOnly functions

cc @jano, @jwatzman

jwatzman:
cf. https://reviews.facebook.net/D38985

jwatzman:
I have a diff up internally to do this, waiting on some cleanup in FB's frontend to percolate through.

PR #5407, opened 2015-05-27
Remove idx hhi

simonwelsh:
`idx` isn't open-sourced, so providing a hhi version of it prevents it being implemented in Hack projects.

Alternatively, open-source `idx` :)


jwatzman:
As on the diff, not going to take this, in favor of just open-sourcing `idx`: https://github.com/facebook/hhvm/issues/4071

PR #5413, opened 2015-05-28
Make building certain extensions with external dependencies optional

Orvid:
This makes it so that the curl, icu, ldap, libxml, mcrypt, memcache, memcached, mysql, simplxml, xml, xmlreader, xmlwriter, and xsl extensions will only build if their dependencies were found, or if they were explicitly enabled. This means there are now 8 less external dependencies required to build a minimal version of hhvm.


Orvid:
Superseded by PR #5571.

PR #5433, opened 2015-06-01
Start on Windows support for hh_client and family

Orvid:
This also adds support for compiling the .c files as C++, due to MSVC's C mode being terrible.
This also implements the filesystem watcher, which is mostly adapted from the OSX implementation.
These ports are targetted towards MSVC 2013.

This doesn't touch the makefile, and I've left out the support for the shared heap, due to my implementation of it known to be broken.
This does not make it possible to build on windows, as I haven't touched the build system in this commit. OCamlbuild on Windows does not play well with the C files (it's possible, but not practical).

As the changes for Windows support will end up being massive, a lot of smaller PRs will be much easier to review than a single massive PR would be.

### DIFF
--- a/hphp/hack/src/fsevents_win/fsevents.ml
+++ b/hphp/hack/src/fsevents_win/fsevents.ml
@@ -1,0 +1,21 @@
+(**
+ * Copyright (c) 2014, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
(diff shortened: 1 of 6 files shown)


steelbrain:
:heart:

Orvid:
I'm closing this for now. Once #5775 is in, there are plans to use some of this as the base for future work, but it will not be used entirely as-is, so I'll let whoever ends up doing that deal with extracting what they need from this and submitting it separately.

Issue #5440, opened 2015-06-02
HACK strict mode natively doesn't allow to use password_hash and password_verify


Arilas:
There's no hhi for password_hash and password_verify, so if you want to use this functions in HACK with strict mode, you need to map it manually.

paulbiss:
You can add an hhi file to your project as an interim fix, we'd also be happy to take a pull request adding these functions to Hack's builtin hhis.

PR #5446, opened 2015-06-03
Add HHI for Password

steelbrain:
Fixes facebook/hhvm#5440


Arilas:
Pull Request doesn't contain two methods also related to Passwords:
password_needs_rehash
password_get_info

paulbiss:
@Arilas you should comment on the diff (see link above).

steelbrain:
@Arilas added them in #5452

PR #5455, opened 2015-06-06
Fix compile fail on class_map.cpp when ENABLE_ASYNC_MYSQL is Off

robocoder:


fredemmott:
@robocoder : why do you need this option? I want to remove it entirely.

robocoder:
Verified that it 3.8.0-dev builds/runs with ENABLE_ASYNC_MYSQL=on.

PR #5467, opened 2015-06-10
Enable option to perform gold linker identical code folding

sathvikl:
adding gold linker option --icf=safe option reduces binary size from 77 MB to 66 MB.
use --print-icf-sections to list folded identical code sections on stderr

from the gold man page:
--icf= [safe,none,all] Folds  ctors,  dtors  and functions whose pointers are definitely not taken


sathvikl:
@JoelMarcey  I updated this diff but I think I messed up when I did the push. I don't see the same diff in fabricator. Can you please help.

JoelMarcey:
@sathvikl I see it now in phabricator.

PR #5483, opened 2015-06-12
Delete configure_generic.sh

simonwelsh:
Hasn't worked since the removal of the libevent patch.

Still targets Ubuntu 12.04 (installs custom GCC) too.


ng0ctrinh:
Sent from Yahoo Mail on Android

Issue #5488, opened 2015-06-14
Typecheck not seeing file updates on 3.8 nightly with ubuntu 4.04


akuseru:
```
$ hhvm --version
HipHop VM 3.8.0-dev (rel)
Compiler: heads/master-0-g60b7aa4dc7f452e630eec9071237060442ea8625
Repo schema: af30432416a745491098cd2a9825533fb420ab33

$  lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 14.04.1 LTS
Release:	14.04
Codename:	trusty

```

I am not  sure if this is related to the typecheck or not it's just what's yelling at me in the logs. I started seeing this when I updated to nightly to fix an issue I was having with pgsql & 3.7

after the update I started seeing this in my log file:
`Hack type error: Parse error: declarations are not supported outside global scope at /var/www/private/functions.php line 4`  (side note: I am 99% sure that these functions are in the global scope. unless a require moves us out of the global scope)

attempting to debug this issue turned out to be challenging as nothing was working. so I just emptied the file:

```php
<?hh

//nothing

```

and I was still seeing the same `Hack type error: Parse error: declarations are not supported outside global scope at /var/www/private/functions.php line 4` 

Here is what I tried:
restarting the server: no change.
touching the file to update timestamp: no change.
stoping the server, checking processes & starting the server: no change.
`hhvm.eval.jit = 0` `eval.jit=0` in both server.ini and php.ini files. no change.

remove .hhconfig file from root directory & restart.: you do not appear to be running the hack typechecker message. ok try the option there.

`Hack.Lang.LookForTypechecker=0` in both server.ini and php.ini files & restart: you do not appear to be running the hack typecherk message. ok, so this option doesn't do anything?

put .hhconfig back in the root directory & restart: ok, progress error is now somewhere else.

Thank you for your help!

akuseru:
Creating a phpinfo file has given me some more information about the configuration options. correctly setting bot jit and typechecker have no affect on this issue.

akuseru:
this appears to have been an issue with virtualbox and timestamps.

JoelMarcey:
Looks like this may not be related given what you found out, but the settings should be:

```
hhvm.jit = 0
```

and

```
hhvm.hack.lang.look_for_typechecker = 0
```

You were forming the setting wrong.

akuseru:
[...]

PR #5490, opened 2015-06-15
look_for_typechecker in error log messages should recommend the proper config key

akuseru:
look_for_typechecker in error log messages should recommend the proper configuration key value even if setting this value is not recommended.

Slightly Related github issue: #5488


JoelMarcey:
@akuseru Hi! Did you see my comment on your pull request? How do you want to proceed?

Thanks!

fredemmott:
See comments on phabricator

PR #5502, opened 2015-06-18
Incorrect return value from eval, Closure generated in first eval pass is returned in the second eval pass

ebernhardson:
This is the smallest reproduction case i could put together. Fails on hhvm 3.7.2, 3.6.1 and possibly others.  The 3.7.2 package is the debian package distributed by facebook for ubuntu trusty.

http://3v4l.org/3qEET

    function reproduce( $code ) {
            echo $code, "\n\n";
            $template = eval( $code );
            return $template( array() );
    }

    echo reproduce('return function ($in, $debugopt = 1) {
        return (function() {return "first";})();
    };');
    echo "\n-----\n\n";
    echo reproduce('return function ($in, $debugopt = 1) {
        return (function() {return "second";})();
    };');

Expected Output:
----
    return function ($in, $debugopt = 1) {
        return (function() {return 'first';})();
    };

    first
    -----

    return function ($in, $debugopt = 1) {
        return (function() {return 'second';})();
    };

    second

Actual output
----
    return function ($in, $debugopt = 1) {
        return (function() {return 'first';})();
    };

    first
    -----

    return function ($in, $debugopt = 1) {
        return (function() {return 'second';})();
    };

    first

### DIFF
--- a/hphp/runtime/vm/bytecode.cpp
+++ b/hphp/runtime/vm/bytecode.cpp
@@ -6762,9 +6762,10 @@ OPTBLD_INLINE void iopEval(IOP_ARGS) {
   auto vm = &*g_context;
   string_printf(
     evalFilename,
-    "%s(%d" EVAL_FILENAME_SUFFIX,
+    "%s(%d)(%s" EVAL_FILENAME_SUFFIX,
     vm->getContainingFileName()->data(),
-    vm->getLine()
+    vm->getLine(),
+    string_md5(code.data(), code.size()).c_str()
   );
   Unit* unit = vm->compileEvalString(prefixedCode.get(), evalFilename.c_str());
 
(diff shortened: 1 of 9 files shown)


ebernhardson:
I pushed a new version of the patch that also puts the above test case into hphp/test/slow/closure/eval_closure.php, but it doesn't seem to have made it into phabricator.

jwatzman:
cc current oncall @JoelMarcey on Phabricator issues, I think he was looking into them earlier today

JoelMarcey:
Yeah, our phabricator syncing is a known problem and I think it is being looked at.

JoelMarcey:
@ebernhardson I imported your diff in this evening. And we got reviewers on it.

Issue #5471, opened 2015-06-10
strtr( $str, $replace_pairs ) performs poorly with certain inputs


atdt:
`strtr()` invoked with an array `$replace_pairs` as the second argument appears to perform much worse under HHVM than PHP5.

Test case: `strtr( $text, $zh2Hant )`, where `$text` is the contents of http://www.gutenberg.org/files/23835/23835-0.txt and $zh2Hant is as defined in https://github.com/wikimedia/mediawiki/blob/master/includes/ZhConversion.php

Working script: https://github.com/atdt/hhvm-strtr

HHVM 3.6.1: 2279.07ms
PHP5: 83.26ms

atdt:
PHP's implementation of multiple-string search was formerly quite naive, which was the subject of [PHP bug](https://bugs.php.net/bug.php?id=63893). In 2013, it was [rewritten](http://git.php.net/?p=php-src.git;a=commitdiff;h=ccf15cf2dc92d11f92ee30c97e2d86b07f81e030) to utilize a clever and efficient algorithm for multi-pattern searching ([Wu-Manber](http://webglimpse.net/pubs/TR94-17.pdf)), and then [further optimized](http://git.php.net/?p=php-src.git;a=commitdiff;h=cddbb98ada6bdba1596ea82386401edf9b680d47). It looks like HHVM's implementation is based on the naive implementation PHP sported before PHP 5.4, which explains the poor performance.

joostshao:
it is amazing!  

HHVM (-vEval.Jit=true -vEval.JitProfileInterpRequests=0):3080.78ms
PHP5:93.76ms

SiebelsTim:
Closed in #5517

PR #5517, opened 2015-06-20
Port strtr from zend 5.6.10

ebernhardson:
Addresses https://github.com/facebook/hhvm/issues/5471

strtr() invoked with an array $replace_pairs as the second argument appears to
perform much worse under HHVM than PHP5.

PHP's implementation of multiple-string search was formerly quite naïve, and it
was the subject of PHP bug 63893. In 2013, it was rewritten to utilize a clever
and efficient algorithm for multi-pattern searching (Wu-Manber), and then
further optimized. It looks like HHVM's implementation is based on the naive
implementation PHP sported before PHP 5.4, which explains the poor performance.

This patch ports over the strtr implementation in php 5.6.10.

### DIFF
--- a/hphp/runtime/ext/string/ext_string.cpp
+++ b/hphp/runtime/ext/string/ext_string.cpp
@@ -17,6 +17,7 @@
 
 #include "hphp/runtime/ext/string/ext_string.h"
 #include "hphp/util/bstring.h"
+#include "hphp/runtime/ext/hash/hash_murmur.h"
 #include "hphp/runtime/base/comparisons.h"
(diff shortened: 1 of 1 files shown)


jwatzman:
@ebernhardson can you have a look at https://github.com/facebook/hhvm/issues/6384 please? It seems likely this change is the cause of that crash. I'm going to spend some time later to confirm, but if it is, we'll have to revert it if you don't send a PR soonish to fix it.

jwatzman:
#6395 is a proposed fix.

Issue #5559, opened 2015-06-29
Catchable fatal error: Hack type error: Unbound name: Redis (an object type)


milesj:
Updated to the latest HHVM nightly and am now receiving this error when running tests or running individual Hack files. The full error is:

Catchable fatal error: Hack type error: Unbound name: Redis (an object type) at /vagrant/src/Titon/Cache/Storage/RedisStorage.hh line 33

And the file in question:

https://github.com/titon/framework/blob/master/src/Titon/Cache/Storage/RedisStorage.hh#L33

Has something changed recently in regards to Redis? No longer supported?

I tried creating a really basic file, and ran it like so `hhvm test.hh`, yet the error still occurs.

```
<?hh
echo 'foo';
```

milesj:
Another weird thing. I removed that file completely and re-started HHVM. I now get this error:

Catchable fatal error: Hack type error: Could not find static method composeLocale in type Locale at /vagrant/src/Titon/Intl/Locale.hh line 239

All of these errors are missing information from HHI/IDL files. What changed recently in HHVM?

fredemmott:
We changed hhvm to automatically invoke the typechecker.

milesj:
So what can I do to mitigate this, just swallow fatal errors?

fredemmott:
Options are:
 - make your code valid hack
 - persuade your users to add 'hhvm.hack.lang.auto_typecheck=0' to their ini files (we will strongly recommend against this though)

Swallowing fatal errors won't work as the typechecker is invoked pre-execution.

milesj:
Redis and the Locale stuff is definitely valid, as seen here:

https://github.com/facebook/hhvm/blob/4498e2b25bc17ace502b4c86c551dbab25f284bf/hphp/runtime/ext/icu/ext_icu_locale.php#L66

The problem looks like the incomplete definitions in HHI:

https://github.com/facebook/hhvm/blob/4498e2b25bc17ace502b4c86c551dbab25f284bf/hphp/hack/hhi/stdlib/builtins_intl.hhi#L236

Doesn't look like something I can just "fix".

paulbiss:
You can send us a PR to add the definitions, in the meantime you have several options, any one of these should work, I'd recommend the first, the second is not unreasonable.
 - add the HHI file to your project until the changes are merged
 - add HH_FIXME's on the lines immediately before the functions are called
 - don't call the functions in // strict files

milesj:
[...]

PR #5570, opened 2015-07-02
New HHI definitions

milesj:
Added HHI definitions for Redis, YAML, Gettext, Intl, and FileInfo. This fixes https://github.com/facebook/hhvm/issues/5559 and related warnings I was getting.

I wrote this generator script which uses reflection to generate the HHI definitions. It works for the most part, but simply required a few edits here and there (adding generics and what not). https://gist.github.com/milesj/1097f6f8a4f7801710e1


milesj:
Just realized I forgot all the SPL stuff. Should I add it to this pull request or do another one?

paulbiss:
It would probably be easier to review as a followup rather than adding to this PR.

milesj:
No problem. I'll submit another pull once this one goes through.

PR #5571, opened 2015-07-02
Made most of the extensions optional.

Orvid:
This very likely breaks FB's internal build due to the relocation of the .idl.json files for most of the extensions. Additional defines for the enabled extensions also need to be added.
This adds configs for all but the gd and imap extensions, allowing most of them (the ones not explicitly marked as REQUIRED) to be disabled via the ENABLE_EXTENSION_* variables.
This also adds configs for the extensions that were originally in the root directory.

This supersedes PR #5413. 

It is possible that I've missed some in this list, but the additional defines needed by a non-cmake build system in order to compile correctly are:
```
ENABLE_EXTENSION_ASIO
ENABLE_EXTENSION_CLOSURE
ENABLE_EXTENSION_COLLECTIONS
ENABLE_EXTENSION_GENERATOR
ENABLE_EXTENSION_FB
ENABLE_EXTENSION_PDO_MYSQL
ENABLE_EXTENSION_PDO_SQLITE
ENABLE_EXTENSION_SIMPLEXML
ENABLE_EXTENSION_XDEBUG
```

And let's see if this is enough for Phabricator to play nice with the description now...


hphp/runtime/ext/collections/ext_collections.h
```diff
@@ -1,9 +1,53 @@
-#ifndef incl_HPHP_EXT_COLLECTIONS_H
-#define incl_HPHP_EXT_COLLECTIONS_H
+/*
```
sgolemon: hphp/runtime/ext/ext_collections.{h,cpp} and hphp/runtime/ext/collections/ext_collections.{h,cpp} are different files on purpose.  Please don't combine them as part of this diff.

If you want to move any file from runtime/ext/ to runtime/ext/FOO/ and there's already a version of that file there, then please rename it to ext_FOO-idl.{h,cpp} otherwise you're going to make IDL->HNI conversions really ugly.

sgolemon:
I'd also suggest making the movement of extensions into their subfolders as separate pull requests (or maybe just I or @paulbiss or @fredemmott could do it directly).  So that the optionality becomes JUST about the cmake infrastructure necessary to make it happen.

sgolemon:
Move(s) up for review (this will be a building list)

Collections: D2214289

Orvid:
I've removed the changes for relocating the extensions out of the root directory, but have left the IDL moves. I've also accounted for closure and generator not being moved before this PR is merged. Hopefully the changes to closure and generator don't need to modify the IDL files, if they do, I can remove the moves of those as well.

This now assumes that the moves have already been performed, except for closure and generator.
This should now rebase on those commits without issue, excluding resolving likely conflicts in `hphp/runtime/ext/ext.h`, due to the current version including the relocated paths.

sgolemon:
The changes to closure and generator will remove their IDL files entirely. :)

hphp/runtime/ext/ext.h
```diff
@@ -26,21 +26,49 @@
 #include "hphp/facebook/extensions/phpmc/ext_phpmc.h"
 #endif
 
+#ifdef ENABLE_EXTENSION_ARRAY
 #include "hphp/runtime/ext/array/ext_array_idl.h"
```
sgolemon: Actually, it looks like a lot of these includes need to be deleted from ext.h

Only extensions which are implemented with IDL need to be included from here.
I'll put up a diff to strip this file down as well.

sgolemon:
So all the diffs mentioned above have landed internally at this point, and as you can see, some have made it out already (the rest will probably be in the next half hour).

We can sort out the stripping down of ext.h by tomorrow morning, then get this pulled in along with some changes to our internal build system to account for the IDL moves (we don't use cmake).

From there,getting this merged in should be pretty simple.  I love what I've reviewed of it so far and eagerly look forward to making it official.

Thanks a ton!

sgolemon:
ext.h stripping up for review as D2216442

You can pre-apply it using this:
```
diff --git a/hphp/runtime/ext/ext.h b/hphp/runtime/ext/ext.h
index 1103c5e..01c968d 100644
--- a/hphp/runtime/ext/ext.h
+++ b/hphp/runtime/ext/ext.h
@@ -26,19 +26,11 @@
 #include "hphp/facebook/extensions/phpmc/ext_phpmc.h"
 #endif
 
-#include "hphp/runtime/ext/array/ext_array_idl.h"
-#include "hphp/runtime/ext/async_mysql/ext_async_mysql.h"
 #include "hphp/runtime/ext/ext_closure.h"
 #include "hphp/runtime/ext/collections/ext_collections-idl.h"
 #include "hphp/runtime/ext/ext_generator.h"
 #include "hphp/runtime/ext/fb/ext_fb.h"
-#include "hphp/runtime/ext/std/ext_std_file.h"
-#include "hphp/runtime/ext/hash/ext_hash.h"
 #include "hphp/runtime/ext/simplexml/ext_simplexml.h"
-#include "hphp/runtime/ext/soap/ext_soap.h"
-#include "hphp/runtime/ext/spl/ext_spl.h"
-#include "hphp/runtime/ext/string/ext_string.h"
-#include "hphp/runtime/ext/mysql/ext_mysql.h"
 
 // half-IDL, half-HNI
 #include "hphp/runtime/ext/asio/ext_asio.h"
```

Orvid:
Rebased on top of the ext.h changes, so it should now be ready.

sgolemon:
btw, just checking in to mention that it's up for review as D2216843 (has been since the weekend, actually), and I believe I've got all the little issues sorted out (a number of filename typos, missing source files, etc...) nothing major.  Hope to be able to land tomorrow.

hphp/runtime/ext/fb/config.cmake
```diff
@@ -0,0 +1,19 @@
...
+    ext_fb.h
+    FBSerialize.h
+    FBSerialize-inl.h
+    VariantController.h
+  EXTENSION_LIBRARY
+    ext_fb.php
+  IDL
+    fb.idl.json
+  DEPENDS_UPON
+    ext_mysql
```
Orvid: As PR #5573 has been merged before this made it in, this dependency should be removed.

Orvid:
Alright, so this is working for the normal OSS build, right? I just dug into the issue I was having, and figured out that it was not caused by MSVC not parsing something right, but rather, because the only target that was getting the `ENABLE_EXTENSION_*` defines was `hphp_runtime_ext`. I suspect that the defines might be getting scoped to just that target for some reason. I'm going to move the configuration of the extensions up to the root directory, and split that off from returning the set of files to compile.

Orvid:
Alright, I've added a commit that relocates where the `ENABLE_EXTENSION_*` defines are added, causing them to be added to all targets, not just `hphp_runtime_ext`. With these changes, things build correctly again.

sgolemon:
FYI, I've altered this PR a lot internally to deal with that and some more stuff including some quantitative changes to your design.  Watch for it to land in the next week and if we need to alter it from there we can work on that progress.

hphp/runtime/ext/mbstring/config.cmake
```diff
@@ -0,0 +1,17 @@
...
+  SOURCES
+    ext_mbstring.cpp
+  HEADERS
+    ext_mbstring.h
+    php_unicode.h
+    unicode_data.h
+  EXTENSION_LIBRARY
+    ext_mbstring.php
+  DEPENDS_UPON
+    ext_process
```
Orvid: Once #5681 is merged, this dependency can be removed.

hphp/runtime/ext/mail/config.cmake
```diff
@@ -0,0 +1,6 @@
+HHVM_DEFINE_EXTENSION("mail"
+  SOURCES
+    ext_mail.cpp
+  EXTENSION_LIBRARY
+    ext_mail.php
+)
```
Orvid: I was going to say that the `ext_process` dependency here can be removed after #5682 is merged, but it appears I had eliminated this dependency before I typed up the configs in the first place.

sgolemon:
Very nearly mergable... Hang tight!

Orvid:
Depending on the specifics of the modifications, I may have a couple of adjustments to make to the dependency resolution's handling of circular dependencies. Currently it's possible for the process to think an extension is enabled when it's really not. If it's still present in the modified version, I'll type up a full explanation with the PR to fix it.

sgolemon:
We sorta took out ext_* dependencies since these are handled in runtime via Extension::getDependencies() and while it can help raise errors earlier, it's duplication of effort and prone to code-rot.  So the circular resolution is a non issue. :)

Orvid:
The ext_* dependencies are the primary reason I wrote this; it makes it a lot easier to disable something like the posix or process extension when compiling on windows. Without that, all the library and OS dependencies have to be manually propogated to the other extensions. Take PDO for instance, the sqlite and mysql adapters are independently enableable, but they will require

Orvid:
Gah, who failed to put an edit comment button in the mobile UI...

I do agree though that they are subject to bit rot... Hrmm... If we assume that all extensions will be in the `runtime/ext` directory, and that the directory name will always be the name of the extension, I could set up an auto-detection mechanism of sorts... It would require the pdo adapters to be separated though.

Orvid:
Also, only one extension actually uses the getDependencies method when I last looked.

sgolemon:
Well, there's a version of this in now.  We can tweak this a bit and maybe re-add compile-time support.  Id just like to find a way to not duplicate that too much. :/

Orvid:
Just realized that the way that the default to EXPLICIT was done goes completely against what the explicit state was for... The EXPLICIT state was intended for extensions that the user has forcibly enabled, not ones which we believe they probably want. Might I ask the reasoning behind switching it to explicit by default rather than implicit by default, as it originally was?

jwatzman:
(Sorry, I ran out of time today, will get a response to this tomorrow morning.)

jwatzman:
I felt pretty strongly that configurations and builds of HHVM should be deterministic. Having the build system magically turn things on/off depending on what it happens to be able to find is, IMO, confusing and hard to debug -- especially when we get a bug report from someone who built HHVM locally and didn't realize that they didn't build half the extensions they cared about and weren't told by cmake. Or, worse, when someone uses an unofficial package from a builder who didn't notice that the packages they were distributing were missing key extensions. There's lots of very, well, implicit things that come from being implicit by default, and the only upside is a little less typing, the very first time you configure. The only downside of explicit-by-default is that you go and explicitly turn a few things off -- but that's very clear what is going on, as opposed to a build system doing complicated and hard-to-debug things.

We made affordances for a couple of really weird things that folks both might not have installed and probably don't care about, but for the most part, I really want to stick to explicit enabling/disabling.

Orvid:
Alright, then I'll look into properly implementing a 'WANT' state, so that `var*` and `os*` dependencies can be made to play nice. Currently you have to explicitly disable them, even if, for instance, you've set `ENABLE_ASYNC_MYSQL` to `OFF`, you still have to explicitly disable the async mysql extension at the moment.

jwatzman:
Can we just kill the `ENABLE_ASYNC_MYSQL`, `ENABLE_MCROUTER`, etc options, and use `ENABLE_EXTENSION_MCROUTER` etc instead?

Orvid:
Yes, I could certainly do it that way. I'd still have to implement the `WANT` state to deal with `os*` dependencies, but killing those options would solve a couple of the issues I anticipated needing to put in sanity warnings in place for.

Issue #5487, opened 2015-06-14
/hphp/hhbbc/main.cpp is included in all CMake builds


Orvid:
It appears that this is intended as an entry point for something or another, but cmake builds hhbbc as a library, so the entry point shouldn't be included in the build.

Orvid:
As pointed out on the PR, the main file is uses by hhvm itself, so doing this is a terrible idea.

PR #5590, opened 2015-07-04
Fixes issue #5487 /hphp/hhbbc/main.cpp is included in all CMake builds

Orvid:
Fixes issue #5487 /hphp/hhbbc/main.cpp is included in all CMake builds

This will conflict with PR #5581. Whichever one is merged second will need to be updated to reflect the other.

### DIFF
--- a/hphp/hhbbc/CMakeLists.txt
+++ b/hphp/hhbbc/CMakeLists.txt
@@ -8,7 +8,7 @@ HHVM_PUBLIC_HEADERS(hhbc ${files})
 
 # remove anything in a test folder
 foreach (file ${CXX_SOURCES})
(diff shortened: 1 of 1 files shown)


wjywbs:
This file is still being used by hhvm/main.cpp. The main function in this file is in the HPHP::HHBBC namespace, so it does not cause conflict.

https://github.com/facebook/hhvm/blob/master/hphp/hhvm/main.cpp#L42
https://github.com/facebook/hhvm/blob/master/hphp/hhvm/main.cpp#L46

Orvid:
Alright, in that case, this is a terrible idea, and I'll close the issue as well.

PR #5596, opened 2015-07-05
Improve make-lexer.sh to not emit local paths

nikic:
* Do sed replacements even if INSTALL_DIR is not specified
* Do replacements on hphp.ll as well as lex.yy.cpp
* Make path replacement regexes global (`@g`)

Otherwise you end up with local paths in the generated lexer.

### DIFF
--- a/hphp/parser/make-lexer.sh
+++ b/hphp/parser/make-lexer.sh
@@ -24,18 +24,19 @@ if [ $? -ne 0 ] ; then
   exit 1
 fi
 
+SED=`which sed`
+if [ ! -x "$SED" ]; then
+  echo "sed not found" 1>&2
+  exit 1
+fi
+
+$SED -i \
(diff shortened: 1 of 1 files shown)


nikic:
@Orvid Sorry for the ambiguous phrasing there, the replacements only happen in `lex.yy.cpp`, but it now also replaces paths of the form `/.*hphp.ll` to just `hphp.ll` (which are present in the `#line` annotations at the bottom of the generated file).

staabm:
@nikic your answer might not have been reached orvid, because it was not sent in Phabricator.

Orvid:
It still reached me, but yes, normally the review of PRs is done on Phabricator.

nikic:
@staabm @Orvid I don't have a working Phabricator account (it doesn't like my mail and it doesn't seem possible to change it), so I can only comment via commit or here on GitHub :/

PR #5614, opened 2015-07-09
Missing SPL HHI

milesj:
Added the rest of the missing SPL HHI definitions.

Also added `invariant_callback_register()` since I use it and it would throw unbound errors.


milesj:
Looks like my previous branch somehow merged in. Going to try and rectify.

PR #5615, opened 2015-07-09
Add Windows ABI support for JIT.

wjywbs:
Windows uses a different calling convension. Arguments passing and
caller/callee saved registers are added. About 30% quick tests pass
in JIT mode after this change.

https://msdn.microsoft.com/en-us/library/zthk2dkh.aspx
https://msdn.microsoft.com/en-us/library/6t169e9c.aspx

### DIFF
--- a/hphp/runtime/vm/jit/abi-x64.h
+++ b/hphp/runtime/vm/jit/abi-x64.h
@@ -72,12 +72,21 @@ constexpr Reg64 rAsm         = reg::r10;
  * translator manages via its RegMap.
  */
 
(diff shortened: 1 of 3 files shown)


Orvid:
Sorry I'm a bit late on the draw here, but there's one more place where the Windows ABI needs to be dealt with, the callee saved registers define in the backend [here](https://github.com/facebook/hhvm/blob/master/hphp/runtime/vm/jit/back-end-x64.cpp#L91).

wjywbs:
@Orvid Thanks! I missed to add that file in this commit.

PR #5616, opened 2015-07-09
Use statically allocated array to get 32-bit address for TC on Windows.

wjywbs:
On Windows, sbrk() returns a 64-bit address. A statically allocated array
will have a 32-bit address for code cache. The drawback is that the
maximum code cache size is pre-defined.


fredemmott:
Changes planned for > 1 week ,closing per https://github.com/facebook/hhvm/wiki/Human-Timeouts

PR #5630, opened 2015-07-10
Convert all places using __thread to ATTRIBUTE_TLS

Orvid:
Because MSVC doesn't support `__thread`, and uses `__declspec(thread)` instead.

This modifies a lot of files, but each of the changes is small.

Depends: #5629


sgolemon:
I know I said make smaller diffs, but this one should have been stacked with #5629 as it doesn't work without it.  Mind merging them?

sgolemon:
P.S. - I do appreciate you taking my earlier comment to heart.  These diffs are fantastically review-sized overall. :D

Orvid:
Better yet, hows about we make it so this isn't even needed? (as I've now done)

PR #5634, opened 2015-07-10
Added a fallback for platforms without strptime

Orvid:
Aka. MSVC. PHP gets away with not using strptime if the system doesn't have it, but I couldn't do that here.

Hopefully the license of `strptime.cpp` is valid for use in HHVM.

Depends: #5633


Orvid:
Superseded by #5886 (D43905).

PR #5638, opened 2015-07-10
Pull in fnmatch, glob, and rand_r from PHP for MSVC

Orvid:
MSVC doesn't support these functions normally, so this pulls in the PHP implementations for them. They are MSVC only.


Orvid:
Superseded by #5886 (D43905).

PR #5650, opened 2015-07-10
winapi: Rename s_host to s_host_string

Orvid:
Because MSVC defines s_host in WinSock.


sgolemon:
How about we change this one to rename the variable to s_host_string or something?

Orvid:
Done.

PR #5659, opened 2015-07-11
Fix resource generation on Windows

Orvid:
I'm not entirely sure why it wasn't working before, but I believe it was caused by an issue with CMake itself. Hopefully this still works on the previous environement as well, but I know it works for me now on Windows with MSVC.
This also makes it work properly with out-of-tree builds.

### DIFF
--- a/CMake/HPHPFunctions.cmake
+++ b/CMake/HPHPFunctions.cmake
@@ -104,9 +104,9 @@ function(append_systemlib TARGET SOURCE SECTNAME)
     # for each library append the following line to embed.rc
     # $sectionname RCDATA "$source"
     add_custom_command(TARGET generate_rc
-      COMMAND echo "${SECTNAME} RCDATA \"${SOURCE}\"" >> embed.rc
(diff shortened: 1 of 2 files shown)


CMake/HPHPFunctions.cmake
```diff
@@ -104,9 +104,9 @@ function(append_systemlib TARGET SOURCE SECTNAME)
     # for each library append the following line to embed.rc
     # $sectionname RCDATA "$source"
     add_custom_command(TARGET generate_rc
-      COMMAND echo "${SECTNAME} RCDATA \"${SOURCE}\"" >> embed.rc
+      COMMAND echo "\"${SECTNAME}\" RCDATA \"${SOURCE}\"" >> embed.rc
       COMMENT "Adding ${SOURCE} as ${SECTNAME} to embed.rc"
-      VERBATIM)
+      )
```
wjywbs: The changes above break cygwin build and result in error like "invalid embed format". This may need an if-else block for cygwin and msvc.

Orvid: The changes above actually break MSVC as well, see #5799 for the updated version.

PR #5662, opened 2015-07-11
Added a set of defines for MSVC that will be used for compatibility

Orvid:
Added a set of defines for MSVC that will be used for compatibility in various places until MSVC supports them.

This commit will be the base of quite a few others, done independently for the sake of reviewability. These defines, in combination with the folly headers, account for the vast majority of the remaining differences between my port and the git HEAD of HHVM.

### DIFF
--- a/hphp/util/portability.h
+++ b/hphp/util/portability.h
@@ -156,4 +156,47 @@
 
 //////////////////////////////////////////////////////////////////////
 
+#ifdef _MSC_VER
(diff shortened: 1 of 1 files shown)


Orvid:
Closing this for now in lieu of a future PR that also includes the changes that this is needed for. Most of these will go away once cleaned up for VS 2015 RTM.

Orvid:
Replaced by #5984.

PR #5667, opened 2015-07-12
Add CMake checks for cxxabi.h, unwind.h, and strfmon

Orvid:
These will be needed by later PRs to disable pieces for windows. These are done together so the changes for FB's internal build can be done together. The uses of these will each be in separate PRs.

This requires that the following defines be added to FB's internal build system:
```
HAVE_CXXABI_H
HAVE_UNWIND_H
HAVE_STRFMON
```
And, hopefully not breaking Phabricator this time...


sgolemon:
Go ahead and save putting in these checks for the diffs which use them.  It's not a problem for me to backfill the fbmake changes at those points.

Issue #5671, opened 2015-07-12
SplFixedArray::fromArray() can take minutes to execute on even moderately-large datasets


jkoudys:
Consider the following basic example:
```
<?php
$t = microtime(true);
$sfa = SplFixedArray::fromArray(array_fill(0, 100000, 'foobar'));
echo 'Time: ' . (microtime(true) - $t) . 's', PHP_EOL;
```
Execution time on Zend: `Time: 0.013945817947388s`
Execution time on HHVM: `Time: 41.857799053192s`

So it's prohibitively expensive on HHVM. Typically when I want to work with datastructures beyond the standard php `array`, I'll move to hack and start Vectoring and Mapping it up, but a lot of my code needs to run on PHP too and the SPL datastructures are the next best thing. Since this example is a simple array of strings, and only has 100,000 elements, but still takes so long to execute, it almost completely limits me from using HHVM on any project that uses SplFixedArray.

When I break while waiting, I usually see it sitting here:
```
$sfa = SplFixedArray::fromArray(array_fill(0, 100000, 'foobar'));
^CPausing program execution, please wait...
Break at SplFixedArray::offsetSet() on line 14017 of systemlib.php
14016   public function offsetSet($index, $newval) {
14017*    $this->validateIndex($index);
14018     $this->data[$index] = $newval;
```
When I attach gdb and break, I usually see it here:
```
(gdb) where
#0  0x0000000002124cb2 in HPHP::PackedArray::Copy(HPHP::ArrayData const*) ()
#1  0x000000000210c0f3 in HPHP::PackedArray::Append(HPHP::ArrayData*, HPHP::Variant const&, bool) ()
#2  0x00000000021285e9 in HPHP::Array::append(HPHP::Variant const&) ()
#3  0x00000000008ac0ac in HPHP::ArrayUtil::Pad(HPHP::Array const&, HPHP::Variant const&, int, bool) ()
#4  0x000000000140b5ed in HPHP::f_array_pad(HPHP::Variant const&, int, HPHP::Variant const&) ()
#5  0x0000000002153d81 in void HPHP::Native::callFunc<false, false>(HPHP::Func const*, void*, HPHP::TypedValue*, HPHP::TypedValue&) ()
#6  0x0000000002153427 in HPHP::TypedValue* HPHP::Native::functionWrapper<false, false>(HPHP::ActRec*) ()
#7  0x0000000002e00a3d in ?? ()
#8  0x000000000210fef8 in enterTCHelper ()
[...]

PR #5679, opened 2015-07-12
Fix issue #5671 SplFixedArray::fromArray is slow

Orvid:
Fix issue #5671 SplFixedArray::fromArray is slow.

This sets the size to the max key value before setting the values, to avoid re-sizing the internal array for every element.


fredemmott:
request for changes on phabricator for > 1 week, closing (https://github.com/facebook/hhvm/wiki/Human-Timeouts)

Will re-open when updated.

PR #5688, opened 2015-07-13
[MSVC] Port ext_std_network to MSVC

Orvid:
Implemented based on the PHP code.


sgolemon:
I worry a bit about the code duplication.  PHP doesn't use a separate implementation for windows (IIRC).  Is there any way we can follow what PHP is doing for windows (this code is largely lifted from php-src as it is).

Orvid:
PHP does use a separate implementation for windows, and this is that implementation. [Here](https://github.com/php/php-src/blob/master/ext/standard/dns_win32.c)

PR #5689, opened 2015-07-13
Only enable money_format if strfmon is available

Orvid:
This disables `money_format` and the chain of functions it depends on if `strfmon` is not available.

This will require `HAVE_STRFMON` to be defined for the internal build.


fredemmott:
Waiting for changes, will re-open when updated. https://github.com/facebook/hhvm/wiki/Human-Timeouts

Orvid:
Updated such that `HAVE_STRFMON` is no longer needed.

PR #5690, opened 2015-07-13
Struct are structs, and classes are classes, so declare them as such

Orvid:
Because otherwise, MSVC won't be able to link correctly.

This was done via a very restricted regex replacement that ignored any names that were defined multiple times in the codebase, as well as any that had multiple parent types, or a parent with template parameters, or is an inner class defined outside of that class, or where the parent was qualified by a namespace. There were also other restrictions, but that's good enough for now.
This does a large amount of the fixes needed for successfully compiling on MSVC.


Orvid:
Requires significant revision.

PR #5693, opened 2015-07-14
[MSVC] Synchronize forward declarations with their actual struct/class declarations

Orvid:
This was done via a very restricted regex replacement that ignored any names that were defined multiple times in the codebase, as well as any that had a parent with template parameters, or is an inner class defined outside of that class. There were also other restrictions, but that's good enough for now.
This does a large amount of the fixes needed for successfully compiling on MSVC.

This shouldn't break anything, but should be the majority of the changes needed for MSVC to work correctly. I know that there are a few places in this where the wrong qualifier is used, but those shouldn't break the GCC build, and will be fixed in later MSVC PRs.


Orvid:
Moved internally as D2861227.

PR #5700, opened 2015-07-14
Add str_getcsv to string stdlib hhi

jazzdan:
According to [php.net](http://php.net/manual/en/function.str-getcsv.php) str_getcsv() takes at least one string (optionally up to four) and returns an array<int, string>.

I tested this with strings that aren't CSVs and it still returned an array<int, string>.

### DIFF
--- a/hphp/hack/hhi/stdlib/builtins_string.hhi
+++ b/hphp/hack/hhi/stdlib/builtins_string.hhi
@@ -55,7 +55,7 @@ function join($glue, $pieces = null);
 function str_split($str, $split_length = 1);
 function chunk_split($body, $chunklen = 76, $end = "\r\n");
 function strtok($str, $token = null);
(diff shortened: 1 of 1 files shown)


hphp/hack/hhi/stdlib/builtins_string.hhi
```diff
@@ -55,6 +55,7 @@ function join($glue, $pieces = null);
 function str_split($str, $split_length = 1);
 function chunk_split($body, $chunklen = 76, $end = "\r\n");
 function strtok($str, $token = null);
+function str_getcsv(string $input, string $delimiter = ",", string $enclosure = "\", string $escape = "\\"): array<int, string>;
```
sgolemon: `string $enclosure = "\", string`   <--- Unbalanced quotes

jazzdan: FIxed, thanks @sgolemon!

PR #5702, opened 2015-07-15
MRB-GC

psmarshall:

This diff almost completely removes reference counting and only uses a 1-bit count. GC is based off the 1-bit count, meaning this diff leaks objects that are incref'd.

It also changes behaviour of reference types by failing to demote them back to values when appropriate, so it will fail some tests, as well as any tests that rely on precise destruction which is not guaranteed with this diff.


sgolemon:
That's a pretty bold diff.  Have you been working with anyone on the HHVM team about this?

cc: @edwinsmith, who was talking about going with something like this once the GC is solid.

edwinsmith:
Yes, Peter is working with me. you can put me as reviewer on everything; these aren't for landing, we just need diffs so we can do some testing.

sgolemon:
Ah, awesome.  Will start the imports then.

fredemmott:

paulbiss:
@fredemmott isn't closing it going to interfere with the import flow? I thought the point here was to be able to run our full test suite against the code.

fredemmott:
It's already re-running, and tests can be re-triggered from our non-OSS internal tools.

This also can't be re-imported and re-ran without a rebase anyway.

PR #5703, opened 2015-07-15
Double-Inc

psmarshall:

This diff is exactly the same as master, except incref's are now by two instead of by one. This emulates the behaviour of a 1-bit refcount, in that objects that are ever incref'd will never return to a refcount of 1, meaning they are not freed.

It also changes behaviour of reference types by failing to demote them back to values when appropriate, so it will fail some tests, as well as any tests that rely on precise destruction which is not guaranteed with this diff.


fredemmott:

PR #5705, opened 2015-07-15
MRB-CoW

psmarshall:

This diff performs CoW checks based on the MRB (1 bit refcount). It therefore copies more arrays and strings that it strictly needs to inside of CoW checks. Every other aspect of the code still uses the regular refcount.


fredemmott:

PR #5706, opened 2015-07-15
Blind-CoW

psmarshall:

This diff will copy every array and string when performing a CoW check, regardless of any refcount or 1-bit refcount. This slows things down massively.


fredemmott:

PR #5707, opened 2015-07-15
Remove return type annotation from odbc_prepare

jazzdan:
According to [php.net](http://php.net/manual/en/function.odbc-prepare.php) `odbc_prepare()` returns false on failure:

```
Returns an ODBC result identifier if the SQL command was prepared successfully. Returns FALSE on error.
```

Remove the type annotation to reflect this.


jazzdan:
Closing pull request, see phabricator for details.

PR #5708, opened 2015-07-15
gmp_cmp returns ints

jazzdan:
According to [php.net](http://php.net/manual/en/function.gmp-cmp.php), `gmp_cmp()` returns int. This example enumerates all the possible return types:

```
$cmp1 = gmp_cmp("1234.5", "1000.3"); // greater than
$cmp2 = gmp_cmp("1000", "1234"); // less than
$cmp3 = gmp_cmp("1234", "1234"); // equal to

var_dump($cmp1); // int(1)
var_dump($cmp2); // int(0)
var_dump($cmp3); // int(-1)
```


WizKid:
It can return false: http://3v4l.org/lvD60

jazzdan:
Good catch @WizKid, looks like the type annotation on this should also be removed then.

PR #5718, opened 2015-07-16
Don't globally require MySQL

Orvid:
Now that it's possible to easily disable extensions, and the only place in the core of HHVM that relied upon MySQL is moved into the MySQL extension where it belongs, MySQL is no longer needed to compile HHVM itself. The async and normal MySQL extensions handle adding the includes and linking themselves.


Orvid:
Moved internally to D2901049.

PR #5723, opened 2015-07-16
Remove the global LibIConv requirement

Orvid:
LibIConv is currently required globally, but only `ext_iconv` and `ext_gd` use it. `ext_iconv` needed no changes, as it already uses the new extension mechanism, but `ext_gd` doesn't. Because of this, the iconv requirement was inlined in `ext_gd`.
This also implements LibIConv in the extension config mechansim.


fredemmott:
Closing as failures haven't been addressed for long time.

PR #5728, opened 2015-07-17
Remove the global check for libpam

Orvid:
The PAM headers are not used anywhere in HHVM or any of it's dependencies as far as I can tell, and no indication is even made that we have libpam in the first place.
This also removes the FindLibpam.cmake file, as it's not needed if we aren't doing this check.


wjywbs:
I think this broke mac build with homebrew libraries:

```
Undefined symbols for architecture x86_64:
  "_pam_acct_mgmt", referenced from:
      _Checkpw in libc-client.a(osdep.o)
  "_pam_authenticate", referenced from:
      _Checkpw in libc-client.a(osdep.o)
  "_pam_end", referenced from:
      _Checkpw in libc-client.a(osdep.o)
      _checkpw_cleanup in libc-client.a(osdep.o)
  "_pam_set_item", referenced from:
      _Checkpw in libc-client.a(osdep.o)
  "_pam_setcred", referenced from:
      _Checkpw in libc-client.a(osdep.o)
      _checkpw_cleanup in libc-client.a(osdep.o)
  "_pam_start", referenced from:
      _Checkpw in libc-client.a(osdep.o)
ld: symbol(s) not found for architecture x86_64
```

Orvid:
Alright, then I'll submit a PR to revert this once I finish submitting my Folly PRs (there are a lot of them; 30 so far, and I still have a ways to go), and ensure that the check is only done on OSX, because I assume it's a dependency of OSX's libc?

PR #5730, opened 2015-07-17
Adjust the tracing behavior under MSVC

Orvid:
The current code tries to define it in such a way as to allow both debug and release objects to be linked in. MSVC can't handle this, and ends up producing errors about multiply defined symbols at link time.


Orvid:
Duplicate of #5767. Not sure how I managed this, but I did.

PR #5736, opened 2015-07-17
Don't expose HHVM by default

### LINKED ISSUE #5691: Remove HHVM Version from headers by default
Currently, content served by HHVM has `X-Powered-By` header by default which not only exposes that it is being served by HHVM but also exposes its version. Of course you can disable this via `expose_php = off` directive. But why show version by default when it can be used to exploit version specific security vulnerabilities. Is there any benefit of this?

Orvid:
As-per [issue #5691](https://github.com/facebook/hhvm/issues/5691), this sets `expose_php` to `false` by default.
Fixes #5691


fredemmott:
Will comment on issue.

PR #5737, opened 2015-07-17
Remove the global FastLZ requirement

Orvid:
The only thing that uses fastlz currently is `ext_memcached`, if we already aren't building that, we don't need it.

This one is a bit more fun than the rest, because fastlz currently falls back to a bundled version if the system doesn't have it.

This also adds the ability to force the use of the bundled version via `HHVM_USE_BUNDLED_FASTLZ`. This ability will not however work until the way the 3rd party repo handles it is also adjusted.

Now, obviously I can't tell CMake to link fastlz against itself, so I use a generator expression to enable the library in the link string only if the target is not named "fastlz". This works fine, however, some versions of CMake complain that I'm still trying to make fastlz dependent upon itself. To get CMake to be quiet about this, we set policy CMP0038 to the OLD behavior, which is to simply ignore that dependency. (the NEW behavior is to error)

For this to work fully, changes to the third-party repo will also be needed (enable fastlz only if `HHVM_USE_BUNDLED_FASTLZ` is both defined and `true`, rather than checking if `FASTLZ_INCLUDE_DIR` is valid), but it still works without any changes to it.

### DIFF
--- a/CMake/HHVMExtensionConfig.cmake
+++ b/CMake/HHVMExtensionConfig.cmake
@@ -694,7 +694,9 @@ function (HHVM_EXTENSION_INTERNAL_HANDLE_LIBRARY_DEPENDENCY extensionID dependen
         # to figure that out, and complains about it. The OLD behavior here tells cmake
         # to simply not emit the dependency for itself. (which doesn't really exist to
         # begin with)
-        cmake_policy(SET CMP0038 OLD)
+        if(POLICY CMP0038)
+          cmake_policy(SET CMP0038 OLD)
+        endif()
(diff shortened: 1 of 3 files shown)


Orvid:
Actually, due to me failing to notice that the extensions were still being configured in `hphp/runtime/ext`, rather than before everything else in the version of my optional extensions PR that got merged, this will not in-fact work in an acceptable manner until that change occurs. I'll be creating a PR for that momentarily.

Orvid:
This won't work in versions of CMake < 3.0, and was too hacky to begin with.

PR #5739, opened 2015-07-18
Make extension configuration occur before anything else

Orvid:
This is needed for two reasons.

The first is that, with the way it is now, the `ENABLE_EXTENSION_*` defines are only valid in the `hphp/runtime/ext` directory, or, in the case of the IDL defines, where the `IDL_DEFINES` variable was used. This means that it's currently impossible for me to disable the `ext_process` extension, as is required on windows (php disables it on windows as well), as it requires disabling a hook used in the runtime. By moving the configuration up, it makes it a simple matter of surrounding the effected lines of code with `#ifdef ENABLE_EXTENSION_PROCESS`.

The second reason, is that, by doing this early, we have the ability to ignore the bundled versions of libraries that we don't need for the core of HHVM, [FastLZ #5737](http://github.com/facebook/hhvm/pull/5737) for instance.


Orvid:
This was reverted in baced5a. Not sure what was causing the errors described, as this shouldn't be forcing the values anywhere.

JoelMarcey:
I think it is more of an ordering problem on when things get Set and Included. For example, `INCLUDE(HHVMExtensionConfig)` caused us to have to add something somewhere else and the rabbit hole began. We had to unblock the build though. So reverting was our best option at this point.

PR #5742, opened 2015-07-19
Port the hash assembly code to MSVC

Orvid:
Well, MASM.
This ports the asm versions of the crc hash functions to MSVC.
This shouldn't break anything even if the ASM_MASM language hasn't been enabled, which is why it's in a separate PR.


Orvid:
Closed pending rewrite.

Issue #5748, opened 2015-07-19
stream_await() and related constants not in HHI


fredemmott:

SiebelsTim:
Resolved in 73dfc36

SiebelsTim:
@fredemmott You're oncall, right? Close this :P

fredemmott:
That commit didn't actually go into master and doesn't show up in a clone :/ Actually fixed by https://github.com/facebook/hhvm/commit/cbc3e821fe07b10aee2e9d8751bcbf86f507fdb4

PR #5751, opened 2015-07-19
Fix issue #5748 stream_await and related constants not in HHI

Orvid:
This adds the constants, and the stream_await function declaration.
This also adds the `STREAM_URL_STAT_LINK` and `STREAM_URL_STAT_QUIET` constants, which were also missing.


fredemmott:
Waiting for changes, will re-open when updated (https://github.com/facebook/hhvm/wiki/Human-Timeouts)

Orvid:
I'll let someone else handle this, it should be easy, I just have to many other PRs to manage.

PR #5768, opened 2015-07-22
Adjust stack-trace for MSVC

Orvid:
We don't have libiberty when under MSVC, so stub it out to just return the input string.

In reality, this should probably be refactored to use <cxxabi.h> instead.


Orvid:
No longer needed.

PR #5778, opened 2015-07-23
Add support to HPHPSetup for generation targets supporting multiple configurations

Orvid:
Previously they would be forcing either debug or release, which isn't needed when the generation target, such as MSVC or XCode, support multiple configuration types.
CMake also already adds the `DEBUG` and `NDEBUG` defines for the appropriate configurations to begin with, so we simply don't add any definitions for multi-config targets.


Orvid:
This needs some work. I was mistaken about which defines CMake adds automatically. The only define CMake adds automatically is `NDEBUG`. I'll re-open this once I've fixed it.

Orvid:
Fixed this properly. I've ignored XCode support for the moment. A later PR can be made to support it.

Issue #5776, opened 2015-07-22
hh_server parser doesn't gracefully recover from missing visibility modifier


jazzdan:
Repro case:

``` hack
<?hh
class Derp {
    function a() {
        echo 'herp';
    }

    function getType() {
        return 'derp';
    }
}
```

Typechecker output:

```
test.hh:3:5,12: Expected modifier (Parsing[1002])
/tmp/hh_server/hhi_f8220b4/stdlib/builtins_variable.hhi:28:10,16: Name already bound: gettype (Naming[2012])
  test.hh:7:14,20: Previous definition getType differs only in capitalization
  test.hh:7:14,20: This appears to be defined in an hhi file included in your project root. The hhi files for the standard library are now a part of the typechecker and must be removed from your project. Typically, you can do this by deleting the "hhi" directory you copied into your project when first starting with Hack.
```

Orvid:
Any chance someone could test if my attempt, #5783 solves this? As the Hack typechecker is still in the process of being ported to Windows, it's a bit difficult for me to test this myself.

jazzdan:
Hey @Orvid, when I try to compile with that patch I get:


```
File "parsing/parser_hack.ml", line 1263, characters 6-26:
Error: This expression has type Ast.class_elt
       but an expression was expected of type Ast.class_elt list
```

lexidor:
I am going over old issues on this repository, to see which ones apply to the current versions of hhvm.

It appears this issue has been fixed.
You repro typechecks with these errors on hhvm 4.56.
```
Naming[2086] Methods need to be marked public, private, or protected.
   --> file.hack
  7 |     function a(): void {
    |              ^

Naming[2086] Methods need to be marked public, private, or protected.
   --> file.hack
 11 |     function getType(): string {
    |              ^^^^^^^

2 errors found.
```

PR #5783, opened 2015-07-24
Attempt to fix issue #5776 hh_server parser doesn't gracefully recover from missing visibility modifier

Orvid:
This is my attempt to fix issue #5776, however, as I'm on windows, I can't test to see if it actually works. If my understanding of how the parser works is correct though, this should achieve the desired effect.


Orvid:
Due to my very limited ability to test this, I'm going to close this and leave the bug for someone else to fix.

PR #5821, opened 2015-07-29
Fix building gd extension with libvpx 1.4.0

glensc:
Summary: In libvpx 1.4.0 the compatibility layer for some of the older constants have been removed [1].
The use of this constants was deprecated since 2010 in favor of the newer VPX_ prefixed names.
The same issue as been reported in the php bugtracker [2].

Thanks!

[1] https://gerrit.chromium.org/gerrit/#/c/70164/3/vpx/vpx_image.h,unified
[2] https://bugs.php.net/patch-display.php?bug_id=69479&patch=php-gd-newvpx.diff&revision=latest
Closes https://github.com/facebook/hhvm/pull/5191

Reviewed By: @fredemmott

Differential Revision: D2005702


glensc:
this is cherry-pick #5191 into 3.6 branch

JoelMarcey:
@glensc 3.9 is the latest LTS out now, and it has this patch. We are not going to backport this change into 3.6 because it is also an LTS release, and this has a non-zero chance of breaking people. If you can use 3.9, then you are good to go.

PR #5824, opened 2015-07-30
Improve HHVM reporting performance

huzhiguang:
HHVM new version create msg varible before compare report level,this way can reduce hhvm more performance.
Found that the problem is my colleague @wwbmmm and I have fixed this problem.

### DIFF
--- a/hphp/runtime/base/runtime-error.cpp
+++ b/hphp/runtime/base/runtime-error.cpp
@@ -52,10 +52,12 @@ static inline bool checkReportLevel(ErrorMode mode) {
 }
 
 #define CHECK_REPORT_LEVEL(mode) \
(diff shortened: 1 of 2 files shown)


Orvid:
If I may suggest a minor change to the title of this PR to, "Improve HHVM reporting performance".
I'd also like to suggest a change in the first sentence of the description of this PR to, "The new version compares the report level before creating the msg variable, improving reporting performance."

JoelMarcey:
@huzhiguang Per: https://github.com/facebook/hhvm/wiki/Human-Timeouts, I am going to close this. Feel free to address the feedback on https://reviews.facebook.net/D43251 and open another request. Thanks.

Issue #5853, opened 2015-08-04
PHPUnit Phar signature not being correctly verified


simonwelsh:
I downloaded the 4.7.7 phar for PHPUnit (https://phar.phpunit.de/phpunit-4.7.7.phar) and HHVM is failing to run it. It runs fine with php 5.6.

PHP:
```
vagrant@homestead:~/projects$ php --version
PHP 5.6.7-1+deb.sury.org~utopic+1 (cli) (built: Mar 24 2015 11:23:10) 
Copyright (c) 1997-2015 The PHP Group
Zend Engine v2.6.0, Copyright (c) 1998-2015 Zend Technologies
    with Zend OPcache v7.0.4-dev, Copyright (c) 1999-2015, by Zend Technologies
    with Xdebug v2.2.5, Copyright (c) 2002-2014, by Derick Rethans
    with blackfire v0.23.0, https://blackfire.io/, by SensioLabs
vagrant@homestead:~/projects$ php ./phpunit.phar --version
PHPUnit 4.7.7 by Sebastian Bergman and contributors.
vagrant@homestead:~/projects$ 
```

HHVM 3.8.1:
```
vagrant@homestead:~/projects$ hhvm --version
HipHop VM 3.8.1 (rel)
Compiler: tags/HHVM-3.8.1-0-g3006bc45691762b5409fc3a510a43093968e9660
Repo schema: 253b3802ce1bcd19e378634342fc9c245ac76c33
vagrant@homestead:~/projects$ hhvm ./phpunit.phar --version

Fatal error: Uncaught exception 'PharException' with message 'phar has a broken signature' in :
Stack trace:
#0 (): Phar->parsePhar()
#1 (): Phar->__construct()
#2 /home/vagrant/projects/phpunit.phar(15): Phar::mapPhar()
#3 {main}
vagrant@homestead:~/projects$ 
```

HHVM nightly (2015.08.03):
```
vagrant@homestead:~/projects$ hhvm --version
HipHop VM 3.9.0-dev (rel)
Compiler: heads/master-0-g71e7b1d166877fc68b9843e5aebcf5d175005a7e
Repo schema: 198b332a71922557779226df56fa055dc7dd6130
vagrant@homestead:~/projects$ hhvm ./phpunit.phar --version

Fatal error: Uncaught exception 'PharException' with message 'phar has a broken signature' in /home/vagrant/projects/phpunit.phar:15
Stack trace:
[...]

PR #5858, opened 2015-08-04
Fix issue #5853 PHPUnit Phar signatures are broken

Orvid:
The signature flags were being used completely wrong, the signature flags aren't flags, as shown in [php-src/phar.c:779](https://github.com/php/php-src/blob/master/ext/phar/phar.c#L779).
Previously only MD5 and SHA1 signatures were being validated correctly, due to how the shifts were setup. (`MD5` is `1`, `SHA1` is `2`)
This checks the "flags" correctly, and errors correctly if an openssl signature was encountered.

Fixes issue [#5853](https://github.com/facebook/hhvm/issues/5853).


steelbrain:
Just got bit by this a few hours ago, Thanks for this :clap:

PR #5872, opened 2015-08-07
Make extension configuration occur before anything else

Orvid:
~~DO NOT MERGE! - This is currently the exact same as what was reverted, and exists to provide a proper base for some other PRs. I need more details on the issue this caused before I can attempt to fix it.~~

This has now been updated to fix the issues encountered previously.

This was opened and merged previously as #5739, however that got reverted due to other issues caused by it. 

And the description of the original PR:
This is needed for two reasons.

The first is that, with the way it is now, the `ENABLE_EXTENSION_*` defines are only valid in the `hphp/runtime/ext` directory, or, in the case of the IDL defines, where the `IDL_DEFINES` variable was used. This means that it's currently impossible for me to disable the `ext_process` extension, as is required on windows (php disables it on windows as well), as it requires disabling a hook used in the runtime. By moving the configuration up, it makes it a simple matter of surrounding the effected lines of code with `#ifdef ENABLE_EXTENSION_PROCESS`.

The second reason, is that, by doing this early, we have the ability to ignore the bundled versions of libraries that we don't need for the core of HHVM, [FastLZ #5737](http://github.com/facebook/hhvm/pull/5737) for instance.


Orvid:
Replaced internally with D2881805.

PR #5878, opened 2015-08-07
Enable interpreter mode of HHVM to PPC64 architecture

lbianc:
These patches enable HHVM interpreter mode for PPC64 machines, adding necessary stubs as not_implemented (used on JIT mode) and fixes for non-x64_64 machines.
This implementation depends on patch sent to folly, which adds PPC64 as supported architecture.


lbianc:
Referencing folly patch: https://github.com/facebook/folly/pull/298

lbianc:
Closing this pull request due conflict with newer code from HHVM. I will recreate it based on newer commit.

hphp/runtime/vm/native.cpp
```diff
@@ -75,7 +77,7 @@ static void populateArgs(const Func* func,
     if (type == KindOfDouble) {
       if (SIMD_count < kNumSIMDRegs) {
         SIMD_args[SIMD_count++] = args[-i].m_data.dbl;
-      } else if (GP_count < numGP) {
+      } else if (GP_count > numGP) {
```
Orvid: This seems odd.

PR #5883, opened 2015-08-07
Make ext_process optional

Orvid:
ext_process, or, to give it it's correct name, ext_pcntl, is not going to be supported on Windows, and even PHP itself doesn't attempt to support it on Windows. This means that it needs to be possible to disable it on Windows, however, it requires a hook in the core runtime.
This disables the hook if ext_process isn't enabled. This also removes the `REQUIRED` flag in the config for the extension.

Requires #8572.


fredemmott:
Closing - will re-open when requested changes happen.

PR #5886, opened 2015-08-08
Implement fnmatch, glob, rand_r, and strptime for MSVC

Orvid:
This supersedes #5638 (D41601) and #5634 (D41577).
This puts them all in `utils/portability`.


JoelMarcey:
I had a conversation with legal and we cannot accept this pull request since it has the 4-clause BSD license. That is not compatible with what we are able to have in our code.

The actual author of the code would have to sign our CLA and contribute the code.

JoelMarcey:
cc @jwatzman

Orvid:
Alright, I'll sre if I can manage to find an implementation we can use. Any idea if extracting the newlib implementation would be able to be used?

jwatzman:
What license is it under?

PR #5902, opened 2015-08-11
Remove some verbose logs from filesystem extension

dpaneda:
Remove some verbose logs from filesystem extension
A couple of functions are throwing some ugly errors (e.g. when the file
does not exist). This is a small incompatibility with PHP, as those functions
are not supposed to log those kind of errors, so I removed them.
I also removed some commented code which has been dead for a long time.

No test added/modified (seems that this extension is barely covered with tests).

I would like this pull request to be on behalf my employer (Tuenti). We already sign the corporate CLA for that.


fredemmott:
We close pull requests after changes are requested on Phabricator for > 1 week; will re-open when test is added.

PR #5925, opened 2015-08-13
[WIP] Port PHP7 CSPRNG functions random_int() and random_bytes()

SammyK:
Hello! This PR is a port of the [CSPRNG functions](http://php.net/csprng) `random_*()` which were [added in PHP7](https://wiki.php.net/rfc/easy_userland_csprng).

This PR is very much a work in progress and I'll need a lot of help from you smarter kids to get this implemented properly. But it works well on my machine thanks to some one-on-one help from @jmikola.

The things that I know need to be addressed:

1. The PHP version [checks for sources of random at compile time](https://github.com/php/php-src/blob/97f159d70288ae94f9a05a370121bcbea760ed9a/Zend/Zend.m4#L406-L412). Not sure how to do this in HHVM.
2. PHP [supports Windows with `arc4random_buf()`](https://github.com/php/php-src/blob/cbcacbb2dad4c97ba5653f42729f7956193d9112/ext/standard/random.c#L86). Not sure if we want to support this as well as HHVM gets more Windows friendly... is that a thing?
3. There is [an open PR](https://github.com/php/php-src/pull/1397) to change the behavior of handling the cases when a reliable source of random cannot be found. It would be better to throw an exception to halt execution since the functions are used in cryptographic contexts (see PR for details). This PR models the existing implementation but I'd rather it model the more secure PR.
4. There is [another open PR](https://github.com/php/php-src/pull/1398) that changes the behavior of the argument validation.
5. The PHP version [caches the file descriptor](https://github.com/php/php-src/blob/cbcacbb2dad4c97ba5653f42729f7956193d9112/ext/standard/random.c#L33-L50). Not sure how to do that in HHVM. (But I'm using `fstream` to read from `/dev/urandom`. Not sure if we just want to use `open()` instead.)
6. The code references `0xffffffffffffffff` since I don't know the HHVM equivalent of `ZEND_ULONG_MAX`
7. The random bytes logic is contained in the `getRandomBytes()` function but it could potentially be moved to a more global scope to serve other parts of the codebase that need crypto-secure bytes.
8. I've written this as an extension called `random` so that I could quickly iterate changes with `hphpize`. But ultimately this would be integrated into the [standard lib like the PHP version](https://github.com/php/php-src/blob/master/ext/standard/random.c). I've put a [skeleton version up for HHVM](https://github.com/php/php-src/blob/cbcacbb2dad4c97ba5653f42729f7956193d9112/ext/standard/random.c#L33-L50) just as an example.

I think that's it off the top of my head. :)

### DIFF
--- a/hphp/runtime/ext/random/config.cmake
+++ b/hphp/runtime/ext/random/config.cmake
@@ -1,0 +1,2 @@
+HHVM_EXTENSION(random ext_random.cpp)
+HHVM_SYSTEMLIB(random ext_random.php)
--- a/hphp/runtime/ext/random/ext_random.cpp
+++ b/hphp/runtime/ext/random/ext_random.cpp
@@ -74,8 +74,7 @@ Variant HHVM_FUNCTION(random_int, int64_t min, int64_t max) {
   }
 
   // Special case where no modulus is required
-  // @TODO Replace literal with HHVM equivalent of ZEND_ULONG_MAX
-	if (umax == 0xffffffffffffffff) {
+	if (umax == std::numeric_limits<uint64_t>::max()) {
     return result;
 	}
 
@@ -84,8 +83,8 @@ Variant HHVM_FUNCTION(random_int, int64_t min, int64_t max) {
 
   // Powers of two are not biased
 	if ((umax & (umax - 1)) != 0) {
-		// Ceiling under which 0xffffffffffffffff % max == 0
-		int64_t limit = 0xffffffffffffffff - (0xffffffffffffffff % umax) - 1;
+		// Ceiling under which std::numeric_limits<uint64_t>::max() % max == 0
+		int64_t limit = std::numeric_limits<uint64_t>::max() - (std::numeric_limits<uint64_t>::max() % umax) - 1;
 
 		// Discard numbers over the limit to avoid modulo bias
 		while (result > limit) {
--- a/hphp/runtime/ext/random/ext_random.cpp
+++ b/hphp/runtime/ext/random/ext_random.cpp
@@ -24,75 +24,70 @@
 namespace HPHP {
 ///////////////////////////////////////////////////////////////////////////////
 
-// @TODO Throw exceptions instead of raising warnings
 static bool getRandomBytes(void *bytes, size_t length) {
 
-  // @TODO Need better error handling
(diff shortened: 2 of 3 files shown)


Rican7:
:clap: :+1:

SammyK:
Also forgot to tag @lt to take a looksie. :)

@Rican7 :D

Orvid:
Unless there's a specific reason that we need to use a separate implementation, folly already implements a version of `getRandomBytes` [here](https://github.com/facebook/folly/blob/master/folly/Random.cpp#L98), and that should be used instead.

SammyK:
@Orvid As long as the source of random is crypto-secure, I don't think there's any real reason we'd need to use the same implementation as PHP. Unless @lt thinks otherwise. :)

fredemmott:
@SammyK : ping - see comments on phabricator

SammyK:
@fredemmott Thanks for the ping! Just subscribed to the thread in phabricator. :)

JoelMarcey:
@SammyK You still interested in this? I //think// if you read the comments on the diff, all that are needed now are lint changes and we can probably land it.

SammyK:
@JoelMarcey I'm :+1: with having you guys finishing it off. I won't be able to touch it until January. I believe we just needed:

1. Lint changes
2. Tests
3. Having it throw the proper exceptions when parse params fails (not sure if that functionality was added yet).
4. Having it throw when source of random byes fails.

I think that's it. :)

JoelMarcey:
I believe @jwatzman and @sgolemon are planning to do just that.

jwatzman:
Yeah, I'll have a version of it landing in a few hours if all goes well. 1 and 2 are done; 3 is going to likely be done in a followup diff today or tomorrow (I need to clean up other places that throw the wrong exn too). I didn't do 4 and will leave that to someone else, since it requires a folly change I have little context on; it's really suboptimal that we might abort in this case, but at least we won't return bad data.

SammyK:
Thanks guys! Sorry for dropping the ball on this one. Things got crazy with work and dance competitions. :)

jwatzman:
No problem! You did the vast majority of the work, just took me about an hour today to get it across the finish line.

The diff is in and should be on github shortly.

Rican7:
Awesome work! Great to see this collaboration here. :+1: :clap:

PR #5938, opened 2015-08-14
Transform SSwitch opcodes on single characters into Switch opcodes

Orvid:
I have no idea if there's a better place for this, but I did want it to be after all possible constant propagation, etc, as this is a transformation of the bytecode, and not in a way that would prevent or allow other optimizations.
The goal of this is to transform the `SSwitch` opcode generated for switches on a single character into `Switch` opcodes, which are significantly faster.

And yes, I know there are other things that would produce an even bigger speed improvement in the test code, especially for large strings, but that's not the goal of this PR.


Orvid:
Closing until I get around to rewriting it.

PR #5941, opened 2015-08-15
Don't generate an ENABLE_EXTENSION_* variable for required extensions

Orvid:
You can't disable required extensions to begin with, so generating these variables, which is only done to make it easy for the user to set them, is useless, as the user will just get an error if they try to disable it.

Due to where the change is, #5872 (D43749) will conflict with this if that PR is merged first. If this PR is merged before that one, there shouldn't be any conflicts when merging, although it will depend on how picky git decides to be.


sgolemon:
I'd like to keep these variables around for consistency.  I can imagine some later part in the configure process where someone tests for ENABLE_EXTENSION_STANDARD gets false and wonder wtf is going on, for example.

Orvid:
Would it be more reasonable if I were to still generate the variable, but mark them as `INTERNAL`, so that they don't show up in the CMake GUI?

sgolemon:
CMake has a GUI  o.O

Yeah, sure.  That sounds reasonable.

JoelMarcey:
What is this GUI you speak of?

Orvid:
It exists and is included in the default install of CMake on Windows, because we're special :P

PR #5942, opened 2015-08-15
Use the correct working directory when generating the build info.

Orvid:
I have no idea how this has worked up 'til now, but it doesn't work on windows.
This sets the working directory for the script to the current source directory.


sgolemon:
Looking at the build script, I'm kinda surprised that working directory matters at all (it ends up doing a chdir relative to dirname $0, but whatevs.

PR #5943, opened 2015-08-15
Use a temporary placeholder as the frame pointer under MSVC

Orvid:
There is no way to get the frame pointer from C/C++ under 64-bit MSVC, and this is only used in the JIT, so stub this out to a nullptr variable for now, so that we can compile.


sgolemon:
Since JIT will fail pretty spectacularly without this being correct, can we make this define do an `always_assert(false);` instead, with a nice big FIXME comment?

Orvid:
Done.

PR #5956, opened 2015-08-15
Use GetThreadIdForTrace in the logger

Orvid:
Because, under MSVC, pthread_t is a struct, and doesn't convert to an unsigned long long.


Orvid:

PR #5958, opened 2015-08-15
Support util/process under MSVC

Orvid:
While this was ported by me before, it was using a pthread library that turned out to be very broken.
This includes the fixes needed to get the header compiling on MSVC with a functional pthread library (where `pthread_t` is a struct).


Orvid:

PR #5961, opened 2015-08-15
Support MSVC in util/ringbuffer.cpp

Orvid:
The pthread impl used on windows defines `pthread_t` as a struct, so we need to use something different for MSVC.


Orvid:

PR #5962, opened 2015-08-15
Fix util/ssl-init.cpp for platforms where pthread_t is a struct

Orvid:
Such as MSVC. The value returned by the callback simply needs to be unique for each thread, which is exactly what `GetThreadIdForTrace` returns.


Orvid:

PR #5963, opened 2015-08-15
Fix util/trace.cpp's pthread interaction under MSVC

Orvid:
Because `pthread_t` is a struct under MSVC.


Orvid:

PR #5970, opened 2015-08-16
Fixup forward declaration mismatches not handled by #5693

Orvid:
This will fixup various places that the tool used for #5693 (D42015) wasn't able to determine the correct definition of.
This isn't dependent on #5693 (D42015), as this is only to fix things that that didn't touch.


Orvid:
I've decided that it's better to do these fixups as part of the main PR, so I'm closing this one.

PR #5971, opened 2015-08-16
Support util/mutex.h under MSVC

Orvid:
This adds a workaround for a limitation of MSVC, and also uses pthread_t_init rather than casting 0 to a pthread_t.

This is dependent upon an issue with how [folly#264](https://github.com/facebook/folly/pull/264) was merged being resolved first.
This is also dependent upon [folly#285](https://github.com/facebook/folly/pull/285) being merged upstream first as well.


Orvid:

PR #5972, opened 2015-08-16
Support AsyncFunc on platforms where pthread_t is a struct

Orvid:
Such as MSVC.
This is dependent upon [folly#285](https://github.com/facebook/folly/pull/285) being merged upstream first.


Orvid:

PR #5985, opened 2015-08-16
Support generating systemlib.php on Windows

Orvid:
We can't set environment variables when invoking a bash scrip ton windows, so get around it by passing them as arguments instead.
This also includes a minor optimization to speed up generating the systemlib on windows, because process creation is painfully slow. Even with this, it can still take 10-15 seconds to generate it due to the number of calls to basename.

This will require changes to the internal build system.


Orvid:
I decided to go a different route and just eliminate the systemlib entirely, closing this in favor of #6295 (D47421).

PR #5990, opened 2015-08-16
[MSVC] Adjust a couple of types to play nice with MSVC's bitfield layout

Orvid:
As described in the define this adds, the issue here is that MSVC will only put two bitfields in the same byte if their types are the same size.
This is done in such a way as to only effect the MSVC build.


fredemmott:
Closing; changes requested on Phabricator for > 3 weeks. WIll re-open when updated.

PR #6002, opened 2015-08-17
[MSVC] Port ext_std_file to MSVC

Orvid:
I do this by disabling `chown`, `lchown`, `chgrp`, `lchgrp`, and `chroot`.
I also do this by switching from direct system calls for anything that MSVC doesn't support as-is, over to use `boost::filesystem` instead. Hopefully this doesn't mess with the error messages, but we'll see.
`touch` got special treatment, and uses a boost implementation on MSVC, and the other implementation elsewhere.


Orvid:
Moved internally to D2880194.

Issue #5261, opened 2015-05-07
Missing memcached::deleteMulti function


r3wt:
```
\nFatal error: Call to undefined method Memcached::deleteMulti() in DBM/Broker.php on line 37
```

r3wt:
< please > < / please >

noony:
This issue can be closed.

PR #6011, opened 2015-08-17
Implement deleteMultiByKey and deleteMulti functions.

noony:
Implement deleteMultiByKey and deleteMulti functions and getMulti must return an empty array instead of false if no key was found.

getMulti must return an array, memcached's documentation is incorrect on php.net :
https://github.com/php-memcached-dev/php-memcached/blob/def716a80f1be732b5dad1c13d3023248db64881/php_memcached.c#L1762

deleteMulti and deleteMultiByKey must return an array, memcached's documentation is incorrect on php.net :
https://github.com/php-memcached-dev/php-memcached/blob/def716a80f1be732b5dad1c13d3023248db64881/php_memcached.c#L691

Fix for #5261.

### DIFF
--- a/hphp/test/slow/ext_memcached/deletemulti.php
+++ b/hphp/test/slow/ext_memcached/deletemulti.php
@@ -20,7 +20,7 @@ $data = array(
     'bar' => 'bar-data',
     'baz' => 'baz-data',
     'lol' => 'lol-data',
-    'kek' => 'kek-data',
+    'kek' => 'kek-data'
 );
 
 $keys = array_keys($data);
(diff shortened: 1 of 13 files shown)


hphp/test/slow/ext_memcached/deletemulti.php
```diff
@@ -0,0 +1,65 @@
...
+    }
+    return true;
+}
+
+$data = array(
+    'foo' => 'foo-data',
+    'bar' => 'bar-data',
+    'baz' => 'baz-data',
+    'lol' => 'lol-data',
+    'kek' => 'kek-data',
```
sgolemon: Lok'Tar Ogar?

JoelMarcey:
@noony Did you see my comment about the tests? Maybe a `.skipif` is all we need?

noony:
@JoelMarcey Modifications done.

PR #6015, opened 2015-08-17
Don't force ext_zend_compat to be enabled with the heap scanning

Orvid:
It is possible to disable ext_zend_compat, and I haven't ported it to MSVC yet, and it will be a while before I get around to doing so.
This just disables the include and the body of the resource scanning stub for zend compat.


Orvid:
Fixed in 4c9ab09, so this PR is no longer needed.

jwatzman:
Sorry, I missed this PR, otherwise would have merged it instead of doing it myself :)

Orvid:
Your fix was done quite a while before I opened this, it just hadn't made it into master yet, and I hadn't realized exactly what you were fixing in your message on the commit.

PR #6063, opened 2015-08-19
[MSVC] Handle the pthread implementation we use under MSVC

Orvid:
This is the changes needed to compile against the pthread implementation we user under MSVC. It includes both changes needed due to `pthread_t` being a struct, and `pid_t` being defined as a `void*`.

This is a combined form of 7 other PRs, #5956 (D44367), #5958 (D44379), #5961 (D44397), #5962 (D44403), #5963 (D44409), #5971 (D44457), and #5972 (D44463), as well as an additional set of changes for the core of the runtime.

This is dependent upon [folly#285](https://github.com/facebook/folly/pull/285) being merged upstream first.
This is also dependent upon an issue with how [folly#264](https://github.com/facebook/folly/pull/264) was merged being resolved first.


Orvid:
Moved internally to D2880839.

Issue #5932, opened 2015-08-13
Bug on casting double to uint64_t


lbianc:
Taking a closer look into the cast function, I could identify a behavior that would not be the correct one when checking if the double value fits in `uint64_t`. 
See code below (https://github.com/facebook/hhvm/blob/master/hphp/runtime/base/type-conversions.h#L69):
```cpp
inline int64_t toInt64(double  v) {
  // If v >= 0 is false and v < 0 is false, then v is NaN. In that case, on
  // Intel, you get 0x800..00, a.k.a. the minimum int64_t. We mimic that on all
  // platforms, though this makes us sad.
  return (v >= 0
          ? (v > std::numeric_limits<uint64_t>::max() ? 0u : (uint64_t)v)
          : (v < 0 ? (int64_t)v : std::numeric_limits<int64_t>::min()));
}
```

Let `18446744073709551616.0` be `2^64` represented as double. Let’s call it **64b_maxd**.
If `v` has a value in the range from (**64b_maxd - 1024**) to (**64b_maxd + 2048**) the statement `v > std::numeric_limits<uint64_t>::max()` returns false, resulting in a cast of `v` to `uint64_t`. There is no error running the code, because coincidentally the cast for that values returns "0", what is expected for a value that does not fit in a "uint64_t". This bug does not generate an error to x86_64, but may to other platforms.
I coded a small test here: https://gist.github.com/lbianc/3d320a1a3f75f487e21b.

Based on that, the correct way to check if `v` value fits in a `uint64_t` wouldn't be verifying if it is smaller than max of "uint64_t"? 
The change would be:

```cpp
  return (v >= 0
          ? (v < std::numeric_limits<uint64_t>::max() ? (uint64_t)v : 0u)
          : (v < 0 ? (int64_t)v : std::numeric_limits<int64_t>::min()));
```

jwatzman:
> If `v` has a value in the range from (64b_maxd - 1024) to (64b_maxd + 2048) the statement `v > std::numeric_limits<uint64_t>::max()` returns false, resulting in a cast of `v` to `uint64_t`.

Can you elaborate on this? Since 64b_maxd is the maximum uint64_t, wouldn't that be the value of `std::numeric_limits<uint64_t>::max()`? Why is it from the range of 1024 below to 2048 above?

Also, your proposed change is, in essence, changing `a > b ? c : d` into `a < b ? d : c`, which I'd expect to be the same unless `a == b` which I don't think is the problem here?

[...]

PR #6067, opened 2015-08-19
Fix bug on casting double to uint_64

lbianc:
When 'v' is equal to 'std::numeric_limits<uint64_t>::max()', it should not be
casted to uint64_t, because there will be an overflow.
For x86_64 machines, the cast above results in overflow, but the result is
'0', i. e., the expected result.
Others platforms may not behave the same way, as tested on PPC64, that returns
'-1' for the same cast.
Looking for make HHVM behave the same way in both platforms, this change
returns '0' if the number going to be casted ('v') is equal to
'std::numeric_limits<uint64_t>::max()'. There is no behavior change for x86_64
platform.
See discussion: https://github.com/facebook/hhvm/issues/5932.
PS: tested on php5 v5.6.9-1 on PPC64 and that result is also supposed to be ‘0’,
not ‘-1’.

### DIFF
--- a/hphp/runtime/base/type-conversions.h
+++ b/hphp/runtime/base/type-conversions.h
@@ -71,7 +71,7 @@ inline int64_t toInt64(double  v) {
   // Intel, you get 0x800..00, a.k.a. the minimum int64_t. We mimic that on all
   // platforms, though this makes us sad.
   return (v >= 0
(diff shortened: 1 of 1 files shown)


lbianc:
Any comment regarding this PR? Can it be merged?

PR #6068, opened 2015-08-19
Adding DNS cache feature (disabled by default)

### LINKED ISSUE #5927: DNS cache feature (and config option)
The DNS cache feature documented on the HDF description seems not to be available anymore.

If you have a lot of threads and the DNS resolutions are common, the performance is degraded because on gethostbyname_r the thread safety is implemented with a global lock (at least on the squeeze version), so all the DNS resolution are serialized, which could lead to some serious contention on the server.

Is there any workaround about this?

We have a DNS cache implemented for the HPHP version, which should be easy to port it to the current version. Would you accept a pull request of that feature?

dpaneda:
This fix #5927

The change is a little intrusive, but it leads to cleaner code.
The calls to safe_gethostbyname are now changed to cached_gethostbyname.
The cache is implemented with a thread-local map and this map is accessed every
time a DNS resolution is issued.
If the cache is deactivated, the map is cleared before each DNS resolution
(effectively negating the cache).

If the cache is enabled, the entries are considered stale after TTL seconds and
the next DNS resolution triggers a reload.

Related config entries (already on the documentation):
  * hhvm.server.dns_cache.enable (default: false)
  * hhvm.server.dns_cache.ttl (default: 60)

### DIFF
--- a/hphp/doc/options.compiled
+++ b/hphp/doc/options.compiled
@@ -314,12 +314,6 @@ These are experimental LFU settings.
 
     }
 
-    # DNS cache
-    DnsCache {
-      Enable = false
-      TTL = 600   # in seconds
-    }
-
     # Light process has very little forking cost, because they are pre-forked
     # Recommend to turn it on for faster shell command execution.
     LightProcessFilePrefix = ./lightprocess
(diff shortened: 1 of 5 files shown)


billf:
close as requested by contributor in https://reviews.facebook.net/D44931#771369

PR #6072, opened 2015-08-19
Add some type hints to builtins_class

simonwelsh:
Test Plan: I checked the return types added to make sure PHP doesn't return false on failures.


hphp/hack/hhi/stdlib/builtins_class.hhi
```diff
@@ -12,17 +12,17 @@ function get_declared_classes();
 function get_declared_interfaces();
 function get_declared_traits();
 function enum_exists(string $class_name, bool $autoload = true): bool;
-function class_exists($class_name, bool $autoload = true): bool;
-function interface_exists($interface_name, bool $autoload = true): bool;
-function trait_exists($trait_name, bool $autoload = true): bool;
-function get_class_methods($class_or_object);
-function get_class_vars($class_name);
+function class_exists(string $class_name, bool $autoload = true): bool;
```
steelbrain: I think hack has a `classname` type right?

SiebelsTim: I'm relatively sure you can't use that type here though. These are php functions. You need to be able to pass anything to them.

fredemmott: This is the hhi, not the hni, so classname should work. Might be overly-strict though.

cc @jwatzman

SiebelsTim: @fredemmott What would you type it, so that you can pass a string? I thought you need `::class` for it.

fredemmott: Why do you need to pass an arbitrary string if you're using these functions from Hack code?

simonwelsh: If you've got a `classname`, then the class exists so there's no point calling this.

I'm using `class_exists` as I can't type user attributes, so can't enforce something being a `classname`, so have to manually check the value is valid rather than relying on the type checker.

PR #6085, opened 2015-08-21
Lock before truncating with file_put_contents()

### LINKED ISSUE #5657: file_put_contents() LOCK_EX flag is useless with advisory locking
HHVM seems to have the same problem as described at https://bugs.php.net/bug.php?id=43182

When `file_put_contents('foo', 'new contents', LOCK_EX);` is called HHVM will open the file in `wb` mode which truncates the existing file and then try to acquire a lock on the file (<https://github.com/facebook/hhvm/blob/f4f799a566d8890efb2818f9b81b9562420088eb/hphp/runtime/ext/std/ext_std_file.cpp#L587-L616>).

bd808:
Ensure that lock is acquired before truncating file when
`file_put_contents('foo', 'new contents', LOCK_EX);` is called. This
requires opening the file in create mode rather than write mode which
automatically truncates, acquiring the lock and then truncating the
file.

See https://bugs.php.net/bug.php?id=43182 for detailed test case.

Closes #5657

### DIFF
--- a/hphp/runtime/ext/std/ext_std_file.cpp
+++ b/hphp/runtime/ext/std/ext_std_file.cpp
@@ -588,9 +588,21 @@ Variant HHVM_FUNCTION(file_put_contents,
                       int flags /* = 0 */,
                       const Variant& context /* = null */) {
   CHECK_PATH(filename, 1);
+
+  char mode[3] = "wb";
+  if (flags & PHP_FILE_APPEND) {
+    mode[0] = 'a';
+  } else if (flags & LOCK_EX) {
+    // Open in "create" mode (writing only, create if needed, no truncate)
+    // so that the file is not modified before we attempt to aquire the
+    // requested lock.
+    mode[0] = 'c';
+  }
+  mode[2] = '\0';
+
   auto file = File::Open(
     filename,
-    (flags & PHP_FILE_APPEND) ? "ab" : "wb",
+    mode,
     flags,
     dyn_cast_or_null<StreamContext>(context)
   );
@@ -611,6 +623,10 @@ Variant HHVM_FUNCTION(file_put_contents,
     if (!file->lock(LOCK_EX)) {
       return false;
     }
+  }
+
+  if (mode[0] == 'c') {
+    file->truncate(0);
   }
 
   int numbytes = 0;


hphp/runtime/ext/std/ext_std_file.cpp
```diff
@@ -588,9 +588,21 @@ Variant HHVM_FUNCTION(file_put_contents,
...
   CHECK_PATH(filename, 1);
+
+  char mode[3] = "wb";
+  if (flags & PHP_FILE_APPEND) {
+    mode[0] = 'a';
+  } else if (flags & LOCK_EX) {
+    // Open in "create" mode (writing only, create if needed, no truncate)
+    // so that the file is not modified before we attempt to aquire the
+    // requested lock.
+    mode[0] = 'c';
```
sgolemon: I don't think 'c' does what you think it does...

```
NOTES
   Glibc Notes
       The GNU C library allows the following extensions for the string speci-
       fied in mode:

       c (since glibc 2.3.3)
              Do not make the open operation, or  subsequent  read  and  write
              operations, thread cancellation points.
```

sgolemon: Nevermind me.  This is caught before being passthrough to posix fopen...

```
'c'	Open the file for writing only. If the file does not exist, it is created. If it exists, it is neither truncated (as opposed to 'w'), nor the call to this function fails (as is the case with 'x'). The file pointer is positioned on the beginning of the file. This may be useful if it's desired to get an advisory lock (see flock()) before attempting to modify the file, as using 'w' could truncate the file before the lock was obtained (if truncation is desired, ftruncate() can be used after the lock is requested).
```

bd808: Yeah, it's special PHP magic. It looks to me like [PlainFile::open](https://github.com/facebook/hhvm/blob/master/hphp/runtime/base/plain-file.cpp#L85-L95) at least does the right thing. I'm not 100% sure about all other stream types.

This patch did fix the reproduction case given in https://bugs.php.net/bug.php?id=43182 for me in local testing but I'm not sure how to translate a test that needs multiple processes into the HHVM test system.

PR #6103, opened 2015-08-22
Don't use the ?: operator

Orvid:
MSVC doesn't support it.


hphp/runtime/vm/jit/mc-generator.cpp
```diff
@@ -546,7 +546,9 @@ MCGenerator::createTranslation(const TranslArgs& args) {
                                   TransFlags().packed);
   } else {
     auto const stubsize = svcreq::stub_size();
-    auto newStart = code.cold().allocInner(stubsize) ?: code.cold().frontier();
+    auto newStart = code.cold().allocInner(stubsize);
+    if (!newStart)
```
sgolemon: nit; Braces, please?

PR #6104, opened 2015-08-22
Explicitly cast kNumActRecCells to int32 when constructing AStack

Orvid:
Because `kNumActRecCells` is defined as a `size_t`, and, although the actual value would fit in an `int32_t`, MSVC attempts to resolve the overload based on the `size_t` type, and fails.


hphp/runtime/vm/jit/memory-effects.cpp
```diff
@@ -823,7 +823,7 @@ MemEffects memory_effects_impl(const IRInstruction& inst) {
           // ActRec, but AliasClass needs an offset to the highest cell it will
           // store.
           spOffset.offset + int32_t{kNumActRecCells} - 1,
-          kNumActRecCells
+          (int32_t)kNumActRecCells
```
sgolemon: c++ style static_cast<int32_t>(), please

PR #6113, opened 2015-08-22
Only output MySQL stats if the MySQL extension is enabled

Orvid:
Because we can't build with MySQL currently under MSVC. Because we can't currently build MySQL 5.6, nor the RC version of 5.7 under MSVC 2015.

This requires #5872 (D43749).


JoelMarcey:
@Orvid Can you resubmit this as a new pull request. I landed it internally forgetting that it required #5872 first -- and it is causing a bit of mess with our review tooling.

Thanks!

PR #6124, opened 2015-08-23
[MSVC] Use the folly event portability header where needed

Orvid:
In order to workaround the fact that the Windows version of libevent expects `HANDLE`'s and HHVM works with file descriptors, I created a portability header in folly for it.
This just includes the header in the only place (currently) in HHVM where it is needed. (it's needed in multiple places in Folly and Proxygen)

This requires [folly#301](https://github.com/facebook/folly/pull/301) to be merged upstream first.


Orvid:
Replaced internally with D2874829.

Issue #6058, opened 2015-08-18
intdiv function


jwatzman:
https://wiki.php.net/rfc/intdiv

SiebelsTim:
Closed in #6129

nikic:
If this gets closed, there should be another issue that tracks the incorrect failure conditions.

jwatzman:
What do you mean "incorrect failure conditions"?

nikic:
@jwatzman Division by zero should throw DivisionByZeroError (instead of a warning) and division of -1 by PHP_INT_MIN should throw ArithmeticError instead of returning false. I assumed this wasn't done yet as it depends on the Exception/Error splitup, which probably requires some extra work.

jwatzman:
Ah yeah, ok. That doesn't look to be part of this RFC -- which RFC was that changed in?

nikic:
@jwatzman This did not go through an RFC. A couple of other operator error conditions were also changed:

* Division by zero with `/` now results in Inf/-Inf/NaN (and still a warning). Previously it returned false.
* Modulo by zero with `%` now throws a `DivisionByZeroError`. Previously it threw a warning and returned false.
* Bitshifts by negative numbers now throw an `ArithmeticError`. This was a warning+false for a short time, as this check was only introduced in PHP 7 anyway.

Hope I didn't miss anything.

SiebelsTim:
Here is a not working start of Throwable implementation:
https://gist.github.com/SiebelsTim/71685ba69a452258d435

I tried to take the Exception implementation into a trait. However that results in errors, because the runtime expects `$previous` to be readable rather than calling `getPrevious()` (and probably more).

That approach might be completely unusable, but maybe it helps setting up class hierarchy or something useful.

jwatzman:
@nikic thanks for the info, going to file a separate issue for that.
@SiebelsTim did you mean to post that on https://github.com/facebook/hhvm/issues/6037?

SiebelsTim:
@jwatzman It would've fit there as well. But I posted it here, because nikic mentioned the error cases which need the Throwable rework.

PR #6129, opened 2015-08-24
Implement intdiv function

noony:
Closes : #6058


hphp/runtime/ext/std/ext_std_math.cpp
```diff
@@ -495,6 +495,16 @@ int64_t HHVM_FUNCTION(mt_rand,
 
 double HHVM_FUNCTION(lcg_value) { return math_combined_lcg();}
 
+Variant HHVM_FUNCTION(intdiv, int64_t numerator, int64_t divisor) {
+  if (divisor == 0) {
+    raise_warning("intdiv(): Division by zero");
```
nikic: Should throw `DivisionByZeroError`.

hphp/runtime/ext/std/ext_std_math.cpp
```diff
@@ -495,6 +495,16 @@ int64_t HHVM_FUNCTION(mt_rand,
...
+    return false;
+  } else if (divisor == -1 && numerator == LONG_MIN) {
+    return 0;
```
nikic: Should throw `ArithmeticError`.

hphp/runtime/ext/std/ext_std_math.cpp
```diff
@@ -495,6 +495,16 @@ int64_t HHVM_FUNCTION(mt_rand,
 
 double HHVM_FUNCTION(lcg_value) { return math_combined_lcg();}
 
+Variant HHVM_FUNCTION(intdiv, int64_t numerator, int64_t divisor) {
+  if (divisor == 0) {
+    raise_warning("intdiv(): Division by zero");
+    return false;
+  } else if (divisor == -1 && numerator == LONG_MIN) {
```
nikic: Instead of `LONG_MIN` it would be better to use the `int64_t` min here (which do not match e.g. on windows). I think in C++ you get this via numeric_limits.

noony: I'll replace this by std::numeric_limits<int64_t>::max()

PR #6132, opened 2015-08-24
Implement JSON_PRESERVE_ZERO_FRACTION

### LINKED ISSUE #6046: JSON preserve fractional part
https://wiki.php.net/rfc/json_preserve_fractional_part

SiebelsTim:
new php7 feature
Closes #6046 (see #6046 for RFC)

Also remove a unnecessary if

The test I added is from the php-src diff.

Test Plan: new test

### DIFF
--- a/hphp/runtime/base/variable-serializer.cpp
+++ b/hphp/runtime/base/variable-serializer.cpp
@@ -242,8 +242,11 @@ void VariableSerializer::write(double v) {
       vspprintf(&buf, 0, "%.*k", precision, v);
       if (m_option & k_JSON_PRESERVE_ZERO_FRACTION
           && strchr(buf, '.') == nullptr) {
-        free(buf);
-        asprintf(&buf, "%.1f", v);
+        int buflen = strlen(buf);
+        buf = (char *) realloc(buf, buflen + 3);
+        buf[buflen++] = '.';
+        buf[buflen++] = '0';
+        buf[buflen] = '\0';
       }
       m_buf->append(buf);
       free(buf);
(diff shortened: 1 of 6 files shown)


hphp/runtime/base/variable-serializer.cpp
```diff
@@ -240,12 +240,15 @@ void VariableSerializer::write(double v) {
     if (!std::isinf(v) && !std::isnan(v)) {
       char *buf;
       vspprintf(&buf, 0, "%.*k", precision, v);
+      if (m_option & k_JSON_PRESERVE_ZERO_FRACTION
+          && strchr(buf, '.') == nullptr) {
+        free(buf);
+        asprintf(&buf, "%.1f", v);
```
nikic: Won't this disregard `precision`? Also, is `asprintf` locale-independent, i.e. will this also use `.` instead of `,`?

I think it would be safer to just do the same as in PHP and do an extra append of `".0"`.

SiebelsTim: I think precision only truncates and does not add trailing zeros.
However, you are correct about the locale. I didn't think of it. Thanks!

nikic: You're right, the comment about precision was bogus.

Rather than reallocating the buffer, maybe move the check one line down and do something like this?

```
m_buf->append(buf);
if (m_option & k_JSON_PRESERVE_ZERO_FRACTION
    && strchr(buf, '.') == nullptr) {
  m_buf->append(".0");
}
```

SiebelsTim: This is perfect. I start to wear blinders when looking at php-src...
You've done most the work for this diff by now. :D

PR #6143, opened 2015-08-25
[MSVC] Use object libraries for hphp_runtime_static and hphp_runtime_ext

Orvid:
DO NOT MERGE! - This WILL break every build except for MSVC. I'm opening this to get feedback on the idea.

This eliminates the need for `--whole-archive` to be passed to the linker, because there is no equivelent for MSVC.
This achieves this by using OBJECT libraries in CMake, which causes CMake to pass the object files as if they are part of the original binary being compiled, rather than being present in a static library.

This is implemented in a quick hacky form, the final version will keep hphp_runtime_static and hphp_runtime_ext as static libraries, which is needed for DSO extension compilation. The final version will also strip the entire --whole-archive structure out of HPHPSetup, and will also convert hphp_ext_zend_compat to an object library as well.


Orvid:
Moved internally to D2901111.

PR #6147, opened 2015-08-25
Fix SplFixedArray::fromArray performance

### LINKED ISSUE #5671: SplFixedArray::fromArray() can take minutes to execute on even moderately-large datasets

SiebelsTim:
It would execute an array_pad for every entry. This would get bad pretty fast.

This diff also handles these tests, raising a warning on bad input.

Fixes #5671

Test Plan: exisiting tests, no new tests because this is perf only
I checked for perf on my machine with given test script from #5671 and it's way faster

### DIFF
--- a/hphp/system/php/spl/datastructures/SplFixedArray.php
+++ b/hphp/system/php/spl/datastructures/SplFixedArray.php
@@ -245,15 +245,20 @@ class SplFixedArray implements \HH\Iterator, ArrayAccess, Countable {
    */
   public static function fromArray($array, $save_indexes = true) {
     $fixed_array = new self;
+    if (!is_array($array)) {
+      trigger_error(
+        sprintf("SplFixedArray::fromArray() expects parameter 1 to be array,".
+                " %s given", gettype($array)),
+        E_WARNING);
+      return $fixed_array;
+    }
     if ($save_indexes) {
+      $fixed_array->setSize(max(array_keys($array)) + 1);
       foreach ($array as $key => $value) {
         if (!is_numeric($key) || $key < 0) {
           throw new InvalidArgumentException(
             'array must contain only positive integer keys'
           );
-        }
-        if ($key >= $fixed_array->count()) {
-          $fixed_array->setSize($key + 1);
         }
         $fixed_array[$key] = $value;
       }


hphp/system/php/spl/datastructures/SplFixedArray.php
```diff
@@ -245,16 +245,21 @@ public function toArray() {
...
         if (!is_numeric($key) || $key < 0) {
           throw new InvalidArgumentException(
             'array must contain only positive integer keys'
           );
         }
-        if ($key >= $fixed_array->count()) {
-          $fixed_array->setSize($key + 1);
-        }
         $fixed_array[$key] = $value;
       }
```
nikic: If somebody is wondering why we don't just reuse `$array` and only do a validation loop, the answer is (probably) that we need to break references, so we have to do a hard copy.

It's not related to your change, but that `is_numeric` check should be `is_int`. We don't want `"1.2"` as a key in a fixed array.

SiebelsTim: > It's not related to your change, but that `is_numeric` check should be `is_int`. We don't want `"1.2" ` as a key in a fixed array.

Yup, but that's probably best for an own diff.

PR #6169, opened 2015-08-28
Fix build on Windows.

OCamlPro-Bozman:
- remove usage of 'strptime' reimplemented in OCaml;
 - update build.ocp (adding recently introduced source files). Please
   update this file when adding or removing source files, if not it
   will break build on Windows.


JoelMarcey:
Replaced by: https://github.com/facebook/hhvm/pull/6436

PR #6171, opened 2015-08-30
Allow implicit conversion of nullptr to Variant

Orvid:
Constructing a Variant representing null is currently quite verbose.
This makes it possible to implicitly construct a Variant from nullptr via the nullptr_t type.


Orvid:
Not needed.

PR #6175, opened 2015-08-31
Proper timeout handling on Windows.

OCamlPro-Bozman:
Add a timeout module with two implementations:
  * `Alarm_timeout` which is used on Unix. It is based on `Unix.alarm`
  to trigger a timeout. On Unix, it is the same behavior as before;
  * `Select_timeout` which is used on Windows. It is based on
  `Unix.select` to trigger a timeout. The usage of `Unix.select` is
  hidden behind a custom `Timeout.in_channel`. It is similar to the
  classical `Pervasives.in_channel`, but all input function receives
  an optionnal `timeout` parameter


JoelMarcey:
This has been merged via af1fad2

haowanaa:
您好，来信我已收到，我将尽快给您回复。

PR #6177, opened 2015-08-31
Connection error now throws RedisException

darkain:
Connection errors now throw a RedisException instead of triggering an error. This allows an application to catch the connection problem and handle it gracefully.

### DIFF
--- a/hphp/system/php/redis/Redis.php
+++ b/hphp/system/php/redis/Redis.php
@@ -1626,10 +1626,10 @@ class Redis {
         $conn    = stream_socket_client(
           $sok, $errno, $errstr, $timeout, 2, $context);
       } else {
-        $conn = pfsockopen($host, $port, $errno, $errstr, $timeout);
+        $conn = @pfsockopen($host, $port, $errno, $errstr, $timeout);
       }
     } else {
-        $conn = fsockopen($host, $port, $errno, $errstr, $timeout);
+        $conn = @fsockopen($host, $port, $errno, $errstr, $timeout);
     }
     $this->last_connect = time();
     $this->host = $host;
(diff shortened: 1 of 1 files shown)


JoelMarcey:
@darkain Hi! Please sign the CLA so we can import this PR for internal testing, etc.

darkain:
@JoelMarcey done!

SiebelsTim:
Is this php5's behaviour? If so, it needs a test in `slow/ext_redis`.

darkain:
This issue effects the built in Redis module for HHVM. If there is a connection issue to the Redis server, the script entirely bails fsockopen without even letting the application handle the problem (unless a custom error handler is set, but this is messy since it is application wide). This change makes it so an exception is thrown that the calling application can catch, instead of an error. 

For example: my application only uses Redis as a caching server, nothing more. So if there is a connection error to the Redis server (such as rebooting it for an upgrade), the application will continue to function, but in a performance degraded state. This is a much better alternative than simply having the entire application fail because it cannot connect to the Redis server.

JoelMarcey:
As @SiebelsTim mentioned in the phabricator diff, this looks like it breaks compatibility with PHP 5; and there has to be a real compelling reason for us to do that. Do you know if this changes in PHP 7?

darkain:
I have this actively running right now on both HHVM 3.9 and PHP 5.6.12. This change shouldn't break any compatibility, it only changes from an error being thrown to an exception being thrown instead.

r3wt:
:+1:

fredemmott:
This is a BC break, however the same BC break was done in the PHP5 extension. This doesn't happen often in HHVM, but 3rd-party extensions aren't as rigid about BC as php-internals.

JoelMarcey:
Ok. I am going to accept the pull request based on the discussion and comments here.

JoelMarcey:
@darkain Actually, can you have a look at the diff again...

https://reviews.facebook.net/D45861#785265

We are getting what seems to be a related test failure.

fredemmott:
We close pull requests after changes are requested on phabricator for > 1 week. Will re-open when changes are made.

Issue #2854, opened 2014-06-04
Memory Leak in SimpleXML parsing an XML


nserban:
SimpleXML leaks memory each time it parses a XML file.

Code to reproduce:

``` php
for ($i=0;$i<10000;$i++) {
        $myFeed = simplexml_load_file('test.xml');
        foreach ($myFeed->entry as $xmlEntry) {
            // Do something for every xml entry
        }
        if ( !($i % 2000) )
            echo round(memory_get_usage(true) / 1024 / 1024) . ' MB' . PHP_EOL;
    }
```

test.xml

``` xml
<?xml version="1.0" encoding="utf-8"?>
<feed>
    <entry>
        <title>Title 1</title>
        <url>http://hhvm.com/</url>
    </entry>
    <entry>
        <title>Title 2</title>
        <url>http://hhvm.com/</url>
    </entry>
    <entry>
        <title>Title 3</title>
        <url>http://hhvm.com/</url>
    </entry>
    <entry>
        <title>Title 4</title>
        <url>http://hhvm.com/</url>
    </entry>
    <entry>
        <title>Title 5</title>
        <url>http://hhvm.com/</url>
    </entry>
</feed>
```

Output:

```
2 MB
25 MB
49 MB
72 MB
95 MB
```

HHVM Version:

```
HipHop VM 3.2.0-dev+2014.06.02 (rel)
Compiler: heads/master-0-gf43ae21551b3e37ec955bce770ca9a99677388fc
Repo schema: b6cc5eef6685b96a2185a5c1950f6c81ad8583b1
```

Although it can be reproduced with the stable version:

```
HipHop VM 3.1.0 (rel)
Compiler: tags/HHVM-3.1.0-0-g71ecbd8fb5e94b2a008387a2b5e9a8df5c6f5c7b
Repo schema: 88ae0db264d72ec2e2eb22ab25d717214aee568b
```

I try to free the memory using unset() or set the values on variables with NULL, but don't work.

Using PHP

```
PHP 5.4.28-1~dotdeb.1 (cli) (built: May  7 2014 22:16:44) 
```

The output is:

```
0 MB
0 MB
0 MB
0 MB
0 MB
```

Machine info: 

```
Debian 7
Linux donald 3.2.0-4-amd64 #1 SMP Debian 3.2.57-3+deb7u1 x86_64 GNU/Linux
Libxml2 version 2.8.0+dfsg1-7+nmu3  amd64 
```

Kozzi11:
[...]

PR #6213, opened 2015-09-09
Fix a memory leak in simplexml

Kozzi11:
https://github.com/facebook/hhvm/issues/2854#issuecomment-138819323


JoelMarcey:
We find the underlying culprit of the leak and it should land today...

internal diff: D2433735

cc: @edwinsmith

PR #6217, opened 2015-09-10
[MSVC] Port ext_std_process to MSVC

Orvid:
This adds MSVC support to the process functions.

~~This moves everything except for the `pcntl_*` functions out of ext_process and into ext_std. This reflects the way PHP is setup, as the `pcntl_*` functions can't really be supported under MSVC.
This also adds support for MSVC to the functions that were brought in.~~

~~It will probably be easier to review the 2 revisions in this PR separately. The first does the actual move, the second has the changes for MSVC support.~~


Orvid:
Moved internally as D2871213.

PR #6221, opened 2015-09-10
Fix #6209 Add IntlChar::NO_NUMERIC_VALUE

### LINKED ISSUE #6209: Missing constant: IntlChar::NO_NUMERIC_VALUE
According to the documentation for `IntlChar::getNumericValue()` ([see source](https://github.com/facebook/hhvm/blob/master/hphp/runtime/ext/icu/ext_icu_uchar.php#L67)), a value of `IntlChar::NO_NUMERIC_VALUE` may be returned if no numeric value exists for the given code point.  This behavior is consistent with [the underlying ICU library](https://ssl.icu-project.org/apiref/icu4c/uchar_8h.html#a36982efc84143b9bb5a11bf54b86af48) and PHP 7 implementation.

However, this constant doesn't actually exist - attempting to reference it results in the following error:

    Fatal error: Couldn't find constant IntlChar::NO_NUMERIC_VALUE in /in/48Bnh on line 5

You can replicate this with the following code:

    <?php

    $value = IntlChar::getNumericValue("x");
    var_dump($value);
    var_dump($value === IntlChar::NO_NUMERIC_VALUE);

([See results on 3v4l.org](https://3v4l.org/48Bnh))

According to the [ICU documentation](https://ssl.icu-project.org/apiref/icu4c/uchar_8h.html#a0358d9cce8fdd3255867affe2a8b21ff) and current implementation, `IntlChar::NO_NUMERIC_VALUE` should have a value of `float(-123456789)`.

I therefore request that you define/expose this missing constant.  [I have filed a similar issue with the PHP folks](https://bugs.php.net/bug.php?id=70455).

Orvid:
This is a fix for [#6209](https://github.com/facebook/hhvm/issues/6209).

Note that PHP doesn't currently have this constant either.

### DIFF
--- a/hphp/runtime/ext/icu/ext_icu_uchar.cpp
+++ b/hphp/runtime/ext/icu/ext_icu_uchar.cpp
@@ -357,6 +357,8 @@ void IntlExtension::initUChar() {
   IC_CONSTL("CODEPOINT_MAX", UCHAR_MAX_VALUE)
   IC_CONSTL("FOLD_CASE_DEFAULT", U_FOLD_CASE_DEFAULT)
   IC_CONSTL("FOLD_CASE_EXCLUDE_SPECIAL_I", U_FOLD_CASE_EXCLUDE_SPECIAL_I)
+  Native::registerClassConstant<KindOfDouble>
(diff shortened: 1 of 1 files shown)


colinodell:
Corresponding PHP bug report: https://bugs.php.net/bug.php?id=70455

PR #6222, opened 2015-09-11
Fix linking errors with static versions of ICU

Orvid:
ICU defines MessageFormatAdapter::getArgTypeList itself internally, and when we link against it as a static lib under MSVC, we get errors because it's defined in multiple places.
To solve this, just name it something different.


cscott:
Hey -- thanks for this, for me it solved a problem with building "normal PHP" with static libicu as well: https://github.com/cscott/node-php-embed/commit/8ad686cb3e5c168bda89e1de95b708762667dc5d#diff-585ff91cd15eca30cc314c18299f62a6

PR #6237, opened 2015-09-13
Don't mark the Systemlib as always out of date

Orvid:
Currently the systemlib is generated via add_custom_target, which is always considered out of date by CMake, which gets annoying when trying to build and debug HHVM in Visual Studio.
This switches systemlib generation to a PRE_BUILD (PRE_LINK for everything except Visual Studio) command of hhvm, eliminating the problem.


Orvid:
I decided to go a different route and just eliminate the systemlib entirely, closing this in favor of #6295 (D47421).

PR #6247, opened 2015-09-14
Mark the 'MemExceededFlag' as not sticky

jmarrama:
This diff changes the surprise flag that denotes that the current
request has run out of memory as not 'sticky', meaning that it will
now be cleared when surprise flags are checked and cleared. It should
_not_ be sticky because this will cause all subsequent calls to OOM,
even after the stack has been unwound and everything cleared (for
example, this causes shutdown handlers to OOM deterministically even
when there is ample memory).

The previous behavior was likely an unintended consequence of commit
40f9d9547b3e13767575534686e5e1b061d2d029.

### DIFF
--- a/hphp/runtime/base/surprise-flags.h
+++ b/hphp/runtime/base/surprise-flags.h
@@ -49,16 +49,17 @@ enum SurpriseFlag : size_t {
   /* Set if a GC should be run at the next safe point. */
   PendingGCFlag        = 1ull << 60,
 
-  /*
-   * Flags that shouldn't be cleared by fetchAndClearSurpriseFlags, because
-   * fetchAndClearSurpriseFlags is only supposed to touch flags related to
-   * PHP-visible signals/exceptions and resource limits.
-   */
   ResourceFlags =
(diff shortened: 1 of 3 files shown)


jmarrama:
Oops, I lied - this behavior was an intended consequence of 1a0b6ab85b6ece5c42ccdbc86a05d461b1c1ec0b. @markw65 - I now am not so sure about this making the OOM flag not sticky. Not sure what the best way forward is here....

jmarrama:
On a third pass, my eyes shift back to 40f9d9547b3e13767575534686e5e1b061d2d029 - this introduced the behavior in handle_exception_helper that flips back on the MemExceeededFlag at scope exit. Before this commit, the MemExceededFlag was cleared in the first line of ExecutionContext::onFatalError (which handle_exception_helper calls on OOMs). Maybe the right solution here is to remove the OOM flag from ResourceFlags?

jmarrama:
Closing in favor of https://github.com/facebook/hhvm/pull/6256

PR #6257, opened 2015-09-16
[MSVC] Use auto for the return type of the ARCH_SWITCH_CALL lambda

Orvid:
Because using the explicit type makes MSVC grumpy and it complains about a function pointer type not being allowed to have default arguments.

Also, this is far cleaner.


Orvid:
No longer needed.

PR #6291, opened 2015-09-22
Use [[noreturn]] rather than ATTRIBUTE_NORETURN

Orvid:
It's shorter, and it's [standardized](http://en.cppreference.com/w/cpp/language/attributes) as part of C++ 11. It's also supported by GCC, Clang and MSVC.

Note that in all 3 compilers, C++ 11 attributes must be before any `__declspec` or `__attribute__` specifiers associated with the function.


Orvid:
It has been decided that `ATTRIBUTE_NORETURN` is more flexible, so, even though this is standardized, this won't be merged.

PR #6295, opened 2015-09-23
Eliminate the systemlib

Orvid:
Well, the generation of it at compile time anyways.

This embeds the manifest file, `hphp/system/php.txt`, and each of the files in it rather than combining them at compile time. At runtime, it loads the manifest, and combines the individual pieces of the systemlib into one. This could eventually be extended to merge the systemlibs for all extensions, including DSO extensions, as well.

This also does a minor refactor of the embedded data to add a `data()` function that retrieves the actual data, otherwise it would have had to be duplicated for the systemlib and extension lib embedded data loading.

My primary motivation for this is that, unless I move the actual generation of the systemlib into `hphp/hhvm/CMakeLists.txt`, where it doesn't belong, it is not possible for me to generate the systemlib in a way that doesn't cause it to always be considered out of date. To add to this, due to the cost of process creation on Windows, generating the systemlib takes 15-20 seconds when using the normal bash script. To work around this, I've been using a custom tool (written in D) that does the generation in < 1 second, but I really don't think a custom tool should be needed for something as simple as this.
It was also suggested to use re-write the script in perl, but, as I don't know perl, and as doing it this way allows the possibility of including the systemlib of the extensions as well, I decided this would be the best way to go.


Orvid:
Brought in internally as D2856418.

PR #6298, opened 2015-09-24
Fix for compiling with GCC 5.2

Aatch:
For some reason, the MovePlan was being initialized with the wrong
values, specifically register #0 instead of InvalidReg like it is
supposed to. This meant that the entire move plan started out
representing a move from register #0 to every other physical register.

The extra copies generated bloated the code blowing, through the
translation caches almost immediately (each copy instruction would turn
into 32 copy instructions). This resulted in HHVM being unable to start
at all.

This patch explicitly sets the map's contents to InvalidReg as a
workaround.

### DIFF
--- a/hphp/runtime/vm/jit/phys-reg.h
+++ b/hphp/runtime/vm/jit/phys-reg.h
@@ -151,8 +151,8 @@ struct PhysReg {
   template<typename T>
   struct Map {
     Map() {
-      // Workaround for a potential GCC 5 bug, value initializing m_elms seems to
-      // use zero-initialization instead of default initialization.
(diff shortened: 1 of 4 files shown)


edwinsmith:
If this really is a bug in gcc, can you put a comment near the workaround with a gcc bug-tracking number? that way when gcc is fixed we can remove the workaround.  if default-constructor initialization is broken in gcc 5.2 there are probably many other bugs lurking.

if on the other hand we're in undefined territory, and gcc 5.2 is taking more liberties (because it can), the workaround isn't a workaround, it's a real fix...

Aatch:
@edwinsmith I can't find a bug and can't reproduce an isolated test case for it to open one.

PR #6307, opened 2015-09-27
All in one extensions

Orvid:
This merges the systemlibs of all extensions into the core systemlib at startup. This includes zend_compat and DSO extensions.

This is dependent upon #6295, and will need to be rebased once that is in.
This is also dependent upon ext_closure being ported to HNI rather than IDL.


hphp/runtime/ext/extension-registry.cpp
```diff
@@ -292,6 +292,17 @@ void scanExtensions(IMarker& mark) {
   }
 }
 
+
+std::vector<std::pair<std::string, std::set<std::string>>>
+getAllExtensionSystemlibSources() {
+  std::vector<std::pair<std::string, std::set<std::string>>> ret;
+  for (auto& ext : s_ordered) {
```
Orvid: Note to self when updating this for ext_closure: This needs to be checking that the extensions are actually enabled before adding them to the list.

Orvid:
Moved internally as D2882035.

PR #6311, opened 2015-09-28
[MSVC] Don't modify the vector as we iterate in vasm-xls

Orvid:
So, MSVC gets a bit grumpy when you try to modify the underlying vector of an iterator you're iterating over, this is actually a check that is only done in debug mode, but it does get tripped when this is run.

Unfortunately, even with this fix, it still fails further down the line in register allocation, however, I believe I've managed to do this correctly, and am primarily creating this PR to make sure that the test suite agrees, and that this is correct.

I'm still not entirely sure how `*i = list.back(); list.pop_back()` iterates through the vector.


Orvid:
No longer needed.

PR #6312, opened 2015-09-28
Support Windows and a non-pcntl mode in the tester

Orvid:
This has two distinct parts to it.
The first is adding support for a non-pcntl serial mode. This mode outputs the results of each test directly.
The second are a few changes for Windows, dealing with the directory separator character, and the timeout script not working under msysgit's bash shell.

### DIFF
--- a/hphp/test/run
+++ b/hphp/test/run
@@ -4,6 +4,8 @@
 * Run the test suites in various configurations.
 */
 
+const NO_PCNTL = !function_exists("fork");
+
 function is_testing_dso_extension() {
   // detecting if we're running outside of the hhvm codebase.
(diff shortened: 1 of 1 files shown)


hphp/test/run
```diff
@@ -935,6 +944,30 @@ class Status {
...
+          echo "Failed\n";
+          break;
+        case self::MSG_TEST_SKIP:
+          echo "Skipped\n";
+          break;
+        case self::MSG_SERVER_RESTARTED:
+          echo "Restarted\n";
+          break;
+      }
+      var_dump($msg);
```
steelbrain: Is this `var_dump` intentional?

Orvid:
Closing for now, as this requires significant reworking by me.

PR #6334, opened 2015-10-06
Fix Timezone transitions

Aatch:
This fixes two bugs, a segfault when there were no transitions for the timezone (e.g. UTC) and an incompatibility with Zend PHP with the transition list when using a starting timestamp (it would always use the first transition, rather than the one before the starting timestamp).


hphp/test/slow/config.ini
```diff
@@ -16,3 +16,5 @@ hhvm.mysql.slow_query_threshold = 0
 hhvm.resource_limit.serialization_size_limit = 134217728
 hhvm.server_variables[ALPHA_CONSOLE] = 1
 hhvm.server_variables[TFBENV] = 16777216
+
+date.timezone = 'UTC';
```
steelbrain: no EOF?

aorenste:
It looks like this issue has been addressed separately.

PR #6363, opened 2015-10-13
Remove nonexistent functions in linker commands.

wjywbs:
This fixes linker error because libevent server was removed in f4cb063 and
register_fastcgi_server was removed in c8340da.


jwatzman:
I've had a diff up to do this for a few days... I'll get someone to stamp it.

PR #6365, opened 2015-10-14
Port most extensions to use the new set of constant registering defines

Orvid:
This ports most extensions, specifically the ones that I build under MSVC and can actually test, to use the new set of macros for registering constants. This is a significant refactor, and, as part of the refactor eliminates a large number of global constants that simply weren't needed (they were the exact same, and the same name as a macro already provided, with the difference that the constant was prefixed with `k_`), I'd be amazed if this didn't break at least some of FB's internal code.

~~An example PR of the changes they would allow was requested, so here is the PR, that can be merged after the main one is.~~

~~I originally thought about adding a pair of `HHVM_RC_SAME_PREF_*` macros to replace the `MW_CONST` macro used by ext_imagick, but decided it's not likely to be a common pattern, so I chose not to add them.~~

~~Depends upon #6343 (D48255)~~ - Already merged


Orvid:
This has been split into #6461, #6462, #6463, #6464, #6465, #6466, #6467, #6468, #6469, #6470, #6471, #6472, #6473, #6474, #6475, #6476, #6477, #6478, #6479, #6480, #6481, #6482, #6483, #6484, #6485, #6486, #6487, #6488, #6489, #6490, #6491, and #6492.

Issue #6372, opened 2015-10-15
mb_encoding_aliases is not implemented


nicolas-grekas:
Looks like this PHP function is missing an implementation in HHVM

jwatzman:
Looks like. There's already `test/zend/bad/ext/mbstring/tests/mb_encoding_aliases.php` if someone wants to look into this.

noony:
This issue can be closed. fixed by 4d2e1091abf0aaa90439e8aa783a21e2d6e17407

nicolas-grekas:
Cool, thanks

PR #6389, opened 2015-10-18
Implement missing function mb_encoding_aliases

noony:
Fix for #6372


hphp/runtime/ext/mbstring/ext_mbstring.cpp
```diff
@@ -1030,6 +1030,25 @@ Array HHVM_FUNCTION(mb_list_encodings) {
...
+  const mbfl_encoding *encoding;
+  encoding = mbfl_name2encoding(name.data());
+  if (!encoding) {
+    raise_warning("mb_encoding_aliases(): Unknown encoding \"%s\"",
+                  name.data());
+    return false;
+  }
+
+  Array ret = Array::Create();;
+  if (encoding->aliases != NULL) {
```
SiebelsTim: Use nullptr instead of NULL

noony: Thanks, modification done.

Issue #6392, opened 2015-10-19
Assertion Fails for Debug Builds


octmoraru:
Steps to reproduce:
* Configure for Debug build:
```
./configure -DCMAKE_BUILD_TYPE=Debug
```
* Create test script:
```
echo "<?php echo 'hello, world';" > test.php
```
* Run it:
```
$ /usr/local/bin/hhvm/hhvm-dbg/bin/hhvm test.php
hhvm: /home/omoraru/git/hhvm-dbg/hphp/runtime/base/class-info.cpp:597: static void HPHP::ClassInfo::Load(): Assertion `!i' failed.
Core dumped: Aborted
Stack trace in /tmp/stacktrace.25031.log
Aborted (core dumped)
$ file /tmp/stacktrace.25296.log
/tmp/stacktrace.25296.log: empty
```

paulbiss:
cc @sgolemon looks IDL related, perhaps it has something to do with some of the recent conversions?

octmoraru:
Found the culprit. Submitting a PR right now.

PR #6393, opened 2015-10-19
Fix assertion failing in debug builds.

octmoraru:
Fixes https://github.com/facebook/hhvm/issues/6392
Commit df393b2 moved the ext_closure extension into its own
directory, also adding an IDL to it. A hard-coded IDL is no
longer needed.


jwatzman:
https://github.com/facebook/hhvm/commit/fcb84bee55b2e6cfcbe22cc78658633de5687417

PR #6071, opened 2015-08-19
fix segfault when input shorter than all patterns

ebernhardson:
When the input string is shorter than the shortest input
pattern the last position to visit is set to -1. Patch
short circuits and returns input string when all patterns
are longer than the input string.

PR #6395, opened 2015-10-19
Fix segfault in strtr using the Wu-Manber algorithm

### LINKED ISSUE #6384: Segmentation fault (core) with 3.10.0 on strtr() on some chars if passed array exeeds 64 items
Hi everyone,

starting with the update 3.9.1 to 3.10.0 dokuwiki starts crashing on Mainpage (doku.php). The hhvm/error.log only reports segmentation fault: core dumped and refers to a tmp-log:

PHP Stacktrace:
#0  utf8_deaccent(:, -1) called at [/var/www/service/wiki/inc/pageutils.php:132]
#1  cleanID(:) called at [/var/www/service/wiki/inc/pageutils.php:294]
#2  wikiFN(:, ) called at [/var/www/service/wiki/inc/pageutils.php:549]
#3  resolve_pageid(, :, ) called at [/var/www/service/wiki/inc/template.php:941]
#4  tpl_youarehere() called at [/var/www/service/wiki/lib/tpl/dokuwiki/tpl_header.php:84]
#5  include(/var/www/service/wiki/lib/tpl/dokuwiki/tpl_header.php) called at [/var/www/service/wiki/lib/tpl/dokuwiki/main.php:33]
#6  include(/var/www/service/wiki/lib/tpl/dokuwiki/main.php) called at [/var/www/service/wiki/inc/actions.php:196]
#7  act_dispatch() called at [/var/www/service/wiki/doku.php:124]

I tested the function utf8_deaccent and it crashes if only a ":" string is passed using the specified array:

````PHP
if(!function_exists('utf8_deaccent')){
    /**
     * Replace accented UTF-8 characters by unaccented ASCII-7 equivalents
     *
     * Use the optional parameter to just deaccent lower ($case = -1) or upper ($case = 1)
     * letters. Default is to deaccent both cases ($case = 0)
     *
     * @author Andreas Gohr <andi@splitbrain.org>
     *
     * @param string $string
     * @param int $case
     * @return string
     */
    function utf8_deaccent($string,$case=0){
        if($case <= 0){
            global $UTF8_LOWER_ACCENTS;
            $string = strtr($string,$UTF8_LOWER_ACCENTS);
        }
        if($case >= 0){
            global $UTF8_UPPER_ACCENTS;
            $string = strtr($string,$UTF8_UPPER_ACCENTS);
        }
        return $string;
    }
}

SiebelsTim:
We would not take the source string length into consideration when calculating
the smallest string. That would result in an unsigned integer overflow
calculating the lastpos in WuManberReplacement::translate. That would later
result in a segfault trying to access a char of a string beyond its size.

Fixes #6384

Test Plan: new test to confirm fixing the segfault. Old slow and zend tests
to confirm I didn't break anything.


SiebelsTim:
Closing in favor of #6071

PR #6226, opened 2015-09-11
General cleanup of ext_pdo

Orvid:
The biggest thing this does is make it so that ext_pdo_sqlite and ext_pdo_mysql aren't used in ext_pdo. This also eliminates the need to declare the individual PDO drivers in pdo_driver.cpp, and instead allows them to be declared in the respective extension, ext_pdo_sqlite / ext_pdo_mysql.

This also eliminates all the global constants in C++ that are not actually needed, and removes the global StaticStrings that were only used once.

This directly depends on #6224 (D46659) and, because this has to include that, will need to be rebased once that is merged.

PR #6398, opened 2015-10-21
General cleanup of ext_pdo

Orvid:
The biggest thing this does is make it so that ext_pdo_sqlite and ext_pdo_mysql aren't used in ext_pdo. This also eliminates the need to declare the individual PDO drivers in pdo_driver.cpp, and instead allows them to be declared in the respective extension, ext_pdo_sqlite / ext_pdo_mysql.

This also eliminates all the global constants in C++ that are not actually needed, and removes the global StaticStrings that were only used once.

This was originally #6226 (D46671) but was mistakenly closed while waiting on the base PR to be merged. Due to Github being grumpy, it couldn't be re-opened, so this new PR had to be created instead.


Orvid:
Replaced internally with D2870465.

PR #6420, opened 2015-10-24
Hhvm 3.9

hariharanravi:


paulbiss:
I don't think you meant to do that

PR #6424, opened 2015-10-25
[MSVC] Synchronize forward declarations by changing everything to structs

Orvid:
DO NOT MERGE - This is nowhere near close to being able to work for MSVC, as there are still many places that have to be changed.

This PR is the start of the alternate method suggested for #5693 (D42015), created to illustrate the difference in the scale of the changes required for the approach.

This was done with a regex and a manual review, and excludes a large number of files where invalid changes where made.


Orvid:
Preferring to go with the synchronization PR instead. (probably)

PR #6441, opened 2015-10-30
Fix compile error for CompactTaggedPtr

zlim:
Drive-by cleanup. This path is not active on x64 builds, which is probably why it was never caught.


zlim:
I've signed CLA. Let me know if there's anything else required.

PR #6456, opened 2015-10-31
Perform dirname at compile time when possible.

gdbentley:
If the argument to dirname() is a string literal (or equivalent), we can call dirname() at compile time and just return the precomputed value at runtime. This is already done for ini_get().

This makes
 dirname(__FILE__) . "/some_relative_file"
as efficient as using a hardcoded absolute path.

I don't see a measurable performance improvement with WordPress, but I don't see a downside to this optimization.


gdbentley:
The underscores got changed to bold in my previous comment. :(

gdbentley:
After measuring again, I do see a slight (~.4%) improvement on WordPress.

PR #6462, opened 2015-10-31
[Const Macros] Switch ext_zip to use the new const registering macros

Orvid:
This was split out of #6365 (D48705) to make reviewing easier.


hphp/runtime/ext/zip/ext_zip.cpp
```diff
@@ -272,116 +272,6 @@ class ZipDirectory: public SweepableResourceData {
 IMPLEMENT_RESOURCE_ALLOCATION(ZipDirectory);
 
 const StaticString s_ZipArchive("ZipArchive");
-const StaticString s_CREATE("CREATE");
-const int64_t k_CREATE = 1;
```
sgolemon: Such hard-coding, many bad, wow.

Thank you for this diff!

PR #6469, opened 2015-11-01
[Const Macros] Switch ext_std to use the new const registering macros

Orvid:
This was split out of #6365 (D48705) to make reviewing easier.

It is worth pointing out that I also moved `k_INF` and `k_NAN` from `ext_std_misc` to `ext_std_math`, which is where I believe it should be, and where it also means we have a macro for registering doubles.
The move of those constants has the potential to be a breaking change for some of FB's internal code. I fixed what I found in what I can build under MSVC, but there might be other places as well.


Orvid:
Moved internally to D2870023.

PR #6516, opened 2015-11-06
Enable NUMA if libnuma.so is found

sathvikl:


sathvikl:
https://github.com/hhvm/hhvm-third-party/pull/96  
This PR is tagged to https://github.com/hhvm/hhvm-third-party/pull/96. 
Both commits are required for changes to take effect.

gdbentley:
hhvm/hhvm-third-party#96 is no longer required.

Issue #6500, opened 2015-11-02
stream_await not working on sockets?


roba269:
Hi, not sure if I understand the usage of stream_await correctly, but when I was trying to send multiple requests to a simple echo server, the program will always be stuck on waiting for the read to be ready, and sometimes it crashes.

Here is the client code:

```
<?hh

async function writeAsync($r, $buf) {
        while (strlen($buf) > 0) {
                echo "before stream_await in writeAsync\n";
                $status = await stream_await($r, STREAM_AWAIT_WRITE, 2);
                echo "after stream_await in writeAsync\n";
                if ($status === STREAM_AWAIT_READY) {
                        $written = stream_socket_sendto($r, $buf);
                        if ($written === -1 || $written === false) {
                                throw new Exception('write error');
                        }
                        $buf = substr($buf, $written);
                } else {
                        throw new Exception('write error');
                }
        }
}

async function readAsync($r) {
        $data = '';
        while (!feof($r)) {
                echo "before stream_await in readAsync\n";
                $status = await stream_await($r, STREAM_AWAIT_READ, 2);
                echo "after stream_await in readAsync\n";
                if ($status === STREAM_AWAIT_READY) {
                        $data .= stream_socket_recvfrom($r, 100);
                } else {
                        throw new Exception('read error');
                }
        }
        return $data;
}

async function doit($content) {
        // $errno = 0; $errstr = '';
        // $r = fsockopen('192.168.33.10', 9034, $errno, $errstr, 10.0);
        $r = stream_socket_client("tcp://192.168.33.10:9034");
        await writeAsync($r, $content);
        $data = await readAsync($r);
        if ($content != $data) {
                echo "error! data unmatch! send_data: $content recv_data: $data\n";
        } else {
                echo $data . "\n";
        }
        fclose($r);
}

$ITER_COUNT = 1000;
for ($i = 0 ; $i < $ITER_COUNT ; ++$i) {
        $handle = doit('testinfo' . $i);
        $handle->join();
}
```

And a typical run result:

```
before stream_await in writeAsync
after stream_await in writeAsync
before stream_await in readAsync
after stream_await in readAsync
testinfo0
before stream_await in writeAsync
after stream_await in writeAsync
before stream_await in readAsync
after stream_await in readAsync
testinfo1
before stream_await in writeAsync
after stream_await in writeAsync
before stream_await in readAsync
after stream_await in readAsync
testinfo2
before stream_await in writeAsync
after stream_await in writeAsync
[...]

Issue #6289, opened 2015-09-22
Audit Asio extensions for usage of runInEventBaseThread()


fredemmott:
Some of these may be correct, but they may also be race conditions that cause crashes. Example crash fixes:

https://github.com/facebook/hhvm/commit/e03f6979122318c63378ad7162aa83569539abc0
https://github.com/facebook/hhvm/commit/484154f619666abf92b609b7ece16157cee65eae

fredemmott:
Did this.

PR #6518, opened 2015-11-06
fix race condition in async stream

roba269:
Summary: See issue #6500
The root cause and the solution looks similar as the fixing of async curl, see https://github.com/facebook/hhvm/commit/e03f6979122318c63378ad7162aa83569539abc0

### DIFF
--- a/hphp/runtime/base/file-await.cpp
+++ b/hphp/runtime/base/file-await.cpp
@@ -33,7 +33,7 @@ FileAwait::FileAwait(int fd, uint16_t events, double timeout) {
   if (timeout_ms > 0) {
     m_timeout = std::make_shared<FileTimeoutHandler>(asio_event_base.get(),
                                                      this);
-    asio_event_base->runInEventBaseThread([this,timeout_ms] {
+    asio_event_base->runInEventBaseThreadAndWait([this,timeout_ms] {
       if (m_timeout) {
         m_timeout->scheduleTimeout(timeout_ms);
       }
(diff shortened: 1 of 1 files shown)


SiebelsTim:
ref #6289

JoelMarcey:
@roba269 Hi! Did you actually sign the CLA? Internally we don't show you has having signed it even though the bot thinks you did.

roba269:
I'm not sure... I checked this page https://github.com/facebook/hhvm/blob/master/CONTRIBUTING.md, but when I'm trying to click "I agree" on this link https://code.facebook.com/cla now, I got a message saying github and email have been registered. I can send the signed pdf to cla@fb.com later.

JoelMarcey:
@roba269 I think the problem is that internally we have you with a former email address and you are now using a different one maybe. I think we need to sync the two some how.

JoelMarcey:
@roba269 Yeah, if you can send the signed pdf to cla@fb.com that would help a lot too. Just so we can be unblocked.

roba269:
Yeah.. I have two emails and maybe somehow I messed up. 

Just sent the singed pdf to cla@fb.com. Let me know if there are still problems. Thanks!

dneiter:
This fix is incorrect, since FileAwait destructor may run from inside AsioEventBase thread (e.g. if php request fatals).

When run from event base thread, runInEventBaseThreadAndWait prints an error message and returns false (and callback never actually runs at all): https://github.com/facebook/folly/blob/master/folly/io/async/EventBase.cpp#L613

PR #6522, opened 2015-11-07
Refactor out spots that directly divide by zero

Orvid:
MSVC produces a warning complaining about the "potential" divide by 0. This may also not be fully portable to other architectures, so only allow it under x86_64 for now.


Orvid:
Moved internally to D2901176.

PR #6523, opened 2015-11-08
Hhvm 3.9

davefreemanHH:
just check the amazin features......


paulbiss:
I don't think this is what you meant to do

PR #6538, opened 2015-11-11
Generate a global config file

Orvid:
There are multiple places in HHVM's source tree where a pre-generated config file is used, and those config files are not always correct for the current platform.

This generates a global config file that will be used in place of the baked versions.
This also switches `ext_fileinfo` to use the global config file rather than a local one. It's worth noting that a large number of the defines in the original config file are no longer generated, because they were never referenced to begin with.
This also swaps the `ENABLE_AVX2`, `ENABLE_ZEND_COMPAT`, and `EXECUTION_PROFILER` defines to be defined only in `hphp/hphp-config.h`, rather than globally.
This also eliminates the `DEBUG_MEMORY_LEAK`, `DEBUG_APC_LEAK`, and `ENABLE_FULL_SETLINE` options, as they did nothing.
This also adds the include of `hphp/hphp-config.h` in any file that either uses one of the 3 defines that were moved, or else use an `ENABLE_EXTENSION_*` define, although the code to generate the `ENABLE_EXTENSION_*` defines in `hphp/hphp-config.h` is not part of this PR.

There are more places in HHVM that used a baked config file that can be converted to use this instead, but this is just the start.

This was originally part of #5872 (D43749), but is a large enough change to deserve its own PR.

Oh, and one last thing I forgot about, because of the way the generation of the config is setup, it should be possible for FB's internal build system to simply call the `HHVM_GENERATE_CONFIG` function via CMake and have it generate the config appropriately. That technique will require a bit of adjustment when the extension defines get added to the config file, but the changes required for that should be minimal.


Orvid:
Replaced internally with D2874920.

PR #6540, opened 2015-11-11
[MSVC] A pair of minor fixes for MSVC

Orvid:
The first is a shift that was causing `RequestMemoryMaxBytes` to be set to `0`, although I have no idea why that never caused an issue.
The second is to only include `monetary.h`, which declares `strfmon`, if not under MSVC, because MSVC will be using a portability implementation.
The final change is to suppress a warning that MSVC generates about comparing against the address of a string literal.


Orvid:
Moved internally as D2845171.

PR #6543, opened 2015-11-11
Fix a memory leak in SSL sockets

### LINKED ISSUE #6537: Memory leak when using SSL sockets
We use persistent SSL sockets on almost every page request. With HHVM, this seems to cause memory usage to increase over time. If we don't use SSL sockets, but still use persistent sockets, the memory leak seems to go away. It's possible my pull request #6329 caused this or perhaps it has always been an issue.

The code bug that's causing this was not immediately obvious to me. I did notice in a small test case (test/slow/ext_socket/persistent_socket.php) that the number of calls to the constructor and destructor of SSLSocket and SSLSocketData is as expected.

The leak is easily noticeable for us. It takes about 2 hours to burn through 48GB of RAM with a request rate of around 100 requests/sec.

apaloma:
Resolves #6537. The SSL_CTX object was not being freed.

### DIFF
--- a/hphp/runtime/base/ssl-socket.cpp
+++ b/hphp/runtime/base/ssl-socket.cpp
@@ -33,7 +33,9 @@ bool SSLSocketData::closeImpl() {
   }
   if (m_handle) {
     SSL_free(m_handle);
+    SSL_CTX_free(m_ctx);
     m_handle = nullptr;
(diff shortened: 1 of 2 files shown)


apaloma:
I'm going to resubmit per our discussion on Phabricator.

PR #6544, opened 2015-11-11
Fix persistent sockets not persisting across page loads

### LINKED ISSUE #6542: Persistent sockets aren't very persistent
Persistent sockets will work when reused in the same page request. However after a page request completes, it appears the persisted socket is closed, and thus a new page request on the same thread will be unable to use the persisted socket and a new connection will be created.

It appears this is sweep's fault. I think File and Socket's implementation of sweep is doing too much. They'll call closeImpl on the data object which will close the file/socket, but that will also happen in the data object's destructor, so I think we can remove that extra call to closeImpl and just rely on that happening when the data object's refcount hits zero.

apaloma:
Sockets were always being closed after a page request. You wouldn't be able to reuse the socket on a new page request. Resolves #6542.

### DIFF
--- a/hphp/runtime/base/file.cpp
+++ b/hphp/runtime/base/file.cpp
@@ -198,8 +198,11 @@ void File::sweep() {
   // sweep() is responsible for closing m_fd and any other non-request
   // resources it might have allocated.
   assert(!valid());
-  File::closeImpl();
+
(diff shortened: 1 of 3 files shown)


apaloma:
I'm going to resubmit per our discussion on Phabricator.

PR #6546, opened 2015-11-12
Define 'zero' vasm instruction to clear register

gut:
Summary: Some architectures might not use xor as an effective way of
zeroing a register. On top of that it forbids xor from being lowered as
the Vxls::insertCopiesAt is called after lowering step. It'd be good to
have a specific vasm instruction for doing that.

This approach will give the flexibility needed for other architectures
that might want to zero differently or lower the xor vasm instructions.


gut:
Will have to update this to match d477d09b799ecd978ae42a34b47173407c363a56 changes

PR #6552, opened 2015-11-15
[AArch64] Fixes compilation issues.

dave-estes:
Adds missing translator helper placeholder labels for AArch64. Updates
some inline assembly for AArch64 which is missing the pause mnemonic.
Enables compact tagged pointers for AArch64. Enables FASTCGI.

Does not fix any correctness issues. They will be addressed in
subsequent commits.


JoelMarcey:

Thanks.

PR #6559, opened 2015-11-17
Eliminate php_compat.h in ext_gd

Orvid:
This was needlessly duplicating functionality present in ext_zend_compat, so change the places that use it to use the HHVM equivalent instead.

Using the HHVM equivalents is both cleaner and shorter.


Orvid:
Replaced internally with D2871289.

PR #6564, opened 2015-11-17
[AArch64] General correctness fixes

dave-estes:
These commits address general correctness issues for AArch64. Most notably, the native func caller interface has been adapted to explicitly pass a return parameter for return-by-reference calls. This affects x86_64 for when the return value is a non-POD, but code generated should be identical.


JoelMarcey:
@dave-estes Hi Dave. It says you haven't signed the CLA. If that's the case, can you sign it before we pull this in?

dave-estes:
Sure thing, @JoelMarcey. We should have a corporate CLA on file already. I've got an email in to cla@fb.com about it, but I haven't heard back yet.

JoelMarcey:

Thanks.

PR #6572, opened 2015-11-19
Remove duplicate defines of HAVE_LIBJPEG and HAVE_LIBPNG.

wjywbs:
They are already defined in HHVMExtensionConfig.cmake.
https://github.com/facebook/hhvm/blob/bbf1a3f443b385af20764cbbd9390605fb9de079/CMake/HHVMExtensionConfig.cmake#L811

### DIFF
--- a/hphp/runtime/ext/gd/libgd/php_compat.h
+++ b/hphp/runtime/ext/gd/libgd/php_compat.h
@@ -34,8 +34,6 @@
 // And start the blasted C stuff again
 extern "C" {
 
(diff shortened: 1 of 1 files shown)


Orvid:
Incidentally, these were probably why the libJpeg and libPng dependencies were originally marked as required, because they were being forced through this. (Also, this is adjusting `HAVE_LIBJPEG` and `HAVE_LIBPNG`, not `HAVE_GD_JPG`.

PR #6574, opened 2015-11-19
[Aarch64] Relaxes FP Precision of Results

dave-estes:
Relaxes the expected results for some floating point tests. Switches key double-to-int casts to use the helper routine.


JoelMarcey:

Thanks.

PR #6575, opened 2015-11-19
Refines a few unit tests to for Ubuntu Wily Werewolf

dave-estes:
Checks for presence of test data file for GD and uses a different timezone to avoid failures from different handling of leap seconds.


JoelMarcey:

Thanks.

PR #6576, opened 2015-11-19
[AArch64] Passes status variable explicitly to ::waitpid.

dave-estes:
The documentation for ::waitpid states that status may be a null
pointer, however libc under wily for AArch64 dereferences status
regardless. This patch is a safety net until changes to libc
are upstreamed.

Fixes test/slow/ext_process/lwp.php


JoelMarcey:

Thanks.

PR #6607, opened 2015-11-26
Remove outdated and unsupported MongoDB driver

derickr:
This old version of the driver has many many issued, and has not been updated
for several years. This legacy driver is also no longer supported by MongoDB,
and has been replaced by a new driver, for which there is a specific HHVM
alternative implementation: https://github.com/mongodb/mongo-hhvm-driver


JoelMarcey:
@derickr Yes, this should be done. Thanks. But we can't do it right now, with this particular pull request. We don't want to just break people who might be using it without some sort of deprecation announcement.

So how about a pull request that disables Mongo by default in master. In the next one or two HHVM releases, we don't have it enabled; if someone does enable it, a deprecation notice is shown. And then we resurrect this pull request?

I am going to close this pull request to get it out of our queue, wait for the new deprecation pull request, accept that, and then you can re-open this one after the deprecation period is over.

If you feel that is unreasonable, let me know why and we can discuss.

derickr:
Done at https://github.com/facebook/hhvm/pull/6644

PR #5240, opened 2015-05-02
Fixed 'last day of this month' issue with DateTime

SuRaMoN:
Fix for issue https://github.com/facebook/hhvm/issues/4681
$a=new DateTime('now'); $a->modify('last day of this month') didnt work. should be fixed now

PR #6637, opened 2015-12-06
Fix 'last day of this month' issue with DateTime for 3.6

nicholas-eden:
Applying patch from pull request https://github.com/facebook/hhvm/pull/5240


fredemmott:
FYI, given that 3.6 has not had a release since August and becomes unsupported on January 28th, it currently seems unlikely that there will be another release.

PR #6657, opened 2015-12-13
Remove warning on srand and mt_srand functions if first parameter is not a long

noony:


fredemmott:
Why? This introduces a silent failure.

noony:
To fit with php behavior.

SiebelsTim:
for which version?
https://3v4l.org/njrWv

SiebelsTim:
The warning itself is valid for most cases. 
The only case where we warn and php doesn't is when passing `true` or `false`.

fredemmott:
Not taking this: inaccurate, and for true/false, the incompatibility seems worth the warning.

PR #6664, opened 2015-12-15
Fix type error using stream_set_blocking.

yukitaka:
Change the second argument of stream_set_blocking to int.

### DIFF
--- a/hphp/system/php/redis/Redis.php
+++ b/hphp/system/php/redis/Redis.php
@@ -1650,7 +1650,7 @@ class Redis {
         E_WARNING);
       return false;
     }
-    stream_set_blocking($this->connection, true);
+    stream_set_blocking($this->connection, 1);
     $this->setOption(Redis::OPT_READ_TIMEOUT, $timeout);
 
     return true;


yukitaka:
All right.
I agreed to Contributor License.

hphp/system/php/redis/Redis.php
```diff
@@ -1650,7 +1650,7 @@ protected function doConnect($host,
         E_WARNING);
       return false;
     }
-    stream_set_blocking($this->connection, true);
+    stream_set_blocking($this->connection, 1);
```
steelbrain: http://php.net/manual/en/function.stream-set-blocking.php <-- docs say it should be bool

yukitaka:
hhvm/hphp/runtime/ext/stream/ext_stream.php <-- Implementation is int

It seems to have different defined by the doc language.

http://php.net/manual/ja/function.stream-set-blocking.php
http://php.net/manual/fr/function.stream-set-blocking.php

php7.0.1
ext/standard/streamsfuncs.c <-- proto comment is int

It is preferable to fix either ?

SiebelsTim:
@yukitaka 
6 days ago: https://github.com/php/php-src/commit/4171a8f6c631e3c34d1d070d553e7611b5474263

---

`int` doesn't make much sense either. So I would suggest sticking with `bool`.

yukitaka:
I prefer bool too.
It looks good to add a bool for backward compatibility .

JohanTan:
Any updates on when this is merging? Redis extension is currently broken on HHVM 3.12 now.

SiebelsTim:
This PR is abandoned. How is redis broken by this?

JohanTan:
Got the following fatal error once `hhvm.php7.all = 1` is enabled:
```Fatal error: Uncaught TypeError: Argument 2 passed to stream_set_blocking() must be an instance of int, bool given```

PR #6675, opened 2015-12-17
Expose HHVM_API_VERSION to hhvm --version

paravoid:
Expose the constant HHVM_API_VERSION to hhvm --version. We've been using
that in the Debian packages to create Debian metadata for HHVM extension
packages that may rely on the API/ABI.


fredemmott:
This constant is useless as it's never updated:

```
// Deprecated: Use HHVM_VERSION_BRANCH for source compat checks
#define HHVM_API_VERSION 20150212L
```

Issue #6685, opened 2015-12-18
3.11.1 Cherry-pick list


fredemmott:
- [x] fix path handling/fastcgi crashes/404s/500s (#6678)
 - [x] fix hh_client crashes (#6671)
 - [x] `hh_server --check` when already running (#6646)

cc @jwatzman @sgolemon @JoelMarcey

fredemmott:
Believe the 404 issues are fixed, waiting for confirmation in #6648

fredemmott:
last one in: #6774

Orvid:
This already went out.

PR #6691, opened 2015-12-22
Redo eb4a0cd without its transitive dependencies

alexchow:
check mode runs typechecker in main process instead of daemon


alexchow:
@int3 @fredemmott  For #6685

PR #6713, opened 2016-01-03
Add missing hhi definitions for the DOM

### LINKED ISSUE #5322: Unbound name: DOMDocument (an object type) (Naming[2049])
``` hack
<?hh // strict
class Example
{
    public function test(): void
    {
        $dom = new \DOMDocument('1.0', 'UTF-8');
    }
}
```

I use DOMDocument/DOMElement/DOMNode classes without problems in my project (Hack in strict mode).
Except that the typechecker says:
`Unbound name: DOMDocument (an object type) (Naming[2049])`
So I add `// UNSAFE`, but that's not beautiful.

```
HipHop VM 3.7.0 (rel)
Compiler: 1430225359_311382370
Repo schema: bab2fb342e0e21b096813ebb19860646b8eaa0d2
```

lapetr:
( based on http://docs.hhvm.com/ 2015-08-10 )
Fixes #5322

### DIFF
--- a/hphp/hack/hhi/stdlib/builtins_dom.hhi
+++ b/hphp/hack/hhi/stdlib/builtins_dom.hhi
@@ -1,4 +1,4 @@
-<?hh     /* -*- php -*- */
+<?hh // decl /* -*- php -*- */
 /**
  * Copyright (c) 2014, Facebook, Inc.
  * All rights reserved.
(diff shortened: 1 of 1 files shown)


aorenste:
Unfortunately this had to be reverted (https://github.com/facebook/hhvm/commit/abca8a) due to an internal conflict.

mofarrell:
Worth noting that the now committed code is not as thoroughly annotated as the originally submitted PR.

Issue #6632, opened 2015-12-04
timezone_open throws an exception rather than returning false


nicholas-eden:
This is a rather obscure one, but reporting it so it is noted.  The timezone_open method returns false in php but throws an exception in hhvm.

```php
php > var_dump(timezone_open('sdf'));
bool(false)
```

```php
hphpd> var_dump(timezone_open('sdf'));
Hit a php exception : exception 'Exception' with message 'DateTimeZone::__construct(): Unknown or bad timezone (sdf)' in :1
```

jurchiks:
Actually, timezone_open('invalid argument') should also output a warning, according to this: https://github.com/php/php-src/blob/master/ext/date/tests/timezone_open_variation1.phpt#L112

PR #6729, opened 2016-01-07
Fixing #6632: timezone_open() should return false on failure

jurchiks:


jurchiks:
It's a bit of a messy fix, but it's certainly better than digging into the DateTimeZone class.

hphp/runtime/ext/datetime/ext_datetime.php
```diff
@@ -876,8 +875,21 @@ function timezone_name_get(DateTimeZone $timezone): string {
...
+    return new DateTimeZone($timezone);
+  }
+  catch (Exception $e) {
+    trigger_error(
+      str_replace(
+        'DateTimeZone::__construct():',
+        __FUNCTION__ . '():',
+        $e->getMessage()
+      ),
+      E_WARNING
```
jurchiks: By the way, trigger_error() does not allow anything except E_USER_* constants for the second parameter, but I noticed that in your code you have plenty of calls that use E_WARNING; is there some kind of exception for these particular files?

fredemmott: Yep, builtins are allowed to use any error constant, even if they're written in PHP or Hack: https://github.com/facebook/hhvm/blob/master/hphp/runtime/ext/std/ext_std_errorfunc.cpp#L291-L312

We needed this to maintain PHP-compatibility, while still allowing parts of the standard library to be implemented in PHP or Hack.

jurchiks: Sounds logical, otherwise I'd have a real pain in the ass to deal with.
Does E_STRICT have no use cases?

Orvid:
Fixed via #7092.

PR #6765, opened 2016-01-19
Refactor unwind translator to be arch independent.

racardoso:
Summary:
Unwind translator have separated files for each architecture like unwind-x64.* . This file are called by debugger, unique stubs, and others arch independent (backend) code. Once these files call's unwind-x64 which is specific for x64 it's necessary to insert #ifdef in order to select the right unwind code when port hhvm for others architectures. This commit creates a arch independent unwind that can be called by arch independent code in the same way that was done with unique-stubs avoiding the use of #ifdef's.


JoelMarcey:
This was abandoned.

Issue #6731, opened 2016-01-08
Readline() has a required parameter but it should be a optional parameter


mbutkereit:
function readline(string $prompt): mixed;

$prompt is in hhvm required and in php optional.

https://github.com/facebook/hhvm/blob/master/hphp/runtime/ext/readline/ext_readline.php 
http://php.net/manual/en/function.readline.php

noony:
Can be closed by https://github.com/facebook/hhvm/commit/2ba88997d3b70bc6e872749f243bd299ebd1c888

PR #6766, opened 2016-01-19
first argument of readline function is optional

noony:
Close: #6731


SiebelsTim:
`String`s can be null, can't they?

jurchiks:
@SiebelsTim not in PHP; scalar types are not objects in PHP, not in the traditional sense anyway.

SiebelsTim:
I was talking about the HNI type String. 
I didn't find confirmation though, so Variant it is I guess.

PR #6767, opened 2016-01-19
Fix another segfault in strtr

adrianheine:
The Wu-Manber implementation in `strtr` used `qsort_r` on a collection of
`std::string`s. However, the C++ standard clearly says that `qsort`'s behavior
is undefined with non-trivial collection elements. `std::string` is not
trivial. While `qsort_r` is not part of the C++ standard, it's behavior should
not be expected to be defined with non-trivial types either.

See http://en.cppreference.com/w/cpp/algorithm/qsort

This change replaces `qsort_r` usage with `std::sort`.


adrianheine:
Downstream Wikimedia bug: https://phabricator.wikimedia.org/T124276

PR #6779, opened 2016-01-22
dns_get_record must return by default an empty array not null

noony:


sgolemon:
Sorry, I had to revert this due to a test failure that snuck through CI.
Should be able to re-land it soon though.

noony:
Ok no problem @sgolemon something doesn't have same comportement than php in my pr ?

sgolemon:
FB-internal code that makes bad assumptions about the return value of dns_get_record().  It's our bug and I'mma fix it.  I just don't want a bunch of web-devs getting sort at me while it's broken. :D

JoelMarcey:
I am going to close this since we have the needed internal fix in and I have a diff up to unrevert the revert.

D2938499

PR #6797, opened 2016-02-01
Fix emitDecRefGeneric stub register saving

gut:
As it's called on a callfaststub context, it cannot allocate any Vreg.
The coded vasm sticks with this constraint but unfortunately it doesn't
know about the vasm lowering phase, and PPC64 requires additional
registers to lower a cmplim so it violates that constraint.

Before vasm lowering phase:
```
B0      main
     stublogue
     cmplim     1, +0xc(r3) => %64
     jcc        E, %64, B3, else B2

B3      main, B0
     push       r3        /* begin of PhysRegSaver */
     push       r4
     ...
```

After vasm lowering phase and register allocation:
```
B0      main
     stublogue
     loadl      +0xc(r3) => r14      /* r14 is a temp register */
     movl       r14 => r14           /* hhvm allocated a caleeSaved reg */
     extsw      r14 => r14
     ldimmw     1 => r15             /* r15 is also a temp register */
     cmpq       r15, r14 => cr0
     jcc        E, cr0, B3, else B2

B3      main, B0
     push       r3        /* begin of PhysRegSaver */
     push       r4
     ...
```

The stub was already intending to save registers, but it was wrongly scoped.

After change, before vasm lowering phase:
```
 B0      main
      stublogue
      push       r3        /* begin of PhysRegSaver */
      push       r4
      ...
      push       r9
      push       r10       /* end of PhysRegSaver */
      cmplim     1, +0xc(r3) => %64
      jcc        E, %64, B3, else B2
```

After change, after vasm lowering phase and register allocation:
```
 B0      main
      stublogue
      push       r3        /* begin of PhysRegSaver */
      push       r4
      ...
      push       r9
      push       r10       /* end of PhysRegSaver */
      loadl      +0xc(r3) => r5        /* r5 is free, it used this one */
      movl       r5 => r5
      extsw      r5 => r5
      ldimmw     1 => r6               /* r6 is also ok to use here */
      cmpq       r6, r5 => cr0
      jcc        E, cr0, B3, else B2
```

This bug was not affecting X64.

### DIFF
--- a/hphp/runtime/vm/jit/unique-stubs.cpp
+++ b/hphp/runtime/vm/jit/unique-stubs.cpp
@@ -768,10 +768,7 @@ TCA emitDecRefGeneric(CodeBlock& cb) {
   return vwrap(cb, [] (Vout& v) {
     v << stublogue{};
 
-    auto const rdata = rarg(0);
-    auto const rtype = rarg(1);
-
-    auto const destroy = [&] (Vout& v) {
+    { /* begin of PhysRegSaver scope */
       // decRefGeneric is called via callfaststub, whose ABI claims that all
       // registers are preserved.  This is true in the fast path, but in the
       // slow path we need to manually save caller-saved registers.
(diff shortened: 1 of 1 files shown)


gut:
As discussed on Phabricator, this will not be the final solution. I'll change the `cmplim{}` vasm lowering phase specifically for PPC64 as a solution. 

The solution will not be sent on this pull request.

PR #6833, opened 2016-02-16
Refactor DecodedInstruction class for arch-specific implementations

gut:
- Moved the code from asm-x64* to decoded-instr-x64*
- Included the 'decoded-instr.h' header on all cpp files that needs it.
- Created 'decoded-instr.h' file that alias the
  HPHP::jit::DecodedInstruction type to an arch specific
  HPHP::jit::(arch)::DecodedInstruction type.

Now each architecture can implement its DecodedInstruction on it's on
file and simply include it on decoded-instr.h for an alias.

Regarding PPC64, the decoded-instr-ppc64.cpp will be sent afterwards
along with its disassembler.


gut:
a new pull request will be done soon with a different solution. See comments on https://reviews.facebook.net/D54207 for more info

PR #6837, opened 2016-02-18
Add -gsplit-dwarf support to CMake build system

tstarling:
Linking hhvm was causing an OOM on my 8GB laptop unless I closed all
open applications. So it's time for -gsplit-dwarf, a.k.a. Debug
Fission, which avoids the link-time overhead of debug symbols by
creating a .dwo file at compile time, which is then ignored by the
linker. This reduces both time and memory usage of the linker
substantially. This requires GCC 4.7+.

A future extension would be to package the .dwo files into a .dwp file
and install it. This should be optional, since packaging forgoes most of
the link time improvement, copying gigabytes of debug symbols. For
development, it's most efficient to just leave the .dwo files in the
build tree.

### DIFF
--- a/CMake/HPHPCompiler.cmake
+++ b/CMake/HPHPCompiler.cmake
@@ -182,18 +182,24 @@ if (${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang" OR ${CMAKE_CXX_COMPILER_ID} STREQU
     endif()
   endif()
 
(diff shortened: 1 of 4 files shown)


Orvid:
Blarg, left it too long after the rebase before merging, now it needs to be rebased again :(

tstarling:
Rebased.

PR #6845, opened 2016-02-19
Fix DatePeriod::constuctor arguments

SiebelsTim:
partially... DatePeriod also allows creating from iso string, which we don't
do currently.

However with the current implementation, passing null to the interval
is always wrong, because it tries to clone the object we passed.


Orvid:
Ping. A rebase and not touching third-party should be enough for this.

SiebelsTim:
Sorry, forgot this.

fredemmott:
^ don't expect 'details' links to work for a day or two

PR #6868, opened 2016-02-29
Fixed arm64 abi and alignment issues

lpathy:


lpathy:
Will resubmit. Closing it.

PR #6870, opened 2016-02-29
Fixed smashable instructions for arm64

lpathy:


lpathy:
Will resubmit. Closing it.

PR #6872, opened 2016-02-29
Fixed decode issues in arm64 cases

lpathy:
1.  Added arm64 jcc decoding in mc-generator.cpp
2.  Added arm64 framePtr save in target-cache.cpp
3.  Fixed stub_size and extract_spoff for arm64 in service-requests.cpp
4.  Fixed emitPredictionsAndPreConditions for arm64 in translate-region.cpp


lpathy:
Will resubmit. Closing it.

PR #6873, opened 2016-02-29
Added missing vasm to arm64 routines

lpathy:
Note:
  Reused code at https://github.com/edwinsmith/hhvm-armjit partially.

1.  Removed unnecessary arm specific vasm codes (hostcall,cbcc,tbcc etc.)
2.  Fixed/Added remaining vasm to arm64 conversion routines
3.  SIMD is pending


lpathy:
Will resubmit. Closing it.

PR #6874, opened 2016-02-29
Avoided relocation for non-x64 platforms and added 16' physreg

lpathy:


lpathy:
Changes in this PR are already in master. Closing it.

PR #6888, opened 2016-03-05
Fixed Datetime comparison with milliseconds involved

wjzijderveld:
I found this bug earlier in PHP and [fixed it there](https://github.com/php/php-src/pull/1653). After that I checked if it was also in HHVM. It was, so I also tried to fix it here :smile: 

Depends on an updated version of [hhvm-third-party](https://github.com/hhvm/hhvm-third-party/pull/102).


mwjames:
Does this accounts for the discrepancy [0] we encounter in our tests with HHVM vs PHP?

```
<?php

$dateTime = new \DateTime( '1300-11-02 12:03:25.8884999' );
echo( $dateTime->format( 'Y-m-d H:i:s.u' ) );
```

Output for hhvm-3.6.1 - 3.12.0
1300-11-02 12:03:25.888499

Output for 5.5.0 - 7.0.4
1300-11-02 12:03:25.888500


https://3v4l.org/s3eRL

wjzijderveld:
Not that I'm aware of, as this only impacts comparing objects ($datetime1 > $datetime2 etc).

mwjames:
> Not that I'm aware of,

Thanks for the quick response. This now tracked with https://github.com/facebook/hhvm/issues/6899.

Orvid:
Could you rebase this?

Orvid:
Alright, the timelib changes have landed, could you rebase this so that the test infra isn't grumpy about the base being so old?

wjzijderveld:
It's rebased, but haven't been able to run the tests myself yet, as I'm unable to actually build HHVM at the moment, it keeps failing here:

```
+ /usr/bin/ocamlc.opt -c -g -w A -warn-error A -w -3-4-6-29-35-44-48-50-52 -I server -I procs -I socket -I emitter -I typing -I find -I globals -I format -I stubs -I heap -I hhi -I deps -I decl -I parsing -I dfind -I fsnotify_linux -I libancillary -I client -I search -I h2tp -I watchman -I hh_matcher -I naming -I monitor -I utils -I injection/default_injector -I third-party/inotify -I third-party/avl -I third-party/core -I h2tp/common -I h2tp/mapper -I h2tp/unparser -I hh_matcher/test -I utils/hh_json -I utils/collections -o server/serverError.cmo server/serverError.ml
File "server/serverError.ml", line 33, characters 2-24:
Error: Unbound value Hh_json.json_to_output
Command exited with code 2.
Compilation unsuccessful after building 789 targets (788 cached) in 00:00:51.
make[3]: *** [_build/hh_server.native] Error 10
make[2]: *** [hphp/hack/CMakeFiles/hack] Error 2
make[1]: *** [hphp/hack/CMakeFiles/hack.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

So I'm curious what the build system thinks of this PR now :smile:

Orvid:
O_o... We shouldn't be pushing if we have build errors... What version of OCamlC are you building with? (also, you can skip the hack build if you just need to test HHVM; after configuring and generating the makefile `make hhvm` will make just HHVM and not build hack)

wjzijderveld:
I didn't meant any kind of release build, just assumed some test build would be triggered in phabricator :-)

```
$ ocaml -version
> The OCaml toplevel, version 4.01.0
```

`make hhvm` worked.
Quick test has 1 failure (`all_type_comparison_test`) due to a segfault.
Datetime tests do pass (both quick and slow).

Haven't run the full slow test suite yet.

Orvid:
Yep, Phabricator does trigger a test build after we import it internally, but we also have the job that syncs to Github setup to build HHVM and run the quick tests before syncing. If HHVM fails to build or a quick test fails, it doesn't push the latest changes to Github and sends many emails to me until it's fixed.

PR #6889, opened 2016-03-06
Fix the Redis client's scan family.

kmaslack:
The PECL Redis client wraps the *scan family of methods
in a less trivial way than most other methods. I've manually
verified that a Redis test suite agrees across PHP5 and HHVM
with these changes.


kmafb:
Any remaining concerns here? We're in the process of setting up build infra for a longer-than-expected-lived branch just because we can't find this fix in nightlies at present.

Orvid:
Any particular reason for closing this? I had just gotten a clean test run, and was doing a final pass over the changes before accepting :(

Orvid:
ping @kmaslack.

kmaslack:
Gah! Sorry about this. The only reason for closing it was that I suck at github and phabricator at the same time. Let me resurrect it.

PR #6926, opened 2016-03-21
Extend support for linux tools perf for jitted code

sathvikl:
Summary: This patch provides better support for two of the perf tools infra

(a) perf record/report/annotate supported limited profiling of
    jitted code. Function/ DSO view worked however assembly profile view
    of jitted code was not supported.
    With this patch and the latest version of perf tools from
    https://git.kernel.org/cgit/linux/kernel/git/tip/tip.git/
    jitted code profile can be viewed at the assembly level.

    It works by generating /tmp/perf-<pid>.dump for perf to read jitted code.
There's two choices for clock source to be used for profiling
(.1) use arch_timestamp; perf tooling requires the env variable
     JITDUMP_USE_ARCH_TIMESTAMP=1 to be set.
     run perf record
(.2) use MONOTONIC_CLOCK;
     run perf record -k 1/mono

     perf inject --jit -i perf.data -o perf.jitted.data
     will inject jit records into perf.data. This will generate an ELF
     file for every trace recorded in perf-<pid>.dump. "PERF_RECORD_MMAP2"
     entry in inserted into perf.data for each one of these ELF files
     created in /tmp.
     By default they will be in /tmp unless specified via
     env variable JITDUMPDIR.

perf report -i perf.jitted.data will generate the results.

The patch defaults to using arch_timestamp unless
env variable is set to JITDUMP_USE_ARCH_TIMESTAMP=0

(b) It extends support for jitted code for
    perf event "intel_pt//" ;PT processor trace recording.
    perf record -e intel_pt//u -p <pid>
    perf inject --jit -i perf.data -o perf.jitted.data
    perf script --itrace=i0ns -F.. -i perf.jitted.data
    will produce a trace of the workload.

This requires 4.1+ version of kernel and the latest version of perf.
 --itrace option of 0ns is supported only in the latest version.

The env var JITDUMP_USE_ARCH_TIMESTAMP=1 must be set and is used by the
linux perf tools infra.


hphp/runtime/vm/debug/perf-jitdump.cpp
```diff
@@ -0,0 +1,287 @@
...
+  asm volatile("rdtsc" : "=a" (low), "=d" (high));
+  return low | ((uint64_t)high) << 32;
+#else
+  return 0;
+#endif
+}
+
+/*On Win32 and linux CLOCK_MONOTONIC is available */
+#if !defined(CLOCK_MONOTONIC) && !defined(__APPLE__)
+#define CLOCK_MONOTONIC 1
```
Orvid: Why do you need this at all? You should be able to do an `#ifdef CLOCK_MONOTONIC` and just fall back to `CLOCK_REALTIME` if it's not defined.

sathvikl: In third-party/folly/src/folly/portability/Time.h
only on win32, CLOCK_MONOTONIC is defined, so I had to keep the # define.  
Without the #define it will fall back to CLOCK_REALTIME even on Linux 

// These aren't generic implementations, so we can only declare them on
// platforms we support.
#if !FOLLY_HAVE_CLOCK_GETTIME && (defined(__MACH__) || defined(_WIN32))
#define CLOCK_REALTIME 0
// The Windows implementation supports a few other
// clock types as well.
#ifdef _WIN32
# define CLOCK_MONOTONIC 1
# define CLOCK_PROCESS_CPUTIME_ID 2
# define CLOCK_THREAD_CPUTIME_ID 3
#endif

Orvid: It should be getting defined when you include <folly/portability/Time.h> on Linux as well, is it not?

sathvikl: @Orvid On the kernel build I have done it is defined in  
> debian/hdrtmp/usr/src/linux-headers-4.3.0-rc4/include/uapi/linux/time.h 
This time.h has _UAPI_LINUX_TIME_H
It's missing here -
> debian/hdrtmp/usr/src/linux-headers-4.3.0-rc4/include/linux/time.h

when I do a headers install, the system's /usr/include/time.h does not have it.

sathvikl: Hi Orvid,

If it defined on FB infrastructure then I can remove the `#define`  Please
advise and I can update it today.

Thanks,
sathvik

Orvid: Remove the define for now, and it should build. Also, drop the change to third-party from this PR, as it's already been updated.

sathvikl: Done. It will probably take overnight for Phabricator to update, when it does could you please re-start the build/tests.

hphp/runtime/vm/debug/perf-jitdump.cpp
```diff
@@ -0,0 +1,276 @@
...
+   * framework as well, in order to collect Intel PT processor trace 
+   * this must be set to 1. 
+   * If use perf in sampling mode, there's 2 clock sources available
+   * Monotonic clock : perf record -k mono ..
+   *                 : perf record -k 1 ..
+   * ARCH clock: perf record .. 
+   * The policy used here, defaults to ARCH clock unless the env variable 
+   * is explicitly set to 0. 
+   */
+  use_arch_timestamp = 1;
```
Orvid: This is a bool, so assign `true`/`false` to it rather than integers.

hphp/runtime/vm/debug/perf-jitdump.cpp
```diff
@@ -0,0 +1,276 @@
...
+   * framework as well, in order to collect Intel PT processor trace 
+   * this must be set to 1. 
+   * If use perf in sampling mode, there's 2 clock sources available
+   * Monotonic clock : perf record -k mono ..
+   *                 : perf record -k 1 ..
+   * ARCH clock: perf record .. 
+   * The policy used here, defaults to ARCH CPU clock unless 
+   * the env variable is explicitly set to 0. 
+   */
+  use_arch_timestamp = true;
```
sathvikl: @Orvid  updated this and on line 136 below.

Issue #6885, opened 2016-03-04
proc_open() does not accept null in 4th parameter


gfloyd:
With `hhvm.php7.scalar_types = 1`, the following code produces a warning:

```
<?php
$process = proc_open('/bin/false', [], $pipes, null);
```

```
Warning: proc_open() expects parameter 4 to be string, null given
```

The docs list `null` as acceptable here: http://php.net/manual/en/function.proc-open.php
It does not produce a warning in php5 or 7, or with `hhvm.php7.scalar_types = 0`.

Note that this currently breaks composer. :)

I am on the nightly release, and also experienced the same in 3.12.

```
$ hhvm --version
HipHop VM 3.13.0-dev (rel)
Compiler: heads/master-0-ga84bdec097fe5adb40587b8a48142d361e4e5911
Repo schema: d3db0494dfde127827da2eb2b308a3fdce8cda83
```

Orvid:
Should just be a matter of updating the function signature to account for possible null values.

satyriasizz:
Hi, It may be bounded to the issue I've just stumbled:

I have HHVM 12.1 using it with nginx on Ubuntu 15.10.
I'm using laravel for my project, so when I tried to run 
`$ php artisan compile`
I got this:
`
root@Peace-EU:/var/www/peacerevolution.net# php artisan compile


Fatal error: Uncaught exception 'ErrorException' with message 'proc_open() expects parameter 4 to be string, null given' in /var/www/peacerevolution.net/vendor/symfony/console/Application.php:898
Stack trace:
#0 (): Illuminate\Foundation\Bootstrap\HandleExceptions->handleError()
#1 /var/www/peacerevolution.net/vendor/symfony/console/Application.php(898): proc_open()
#2 /var/www/peacerevolution.net/vendor/symfony/console/Application.php(701): Symfony\Component\Console\Application->getSttyColumns()
#3 /var/www/peacerevolution.net/vendor/symfony/console/Application.php(662): Symfony\Component\Console\Application->getTerminalDimensions()
#4 /var/www/peacerevolution.net/vendor/symfony/console/Application.php(595): Symfony\Component\Console\Application->getTerminalWidth()
#5 /var/www/peacerevolution.net/vendor/laravel/framework/src/Illuminate/Foundation/Exceptions/Handler.php(141): Symfony\Component\Console\Application->renderException()
[...]

PR #6929, opened 2016-03-22
Allow null in $cwd parameter of proc_open()

gfloyd:
PHP docs list `null` as acceptable in the `$cwd` parameter of `proc_open()`, but currently a warning is produced when php7 mode is enabled.

http://php.net/manual/en/function.proc-open.php

This should address #6885.


hphp/runtime/ext/std/ext_std_process.php
```diff
@@ -121,7 +121,7 @@ function system(string $command, mixed &$return_var = null): string;
 function proc_open(string $cmd,
                    array $descriptorspec,
                    mixed &$pipes,
-                   string $cwd = "",
+                   string $cwd = null,
```
SiebelsTim: This should be of type `?string` now.

PR #6943, opened 2016-03-28
Pooled curl runtime management

ebernhardson:
Includes a breaking change. Using curl_init_pooled to request
a handle from an unknown pool will now return false rather than an
unpooled handle. This gives the application the opportunity to create
the pool with appropriate settings.

* Adds three new methods, curl_create_pool, curl_destroy_pool,
  and curl_list_pools to allow the rutime configuration of
  connection pools.
* Required adding a ReadWriteMutex on all pool operations due to
  namedPools no longer being immutable.
* Changes curl_init_pooled to return false instead of an unpooled handle
  when the requested pool does not exist
* Adds stats collection for number of uses, number of times empty, and
  us spent waiting for pooled handles per pool.
* Renames a couple CurlHandlePool function arguments and members to
  provide consistent naming of the parameters from the ini and back
  through the code.
* Fix bug where the time spend waiting for per-pool lock was not
  included in the maximum time allowed waiting for an available handle.

### DIFF
--- a/hphp/runtime/ext/curl/curl-pool.cpp
+++ b/hphp/runtime/ext/curl/curl-pool.cpp
@@ -43,12 +43,14 @@ void PooledCurlHandle::resetHandle() {
 /////////////////////////////////////////////////////////////////////////////
 // CurlHandlePool
 
-CurlHandlePool::CurlHandlePool(int poolSize,
-                               int waitTimeout,
-                               int numConnReuses)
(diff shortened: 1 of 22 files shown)


ebernhardson:
The first commit in this PR is separate, in https://github.com/facebook/hhvm/pull/6920. Wasn't sure how to make github keep them separate though.

PR #6981, opened 2016-04-12
Improve strcoll test portability

igorsnunes:
Accordingly to the man page [1] and php documentarion [2], the
strcoll function does not specify the return value for strcoll
when the operands are not equal. It just specifies if the return
value must be greater, or lower than zero in case the operands are
different.

For ppc64, when running the following chunk of code in a cpp
program (i.e. using string.h's strcoll function, like hhvm does),
I get "1" as result:

  setlocale(LC_COLLATE,"C");
  printf("%d\n", strcoll("a","A"));

However on x86 I get "32".

The results are different, but according to [1] and [2] they are
not wrong. The zend test [3] is done in a portable way.

[1] http://linux.die.net/man/3/strcoll
[2] http://php.net/manual/en/function.strcoll.php
[3] hphp/test/zend/good/ext/standard/tests/strings/strcoll.php


leitao:
I acknowledge this patch.  It shouldn't assume what is going to be the absolute number returned by `strcoll()`, but just if the return value is a negative number, zero or a positive number.

Other than that, It is important to say that `strcoll()` and `strcmp()` is consistency on POWER, as using the code below:

```
int main () {
	setlocale(LC_COLLATE,"C");
	printf("%d\n", strcoll("a","A"));
	printf("%d\n", strcmp("a","A"));
	return 0;
}
```
and the result is:
```
1
1
```

Meanwhile, the same test shows different results on x86, as:

```
32
1
```

Anyway, this is not wrong also, according to the documentation.

PR #7006, opened 2016-04-20
Fix for copying empty file

nazar-pc:
Basically, when we've copied 0 bytes - let's make sure that there was actually something to copy.
Backport of #6972.


Orvid:
Unless you actually have a reason to need this backported to an LTS release, then we don't need this PR.

nazar-pc:
I my project I have a bunch of hacks `copy() + file_exists()`, which I'd like to replace with now fixed `copy()` soon. However, this will mean that current LTS versions will not be supported, while this is the only change needed to make it working.
So backport is far from critical and I'll understand if this is not strong enough reason for you.

Orvid:
As it's not critical, we won't backport it (we usually only backport security issues)

PR #7007, opened 2016-04-20
Fix for copying empty file

nazar-pc:
Basically, when we've copied 0 bytes - let's make sure that there was actually something to copy.
Backport of #6972.


Orvid:
Unless you actually have a reason to need this backported to an LTS release, then we don't need this PR.

PR #7011, opened 2016-04-21
Fix for iterating Tar archives that contain type flag `5`

nazar-pc:
Test case included file without entries with such flag.

`Archive_Tar-1.3.11.tar.gz` has exactly the same contents as `Archive_Tar-1.3.11.tgz `, but with 3 directory entries (created using popular GUI tool `file-roller`).

Here is `\__SystemLib\TarArchiveHandler::getContentsList()` returned value for added test file:
```
object(HH\Map)#9 (6) {
  ["Archive_Tar-1.3.11/Archive/Tar.php"]=>
  object(__SystemLib\ArchiveEntryStat)#10 (4) {
    ["crc"]=>
    NULL
    ["size"]=>
    int(66942)
    ["compresedSize"]=>
    NULL
    ["timestamp"]=>
    int(1360410257)
  }
  ["Archive_Tar-1.3.11/Archive/"]=>
  object(__SystemLib\ArchiveEntryStat)#11 (4) {
    ["crc"]=>
    NULL
    ["size"]=>
    int(0)
    ["compresedSize"]=>
    NULL
    ["timestamp"]=>
    int(1461241342)
  }
  ["Archive_Tar-1.3.11/docs/"]=>
  object(__SystemLib\ArchiveEntryStat)#12 (4) {
    ["crc"]=>
    NULL
    ["size"]=>
    int(0)
    ["compresedSize"]=>
    NULL
    ["timestamp"]=>
    int(1461241342)
  }
  ["Archive_Tar-1.3.11/docs/Archive_Tar.txt"]=>
  object(__SystemLib\ArchiveEntryStat)#13 (4) {
    ["crc"]=>
    NULL
    ["size"]=>
    int(18363)
    ["compresedSize"]=>
    NULL
    ["timestamp"]=>
    int(1360410257)
  }
  ["package.xml"]=>
  object(__SystemLib\ArchiveEntryStat)#14 (4) {
    ["crc"]=>
    NULL
    ["size"]=>
    int(11010)
    ["compresedSize"]=>
    NULL
    ["timestamp"]=>
    int(1360410257)
  }
  ["Archive_Tar-1.3.11/"]=>
  object(__SystemLib\ArchiveEntryStat)#15 (4) {
    ["crc"]=>
    NULL
    ["size"]=>
    int(0)
    ["compresedSize"]=>
    NULL
    ["timestamp"]=>
    int(1461241342)
  }
}
```

There are 6 elements, while file for original test case includes only 3 (without directories, which are `size => 0` in output above).


nazar-pc:
Can anyone fix @facebook-github-bot so that it will send only one message?)

nazar-pc:
Just found that similar check actually present in Zip format handling: https://github.com/facebook/hhvm/blob/master/hphp/system/php/archive_handler/ZipArchiveHandler.ns.php#L32

Orvid:
Feel free to add that check in this PR, or another if you want.

nazar-pc:
No, I mean that check I've added in this PR is already present in Zip format handling. Someone who was working on it just forgot to add it for Tar as well.

nazar-pc:
I'll better include it into upcoming big patch to avoid unnecessary rebasing

Issue #6970, opened 2016-04-07
extended ascii heredoc labels not accepted


nikic:
### HHVM Version

3v4l all

### Standalone code, or other way to reproduce the problem

```
var_dump(eval("return <<<\xff\nXYZ\n\xff\n;"));
```

https://3v4l.org/fWX9I

### Expected result

```
string(3) "XYZ"
```

### Actual result

```
Fatal error: Unterminated heredoc at end of file (Line: 1, Char: 18) syntax error, unexpected $end in /in/fWX9I(3)(11cd92606aa2a975a0958cb20ba5a4f0) : eval()'d code on line 1 bool(false)
```

Probably some kind of signed unsigned mismatch. Maybe IS_LABEL_START?

Orvid:
Fix already landed.

PR #7014, opened 2016-04-22
Fix handling of extended ascii heredoc labels

nikic:
Caused by signed-unsigned-mismatch in IS_LABEL_START().

Fixes issue #6970.


nikic:
@Orvid Is it possible to get a diff on the failing test? It works locally, so not sure what the issue is.

nikic:
Nevermind, Sara pointed out what the problem is.

Issue #4263, opened 2014-11-18
HHVM doesn't support tar, tar.gz, tar.bz2 and zip versions of Phar archives


nazar-pc:
tar.gz and tar.bz2 fails with:
> \nFatal error: Uncaught exception 'PharException' with message 'Corrupt phar, can't read 4 bytes starting at offset 1140' in ...

tar fails with:
> \nFatal error: Uncaught exception 'PharException' with message 'Malformed manifest. Expected 0 bytes, got 40' in ...

zip fails with:
> \nFatal error: Uncaught exception 'PharException' with message 'Not a phar: phar:///text.phar.php/text' in ...

Since all this variant supposed to be supported it is a huge incompatibility with PHP5 and should be fixed in next version.

mwjames:
Do we see any chance that in near future phar archives will receive fully support and PHP compatibility?

jwatzman:
No immediate plans that I'm aware of. I might have time to take a look in the next couple days, but other stuff might come up, we'll see how it goes.

jwatzman:
And it does *not* repro with an internal build of master, so sounds like some difference there. Exciting.

jwatzman:
I looked at this briefly; we're getting stuck in a `read` syscall, called as part of `fread`, when first reading from what looks to be a pipe into the child process `timeout`. AFAICT we call `read` once and it never returns. It was not immediately clear what would cause that, and I didn't have time to look into it any deeper. Will comment again if I get more time to dig in further, this is pretty interesting.

jwatzman:
[Above two comments deleted, they were on the wrong task, sorry.]

andrerom:
Duplicate: #5139 "Phing phar outputs gibberish on hhvm"
Related: #1854 "ob_gzhandler not implemented"

This is blocking use of compression by default in Composer: https://github.com/composer/composer/pull/3890 _(typically wasting quite a lot of bandwith, at least when using satis)_

nazar-pc:
Not sure that #1854 is really related, looks like totally different area.

andrerom:
Yes, it was regarding missing compression support in general, but deleted that part now.

bauerj:
Also, it does not seem that #5139 is related as the file in question seems not to be compressed.

nazar-pc:
There is https://pear.php.net/package/PHP_Archive/ licensed under PHP License.

Is it possible to just to include it into HHVM, register as stream wrapper for `phar` and forget about all issues with Phar?
[...]

PR #7019, opened 2016-04-23
Trivial tweaks

nazar-pc:
This is preparation for landing support for compressed Phar files.
As discussed in #4263 changes that doesn't affect code logic separated for easier review.
This PR primarily includes manual edits of PhpDoc sections to conform with arguments names from PHP documentation and to make them understandable to IDE. Also some Exceptions classes changed as it is defined in PHP docs.


Orvid:
This seems to have dropped all of the summaries from the doc comments, could we not do that?

nazar-pc:
Initially I saw most functions without them, so for consistency I've also removed some. My goal was to get sane IDE support, comments are available in PHP docs anyway, but it is handy to have arguments types in source code.

Should I bring them back?

Orvid:
If the summaries were already there, as is the case here, keep them.

nazar-pc:
Restored what was already present

PR #7023, opened 2016-04-26
Adding start transaction and commit as keywords for mysql command tracking.    We happen to use different keywords for transactions than facebook does, this makes it work properly for us.

joelastpass:


facebook-github-bot-1:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D57261

PR #7024, opened 2016-04-26
Add scrypt extension to hhvm

joelastpass:
This is a HHVM extension wrapper of Colin Percival's scrypt. 
William Xu : williamxu15@gmail.com was the primary developer on this 
project here at LastPass.

### DIFF
--- a/hphp/runtime/ext/scrypt/config.cmake
+++ b/hphp/runtime/ext/scrypt/config.cmake
@@ -1,0 +1,14 @@
+HHVM_DEFINE_EXTENSION("scrypt"
+  SOURCES
+    ext_scrypt.cpp
+    crypto/crypto_scrypt-sse.cpp
+    crypto/crypto_scrypt-nosse.cpp
+    crypto/sha256.cpp
+    crypto/params.cpp
+  HEADERS
+    crypto/crypto_scrypt.h
+    crypto/params.h
+    crypto/sha256.h
+  SYSTEMLIB
+    ext_scrypt.php
+)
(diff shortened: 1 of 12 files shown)


facebook-github-bot-9:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D57267

Orvid:
The mysql commit seems to have made it into this PR.

Orvid:
ping?

joelastpass:
Thanks, resubmitted without that commit.

joelastpass:
Just noticed the phabricator comments as well, will address and resend.

joelastpass:
I think this addresses all the phabricator comments, please take another look.

joelastpass:
I believe all phabricator comments have been addressed, can you take another look?

Orvid:
Yep, I was just looking at it yesterday. I'm trying to decide how to suggest handling the SSE vs. no SSE support (it's worth keeping the SSE), because all x64 targets that HHVM supports have at least SSE2. The problem comes from ARM and PPC, which (for the most part) don't have SSE (at the very least, the intrinsics are different). I'll make a couple of comments on phabricator.

joelastpass:
Can we provide any assistance on getting this unstuck in your internal build process?  We're not sure how to help here.

Orvid:
I was just looking at this, will give an update in a bit.

Orvid:
It took me a bit to realize it, but we have an internal extension that was also named scrypt (it doesn't implement the same API that PHP exposes, and is dependent on some other internal stuff), so I have a diff up to rename the internal extension. Once that lands, this should build just fine.

joelastpass:
Okay great.   Do the successful checks mean this is now ready to be merged?

PR #7025, opened 2016-04-28
Call __IsFoldable functions at jit time when possible.

gdbentley:
This change improves WordPress by 2.5% when run in non-repo mode. MediaWiki improves 0.5%.

FCallBuiltin invocations get optimized but not FCall invocations. Those can be handled in a follow-on.

### DIFF
--- a/hphp/runtime/base/execution-context.h
+++ b/hphp/runtime/base/execution-context.h
@@ -128,11 +128,23 @@ struct DebuggerSettings {
   int printLevel = -1;
 };
 
+struct ThrowAllErrorsSetter {
+  ThrowAllErrorsSetter();
+  ~ThrowAllErrorsSetter();
+
+  private:
+  int64_t m_noticeFreq;
+  int64_t m_warningFreq;
+  bool m_throwAllErrors;
+};
+
 using InvokeArgs = folly::Range<const TypedValue*>;
 
 ///////////////////////////////////////////////////////////////////////////////
 
 struct ExecutionContext {
+  friend ThrowAllErrorsSetter;
+
   enum ShutdownType {
     ShutDown,
     PostSend,
@@ -286,7 +298,6 @@ public:
   void setTimeZone(const String&);
 
   bool getThrowAllErrors() const;
-  void setThrowAllErrors(bool);
 
   Variant getExitCallback();
   void setExitCallback(Variant);
@@ -325,6 +336,7 @@ private:
   // helper functions
   void resetCurrentBuffer();
   void executeFunctions(ShutdownType type);
+  void setThrowAllErrors(bool);
 
 public:
   void requestInit();
--- a/hphp/runtime/base/execution-context.h
+++ b/hphp/runtime/base/execution-context.h
@@ -133,8 +133,6 @@ struct ThrowAllErrorsSetter {
   ~ThrowAllErrorsSetter();
 
   private:
-  int64_t m_noticeFreq;
-  int64_t m_warningFreq;
   bool m_throwAllErrors;
 };
 
--- a/hphp/runtime/base/execution-context.h
+++ b/hphp/runtime/base/execution-context.h
@@ -132,7 +132,7 @@ struct ThrowAllErrorsSetter {
   ThrowAllErrorsSetter();
   ~ThrowAllErrorsSetter();
 
-  private:
+private:
   bool m_throwAllErrors;
 };
 
--- a/hphp/runtime/base/request-injection-data-inl.h
+++ b/hphp/runtime/base/request-injection-data-inl.h
@@ -35,7 +35,6 @@ inline bool RequestInjectionData::getJit() const {
 
 inline bool RequestInjectionData::getJitFolding() const {
   return m_jitFolding;
-  return true;
 }
 
 inline void RequestInjectionData::setJitFolding(bool flag) {
--- a/hphp/runtime/base/request-injection-data-inl.h
+++ b/hphp/runtime/base/request-injection-data-inl.h
@@ -33,6 +33,15 @@ inline bool RequestInjectionData::getJit() const {
   return m_jit;
 }
 
+inline bool RequestInjectionData::getJitFolding() const {
+  return m_jitFolding;
+  return true;
+}
+
+inline void RequestInjectionData::setJitFolding(bool flag) {
+  m_jitFolding = flag;
+}
+
 inline bool RequestInjectionData::getCoverage() const {
   return m_coverage;
 }
--- a/hphp/runtime/base/runtime-error.cpp
+++ b/hphp/runtime/base/runtime-error.cpp
@@ -128,9 +128,8 @@ void raise_recoverable_error(const char *fmt, ...) {
 static int64_t g_notice_counter = 0;
 
 static bool notice_freq_check(ErrorMode mode) {
-  if ((RuntimeOption::NoticeFrequency <= 0 ||
-      g_notice_counter++ % RuntimeOption::NoticeFrequency != 0) &&
-      !g_context->getThrowAllErrors()){
+  if (RuntimeOption::NoticeFrequency <= 0 ||
+      g_notice_counter++ % RuntimeOption::NoticeFrequency != 0) {
     return false;
   }
   return g_context->errorNeedsHandling(
--- a/hphp/runtime/base/runtime-error.cpp
+++ b/hphp/runtime/base/runtime-error.cpp
@@ -128,8 +128,9 @@ void raise_recoverable_error(const char *fmt, ...) {
 static int64_t g_notice_counter = 0;
 
 static bool notice_freq_check(ErrorMode mode) {
-  if (RuntimeOption::NoticeFrequency <= 0 ||
-      g_notice_counter++ % RuntimeOption::NoticeFrequency != 0) {
+  if ((RuntimeOption::NoticeFrequency <= 0 ||
+      g_notice_counter++ % RuntimeOption::NoticeFrequency != 0) &&
+      !g_context->getThrowAllErrors()){
     return false;
   }
   return g_context->errorNeedsHandling(
--- a/hphp/runtime/base/runtime-error.cpp
+++ b/hphp/runtime/base/runtime-error.cpp
@@ -128,8 +128,9 @@ void raise_recoverable_error(const char *fmt, ...) {
 static int64_t g_notice_counter = 0;
 
 static bool notice_freq_check(ErrorMode mode) {
-  if (RuntimeOption::NoticeFrequency <= 0 ||
-      g_notice_counter++ % RuntimeOption::NoticeFrequency != 0) {
+  if (!g_context->getThrowAllErrors() &&
+      (RuntimeOption::NoticeFrequency <= 0 ||
+       g_notice_counter++ % RuntimeOption::NoticeFrequency != 0)) {
     return false;
   }
   return g_context->errorNeedsHandling(
@@ -185,8 +186,9 @@ void raise_strict_warning(const char *fmt, ...) {
 static int64_t g_warning_counter = 0;
 
 bool warning_freq_check() {
-  if (RuntimeOption::WarningFrequency <= 0 ||
-      g_warning_counter++ % RuntimeOption::WarningFrequency != 0) {
+  if (!g_context->getThrowAllErrors() &&
+      (RuntimeOption::WarningFrequency <= 0 ||
+       g_warning_counter++ % RuntimeOption::WarningFrequency != 0)) {
     return false;
   }
   return g_context->errorNeedsHandling(
--- a/hphp/runtime/ext/fb/ext_fb.php
+++ b/hphp/runtime/ext/fb/ext_fb.php
@@ -91,7 +91,7 @@ function fb_rename_function(string $orig_func_name,
  * @param mixed $input - What string to sanitize.
  * @return bool - Sanitized string.
  */
-<<__HipHopSpecific, __Native>>
+<<__HipHopSpecific, __Native, __IsFoldable>>
 function fb_utf8ize(mixed &$input): bool;
 
 /* Count the number of UTF-8 code points in string or byte count if it's not
@@ -100,7 +100,7 @@ function fb_utf8ize(mixed &$input): bool;
  * @return int - Returns the count of code points if valid UTF-8 else byte
  * count.
  */
-<<__HipHopSpecific, __Native>>
+<<__HipHopSpecific, __Native, __IsFoldable>>
 function fb_utf8_strlen_deprecated(string $input): int;
 
 /* Count the number of UTF-8 code points in string, substituting U+FFFD for
@@ -109,7 +109,7 @@ function fb_utf8_strlen_deprecated(string $input): int;
  * @return int - Returns the number of code points interpreting string as
  * UTF-8.
  */
-<<__HipHopSpecific, __Native>>
+<<__HipHopSpecific, __Native, __IsFoldable>>
 function fb_utf8_strlen(string $input): int;
 
 /* Cuts a portion of str specified by the start and length parameters.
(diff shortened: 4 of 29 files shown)


facebook-github-bot-8:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D57369

aorenste:
New test failures (4) that were not failing on base rev:
```
hphp/facebook/test/push-blocking:verify_quick_push_blocking - hphp/test/quick/dict/reflect.php: failed
hphp/facebook/test/push-blocking:verify_quick_push_blocking - hphp/test/quick/keyset/reflect.php: failed
hphp/facebook/test/push-blocking:verify_quick_push_blocking - hphp/test/quick/reflect.php: failed
hphp/facebook/test/push-blocking:verify_quick_push_blocking - hphp/test/quick/vec/reflect.php: failed
```

aorenste:
Latest version looks like it has some compilation errors:
```
hphp/runtime/vm/bytecode.cpp:1478:17: error: ‘HPHP::jit::Translator’ has not been declared
     !HPHP::jit::Translator::WriteLease().amOwner();
                 ^
```

hphp/runtime/base/execution-context.h
```diff
@@ -128,11 +128,21 @@ struct DebuggerSettings {
   int printLevel = -1;
 };
 
+struct ThrowAllErrorsSetter {
+  ThrowAllErrorsSetter();
+  ~ThrowAllErrorsSetter();
+
+  private:
```
mxw: nit: Unindent.

hphp/runtime/base/request-injection-data-inl.h
```diff
@@ -33,6 +33,15 @@ inline bool RequestInjectionData::getJit() const {
   return m_jit;
 }
 
+inline bool RequestInjectionData::getJitFolding() const {
+  return m_jitFolding;
+  return true;
```
mxw: Kill this.

hphp/runtime/base/runtime-error.cpp
```diff
@@ -128,8 +128,9 @@ void raise_recoverable_error(const char *fmt, ...) {
 static int64_t g_notice_counter = 0;
 
 static bool notice_freq_check(ErrorMode mode) {
-  if (RuntimeOption::NoticeFrequency <= 0 ||
-      g_notice_counter++ % RuntimeOption::NoticeFrequency != 0) {
+  if (!g_context->getThrowAllErrors() &&
```
mxw: Nice.

hphp/runtime/ext/std/ext_std_variable.cpp
```diff
@@ -247,11 +247,15 @@ String HHVM_FUNCTION(serialize, const Variant& value) {
     case KindOfPersistentString:
     case KindOfString: {
       StringData *str = value.getStringData();
+      auto size = str->size();
+      if (size >= RuntimeOption::MaxSerializedStringSize) {
+        throw Exception("Size of serialized string (%d) exceeds max", size);
+      }
```
mxw: Can you split this off into a separate PR?

hphp/runtime/ext/std/ext_std_variable.cpp
```diff
@@ -247,11 +247,15 @@ String HHVM_FUNCTION(serialize, const Variant& value) {
     case KindOfPersistentString:
     case KindOfString: {
       StringData *str = value.getStringData();
+      auto size = str->size();
```
mxw: (For the PR in question below, `auto const` here.)

hphp/runtime/vm/jit/irgen-builtin.cpp
```diff
@@ -497,6 +586,9 @@ bool optimizedFCallBuiltin(IRGS& env,
   auto const result = [&]() -> SSATmp* {
 
     auto const fname = func->name();
+
+    SSATmp* retVal = opt_foldable(env, func, numArgs, numNonDefault);
+    if (retVal) return retVal;
```
mxw: `if (auto const retVal = opt_foldable(...)) return retVal;`

hphp/runtime/vm/jit/irgen-builtin.cpp
```diff
@@ -488,6 +473,110 @@ SSATmp* opt_set_frame_metadata(IRGS& env, uint32_t numArgs) {
   return cns(env, TInitNull);
 }
 
+SSATmp* opt_foldable(IRGS& env,
+                     const Func* func,
+                     uint32_t numArgs,
+                     uint32_t numNonDefaultArgs) {
+  auto const constAsCell = [] (SSATmp* tmp) {
```
mxw: `const SSATmp* tmp`

hphp/runtime/vm/jit/irgen-builtin.cpp
```diff
@@ -488,6 +473,110 @@ SSATmp* opt_set_frame_metadata(IRGS& env, uint32_t numArgs) {
...
+    // Other Kinds are not expected to be ConstVal
+    not_reached();
+  };
+
+  if (!func->isFoldable()) return nullptr;
+
+  // Don't pop the args yet---if the builtin throws at compile time (because
+  // it would raise a warning or something at runtime) we're going to leave
+  // the call alone.
+  auto const numNonVariadicArgs = (int32_t)func->numNonVariadicParams();
```
mxw: The `(int32_t)` cast here is unnecessary.  (Also the variable is not necessary; just do `func->numNonVariadicParams()` in the condition below.)

hphp/runtime/vm/jit/irgen-builtin.cpp
```diff
@@ -488,6 +473,110 @@ SSATmp* opt_set_frame_metadata(IRGS& env, uint32_t numArgs) {
...
+
+#ifdef DEBUG
+    auto const savedPC = vmpc();
+    // ensure that vmpc is valid for later asserts within the invoke
+    vmpc() = vmfp()->m_func->getEntry();
+    SCOPE_EXIT{ vmpc() = savedPC; };
+#endif
+    SCOPE_EXIT{ RID().setJitFolding(false); };
+    RID().setJitFolding(true);
+    Variant v = invoke(func->name()->data(), args.toArray(), -1, true, true,
```
mxw: Is there a reason you can't call `g_context->invokeFunc()` directly?  `invoke()` doesn't seem to want to be called, based on its comments.

hphp/runtime/vm/jit/irgen-builtin.cpp
```diff
@@ -488,6 +473,110 @@ SSATmp* opt_set_frame_metadata(IRGS& env, uint32_t numArgs) {
...
+    // function would normally get called, so be safe and don't
+    // optimize any calls that COULD generate notices.
+    ThrowAllErrorsSetter taes;
+
+#ifdef DEBUG
+    auto const savedPC = vmpc();
+    // ensure that vmpc is valid for later asserts within the invoke
+    vmpc() = vmfp()->m_func->getEntry();
+    SCOPE_EXIT{ vmpc() = savedPC; };
+#endif
```
mxw: Are the vmRegs always clean when we get here?  If not, you need a VMRegAnchor; if so, it's probably not so bad to do this even in opt builds, since it's not very expensive.

Also, use `if (debug) { ... }` instead of the ifdefs.

hphp/runtime/vm/jit/irgen-builtin.cpp
```diff
@@ -488,6 +473,110 @@ SSATmp* opt_set_frame_metadata(IRGS& env, uint32_t numArgs) {
...
+
+  if (!func->isFoldable()) return nullptr;
+
+  // Don't pop the args yet---if the builtin throws at compile time (because
+  // it would raise a warning or something at runtime) we're going to leave
+  // the call alone.
+  auto const numNonVariadicArgs = (int32_t)func->numNonVariadicParams();
+  PackedArrayInit args(numArgs);
+  for (auto bcOff = BCSPRelOffset{(int32_t)numArgs - 1};
+       bcOff > BCSPRelOffset{(int32_t)(numArgs - numNonDefaultArgs - 1)};
```
mxw: Use `static_cast`.

hphp/runtime/vm/jit/irgen-builtin.cpp
```diff
@@ -488,6 +473,110 @@ SSATmp* opt_set_frame_metadata(IRGS& env, uint32_t numArgs) {
...
+    ThrowAllErrorsSetter taes;
+
+#ifdef DEBUG
+    auto const savedPC = vmpc();
+    // ensure that vmpc is valid for later asserts within the invoke
+    vmpc() = vmfp()->m_func->getEntry();
+    SCOPE_EXIT{ vmpc() = savedPC; };
+#endif
+    SCOPE_EXIT{ RID().setJitFolding(false); };
+    RID().setJitFolding(true);
```
mxw: Swap these two lines (for consistency with the above).

hphp/runtime/vm/jit/irgen-builtin.cpp
```diff
@@ -488,6 +473,110 @@ SSATmp* opt_set_frame_metadata(IRGS& env, uint32_t numArgs) {
...
+    Variant v = invoke(func->name()->data(), args.toArray(), -1, true, true,
+                       !func->unit()->useStrictTypes());
+    auto const retVal = *v.asCell();
+
+    if (isStringType(retVal.m_type)) {
+      return cns(env, makeStaticString(retVal.m_data.pstr));
+    } else if (isArrayLikeType(retVal.m_type)) {
+      return cns(env, make_array_like_tv(
+                     ArrayData::GetScalarArray(retVal.m_data.parr)));
```
mxw: nit: Something like

```
auto const arr = ArrayData::getScalarArray(retVal.m_data.parr);
return cns(env, make_array_like_tv(arr));
```

hphp/runtime/ext/hash/ext_hash.php
```diff
@@ -33,7 +33,7 @@ function hash_algos(): array<string>;
  *
  * @param string $algo     - Name of selected hashing algorithm
  *                           (i.e. "md5", "sha256", "haval160,4", etc..)
- * @param string $filename - File who's contents are to be hashed.
+ * @param string $filename - File whose contents are to be hashed.
```
mxw: Thanks for cleaning these up while you're here :]

mxw:
Running internal performance tests as well.  Sorry that getting this diff in has taken so long—I just now saw that it's been open since April :\

hphp/runtime/ext/std/ext_std_options.php
```diff
@@ -295,7 +295,7 @@ function php_sapi_name(): string;
  * displaying the OS PHP was built on. This will only happen if your uname()
  * library call either doesn't exist or doesn't work.
  */
-<<__Native>>
+<<__Native, __IsFoldable>>
```
simonwelsh: I don't think this can be foldable. `php_uname('n')` can change are run time (if the hostname of the machine changes without restarting HHVM)

hphp/runtime/vm/jit/irgen-builtin.cpp
gdbentley: I was simply copying the IsFoldable implementation from repo authoritative mode which uses `invoke().` `g_context->invokeFunc()` is better though. It avoid the Variant -> Cell ugliness and the unnecessary func -> name -> func.

mxw:
@gdbentley—Looks like there's a merge conflict.  Can you rebase over master and re-push?  If the push doesn't work, you can try `push -f`.

gdbentley:
@mxw Completed rebase number 3.

mxw:
This seems to fail catastrophically on our internal integration tests.  I'll look into why today or Monday and report back.

mxw:
Okay, I'm pretty sure most of the failures were related to our test-infra.  I believe those have been fixed, so I'm testing again.

hphp/runtime/ext/fb/ext_fb.php
```diff
@@ -91,7 +91,7 @@ function fb_rename_function(string $orig_func_name,
  * @param mixed $input - What string to sanitize.
  * @return bool - Sanitized string.
  */
-<<__HipHopSpecific, __Native>>
+<<__HipHopSpecific, __Native, __IsFoldable>>
 function fb_utf8ize(mixed &$input): bool;
```
markw65: I don't see where you're dealing with reference parameters... won't this just "optimize" it to the bool result, while failing to sanitize $input?

hphp/runtime/ext/string/ext_string.php
```diff
@@ -568,7 +569,7 @@ function str_pad(string $input,
  * @return string - Returns the repeated string.
  *
  */
-<<__Native>>
+<<__IsFoldable, __Native>>
 function str_repeat(string $input, int $multiplier): string;
```
markw65: This seems risky - what about str_repeat('x', 2000000000) ? The resulting static string is going to live forever...

hphp/runtime/ext/string/ext_string.php
```diff
@@ -1676,7 +1677,7 @@ function levenshtein(string $str1,
  * @return int - Returns the number of matching chars in both strings.
  *
  */
-<<__Native>>
+<<__IsFoldable, __Native>>
 function similar_text(string $first,
                       string $second,
                       mixed &$percent = null): int;
```
markw65: more reference parameters... I guess this would be ok as long as you check that no parameter is passed for this arg...

hphp/runtime/vm/jit/irgen-builtin.cpp
```diff
@@ -488,6 +473,113 @@ SSATmp* opt_set_frame_metadata(IRGS& env, uint32_t numArgs) {
...
+       bcOff--) {
+    auto const t = topType(env, bcOff);
+    if (!t.hasConstVal() && !t.subtypeOfAny(TUninit, TInitNull, TNullptr)) {
+      return nullptr;
+    } else {
+      args.append(cellAsCVarRef(constAsCell(topC(env, bcOff))));
+    }
+  }
+  if (numArgs != func->numNonVariadicParams()) {
+    if (!topType(env).hasConstVal()) return nullptr;
```
markw65: You should check hasConstVal(TArr), or (since I think the emitter should guarantee that) add an assert that constAsCell(topC(env)) is an array.

hphp/runtime/vm/jit/irgen-builtin.cpp
```diff
@@ -488,6 +473,113 @@ SSATmp* opt_set_frame_metadata(IRGS& env, uint32_t numArgs) {
...
+    SCOPE_EXIT{ vmpc() = savedPC; };
+#endif
+    RID().setJitFolding(true);
+    SCOPE_EXIT{ RID().setJitFolding(false); };
+
+    Cell retVal;
+    g_context->invokeFunc(&retVal, func, args.toArray(), nullptr, nullptr,
+                          nullptr, nullptr, ExecutionContext::InvokeNormal,
+                          !func->unit()->useStrictTypes());
+
```
markw65: When we start jitting in separate threads (real soon now), this will almost certainly break. It might already break due to some of the guards that have recently been added to try to enforce that...

hphp/runtime/vm/jit/irgen-builtin.cpp
gdbentley: Added assert.

hphp/runtime/vm/jit/irgen-builtin.cpp
gdbentley: What specifically is the concern? That the thread doing the jitting will not have a valid g_context? Should I switch back to the (deprecated) invoke() code?

mxw:
@gdbentley—Okay, sorry for the delay; we went through several iterations of the protection mechanism before it was fully correct.  The issue is resolved by a one-liner (`VMProtect::Pause foo;` above your `VMRegAnchor`), which I've already added to the internal copy of the diff.  Tests seem to be passing just fine, so I'm going to rebase and start them again just to be sure.  Unless there's something you want to change, no action is needed on your part—I'll land the diff with that update via our usual internal mechanism.

gdbentley:
@mxw Sounds good. Thanks.

PR #7028, opened 2016-05-01
Support for compressed phar

### LINKED ISSUE #4263: HHVM doesn't support tar, tar.gz, tar.bz2 and zip versions of Phar archives

nazar-pc:
This is a big PR, so it needs extended description.

Features:
* Support for `.phar.gz` `.phar.bz2`, `*.phar.tar`, `*.phar.tar.gz`, `*.phar.tar.bz2`, `*.phar.zip` kinds of Phar archives regardless of their extension
* Implemented methods:
  * `Phar::isValidPharFilename()` with tests ported from PHP's sources
  * `Phar::isCompressed()` returns real value now instead of hardcoded `false`
  * `Phar::setAlias()` implemented as proxy method with necessary checks, but not actually used until write support implemented properly
  * `Phar::setStub()` implemented as proxy method, but not actually used until write support implemented properly
  * `Phar::isFileFormat()` was checking only whether file is plain Phar, now also checks for Tar and Zip
  * `Phar::getSupportedCompression()` returns supported compression formats instead of empty array
  * `Phar::getSupportedSignatures()` returns supported signatures instead of empty array
  * `PharFileInfo::isCRCChecked()` is just a stub for now
* Tar archives support was fixed to support archives that contain type flag `5`, it happened when creating archives with popular `file-roller`, I suspect GNU Tar is used under the hood
* Added support for `include 'phar:///path/to.phar'`, `include 'phar:///path/to.phar.gz'`, `include 'phar:///path/to.phar.bz2'` (gz and bz2 variations do not work without `phar://` protocol yet)
* For `include 'phar:///path/to.phar'` stub will be used, even though path to it is not specified.

Architecture decisions:
* There was initial Tar and Zip support in `PharData` class using `__SystemLib\TarArchiveHandler` and `__SystemLib\ZipArchiveHandler` classes.
Similarly to mentioned classes new class `__SystemLib\PharArchiveHandler` was added, it provides support for plain Phar format (absorbed Phar-specific functionality from `Phar` class), all three classes refactored and unified internally.

* `Phar` and `PharData` classes are very similar, but had different implementation for similar methods (`offset*()` methods especially), now they are unified and share much more common code.

* There was major memory waste by storing all contents of archives in memory, now streams are used instead of strings.
Zip-based implementation relies on `ZipArchive::getStream()` and its efficiency.
Tar-based implementation was rewritten to be stream-based, but for maximum efficiency needs some sort of stream slicing mechanism (otherwise there is still some overhead when copying bytes to `php://temp`, which is not too bad though and very unlikely to eat all memory).
Phar-based implementation was also rewritten for streams usage and similarly to Tar-based needs stream slicing for maximum efficiency, but is good enough already.

Fixes #4263


facebook-github-bot-6:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D57495

PR #7036, opened 2016-05-04
Move emitDecRefGeneric to arch specific file

lbianc:
The method `emitDecRefGeneric` creates an indirect fixup providing an `x64`
specific offset, so this method must be in an architecture specific file.


facebook-github-bot-1:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D57627

Orvid:
Diff abandoned.

PR #7041, opened 2016-05-07
Do not use folly::Optional on iterator in store-elim.

wjywbs:
Clang 3.6 and 3.7 on mac crash with assertion "move constructor should not be
deleted".

https://github.com/llvm-mirror/clang/blob/e1d7b3bd4c2f9d0866ca7eaa4178a0c06815b607/include/clang/AST/DeclCXX.h#L919-L924


facebook-github-bot-0:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D57855

Orvid:
Perhaps it would be better to fix this in Folly rather than here?

wjywbs:
This is very likely a clang bug because clang shouldn't crash.

Orvid:
@wjywbs Ping on the review.

PR #7053, opened 2016-05-12
Allowed TLS address generation for non-x64 platforms also

lpathy:


Orvid:
Should probably poke the PPC guys before changing this, as this check is catching that platform as well.

facebook-github-bot-0:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D58131

PR #7055, opened 2016-05-12
Removed architecture check for call to irgen::prepareEntry()

lpathy:
irgen::prepareEntry() should be called on all platforms.


facebook-github-bot-2:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D58143

PR #7056, opened 2016-05-12
If target is ARM, replaced push/pop calls with pushp/popp (pair)

lpathy:
This will reduce the number of instructions for saving and restoring
the registers.


facebook-github-bot-6:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D58161

PR #7058, opened 2016-05-12
Added arm specific restrictions in copy prop. and imm folding optimizations

lpathy:
1.  After lowering, only [base, index lsl #scale] and [base, #imm] are
    allowed, where the range of #imm is [-256 .. 255]. So, if the Vptr
    is not in one of above two forms, avoided rewrite in vasm-copy
2.  Removed safe_cast<> in imm folding as IsImmArithmetic() and
    IsImmLogical() perform the checks on 64 bit value


facebook-github-bot-3:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D58149

PR #7061, opened 2016-05-13
Add array access to constants

### LINKED ISSUE #6992: Array access on constants not supported
### HHVM Version

3.13.1

```php
<?php

define('great','my');

echo great[1];
```

### Expected result

`y`

### Actual result

`Fatal error: syntax error, unexpected '[', expecting ',' or ';' `

Support for this was added in PHP 5.6 (https://3v4l.org/0EiYf)

apeabody:
Add array access to constants

-Adds array access to constants like PHP 5.6+/7+, includes test cases
  https://3v4l.org/0EiYf

-Updates a related test cases which as a result now more closely matches PHP 5.6+/7+ behavior
  https://3v4l.org/IBW3v

-The test case is currently disabled in RepoAuth mode as it makes an incompatible assumption about constants - I plan to open a specific issue for this incompatibility when/if this is committed.

-Includes a small fix to make-parser.sh for when building hphp.{5,7}.cpp on hphp.tab.hpp outside FB's environment.

NOTE: hphp.5.tab.cpp, hphp.7.tab.cpp, and hphp.tab.hpp files are not included as they will be built on FB's Bison version.

Closes #6992


facebook-github-bot-8:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D58179

PR #7063, opened 2016-05-13
ADD: stream_set_read_buffer()

### LINKED ISSUE #6977: PHP function not implemented: stream_set_read_buffer
### HHVM Version
All, latest checked: 3.12.0

[See at 3v4l](https://3v4l.org/Lq3hf)
```php
<?php
var_dump(function_exists('stream_set_read_buffer'));
```

### Expected result
`bool(true)`

### Actual result
`bool(false)`

apeabody:
Closes #6977 

stream_set_read_buffer() implemented in-line with the PHP spec (https://secure.php.net/manual/en/function.stream-set-read-buffer.php) for file streams.  If a buffer size of 0 is provided then file operations will be unbuffered.  Returns 0 on success, or another number on failure.

Adds one new test, and enables test bug61019.


facebook-github-bot-5:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D58209

PR #7066, opened 2016-05-15
Add stream_set_chunk_size() as no-op

### LINKED ISSUE #6573: PHP function not implemented: stream_set_chunk_size
It seems like `stream_set_chunk_size` isn't around in HHVM (I'm using 3.10.1).

```hack
echo (function_exists('stream_set_chunk_size') ? 'true' : 'false') . PHP_EOL;
```
Result: `false`

Searching for this method in the repository didn't yield any C++ results.

PHP introduced this method in 5.4.0.

apeabody:
Closes #6573 

Implements stream_set_chunk_size() as a call to the new File::setChunkSize() no-op.  This does properly return the current chunk size when appropriate, but won't change the Chunk Size as it's currently a constant.

Also adds a new File::getChunkSize() and tests for stream_set_chunk_size().

### DIFF
--- a/hphp/runtime/base/file.cpp
+++ b/hphp/runtime/base/file.cpp
@@ -764,13 +764,9 @@ int64_t File::getChunkSize() const{
   return CHUNK_SIZE;
 }
 
-Variant File::setChunkSize(int64_t chunk_size) {
-  if (chunk_size > 0) {
-    //no-op currently, return the current chunk size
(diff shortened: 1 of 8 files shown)


facebook-github-bot-9:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D58233

PR #7073, opened 2016-05-18
Re-implement stream_set_write_buffer()

apeabody:
stream_set_write_buffer() re-implemented in-line with the PHP spec (https://secure.php.net/manual/en/function.stream-set-read-buffer.php) for file streams. If a buffer size of 0 is provided then file operations will be unbuffered. Returns 0 on success, or another number on failure.  See related #7063 

Adds additional cases to the slow stream_set_write_buffer test.


facebook-github-bot-2:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D58371

PR #7077, opened 2016-05-18
FIX: preg_split()'s limit option now enabled

### LINKED ISSUE #6815: preg_split() does not accept $limit NULL
`preg_split("/[\s,]+/", "hypertext language, programming", NULL);`

results in:

> PHP Warning: preg_split() expects parameter 3 to be integer, null given

Although PHP documentation states that:

> A limit of -1, 0 or NULL means "no limit" and, as is standard across PHP, you can use NULL to skip to the flags parameter.

PHP 5.6 and 7.0 accept NULL as $limit parameter.

HipHop VM 3.11.1 (Jessie)

apeabody:
preg_split()'s limit will now accept -1, 0, or null as no limit
preg_split()'s limit will now honor provided limit
tests added for preg_split()'s limit null, as well as non-zero

https://secure.php.net/manual/en/function.preg-split.php

Closes #6815


facebook-github-bot-9:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D58401

t8erjoe:
joe was here

PR #7081, opened 2016-05-19
stream_set_chunk_size() now functional

### LINKED ISSUE #6573: PHP function not implemented: stream_set_chunk_size

apeabody:
FIX: stream_set_chunk_size() no longer a no-op

-File::setChunkSize() now implemented
-Moved the chunkSize into the m_data instance as m_chunkSize
-correct stream_set_chunk_size() to return the *previous* chunk size
-New DEFAULT_CHUNK_SIZE is set to 8k
-Updates/passes test/slow/stream

Closes #6573

### DIFF
--- a/hphp/runtime/base/file.cpp
+++ b/hphp/runtime/base/file.cpp
@@ -772,7 +772,7 @@ void File::setChunkSize(int64_t chunk_size) {
   m_data->m_chunkSize = chunk_size;
 
   if (m_data->m_buffer != nullptr && m_data->m_chunkSize > m_data->m_bufferSize) {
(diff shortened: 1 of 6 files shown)


facebook-github-bot-9:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D58491

PR #7084, opened 2016-05-21
stream_set_blocking() arguments now match current PHP spec

### LINKED ISSUE #6985: Redis fails when hhvm.php7.all enabled
With hhvm.php7.all=1 set in the INI file, the type checker breaks with the internal Redis implementation. 

Looking at https://github.com/facebook/hhvm/blob/df46f6565d53081be567ac7d7018a064fff387c0/hphp/system/php/redis/Redis.php @ 1653:

stream_set_blocking($this->connection, true);

The second parameter is being passed in as boolean.

Checking the PHP docs: http://php.net/manual/en/function.stream-set-blocking.php - the second param SHOULD be a boolean. So HHVM is expecting the wrong data type.

### HHVM Version
3.13.1

Set hhvm.php7.all=1 in INI file

$redis = new Redis;
$redis->connect($server, -1, 0.1);

### Expected result
proper connection to a Redis server

### Actual result
Fatal error: Uncaught TypeError: Argument 2 passed to stream_set_blocking() must be an instance of int, bool given in /var/www/test/redis.php:6
Stack trace:
#0 /var/www/test/redis.php(6): Redis->connect()
#1 {main}

apeabody:
stream_set_blocking() arguments now match current PHP spec

-stream_set_blocking() 2nd argument is now bool instead of int per https://secure.php.net/manual/en/function.stream-set-blocking.php
-includes new slow tests for stream_set_blocking()

Closes #6985


facebook-github-bot-4:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D58575

PR #7088, opened 2016-05-22
Fix for PHP5.6 compatibility of multiple arguments unpacking during function call

### LINKED ISSUE #7087: Only the last parameter in a function call is allowed to use ...
### HHVM Version

```
HipHop VM 3.14.0-dev (rel)
Compiler: heads/master-0-gd8285fbc906f340182a2a94b1fd0cf40787a40b9
Repo schema: 093188019b394a8a53aee6c073772c7db4ffc5e6
```

```
$foo = ['x', 'y', 'z'];
var_dump(...$foo, ...$foo);
```
### Expected result

```
~> php index.php
string(1) "x"
string(1) "y"
string(1) "z"
string(1) "x"
string(1) "y"
string(1) "z"
```
### Actual result

```
~> hhvm index.php

Fatal error: Only the last parameter in a function call is allowed to use ... in * on line *
```

There are 2 tests: [hphp/test/slow/unpack_args/unpack_call_error.php](https://github.com/facebook/hhvm/blob/master/hphp/test/slow/unpack_args/unpack_call_error.php) and [hphp/test/slow/unpack_args/unpack_call_multi.php](https://github.com/facebook/hhvm/blob/master/hphp/test/slow/unpack_args/unpack_call_multi.php), but they are both wrong, such usage is completely valid in PHP7

nazar-pc:
Also corrected error message to match such in PHP
Fixes #7087

### DIFF
--- a/hphp/compiler/analysis/emitter.cpp
+++ b/hphp/compiler/analysis/emitter.cpp
@@ -2912,7 +2912,7 @@ void EmitterVisitor::emitCall(Emitter& e,
   auto const unpack = func->hasUnpack();
   if (!func->checkUnpackParams()) {
     throw IncludeTimeFatalException(
-      func, "Only the last parameter in a function call is allowed to use ...");
+      func, "Cannot use positional argument after argument unpacking");
   }
   {
     FPIRegionRecorder fpi(this, m_ue, m_evalStack, fpiStart);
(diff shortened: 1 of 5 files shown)


facebook-github-bot-2:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D58605

nazar-pc:
As pointed out in Phabricator, the issue is by far more complex that I can handle it

PR #7092, opened 2016-05-23
Correct behavior of timezone_open() to match PHP spec

### LINKED ISSUE #6632: timezone_open throws an exception rather than returning false

apeabody:
Correct behavior of timezone_open() to match PHP spec
-returns bool false and warning instead of exception
-add new test for this behavior

Example:
https://3v4l.org/qKNug
https://secure.php.net/manual/en/datetimezone.construct.php

Closes #6632

### DIFF
--- a/hphp/runtime/ext/datetime/ext_datetime.php
+++ b/hphp/runtime/ext/datetime/ext_datetime.php
@@ -876,8 +876,16 @@ function timezone_name_get(DateTimeZone $timezone): string {
 }
 
 <<__ParamCoerceModeFalse>>
-function timezone_open(string $timezone): DateTimeZone {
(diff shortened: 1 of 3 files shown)


facebook-github-bot-5:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D58647

PR #7099, opened 2016-05-25
Fix flag's type from char to int8_t

lbianc:
The sign of a char type is not defined on C standard. On X64, the char expands to "signed char" however PPC64 expands it to "unsigned char", which is a common source of mistake found when porting applications.
The char type is designed to hold an ASCII character and arithmetic should be avoided.
In order to make the code platform independent, the type char was replaced to int8_t, which is explicit signed.


facebook-github-bot-1:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D58761

PR #7100, opened 2016-05-25
Fix calling convention when GP after FP is needed for PPC64

igorsnunes:
Register allocation on ppc64 is not easy: If a fixed argument function
uses FP arguments, the aligned GP arg needs also to be reserved.

Here, the number_format function ("double, int, int, int" arguments,
with storage buffer) should use for it's first int argument the $r5, not
the $r4 how it was as $r3 is the storage buffer and $r4 is reserved due
to $f1 being used for the double.

I tried to make it nicely before by removing simdArgs from other places,
but then there was a lea trying to move a GP to FP, odd... this one is
neater despite wasting (possibly) a copy to an unused GP register
argument.

### DIFF
--- a/hphp/runtime/vm/jit/arg-group.h
+++ b/hphp/runtime/vm/jit/arg-group.h
@@ -16,6 +16,8 @@
 #ifndef incl_HPHP_JIT_ARG_GROUP_H
 #define incl_HPHP_JIT_ARG_GROUP_H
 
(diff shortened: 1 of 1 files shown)


facebook-github-bot-0:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D58767

PR #7102, opened 2016-05-25
Use specific emitTLSAddr for powerpc64

igorsnunes:
For ppc64, there is no fs register. To refer to a thread content,
the ppc64-abi uses a displacement relative to r13 (thread pointer).


facebook-github-bot-5:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D58791

PR #7107, opened 2016-05-27
FIX: strip null char from end of getSortKey() return

### LINKED ISSUE #7106: Collator::getSortKey() wrongly includes null terminator in return value
HHVM includes the null terminator byte in the result from Collator::getSortKey(), whereas recent versions of PHP do not. PHP's behavior changed when [PHP Bug #62070](https://bugs.php.net/bug.php?id=62070) was fixed, which happened four years ago (in php/php-src@51286bd8e5a7acd75326fae497313725e4acf234).

### HHVM Version
3.14.0-dev (040c9d42da91ea40da19006abbf0e5120b2b9ac6), as well as 3.12.0 (see <https://3v4l.org/ofp14>)
Use PHP's "bug62070.phpt" or "bug62070_2.phpt", depending on ICU version. Alternatively, use this variation, which, according to the [ICU User Guide](http://userguide.icu-project.org/collation/api#TOC-GetSortKey), should be independent of ICU version:

```
<?php
$coll = new Collator('root');
var_dump(substr_count($coll->getSortKey('Hello'), "\0"));
```
### Expected result
```
int(0)
```
### Actual result
```
int(1)
```

apeabody:
-Correct String ret length as HHVM's String type auto pads +1 for NULL termination
-Reduce ret.setSize() 1 to truncate NULL termination from ucol_getSortKey() as "Unlike ICU functions that write a string to an output buffer, the terminating zero byte is counted in the sort key length."
-Enable bug62070 test
-New test case based on issue #7106 

Closes #7106


facebook-github-bot-9:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D58887

PR #7108, opened 2016-05-27
Add asm instructions for PPC64

lbianc:
yield instruction added for PPC64 to provides a hint that performance will
probably be improved if shared resources dedicated to the executing
processor are released for use by other  processors.
cntlzw instruction added for count leading zeros.


facebook-github-bot-9:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D58899

PR #7110, opened 2016-05-27
Fix quantity of registers on reg algorithms

lbianc:
After the change that increases the number of possible registers in the
HHVM (from 64 to 128), the constants of this files got outdated. Changing
it to fit the current number of registers.


facebook-github-bot-1:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D58947

PR #7111, opened 2016-05-27
FIX: hphp_get_static_property() & hphp_set_static_property() parameters

### LINKED ISSUE #6824: Differing number of parameters for hphp_{get,set}_static_property in runtime vs HHI
Runtime:

```
/**
 * hphp_set_property() - Used by ReflectionProperty to set the value of a
 *                       property on an instance of a class.
 *
 * @param object $obj  - The object to set the property on.
 * @param string $cls  - The name of the class that the property is accessible
 *                       in or null to only set a public property.
 * @param string $prop - The name of the property.
 * @param mixed $value - The value to set the property to.
 */
<<__Native("NoInjection"), __HipHopSpecific>>
function hphp_set_property(
  object $obj,
  string $cls = "", // null will convert to this and do what we expect
  string $prop,
  mixed $value): void;

/**
 * hphp_get_static_property() - Used by ReflectionProperty to get the value of a
 *                              static property on a class.
 *
 * @param string $cls  - The name of the class.
 * @param string $prop - The name of the static property
 * @param bool $force  - Whether or not to get protected and private properties
 *                       (true) or only public ones (false)
 * @return mixed - The value of the property
 */
<<__Native("NoInjection"), __HipHopSpecific>>
function hphp_get_static_property(string $cls, string $prop,
                                  bool $force): mixed;
```

HHI:

```
function hphp_get_static_property($cls, $prop) { }
function hphp_set_static_property($cls, $prop, $value) { }
```

Found by docs site build process

apeabody:
Add missing 'force' parameters for hphp_get_static_property() & hphp_set_static_property() in .hhi

Closes #6824

### DIFF
--- a/hphp/hack/hhi/stdlib/builtins_reflection.hhi
+++ b/hphp/hack/hhi/stdlib/builtins_reflection.hhi
@@ -16,6 +16,6 @@ function hphp_instanceof($obj, $name) { }
 function hphp_create_object_without_constructor($name) { }
 function hphp_get_property($obj, $cls, $prop) { }
 function hphp_set_property($obj, $cls, $prop, $value) { }
-function hphp_get_static_property($cls, $prop) { }
-function hphp_set_static_property($cls, $prop, $value) { }
+function hphp_get_static_property($cls, $prop, $force) { }
+function hphp_set_static_property($cls, $prop, $value, $force) { }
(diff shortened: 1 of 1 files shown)


facebook-github-bot-4:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D58959

Issue #7112, opened 2016-05-27
MediaWiki segfaults in oss-performance (repo authoritative)


gdbentley:
### HHVM Version
Reproduced on current master branch (e232b86)
### Standalone code, or other way to reproduce the problem
hhvm perf.php --mediawiki --hhvm=/mypath/hhvm
### Expected result
No segfault
### Actual result
Segfault

gdbentley:
Fixed by #7113

Issue #7104, opened 2016-05-26
Coredump running composer with hhvm-nightly on Ubuntu 14.04


sgolemon:
### HHVM Version

3.14.0-dev (rel) hhvm-nightly installed on 2016-05-25
### Standalone code, or other way to reproduce the problem

Standard composer phar installed from get-composer
$ cat composer.json
{ "name": "test/test" }
$ hhvm /usr/local/bin/composer install
### Expected result

Composer runs, finds nothing to install, exists
### Actual result

Segmentation Fault

```
#0  0x000000000355dcec in HPHP::jit::FixupMap::fixupWork(HPHP::ExecutionContext*, HPHP::ActRec*) const ()
#1  0x000000000355dc9b in HPHP::jit::MCGenerator::syncWork() ()
#2  0x00000000035d3199 in HPHP::createBacktrace(HPHP::BacktraceArgs const&) ()
#3  0x0000000002821a39 in HPHP::debug_string_backtrace(bool, bool, long) ()
#4  0x00000000015aa3e7 in HPHP::bt_handler(int) ()
#5  <signal handler called>
#6  0x00000000035428ac in HPHP::StringData::release() ()
#7  0x00000000048002e3 in ?? ()
#8  0x00007fe40437ef60 in ?? ()
#9  0x0000000000000034 in ?? ()
#10 0x00000000048003bf in ?? ()
#11 0x000000000844dde0 in ?? ()
#12 0x000000000480004f in ?? ()
#13 0x00007fe4124e8110 in ?? ()
#14 0x00007ffde95fcfd0 in ?? ()
#15 0x00000000023ea00b in HPHP::jit::x64::enterTCImpl(unsigned char*, HPHP::ActRec*) ()
```

Orvid:
Humm... Which OS/GCC version? I just tried trunk on Debian 8 (after fixing an issue with the build that has apparently been there since the conversion from IDL...) and it worked fine.

sgolemon:
OS is Ubuntu 14.04 (per the issue title).
gcc version is whatever is on the Mac build VM near your desk.  gcc 4.8.something iirc.

Orvid:
Well, I got it reproducing... Now to figure out why...

Orvid:
:((((((( It isn't reproducing in a debug build :(((((

Orvid:
I have it reproducing consistently and am working to identify the exact cause.

nazar-pc:
[...]

PR #7113, opened 2016-05-28
Align thread locals to 16 bytes.

gdbentley:
Commit 8863fc1 changed the order of linking libraries. This caused the layout of thread-local variables to change, revealing a bug in the specification of the necessary alignment for MemoryUsageStats. The compiler was generating a movdqa instruction to access the m_usage field of that object, expecting that field to be 16 byte aligned since it presumed that the beginning of the object would be 16 byte aligned. This change specifies the stricter 16 byte alignment rather than sizeof(void *) (i.e. 8 byte) alignment.


gdbentley:
This fixes #7112

facebook-github-bot-7:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D58983

Orvid:
It's worth noting that there is a known issue (#7104) with trunk on Ubuntu 14.04's version of GCC (I suspect we managed to hit a compiler bug, as it only manifests on 14.04 and release mode), and I'd narrowed it down (manual bisecting) to be caused by one of 10 commits, almost a month before the object libraries change, and I believe @paulbiss kicked off a proper git bisect before leaving yesterday, so we'll see if that fixes the Mediawiki issue before messing with this. It's possible that this is just masking the problem.

gdbentley:
I reproduced this on CentOS 7.

apeabody:
As a note, I have seen some other unexpected segfaults in release builds (but not debug builds) on CentOS 7.  #7115 #7116

CentOS 7:
gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-4)

apeabody:
I just finished a compile & quick/test, and this commit/pr DOES appear to resolve #7115 and #7116 as well.

Orvid:
The thing is that it shouldn't be happening in the first place :( The best I can guess is that we're hitting either a compiler or linker bug and this may just be masking that.

apeabody:
The issues (#7115 & #7116) I had noted no longer manifests in the current master.  My laptop is too slow to track down the exact commit that resolved them, however I do see changes to MemoryManager which StringData::release() heavily utilizes.

gdbentley:
The root issue was alignment of thread locals. The change to the libraries caused the MemoryManager thread locals to be 8 byte aligned where they had been 16 byte aligned. This caused a movdqa (move double quadword aligned) to have an alignment error manifesting as the segfault.

gdbentley:
088f311 is the commit which inadvertantly fixed #7115 and #7116. It removed the m_stats.usage variable, which was the one getting misaligned.

jwatzman:
@gdbentley was it our JIT incorrectly generating a `movdqa` or a clang/ld bug? I wonder if this is related to 967a98fb807651d585dfe9dfd57e8216fe244c80 where I hit a similar issue, where clang asked for a variable to be 16-byte aligned and generated code assuming that, but OS X ld didn't actually respect that.

gdbentley:
@jwatzman gcc produced the movdqa. I'm not sure that it was incorrect to do so though. I'm not familiar enough with the C++ specification to be sure. If the MemoryUsageStats object were allocated starting at a 16 byte alignment (which it would have been for an ordinary global or a local), then the offset for m_usage would have been 16 byte aligned.

Orvid:
Could you update the summary of the PR in as well?

gdbentley:
@Orvid I presume you meant in phabricator. Done.

apeabody:
@gdbentley - In my experience you will want to change the PR summary in Github (first post).  Unless something has changed recently, if you do it in Phabricator the fb-github-bot will at some point sync over it from github.

gdbentley:
@apeabody Thanks.

PR #7117, opened 2016-05-30
Fix hardcoded value for memfd_create.

racardoso:
Summary:

After merge with upstream our build for PPC64 port brokes.
That's because an #ifndef x86_64 error called from hh_shared.

memfd_create syscall is 319 on x86_64, 360 on POWER and, 385 for
ARM. But we can call __NR_memfd_create defined on unistd.h
instead use a hardcoded  value and test if Linux version support
this syscall.

HHVM is currently being ported to architectures others architectures
(POWER and ARM) and must avoid introduces things "only for x86"
that breaks builds on those architectures.

### DIFF
--- a/hphp/hack/src/heap/hh_shared.c
+++ b/hphp/hack/src/heap/hh_shared.c
@@ -130,9 +130,15 @@
  * appeared in Linux 3.17.
  ****************************************************************************/
 #if !defined __APPLE__ && !defined _WIN32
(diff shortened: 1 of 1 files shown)


facebook-github-bot-4:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D58995

PR #7118, opened 2016-05-30
Fixes compilation error due updates on Intel XED.

racardoso:
Summary:

Dissasembler for x86_64 uses Intel XED to disasm code. At some point the interfaces has 
changed causes a compilation error with new versions of XED (2015-09-10). 
So i think it's time to drop the support to old versions.


facebook-github-bot-2:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D59001

PR #7119, opened 2016-05-31
ADD: enable ccache when detected

apeabody:
-Dramatically and safely improve the speed of C/C++ recompiles by utilizing ccache when detected.  Very helpful after a 'make clean', when changing between release and debug builds, or when maintaining builds in multiple directories.  Caches can also be shared locally or via NFS for environments with multiple developers.

ccache: https://ccache.samba.org/


facebook-github-bot-4:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D59025

PR #7124, opened 2016-06-02
FIX: remove orphan call to FindMemfd.cmake removed in <c168d42>

apeabody:
Remove orphan call to FindMemfd.cmake which was removed in https://github.com/facebook/hhvm/commit/c168d42be61f7f13c15a3a6dcf572c7635869af5


facebook-github-bot-3:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D59133

PR #7127, opened 2016-06-04
FIX: correct object conversion error for floats

### LINKED ISSUE #5408: object conversion, err msg wrong, int instead of float
````php
<?php
class A{}
var_dump((float) new A());
?>
````
The output should be: "Notice: Object of class A could not be converted to float in /in/jD88b on line 4
float(1)"

Current output: "Notice: Object of class A could not be converted to int in /in/jD88b on line -1
float(1)"

apeabody:
correct object conversion error message for floats to name float instead of int

Closes #5408


facebook-github-bot-5:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D59211

PR #7139, opened 2016-06-07
FIX: behavior of TempFile's (php://{memory,temp}) fseek(), ftruncate(), & ftell()

### LINKED ISSUE #6830: Ftruncate behaves differently
The following snippet behaves differently on HHVM 3.11 to PHP 5.6 and 7

```php
$fp = fopen('php://memory', 'r+b');
fwrite($fp, "1234567890");
ftruncate($fp, 0);
fwrite($fp, "abcdefghij");
rewind($fp);
echo "'" . fread($fp, 10) . "'";
```

```
 php prints 'abcdefghij'
hhvm prints ''
```

I suspect the behaviour of hhvm is actually correct, because ftruncate is not meant to reset the pointer, however I point this out due as a inconsistency.

The workaround is to add rewind after the ftruncate.

### LINKED ISSUE #6916: fseek beyond the end of the stream is allowed
It appears that we allow seeking beyond the end of a file. PHP does not. This is likely part of the cause of #6830.

https://3v4l.org/MTMtH

apeabody:
NOTE: HHVM implements php://memory and php://temp as the TempFile class 

- correct return of fseek() past eof for TempFile to int(-1)
Closes #6916 
- now adjusts position for ftruncate() for TempFile
Closes #6830 
- correct return of ftell() past eof for TempFile streams to bool(false)
https://3v4l.org/MTMtH
- enable correct behavior of SEEK_END for fseek() for TempFile
- add tests for above corrections

### DIFF
--- a/hphp/runtime/base/file.cpp
+++ b/hphp/runtime/base/file.cpp
@@ -777,6 +777,10 @@ void File::setChunkSize(int64_t chunk_size) {
   }
 }
 
+std::string File::getWrapperProtocol() const {
+  return Stream::getWrapperProtocol(m_data->m_name.c_str()).toCppString(); 
+}
+
 ///////////////////////////////////////////////////////////////////////////////
 // csv functions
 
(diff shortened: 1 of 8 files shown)


facebook-github-bot-9:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D59289

PR #7143, opened 2016-06-09
FIX: update description of HH\Asio\curl_exec

### LINKED ISSUE #6236: curl_errno() always returns 0 with HH\Asio\curl_exec
When switching from curl_exec to HH\Asio\curl_exec the error handling properties change.  curl_exec return false on failure and reports errors through curl_errno(). HH\Asio\curl_exec returns the empty string on failure and does not appear to offer up error codes in any way.  The codes are available within HH\Asio\curl_exec while it has access to the multi handle, through curl_multi_info_read, but this is not exposed to the programmer afaict.

    <?hh

    function createHandle() : resource {
        $ch = curl_init("http://localhost:54321");
        curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
        return $ch;
    }

    function main() : void {
        $ch = createHandle();
        curl_exec($ch);
        echo 'plain curl_exec: ';
        var_dump(curl_errno($ch));

        $ch = createHandle();
        HH\Asio\curl_exec($ch)->getWaitHandle()->join();
        echo 'async curl_exec: ';
        var_dump(curl_errno($ch));
    }

    main();

The expected result is:
    plain curl_exec: int(7)
    async curl_exec: int(7)

But the actual result is:
    plain curl_exec: int(7)
    async curl_exec: int(0)

apeabody:
Note that (unlike the traditional curl_exec) async curl_exec does not work with curl_errno, instead curl_multi_info_read must be used to retrieve error codes.

Closes #6236 
Closes hhvm/user-documentation#335

https://bugs.php.net/bug.php?id=60660

### DIFF
--- a/hphp/runtime/ext/curl/ext_curl.php
+++ b/hphp/runtime/ext/curl/ext_curl.php
@@ -450,7 +450,10 @@ namespace HH\Asio {
  *
  * Pass a cURL handle, or, more simply, a string containing a URL (and the
  * cURL handle will be created for you), and the cURL request will be executed
- * via async and the `string` result will be retuned.
+ * via async and the `string` result will be returned.
+ *
+ * curl_multi_info_read must be used to retrieve error information,
+ * curl_errno can't be used as this function is a wrapper to curl_multi_await.
  *
  * @param $urlOrHandle - An existing cURL handle or a URL as a `string`. String
  *                       URLs will create a default cURL GET handle.


facebook-github-bot-8:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D59367

PR #7147, opened 2016-06-10
improve ccache handling by cmake

apeabody:
- New cmake option "-DSKIP_CCACHE=TRUE" to skip detecting/enabling ccache if desired.
- Only enable ccache if present, and NOT already in use as a wrapper for the system C/CXX compiler - previously ccache would be called recursively in such cases (ccache handles that transparently, but other wrappers such as distcc do not).
- When ccache is enabled, distcc *should* now be easily used during builds with:  "CCACHE_PREFIX=distcc pump make -j{core}" - I plan to test this further and develop documentation as time permits.
- Remove RULE_LAUNCH_LINK as ccache and distcc simply pass through link requests.  While this does invalid the "ccache -s" stats for link request, it does avoid the unneeded pass through calls (might be marginally faster).


facebook-github-bot-8:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D59517

PR #7158, opened 2016-06-15
Fix hf-prod-collect.sh for newer versions of perf

gdbentley:
Changed the perf script command to use "--fields" to be compatible with more versions of perf. The script was using "-f" which has been changed to "-F" in newer versions of perf. "--fields" means the same thing in both versions of perf.


facebook-github-bot-5:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D59661

PR #7162, opened 2016-06-15
ADD: gc_mem_caches() as wrapper to flush_thread_caches()

### LINKED ISSUE #7155: PHP7 support: Undefined function gc_mem_caches()
### HHVM Version
= 3.14.0

    # hhvm.php7.all=1
    echo '<?php echo gc_mem_caches();' | hhvm --php

### Expected result
some integer (or if this has no equivalent in hhvm, 0 and no side effects)

### Actual result
```
Fatal error: Uncaught Error: Call to undefined function gc_mem_caches() in /tmp/php-wrap-MI6Igh:1
Stack trace:
#0 {main}
```

See also: https://secure.php.net/manual/en/function.gc-mem-caches.php

apeabody:
Add gc_mem_caches() as wrapper to flush_thread_caches(), always returns 0

Closes #7155

### DIFF
--- a/hphp/runtime/ext/std/ext_std_gc.cpp
+++ b/hphp/runtime/ext/std/ext_std_gc.cpp
@@ -35,6 +35,10 @@ static int64_t HHVM_FUNCTION(gc_collect_cycles) {
   return 0; // seriously, count cycles?
 }
 
+static int64_t HHVM_FUNCTION(gc_mem_caches) {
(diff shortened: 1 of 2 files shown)


facebook-github-bot-0:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D59733

PR #7175, opened 2016-06-21
FIX: properly rtrim "output" of exec() including \f

### LINKED ISSUE #5096: exec() does not strip \f and \v from the end of the output
php5 seems to strip the \f ( 0x0C, aka NP form feed )  on the end of each line for the result of exec()
hhvm, the following code demonstrates that difference when executed in php5 and hhvm

exec's documentations specifies that whitespace on the end of the line is stripped, but there is a difference in the definition of 'whitespace'
 
<?php
	exec('/bin/echo -n -e "line1\x0Cmoreline1\r\nline2\x0C"', $result, $ret);
	var_dump($result);
	$content = implode($result,"\n");
	var_dump($content);
?>

apeabody:
-Fix the "output" of exec() to properly rtrim it's output of trailing whitespace.
-Mimics PHP's handling of exec() "output"
-Includes test case also tested against PHP 5.4.16

https://secure.php.net/manual/en/function.exec.php

Closes #5096
Closes #5107
Closes https://reviews.facebook.net/D36303


facebook-github-bot-7:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D59907

PR #7190, opened 2016-06-28
ADD: honor verify_peer_name in stream_context

### LINKED ISSUE #7189: verify_peer_name has no effect
### HHVM Version

HipHop VM 3.12.1 (rel)

```
$context = stream_context_create(array(
                'ssl' => array(
                    'verify_peer' => false,
                    'verify_peer_name' => false,
                    'allow_self_signed' => true
                )
            ));

echo file_get_contents('https://192.168.168.10/path/', false, $context);
```

### Expected result

Contents of the url

### Actual result

```
Warning: Failed to open https://self-signed-cert.domain/path/ (SSL: certificate subject name 'xxx' does not match target host name '192.168.168.10') in /srv/www/test.php on line 123
```

apeabody:
Honor verify_peer_name in stream_context
- Default is true based on behavior in PHP >= 5.6.0
- Sets CURLOPT_SSL_VERIFYHOST based on stream_context option verify_peer_name
- For CURLOPT_SSL_VERIFYHOST "TRUE" is actually 2
- If libcurl is built with NSS, verify_peer = FALSE will set verify_peer_name = FALSE

References:
- https://secure.php.net/manual/en/context.ssl.php
- https://curl.haxx.se/libcurl/c/CURLOPT_SSL_VERIFYHOST.html

Closes #7189

### DIFF
--- a/CMake/HPHPFindLibs.cmake
+++ b/CMake/HPHPFindLibs.cmake
@@ -15,6 +15,9 @@
 #   +----------------------------------------------------------------------+
 #
 
+#Increased minimum to 2.8.8 to support CURL_VERSION_STRING
+CMAKE_MINIMUM_REQUIRED(VERSION 2.8.8)
+
 include(CheckFunctionExists)
 
 # libdl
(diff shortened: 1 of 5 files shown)


facebook-github-bot-0:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D60129

Issue #7185, opened 2016-06-27
Fatal error: Uncaught Error: Call to undefined function gc_mem_caches()


donutloop:
HHVM Version 3.14.1 and Nightly

i used that "hhvm --no-config --php composer.phar install" command.

Stacktrace:

`Fatal error: Uncaught Error: Call to undefined function Symfony\Component\ClassLoader\gc_mem_caches() in /var/www/webapplication/vendor/symfony/symfony/src/Symfony/Component/ClassLoader/ClassCollectionLoader.php:198
Stack trace:
#0 /var/www/webapplication/vendor/symfony/symfony/src/Symfony/Component/ClassLoader/ClassCollectionLoader.php(109): Symfony\Component\ClassLoader\ClassCollectionLoader::fixNamespaceDeclarations()
#1 /var/www/webapplication/vendor/sensio/distribution-bundle/Composer/ScriptHandler.php(282): Symfony\Component\ClassLoader\ClassCollectionLoader::load()
#2 /var/www/webapplication/vendor/sensio/distribution-bundle/Resources/bin/build_bootstrap.php(59): Sensio\Bundle\DistributionBundle\Composer\ScriptHandler::doBuildBootstrap()
#3 {main}
Script Sensio\Bundle\DistributionBundle\Composer\ScriptHandler::buildBootstrap handling the post-install-cmd event terminated with an exception`

Best regards,
Marcel

Orvid:
The nightly for which OS and compiled when? 24bdc1458bdf7a0da00bd03ca09bb96bd5e978e8 already implemented the function.

donutloop:
OS: Debian 8 Jessie
i installed yesterday the nightly version.

`sudo apt-key adv --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 0x5a16e7281be7a449
echo deb http://dl.hhvm.com/debian jessie main | sudo tee /etc/apt/sources.list.d/hhvm.list
sudo apt-get update
sudo apt-get install hhvm-nightly`

jwatzman:
With the nightly installed, what's the output of `hhvm --version`?

@Orvid it looks like the builds are all broken, at least the internal cron checker has them all failing for like 8 days...

donutloop:
HipHop VM 3.14.0-dev (rel)
Compiler: heads/master-0-gb6c077acfd520ef0bed64407d134ed1bcbd1dd9c
Repo schema: cde63aaf1ca99dcbe9174acf40af08707725d957

[...]

PR #7194, opened 2016-06-29
ADD: test/slow for gc_mem_caches()

apeabody:
- VERY simple test to confirm gc_mem_caches() is present and properly returns the hardcoded value of int(0) after it executes flush_thread_caches().

- Hopefully reduce future confusion around gc_mem_caches()
Reference #7185


facebook-github-bot-6:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D60165

Issue #7178, opened 2016-06-22
Fatal error: Uncaught TypeError: Argument 1 passed to unserialize() must be an instance of string, null given


donutloop:
HHVM Version: 3.14.1 

Hello,
i loaded the binary (Latest Snapshot and 1.1.2) with wget. Then i tried to install with "hhvm --php composer.phar install" the Dependencies. Now i get these error:

#Fatal error: Uncaught TypeError: Argument 1 passed to unserialize() must be an instance of string, null given in /var/www/app/composer.phar:23
Stack trace:
#0 (): unserialize()
#1 (): __SystemLib\PharArchiveHandler->parsePhar()
#2 (): __SystemLib\PharArchiveHandler->__construct()
#3 (): Phar->__construct()
#4 /var/www/blog-api/composer.phar(23): Phar::mapPhar()
#5 {main}

Composer 
 #7171 

Best regards,
Marcel

Orvid:
@nazar-pc Could this be related in-part to the small refactor that was done while adding support for compressed Phars?

nazar-pc:
Theoretically it can be the case, because significant amount of code was refactored, but I'm not sure how exactly.

@donutloop, I've tried latest nightly build on Ubuntu 16.10 and I do not see this error on my machine.
Can you provide your `composer.json` and distro/version so that it is possible to reproduce exactly the same test case?

Orvid:
We have a nightly on 16.10? O_o....

donutloop:
PHP version (7.0.99-hhvm)
Debian version: jessie

donutloop:
composer.json

`{
  "name": "",
  "license": "MIT",
  "type": "project",
  "description": "",
  "autoload": {
    "psr-4": { "": "src/" },
    "classmap": [ "app/AppKernel.php", "app/AppCache.php" ]
  },
  "autoload-dev": {
    "psr-4": { "Tests\\": "tests/" }
  },
  "require": {
    "php": ">=7.0",
    "symfony/symfony": "3.1.*",
    "doctrine/orm": "^2.5",
    "doctrine/doctrine-bundle": "^1.6",
    "doctrine/doctrine-cache-bundle": "^1.2",
    "symfony/swiftmailer-bundle": "^2.3",
    "symfony/monolog-bundle": "^2.8",
    "sensio/distribution-bundle": "^5.0",
    "sensio/framework-extra-bundle": "^3.0.2",
    "nelmio/api-doc-bundle": "^2.12",
[...]

PR #7197, opened 2016-07-01
Check the Phar has metadata before attempting to unserialize() it

apeabody:
Phar metadata is optional with the metadata length set to 0 if not present.  This confirms patch checks that the metadata length is greater than 0 before attempting to unserialize() it.
- https://secure.php.net/manual/en/phar.fileformat.phar.php

This is in response to the listed issue in #7178, however there are other outstanding/unrelated type checking warnings when running in PHP7 scalar mode (null instead of string).


facebook-github-bot-6:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D60255

PR #7206, opened 2016-07-05
Fix compare double condition codes for arm64.

sunil-sk:
Summary:
When generating code for comparing floating points, use appropriate
condition codes for ARM to resolve ambiguity with "unordered"
comparisons.

Code generator that emits floating point comparisons for X64 uses only "A" (>) and "AE"(>=) 
condition codes. Inversion of these conditions are used to implement < and <=. This is done 
since other condition codes could be potentially set in "unordered" situation too.
 
However, ARM has condition codes that test >, >=, < and <= conditions without the 
"unordered" ambiguity. Also, the X64 approach used above leads to wrong results on ARM
since condition code conversion between X64 and ARM are not consistent across integer 
and floating point comparisons.

This PR fixes the ARM case by using CC_G, CC_GE, CC_B and CC_BE as these map to 
correct ARM conditions codes. (See ConvertCC() in abi-arm.h)

### DIFF
--- a/hphp/runtime/vm/jit/irlower-cmp.cpp
+++ b/hphp/runtime/vm/jit/irlower-cmp.cpp
@@ -194,12 +194,24 @@ void cgNeqDbl(IRLS& env, const IRInstruction* inst) {
  *
  * NB: This is clearly an x64-ism, and likely needs to be done differently for
  * other architectures.
+ *
+ * ARM has condition codes that test Gt, Gte, Lt, Lte without running into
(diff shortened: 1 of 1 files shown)


facebook-github-bot-1:

Orvid:
Let's do that for now while we figure out why the system isn't registering it.

Orvid:
While waiting for the system to catch up, I've imported it to the external phabricator manually for now: https://reviews.facebook.net/D60519

PR #7208, opened 2016-07-05
ext_gd: exif_process_IFD_TAG: Use the right offset if reading from stream

MatmaRex:
When the location of the data is outside of the range we have
preloaded (for example, if it's before the beginning of the IFD
structure), we have to read it from the stream into a separate buffer.
The offset calculations in this case were incorrect, resulting in
bogus values being read for the affected fields (sometimes parts of
other fields, sometimes binary data).

The included test image, sourced from [1], is in the public domain.

[1] https://commons.wikimedia.org/wiki/File:U.S._Marines_Prepare_to_board_an_MV-22_Osprey_160509-M-AF202-041.jpg

(This is the same fix as PHP commit c794d53c0377be960a17c3279715436e405b83f4 / php/php-src#1943.)


MatmaRex:
I have signed the CLA a couple minutes before submitting this pull request…

facebook-github-bot-5:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D60363

facebook-github-bot-5:

MatmaRex:
(For reference, I worked on this for <https://phabricator.wikimedia.org/T97253>.)

Orvid:
Yep, just took a bit to get it imported as I spent most of last week getting the OSS build of HHVM working correctly, which is why Github has a whole bunch of commits marked as being pushed on Saturday when I got the last of the issues resolved.

MatmaRex:
Thanks!

PR #7209, opened 2016-07-05
Remove HHVM version from 'X-Powered-By' header

raeesiqbal:
Ref #5691

### DIFF
--- a/hphp/runtime/server/transport.cpp
+++ b/hphp/runtime/server/transport.cpp
@@ -766,7 +766,7 @@ void Transport::prepareHeaders(bool compressed, bool chunked,
   }
 
   if (RuntimeOption::ExposeHPHP) {
-    addHeaderImpl("X-Powered-By", (String("HHVM/") + HHVM_VERSION).c_str());
+    addHeaderImpl("X-Powered-By", (String("HHVM")).c_str());
   }
 
   if ((RuntimeOption::ExposeXFBServer || RuntimeOption::ExposeXFBDebug) &&


facebook-github-bot-1:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D60369

simonwelsh:
I'm against this change as it stands. It means the header is no longer useful for automated scripts check if an upgrade is needed and also no longer matches PHP.

If anything, this should be controlled by an INI setting akin to Apache's ServerTokens.

raeesiqbal:
This already did not match PHP, PHP shows the language version but HHVM shows its own version rather than the PHP version.

raeesiqbal:
And exposing your version in the headers makes bad guys' job easy because whenever there is a security release for HHVM. They can just use 'automated scripts' to identify and target vulnerable websites.
And only the person who has access to the server should be able to check or be notified if an update is needed. Which is will be the case after this change.

simonwelsh:
PHP shows the engine version, not the language version. You can tell because the language (i.e. the spec) doesn't have patch releases. HHVM also shows the engine version.

Automated vulnerability detection doesn't care about what you say you're running (way too easy to remove/fake), it just tries everything. You've probably seen this in your access logs.

Targeted attacks are unlikely to trust any versions in headers either and they're much more likely to attack your code than the engine.

aorenste:
Sounds like we're not prepared to take this in it's current form - closing.

PR #7210, opened 2016-07-06
Add support to build Hack independently

raeesbhatti:
This change sets CMAKE_MODULE_PATH in hphp/hack/CMakeLists.txt. This is to allow Hack to be built independently from HHVM

Ref: hhvm/homebrew-hhvm#45


facebook-github-bot-2:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D60411

PR #7217, opened 2016-07-12
Make locations reported for parameter type errors more sensible

### LINKED ISSUE #7196: Invalid argument errors have wrong location reported
### HHVM Version
3.14.1 (rel) and everything else on 3v4l

```
<?php

class Foo {
	static function bar(array $baz) {}
}

Foo::bar('fail');
````

### Expected result
`Catchable fatal error: Argument 1 passed to Foo::bar() must be an instance of array, string given in file.php on line 7`

### Actual result
`Catchable fatal error: Argument 1 passed to Foo::bar() must be an instance of array, string given in file.php on line 4`

Reporting the location of the function being called instead of the caller makes finding the cause much harder. For the reference, PHP outputs a much more helpful `Argument 1 passed to Foo::bar() must be of the type array, string given, called in file.php on line 7 and defined in file.php on line 4`

[On 3v4l](https://3v4l.org/EtCml)

### PR DESCRIPTION
**Pull request by MaxSem**
Fixes #7196

### DIFF
--- a/hphp/runtime/base/runtime-error.cpp
+++ b/hphp/runtime/base/runtime-error.cpp
@@ -86,7 +86,7 @@ void raise_typehint_error(const std::string& msg) {
     VMRegAnchor _;
     SystemLib::throwTypeErrorObject(msg);
   }
-  raise_recoverable_error(msg);
+  raise_recoverable_error_without_first_frame(msg);
   if (RuntimeOption::RepoAuthoritative && Repo::global().HardTypeHints) {
     raise_error("Error handler tried to recover from typehint violation");
   }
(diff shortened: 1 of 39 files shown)

### CODE REVIEW

**Comment by facebook-github-bot-7** [comment]
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D60657

**Comment by fredemmott** [comment]
We've just discontinued use of reviews.facebook.net for HHVM - further review will be on GitHub: https://www.facebook.com/groups/hhvm.general/permalink/679913862162010/

I've just retriggered a test run for this.

**Comment by aorenste** [comment]
Most recent comment from phabricator:
```
It looks like there are multiple failures that you missed, you'll want to also test running the various tests in repo mode -r.
This is just one of the failing tests:
hphp/test:verify_slow_jit_repo - hphp/test/slow/serialization/repo_types_003.php: failed
```

PR #7222, opened 2016-07-13
Implement cvttsd2siq with X64 semantics on ARM64

sunil-sk:
Summary:
'cvttsd2siq' VASM instruction follows X64 semantics of returning
special error code upon float to int conversion errors. Follow
the same convention for ARM64.

Changes:
- Marked 'msr' instruction 'effectful' to prevent it from being optimized by DCE.
- Added encoding for FPSR in vixl.
- Added 'fcvtzs' ARM instruction to be used in 'cvttsd2siq'.
- Implemented 'cvttsd2siq' to be consistent with X64 behaviour.

### DIFF
--- a/hphp/runtime/vm/jit/vasm-arm.cpp
+++ b/hphp/runtime/vm/jit/vasm-arm.cpp
@@ -214,7 +214,6 @@ struct Vgen {
   void emit(const cmpq& i) { a->Cmp(X(i.s1), X(i.s0)); }
   void emit(const cmpqi& i) { a->Cmp(X(i.s1), i.s0.q()); }
   void emit(const cvtsi2sd& i) { a->Scvtf(D(i.d), X(i.s)); }
-  void emit(const cvttsd2siq& i) { a->Fcvtzs(X(i.d), D(i.s)); }
   void emit(const decl& i) { a->Sub(W(i.d), W(i.s), 1, SetFlags); }
   void emit(const decq& i) { a->Sub(X(i.d), X(i.s), 1, SetFlags); }
   void emit(const decqmlock& i);
@@ -301,6 +300,7 @@ struct Vgen {
   void emit(const cmplims& i);
   void emit(const cmpsds& i);
   void emit(const fabs& i) { a->Fabs(D(i.d), D(i.s)); }
+  void emit(const fcvtzs& i) {a->Fcvtzs(X(i.d), D(i.s));}
   void emit(const lslwi& i);
   void emit(const lslwis& i);
   void emit(const lslxi& i);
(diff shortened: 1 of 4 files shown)


facebook-github-bot-3:

facebook-github-bot-0:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D60729

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -1257,6 +1257,35 @@ Y(incwm, incw, loadw, storew, m)
...
+  auto res = v.makeReg();
+  auto const sf = v.makeReg();
+  auto err = v.makeReg();
+  auto tmp1 = v.makeReg();
+  auto tmp2 = v.makeReg();
+
+  // Clear FPSR IOC flag
+  v << mrs{FPSR, tmp1};
+  v << andqi{~0x01, tmp1, tmp2, sf};
+  v << msr{tmp2, FPSR};
```
edwinsmith: will it be a big penalty to modify FPSR every time?

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -1257,6 +1257,35 @@ Y(incwm, incw, loadw, storew, m)
...
+  // Do ARM64's double to signed int64 conversion.
+  v << fcvtzs{i.s, res};
+
+  // Check if there was a conversion error
+  v << mrs{FPSR, fpsr};
+  v << testqi{1, fpsr, sf};
+
+  // Move converted value or error
+  v << cmovq{CC_NZ, sf, res, err, i.d};
+  });
```
edwinsmith: While this lowering is certianly fine for getting the port working, the HHIR->VASM lowering code for double->int conversion is probably too dependent on the behavior of X64's cvttsd2siq for the edge cases.

IMHO it would be acceptable to do an arch() check before emitting cvttsd2siq, and just generate different code to check/handle the error conditions. (but cc @mxw)

a nicer solution would be to define a higher level vasm instruction then lower it customly for x64/ppc/arm, however I don't have any good suggestions for how to define that instruction. it could encapsulate *all* the error handling, but that feels too high-level and PHP dependent. and it seems hard to abstractly define the way it would signal errors.

hphp/runtime/vm/jit/vasm-arm.cpp
sunil-sk: Thanks for the review.
AFAIK, there should not be any extra penalty other than the usual cost of executing the register copy. It is possible to just write zero to FPSR instead of read-modify-write to clear the IOC bit (bit 0) alone.

hphp/runtime/vm/jit/vasm-arm.cpp
sunil-sk: Agree.

hphp/runtime/vm/jit/vasm-arm.cpp
apinski-cavium: Actually there can be a huge penalty depending on the processor.

facebook-github-bot-8:

hphp/runtime/vm/jit/vasm-arm.cpp
sunil-sk: OK, thanks for the inputs. Then, this may need some rework along the lines of what Edwin has suggested.

PR #7227, opened 2016-07-14
[PPC64] Use rsfp, rtoc and rfuncentry instead of r1, r2 and r12

gut:


facebook-github-bot-0:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D60801

PR #7234, opened 2016-07-18
Fix compiler warnings

gut:
- signed vs unsigned comparison (int vs size_t)
- not using return value of getcwd.


facebook-github-bot-6:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D60867

gut:
abandoning this change as it'll be sent internally by FB

PR #7242, opened 2016-07-21
PDOEntity Interface to handle column mapping PDO::FETCH_CLASS

tugrul:
There were no feature to handle column mapping on set object properties.

Entity classes able to handle column mapping with implement \PDOEntity interface. In other case classic behavior.

So it is able to hydrate mixed resultset like of JOIN queries.

``` hack
class User implements \PDOEntity
{
    protected static $map = ['first_name' => 'firstName'];

    public $firstName;
    public $lastName;

    public function setField(string $name, mixed $value, array $meta)
    {
        // $name = static::$map[$name]; or..
        $name = static::$map[$meta['org_name']];
        $this->$name = $value;
    }
}


$pdo = new \PDO('mysql:host=localhost;dbname=sakila;charset=utf8', 'root', 'n0tr341');


$stmt = $pdo->query('SELECT `first_name` AS FN, `last_name` AS LN `users` MT');
$stmt->setFetchMode(\PDO::FETCH_CLASS, 'User');


$row = $stmt->fetch();
var_dump($row);
```


facebook-github-bot-7:

facebook-github-bot-6:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D61011

facebook-github-bot-6:

hphp/runtime/ext/pdo/ext_pdo.cpp
```diff
@@ -2037,7 +2038,31 @@ static bool do_fetch(sp_PDOStatement stmt,
...
+          if (!stmt->getColumnMeta(idx, meta)) {
+            PDO_HANDLE_STMT_ERR(stmt);
+            return false;
+          }
+            
+          Array params = Array::Create();
+          params.append(name);
+          params.append(val);
+          params.append(meta);
+          ret.asCObjRef()->o_invoke("setfield", params);
```
paulbiss: https://github.com/facebook/hhvm/blob/14118b4d51011c21b6d026420d4a99c18fbb01cb/hphp/runtime/base/object-data.h#L258-L263

hphp/runtime/ext/pdo/ext_pdo.php
```diff
@@ -1,6 +1,15 @@
 <?hh
 // generated by idl-to-hni.php
 
+/**
+ * Objects implementing PDOEntity can handle value by column name and meta
+ */
+interface PDOEntity {
```
paulbiss: I see that this feature is not yet in PHP either. Has there been any discussion on the internals list about it? I'm not opposed to us adding it without waiting for PHP support, though if they end up using a different interface we will need to change ours to match what they implement. Until this feature is in PHP we will probably want this interface to live in the `HH` namespace.

I'm not entirely sure I follow the rationale in your commit summary, could you add some more detail?

https://bugs.php.net/bug.php?id=69016

hphp/runtime/ext/pdo_mysql/pdo_mysql.cpp
```diff
@@ -1281,7 +1285,11 @@ bool PDOMySqlStatement::getColumnMeta(int64_t colno, Array &ret) {
     ret.set(s_native_type, str);
   }
   ret.set(s_flags, flags);
+  ret.set(s_name, String(F->name, CopyString));
+  ret.set(s_org_name, String(F->org_name, CopyString));
   ret.set(s_table, String(F->table, CopyString));
+  ret.set(s_org_table, String(F->org_table, CopyString));
+  ret.set(s_length, F->length);
```
paulbiss: How are these changes related?

tugrul: `org_name`, `org_table` fields look like available since MySQL 4.0.3 merge. https://github.com/mysql/mysql-server/commit/7134ffec210edde21860a2b2c2654be481de49b4#diff-6391e1a12cc78a7dcd7a52a1fbb713fcR70

However not implemented in PHP yet. We can decision to implement what name used for this fields. There are some choices in my mind.

- Leave the property names like now.
- Change property names to common name like `original_name`, `original_table`
- Add prefix to property names like `hh_org_name`, `hh_org_table`
- Else?

But it should not matter because PHP documentation says that:
> This function is EXPERIMENTAL. The behaviour of this function, its name, and surrounding documentation may change without notice in a future release of PHP. This function should be used at your own risk.

http://php.net/manual/en/pdostatement.getcolumnmeta.php

hphp/runtime/ext/pdo/ext_pdo.php
tugrul: I don't have conversation with anybody who in PHP team. So I don't know what idea on the internals.

I have opened feature request last year. There is no action about this issue from PHP team.

Yes we can move into HH namespace to avoid from hubbub for future situation.

I'm going to move into HH namespace if your decision about this way.

hphp/runtime/ext/pdo/ext_pdo.cpp
tugrul: >TODO Task #2584896: o_invoke and o_invoke_few_args are deprecated. 
>These APIs don't properly take class context into account when looking up the method ...

Where can i see task #2584896 to understand problem. `o_invoke` looks like me too easy and best helper. I can't imagine a structure to call invokeFunc method because it needs too verbose informations. Do `Func` and `Class` already belong to `ObjectData`? Why i need to determine references their seperate ways? Late static calls?

So I don't have an idea about `VarEnv* varEnv = nullptr, StringData* invName = nullptr, InvokeFlags flags = InvokeNormal`

Or can you help me for this? I allowed maintainers to edit

paulbiss: o_invoke looks simple because it's lacking important inputs, hence its deprecation. Sorry, internal task number, but it says exactly what's in the comment, there's no extra context there.

The defaults for those params (null, null, and normal) should be fine for this use case. The key difference is that you'll need to setup a CallCtx to make the call.

hphp/runtime/ext/pdo/ext_pdo.php
paulbiss: I think if we stash this in the HH namespace and update the comment to note that it's experimental I'll be fine with adding it. I do have two requests though, (1) add the link to the PHP bug in a comment explaining that this implementation is subject to whatever the PHP community ultimately does, and (2) send an email to the PHP internals mailing list linking to this pull request and explaining your change (just to see if anyone over there wants to pick it up).

hphp/runtime/ext/pdo_mysql/pdo_mysql.cpp
paulbiss: This feels like a breaking change, is there any way that we could do this that would allow the default behavior to continue matching PHP? Also is there another bug filed against PHP about this? Has there been any discussion of it?

One possible option here would be to add an optional param to this function which indicate that we want these extra fields added. Another would be to only add them in HH mode. I'm not sure if we really need to bother with those though, maybe this isn't a big deal.

mofarrell:
This PR has been waiting for more info for a follow up for a while now.  I'm closing it for now, but feel free to re-open it if you respond to the review comments.

Issue #7133, opened 2016-06-07
Slow string operations in tight loops (ucwords, strtoupper, strtolower)


daxtens:
Calling built in string operations such as `ucwords`, `strto(upper|lower)` and friends are incredibly slow in tight loops.

### HHVM Version
```
HipHop VM 3.14.0-dev (rel)
Compiler: heads/master-0-g6566f01044cda3bec18ac2c4cc8f1d54e7363f34
Repo schema: 1cd15dba5ef1cae0e98620603280c5eedc5e6a37
```
#### PHP 5 version
```
PHP 5.6.4-4ubuntu6.4 (cli) (built: Oct 28 2015 01:21:29) 
Copyright (c) 1997-2014 The PHP Group
Zend Engine v2.6.0, Copyright (c) 1998-2014 Zend Technologies
    with Zend OPcache v7.0.4-dev, Copyright (c) 1999-2014, by Zend Technologies
```
#### PHP 7 version
```
PHP 7.0.4-7ubuntu2.1 (cli) ( NTS )
Copyright (c) 1997-2016 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2016 Zend Technologies
    with Zend OPcache v7.0.6-dev, Copyright (c) 1999-2016, by Zend Technologies
```

### Standalone code, or other way to reproduce the problem

```
<?php
function ouch($count = 900000) {
        $time_start = microtime(true);
        $string = str_repeat("the quick brown fox jumps over the lazy dog", 100);
        for ($i=0; $i < $count; $i++) {
                $r = ucwords($string);
        }
        return number_format(microtime(true) - $time_start, 3);
}

for ($i=0; $i<2; $i++) {
        print("Time taken (run $i): " . ouch() . "\n");
}

```

### Expected result

HHVM runs no slower than php5.

### Actual result

HHVM runs several times slower than php5.

#### PHP 5

```
Time taken (run 0): 17.947
Time taken (run 1): 15.178
```

#### HHVM

```
hhvm/hhvm -vEval.Jit=true ~/hhvm.php 
Time taken (run 0): 56.302
Time taken (run 1): 55.826
```

#### PHP 7
````
php ~/hhvm.php 
Time taken (run 0): 18.485
Time taken (run 1): 16.711
````

jwatzman:
[...]

PR #7246, opened 2016-07-23
Replace bitset with string_charmask() in ucwords.

wjywbs:
Also use the same logic in php to process the string. Overall this is 30% faster
than the bitset version.

Part of #7133.


facebook-github-bot-3:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D61095

PR #7250, opened 2016-07-25
Fix ext/soap find_encoder_by_type_name()

y3llowcake:
This function is incorrectly comparing fully name spaced wsdl type names (e.g. urn:blah.thing.TypeName) to top level type names (e.g. just "TypeName").

Compare with the equivalent Zend function here: https://github.com/php/php-src/blob/master/ext/soap/php_encoding.c#L272

I'm not sure when this bug was introduced, but it seems like it's been around for a while.

### DIFF
--- a/hphp/runtime/ext/soap/encoding.cpp
+++ b/hphp/runtime/ext/soap/encoding.cpp
@@ -451,9 +451,10 @@ void whiteSpace_collapse(xmlChar* str) {
 
 static encodePtr find_encoder_by_type_name(sdl *sdl, const char *type) {
   if (sdl) {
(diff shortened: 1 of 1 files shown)


facebook-github-bot-6:

facebook-github-bot-9:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D61107

facebook-github-bot-0:

kmaslack:
Just to add some color here, this fixes a real bug for anyone trying to use Netsuite to handle invoicing. And my suspicion is that Netsuite is a fair percentage of the entire use case of SOAP clients in PHP in the year 2016.

PR #7252, opened 2016-07-26
ADD: escape character to return array of SplFileObject::getCsvControl()

### LINKED ISSUE #7249: SplFileObject::getCsvControl does not return the escape character
`SplFileObject::setCsvControl` takes 3 parameters 
- the delimiter character
- the enclosure character
- the escape character

But the equivalent `SplFileObject::getCsvControl` returns an array with only two of those parameters. The escape character is missing.

> The same bug was also reported to PHP bug system and it is already resolved 
https://bugs.php.net/bug.php?id=72646
This bug fix will lend on the next stable version of PHP 5.6.25 and 7.0.10
The [php.net documentation was also updated accordingly](http://be2.php.net/manual/en/splfileobject.getcsvcontrol.php)

### HHVM Version

all

```php
<?php

$file= new SplTempFileObject();
$file->getCsvControl();
print_r($controls);
```
### Expected result

```php
<?php
Array
(
 [0] => ,
 [1] => "
 [2] => \
)
```
### Actual result

```php
<?php
Array
(
 [0] => ,
 [1] => "
)
```

apeabody:
Add the escape character to the return array of SplFileObject::getCsvControl() matching PHP 5.6.25 and 7.0.10 behavior
- https://secure.php.net/manual/en/splfileobject.getcsvcontrol.php
- Includes new test case

Closes #7249


facebook-github-bot-3:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D61203

PR #7257, opened 2016-07-27
Inform vtune, gdb, and perf of additional generated code.

gdbentley:
The unique stubs and pcre jitted code were not reported to any of the debugging or profiling tools, so addresses rather than symbolic names appeared. Now all executable addresses should be mapped to function names.


facebook-github-bot-9:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D61251

PR #7267, opened 2016-08-05
setting jit prefix to be the correct width

drewklein:
based on http://lxr.free-electrons.com/source/tools/perf/util/jitdump.h#L57 this field should be 32 bits, not 8, this allows more complete perf analysis


facebook-github-bot-7:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D61587

PR #7269, opened 2016-08-05
Add two cotire header exclusions

tstarling:
* Exclude resolv.h from the cotire prefix header. This fixes a compile
  error in thread-local.cpp due to resolv.h defining a p_type macro,
  which conflicts with the p_type member of ElfW(Phdr). Perhaps this
  conflict is the reason why elf.h was already excluded.
* Exclude the XED headers, since they need to be wrapped with extern "C"
  and cotire doesn't do that.


facebook-github-bot-4:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D61611

PR #7275, opened 2016-08-09
Fix the jsonc link.

kmaslack:
cmake -DUSE_JSONC . fails to link without these symbols defined.


facebook-github-bot-6:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D61785

PR #7285, opened 2016-08-19
Fix SIMD bug on PPC64

brunoalr:
Revert loadsCell to previous behavior.


facebook-github-bot-4:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D62283

PR #7289, opened 2016-08-24
Fix cmp to test vasm simplication for PPC64

gut:
The improvement implemented on fa442471676c9550691c28fe31c9fee2edda0a05 is
arch-specific as a test may not be better than a cmp on every arch. So it
should actually be opt-in for other archs to make use of it.

Moreover PPC64 couldn't handle the test vasm after it's lower phase, therefore
a second call to lower was added to guarantee that the vasms emitted by
simplify and optimizeCopies are actually handled by PPC64 emitter. 

No significant overhead was noticed.

### DIFF
--- a/hphp/runtime/vm/jit/vasm-ppc64.cpp
+++ b/hphp/runtime/vm/jit/vasm-ppc64.cpp
@@ -1581,8 +1581,10 @@ void optimizePPC64(Vunit& unit, const Abi& abi, bool regalloc) {
   lowerForPPC64(unit);
 
   simplify(unit);
-
   optimizeCopies(unit, abi);
+
+  // 2nd lower call to affect new vasms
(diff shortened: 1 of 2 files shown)


facebook-github-bot-4:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D62463

ricklavoie:
Its not so much that test is faster than cmp, but that it encodes in less bytes, increasing code density.

gut:
thanks for pointing it out. I didn't know the exact reason, just that X64's backend devs prefers it.

gut:
Is this "fb/hphp-buck — Tests Failed" a concern?

Orvid:
Nope, it's just a flaky internal test.

PR #7299, opened 2016-08-29
Fix relocation adjustment when a stub is reused

gut:
Service-request's emit_ephemeral can emit code by reusing stub and as
nothing new is emitted on main, cold or frozen, it would not be tracked
for adjustment.


gut:
This bug was reproducible by running oss-performance's wordpress benchmark on ppc64. There the JIT was jumping to a location with `0xcc` content.

gut:
@markw65 : Any comments on this one?

gut:
I just rebased this change and after paulbiss's changes this doesn't seem to be happening anymore, hence no patch needed. It was somehow fixed on his refactor.
Thanks

PR #7301, opened 2016-08-30
[PPC64] Fix segfault when using ActRec

lbianc:
As PPC64 has one frame closed in advance, the builtin function returns the
current frame when the one being searched is the previous, which is the
first after the last VM frame.

### DIFF
--- a/hphp/util/portability.h
+++ b/hphp/util/portability.h
@@ -162,7 +162,7 @@
 #  error Clang implementation not done for PPC64
 # endif
 # define DECLARE_FRAME_POINTER(fp) \
-  auto const fp = (ActRec*) __builtin_frame_address(0)
+  auto const fp = ((ActRec*) __builtin_frame_address(0))->m_sfp
 # define FRAME_POINTER_IS_ACCURATE
 
 #else


facebook-github-bot-4:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D62889

Orvid:
@fredemmott I thought the sync workflow was disabled?

gut:
It was correct on target-cache as it's loading the content of the frame pointer (follow backchain once) to the destiny register, and not just copying it:
https://github.com/facebook/hhvm/blob/c6b96eab21be48f88d19241aab5b78b482aacb05/hphp/runtime/vm/jit/target-cache.cpp#L495

But the one you pointed out was missing. Too bad it was only detected by running oss-performance's drupal 8 benchmark.

markw65:
Can you post the backtrace for the crash? Also, I can't understand the explanation you give in the summary... DECLARE_FRAME_POINTER is used in a number of places - how can you possibly know where the last vm frame is?

I'm concerned that this is going to break VMRegGuard - unless this is to specifically fix VMRegGuard - in which case I'd really like to see a picture of the state of the stack and frames, and how things went wrong...

lbianc:
This is the backtrace when crashing:

```
frame 0x3fff7dff86a0 => 0x15220094 <HPHP::Class::classofNonIFace(HPHP::Class const*) const+128>:        lwz     r9,244(r9)
frame 0x3fff7dff86f0    0x15220250 <HPHP::Class::classof(HPHP::Class const*) const+280>:        nop
frame 0x3fff7dff8760    0x1530d864 <HPHP::ExecutionContext::lookupMethodCtx(HPHP::Class const*, HPHP::StringData const*, HPHP::Class const*, HPHP::CallType, bool)+1204>:       nop
frame 0x3fff7dff8840    0x15547c80 <HPHP::vm_decode_function(HPHP::Variant const&, HPHP::ActRec*, bool, HPHP::ObjectData*&, HPHP::Class*&, HPHP::StringData*&, bool)+4336>:     nop
frame 0x3fff7dff8a10    0x16c86714 <HPHP::jit::decodeCufIterHelper(HPHP::Iter*, HPHP::TypedValue)+384>: nop
frame 0x3fff7dff8aa0    0x1b569e30:     ld      r2,24(r1)
frame 0x3fff953bf640    0x22e00258:     std     r3,0(r29)
frame 0x3fff953bf710    0x22e00258:     std     r3,0(r29)
frame 0x3fff953bf760    0x1fab9e18:     nop
frame 0x3fff953bf820    0x1f9d601c:     nop
frame 0x3fff953bf890    0x1f9d2164:     nop
frame 0x3fff953bf950    0x1f9a39d4:     nop
frame 0x3fff953bf9e0    0x1b2008b8:     addi    r27,r27,8
frame 0x3fff7dff8ad0    0x16ecef08 <HPHP::jit::enterTCImpl(unsigned char*, HPHP::ActRec*)+404>: ld      r2,24(r1)
frame 0x3fff7dff8af0    0x16ecef08 <HPHP::jit::enterTCImpl(unsigned char*, HPHP::ActRec*)+404>: ld      r2,24(r1)
frame 0x3fff7dff8c80    0x168d1734 <HPHP::jit::MCGenerator::enterTC(unsigned char*, HPHP::ActRec*)+1140>:       nop
frame 0x3fff7dff8d00    0x152fcc8c <HPHP::jit::MCGenerator::enterTCAfterPrologue(unsigned char*)+260>:  nop
frame 0x3fff7dff8d70    0x15bb7640 <HPHP::enterVMAtFunc(HPHP::ActRec*, HPHP::StackArgsState, HPHP::VarEnv*)+1328>:      nop
frame 0x3fff7dff8e40    0x153115e8 <HPHP::ExecutionContext::<lambda(HPHP::ActRec*)>::<lambda()>::operator()(void) const+116>:   nop
...
```

This issue there is that the `m_func` information, which is in the `ActRec` structure, is in the last VM frame (`0x3fff953bf640`), not in the `0x3fff7dff8aa0` (native frame).

After this change, all quick tests are still passing and now drupal8 as well.

I wrote "last VM frame" because for the current issue this was the case, I do not mean it will happen for all cases that use the `DECLARE_FRAME_POINTER`.

If you have a situation that this change can break, please, let me know that I will check.

markw65:
The reason I think it might break things is that you're changing the meaning of DECLARE_FRAME_POINTER *everywhere* because of one use thats apparently broken.

And I still don't understand why its broken. decodeCufIterHelper is called directly from the tc, so the frame outer to that should be the last vm frame. Can you explain why it isn't? Is there always an extra, non-vm frame on ppc between the last vm frame, and the first c++ function (thats the only explanation I can think of)? If so, this /still/ seems like the wrong fix - I think its all gone now, but we have had code in the past that inspected the current c++ frame using DECLARE_FRAME_POINTER... thats obviously not going to work with your change...

gut:
@markw65 : Hi it's been a while, but on Jan 29th 2016, we had the following discussion on \#hhvm-dev:
```
< mwilliams> gut_: oh - does the frame pointer point below the frame on ppc?
< gut_> mwilliams: yes, fixupWork (112) is bigger than fixup (48)
< mwilliams> gut_: right, but doesn't the frame pointer point at the top (like it does on x64)
< mwilliams> gut_: the frame size doesn't matter on x64. if a tail calls b, their frames will be at the same address, regardless of the sizes...
< mwilliams> gut_: ok, just found a description of the ppc stackframe layout, and yes, thats the problem. the frame pointer points to the "bottom" of the frame (ie       it includes all the locals etc) rather than the top                               
```

I guess this is where this difference came from: X64's frame pointer (rbp) points to the end of the current frame whereas PPC's stack frame pointer (r1) points to the beginning of the current frame.

We're still shocked how we didn't come across this problem earlier as it seems to affect several tests, but they were all working good.

lbianc:
Thanks @gut!

@markw65, as @gut said, all tests are working, and as PPC64 has this difference when compared with x64, `DECLARE_FRAME_POINTER` need to return the previous frame.
Let me know if you want some specific test or anything else.

lbianc:
@mxw @markw65 ^

lbianc:
@mxw this PR has on check pending since I have created it, is there any issue with that?

mxw:
I think we discussed this on IRC, but to leave a paper trail—the correct solution here is to stop assuming that this code sequence, when found in a C++ helper called directly from the JIT:

```cpp
DECLARE_FRAME_POINTER(fp);
auto const ar = fp->m_sfp;
```

will obtain the caller PHP frame.  We could create a separate helper routine to obtain the PHP frame in this case, which does `fp->m_sfp` on x64 and ARM, while doing `fp->m_sfp->m_sfp` on ppc64, but a better solution is to just to pass in the ActRec* as an argument to the one C++ helper which uses this code sequence.  This has a nonzero cost, since it adds an extra reg-reg move to the TC for every such call, but that's probably fairly negligible.

mxw:
(I'll put up that fix myself and test internally, and then will close out this diff.  There's no more action needed here, except from me.)

mxw:
Fixed on master by passing `rvmfp()` as an argument to `decodeCufIterHelper()`.

PR #7313, opened 2016-09-03
[AArch64] Favors use of toInt64(double) helper routine.

dave-estes:
The behavior when casting doubles with values of min/max uint64_t
is unspecified and therefore can vary amongst architectures.
toInt64(double) in type-conversion.h avoids this problem.

Fixes test/slow/ext_array/1766.php


facebook-github-bot-8:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D63237

PR #7314, opened 2016-09-03
Properly re-attaches the LightProcess signal action after exec.

dave-estes:
LightProcess registers a sig action handler with sigaction(), and therefore it must be restored through sigaction in the destructor of ShellExecContext. This patch implements that as well as
updating ShellExecContext to use sigaction when it registers the default signal handler in the contructor.

All platforms were affected by this bug to the effect of LightProcess::SigChldHandler() was being called with only one argument and trash in the other two.

Fixes test/slow/ext_process/lwp.php for AArch64.

### DIFF
--- a/hphp/runtime/ext/std/ext_std_process.cpp
+++ b/hphp/runtime/ext/std/ext_std_process.cpp
@@ -193,7 +193,7 @@ struct ShellExecContext final {
     // saving the previous signal action.
     struct sigaction sa = {};
     sa.sa_handler = SIG_DFL;
-    if (sigaction(SIGCHLD, &sa, &old_sa) != 0) {
+    if (sigaction(SIGCHLD, &sa, &m_old_sa) != 0) {
       Logger::Error("Couldn't install SIGCHLD handler");
       abort();
     }
@@ -207,8 +207,8 @@ struct ShellExecContext final {
       LightProcess::pclose(m_proc);
 #endif
     }
-    if (old_sa.sa_handler != SIG_DFL) {
-      if (sigaction(SIGCHLD, &old_sa, nullptr) != 0) {
+    if (m_old_sa.sa_handler != SIG_DFL) {
+      if (sigaction(SIGCHLD, &m_old_sa, nullptr) != 0) {
         Logger::Error("Couldn't restore SIGCHLD handler");
         abort();
       }
@@ -250,7 +250,7 @@ struct ShellExecContext final {
   }
 
 private:
-  struct sigaction old_sa = {};
+  struct sigaction m_old_sa = {};
   FILE *m_proc{nullptr};
 };
 
--- a/hphp/runtime/ext/std/ext_std_process.cpp
+++ b/hphp/runtime/ext/std/ext_std_process.cpp
@@ -191,7 +191,12 @@ struct ShellExecContext final {
   ShellExecContext() {
     // Use the default handler while exec'ing in a shell,
     // saving the previous signal action.
-    m_sig_handler = signal(SIGCHLD, SIG_DFL);
+    struct sigaction sa = {};
+    sa.sa_handler = SIG_DFL;
+    if (sigaction(SIGCHLD, &sa, &old_sa) != 0) {
+      Logger::Error("Couldn't install SIGCHLD handler");
+      abort();
+    }
   }
 
   ~ShellExecContext() {
@@ -202,14 +207,11 @@ struct ShellExecContext final {
       LightProcess::pclose(m_proc);
 #endif
     }
-    if (m_sig_handler) {
-      // If we saved the previous signal action, then re-attach it now.
-      // Note: We're using LightProcess::AttachHandler() for this since
-      //       instead of the the signal(SIGCHLD, m_sig_handler) that's
-      //       typically seen in example code because we're re-attaching
-      //       a signal action and not a simple signal handler, and the
-      //       latter approach only supports simple signal handlers.
-      LightProcess::AttachHandler(m_sig_handler);
+    if (old_sa.sa_handler != SIG_DFL) {
+      if (sigaction(SIGCHLD, &old_sa, nullptr) != 0) {
+        Logger::Error("Couldn't restore SIGCHLD handler");
+        abort();
+      }
     }
   }
 
@@ -248,7 +250,7 @@ struct ShellExecContext final {
   }
 
 private:
-  void (*m_sig_handler)(int);
+  struct sigaction old_sa = {};
   FILE *m_proc{nullptr};
 };
 
--- a/hphp/runtime/ext/std/ext_std_process.cpp
+++ b/hphp/runtime/ext/std/ext_std_process.cpp
@@ -189,6 +189,8 @@ namespace {
 
 struct ShellExecContext final {
   ShellExecContext() {
+    // Use the default handler while exec'ing in a shell,
+    // saving the previous signal action.
     m_sig_handler = signal(SIGCHLD, SIG_DFL);
   }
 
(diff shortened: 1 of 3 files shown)


facebook-github-bot-7:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D63249

dave-estes:
Just wanted to post and update. This is an old patch that we've had laying around. Previously, it looked like a failure in libc. While this patch still "fixes" the issue we're seeing, the issue is actually manifesting itself in a weird and different way. I think we missed the true root caused when we fixed it earlier. We're still looking into this. Will keep you posted.

In the meantime, I wonder if @lpathy or others are seeing a failure in test/slow/ext_process/lwp.php.

mxw:
Can you rename this PR?  "You" here refers to anybody who's commented so far.

hphp/runtime/ext/std/ext_std_process.cpp
```diff
@@ -201,7 +203,13 @@ struct ShellExecContext final {
 #endif
     }
     if (m_sig_handler) {
```
mxw: This doesn't seem right... why are we ignoring the old handler value?

dave-estes: These handlers are of the type signalhandler_t, which is a function pointer.

    typedef void (*__sighandler_t) (int);

The interesting thing is that they can be attached as a plain signal or as a `sigaction`. In the case of `light-process`, it's attached as a `sigaction`. `sigaction`s take 3 parameters and not one. Here's the `sigaction` structure that holds the details about an attached signal or `sigaction`

    struct sigaction {
      void     (*sa_handler)(int);
      void     (*sa_sigaction)(int, siginfo_t *, void *);
      sigset_t   sa_mask;
      int        sa_flags;
      void     (*sa_restorer)(void);
    };

Some platforms will union the `sa_handler` and `sa_action` above since they're mutually exclusive. So when the `ShellExecContext` ctor temporarily attaches the generic handler with `signal(SIGCHLD, SIG_DFL)`, the old handler that is returned is a function pointer to `LightProcess::SigChldHandler()`, which is a `sigaction` taking three parameters. This means that we can't restore this handler using `signal(SIGCHLD, m_sig_handler)`, because doing so would then register that function pointer as a simple `signal` and not a `sigaction`. So what's happening here is that `LightProcess::SigChldHandler()` is being called as if it were a simple `signal` with only one parameter instead of three, leading to garbage in the other two argument registers being used.

Sorry if you already got the above, but I'm finally answering your question now. :-) We can use the old handler value here, but the way to use it is to do something like the following:

    struct sigaction sa = {};
    sa.sa_sigaction = m_sig_handler;
    sa.sa_flags = SA_SIGINFO | SA_NOCLDSTOP;
    if (sigaction(SIGCHLD, &sa, nullptr) != 0) {
      Logger::Error("Couldn't install SIGCHLD handler");
      abort();
    }

This is basically a duplicate of the code in `light-process` that attached the `sigaction` originally. We just removed duplication by putting it in a new member method, `void LightProcess::AttachHandler(void)`. It's effectively the same thing as using the old handler value because `m_sig_handler` equals `&LightProcess::SigChldHandler`.

mxw: Right—that last sentence is what I was asking about.  This works _iff_ that's true.  You should either assert that it is, or clarify the behavior when it isn't—i.e., we always override a competing installed sigaction, or we never do, or whatever the appropriate behavior is.

aorenste:
Errors:
```
In file included from folly/Format.h:433:0,
                 from hphp/util/assertions.h:28,
                 from hphp/util/mutex.h:37,
                 from hphp/util/lock.h:20,
                 from hphp/util/light-process.h:34,
                 from hphp/util/light-process.cpp:16:
folly/Format-inl.h: In instantiation of 'void folly::FormatValue<T*, typename std::enable_if<((! std::is_same<char, typename std::decay<_Tp>::type>::value) && $
folly/Format.h:195:5:   required from 'void folly::Formatter<containerMode, Args>::doFormatArg(folly::FormatArg&, Callback&) const [with long unsigned int K = $
folly/Format.h:119:7:   required from 'typename std::enable_if<(K < folly::BaseFormatter<Derived, containerMode, Args>::valueCount)>::type folly::BaseFormatter$
folly/Format.h:127:24:   required from 'void folly::BaseFormatter<Derived, containerMode, Args>::doFormat(size_t, folly::FormatArg&, Callback&) const [with Cal$
folly/Format-inl.h:272:32:   required from 'void folly::BaseFormatter<Derived, containerMode, Args>::operator()(Output&) const [with Output = folly::BaseFormat$
folly/Format.h:77:21:   required from 'typename std::enable_if<folly::IsSomeString<Str>::value>::type folly::BaseFormatter<Derived, containerMode, Args>::appen$
folly/Format.h:85:15:   required from 'std::string folly::BaseFormatter<Derived, containerMode, Args>::str() const [with Derived = folly::Formatter<false, void$
hphp/util/light-process.cpp:442:303:   required from here
folly/Format-inl.h:800:43: error: pointer to a function used in arithmetic [-Werror=pointer-arith]
cc1plus: all warnings being treated as errors
```

hphp/util/light-process.cpp
```diff
@@ -436,6 +436,23 @@ void LightProcess::SigChldHandler(int sig, siginfo_t* info, void* ctx) {
   }
 }
 
+void LightProcess::AttachHandler(sighandler_t sighandler)
+{
```
mxw: Move this to the same line as the above.

hphp/util/light-process.cpp
```diff
@@ -436,6 +436,23 @@ void LightProcess::SigChldHandler(int sig, siginfo_t* info, void* ctx) {
...
+  if (sighandler) {
+    always_assert_flog(sighandler == (sighandler_t)&SigChldHandler,
```
mxw: This seems even more circuitous than before.  If the argument always has to be a constant value, just don't take an argument...

dave-estes: I see what you mean, but the comparison has to be done within the LightProcess class since it is the only class that know explicitly about its SigChldHandler. So I added this parameter which can be null or a sighandler_t. If it is a sighandler_t (not null), then the assert is there to ensure that we're only attaching SigChldHandler again if the saved handler was the same. If it is null, then we're attaching for the first time in LightProcess which won't check.

I'm happy to try something a different approach if you have any ideas.

hphp/runtime/ext/std/ext_std_process.cpp
```diff
@@ -189,6 +189,8 @@ namespace {
 
 struct ShellExecContext final {
   ShellExecContext() {
+    // Use the default handler while exec'ing in a shell,
+    // saving the previous signal action.
     m_sig_handler = signal(SIGCHLD, SIG_DFL);
```
mxw: Can you use `sigaction()` here, too?  It makes me a bit woozy to be defining the handler with `sigaction()`, but then resetting and retrieving it with `signal()`.

dave-estes: I *think* this has to be `signal()`, because we're attaching the default handler, `SIG_DLF`, which is a general handler and not a `sigaction`.

mxw: Yeah, you can just set the `sa_action` member to `SIG_DFL` and that should work.

hphp/runtime/ext/std/ext_std_process.cpp
```diff
@@ -191,7 +191,12 @@ struct ShellExecContext final {
   ShellExecContext() {
     // Use the default handler while exec'ing in a shell,
     // saving the previous signal action.
-    m_sig_handler = signal(SIGCHLD, SIG_DFL);
+    struct sigaction sa = {};
+    sa.sa_handler = SIG_DFL;
```
dave-estes: Does anyone think any sa_flags need to be set?

hphp/runtime/ext/std/ext_std_process.cpp
```diff
@@ -240,7 +250,7 @@ struct ShellExecContext final {
   }
 
 private:
-  void (*m_sig_handler)(int);
+  struct sigaction old_sa = {};
```
mxw: Name this `m_old_sa`, since it's private.

dave-estes: Done.

mxw:
Cool, this is good to go—just need to wait for tests to finish before landing.

PR #7315, opened 2016-09-03
[AArch64] Syncs I$ after vasm patching and padding.

dave-estes:
Previous patches that synced the I$ missed these two places which
also modify the translation cache. Similar to previous patches,
modifications were not caught by the Vgen destructor. This changes
increases the stability of OSS, particularly when using concurrent
jit.


facebook-github-bot-4:
This pull request has been imported into Phabricator, and discussion and review of the diff will take place at https://reviews.facebook.net/D63255

PR #7327, opened 2016-09-09
[PPC64] Fix PPC64 backchain

lbianc:
When analyzing the PPC64 backchain with gdb or perf, it has an infinite
loop caused by the missing first VM frame as expected by PPC64 ABI.
This change adds this missing frame and removes the workaround changing the
value that r31 will receive.

### DIFF
--- a/hphp/runtime/vm/jit/unique-stubs.cpp
+++ b/hphp/runtime/vm/jit/unique-stubs.cpp
@@ -1073,13 +1073,18 @@ TCA emitDecRefGeneric(CodeBlock& cb, DataBlock& data) {
 
 TCA emitEnterTCExit(CodeBlock& cb, DataBlock& data, UniqueStubs& us) {
   return vwrap(cb, data, [&] (Vout& v) {
+    bool saveFrame = true;
+
     // Eagerly save VM regs and realign the native stack.
     storeVMRegs(v);
 
     // Realign the native stack.
     switch (arch()) {
-      case Arch::X64:
       case Arch::PPC64:
+        // Do not load the FP when leaving the TC, as it was not saved in
+        // the stublogue.
+        saveFrame = false;
+      case Arch::X64:
         v << lea{rsp()[8], rsp()};
         break;
       case Arch::ARM:
@@ -1098,7 +1103,7 @@ TCA emitEnterTCExit(CodeBlock& cb, DataBlock& data, UniqueStubs& us) {
     storeReturnRegs(v);
 
     // Perform a native return.
-    v << stubret{RegSet(), true};
+    v << stubret{RegSet(), saveFrame};
   });
 }
 
@@ -1118,11 +1123,25 @@ TCA emitEnterTCHelper(CodeBlock& cb, DataBlock& data, UniqueStubs& us) {
 #endif
 
   return vwrap2(cb, cb, data, [&] (Vout& v, Vout& vc) {
+    bool saveFrame = true;
+
     // Architecture-specific setup for entering the TC.
     v << inittc{};
 
+    // Set if FP need to be saved.
+    switch (arch()) {
+      case Arch::ARM:
+      case Arch::X64:
+        break;
+      case Arch::PPC64:
+        // Do no save the FP for PPC64, as no new frame is created in the
+        // stublogue
+        saveFrame = false;
+        break;
+    }
+
     // Native func prologue.
-    v << stublogue{true};
+    v << stublogue{saveFrame};
 
 #if defined(__CYGWIN__) || defined(__MINGW__) || defined(_MSC_VER)
     // Windows hates argument registers.
--- a/hphp/runtime/vm/jit/unique-stubs.cpp
+++ b/hphp/runtime/vm/jit/unique-stubs.cpp
@@ -1073,18 +1073,13 @@ TCA emitDecRefGeneric(CodeBlock& cb, DataBlock& data) {
 
 TCA emitEnterTCExit(CodeBlock& cb, DataBlock& data, UniqueStubs& us) {
   return vwrap(cb, data, [&] (Vout& v) {
-    bool saveFrame = true;
-
     // Eagerly save VM regs and realign the native stack.
     storeVMRegs(v);
 
     // Realign the native stack.
     switch (arch()) {
-      case Arch::PPC64:
-        // Do not load the FP when leaving the TC, as it was not saved in
-        // the stublogue.
-        saveFrame = false;
       case Arch::X64:
+      case Arch::PPC64:
         v << lea{rsp()[8], rsp()};
         break;
       case Arch::ARM:
@@ -1103,7 +1098,9 @@ TCA emitEnterTCExit(CodeBlock& cb, DataBlock& data, UniqueStubs& us) {
     storeReturnRegs(v);
 
     // Perform a native return.
-    v << stubret{RegSet(), saveFrame};
+    // On PPC64, as there is no new frame created when entering the VM, the FP
+    // must no be saved.
+    v << stubret{RegSet(), arch() != Arch::PPC64};
   });
 }
 
@@ -1123,25 +1120,11 @@ TCA emitEnterTCHelper(CodeBlock& cb, DataBlock& data, UniqueStubs& us) {
 #endif
 
   return vwrap2(cb, cb, data, [&] (Vout& v, Vout& vc) {
-    bool saveFrame = true;
-
     // Architecture-specific setup for entering the TC.
     v << inittc{};
 
-    // Set if FP need to be saved.
-    switch (arch()) {
-      case Arch::ARM:
-      case Arch::X64:
-        break;
-      case Arch::PPC64:
-        // Do no save the FP for PPC64, as no new frame is created in the
-        // stublogue
-        saveFrame = false;
-        break;
-    }
-
     // Native func prologue.
-    v << stublogue{saveFrame};
+    v << stublogue{arch() != Arch::PPC64};
 
 #if defined(__CYGWIN__) || defined(__MINGW__) || defined(_MSC_VER)
     // Windows hates argument registers.
--- a/hphp/runtime/vm/jit/unique-stubs.cpp
+++ b/hphp/runtime/vm/jit/unique-stubs.cpp
@@ -1118,11 +1118,11 @@ TCA emitEnterTCHelper(CodeBlock& cb, DataBlock& data, UniqueStubs& us) {
 #endif
 
   return vwrap2(cb, cb, data, [&] (Vout& v, Vout& vc) {
-    // Native func prologue.
-    v << stublogue{true};
-
     // Architecture-specific setup for entering the TC.
     v << inittc{};
+
+    // Native func prologue.
+    v << stublogue{true};
 
 #if defined(__CYGWIN__) || defined(__MINGW__) || defined(_MSC_VER)
     // Windows hates argument registers.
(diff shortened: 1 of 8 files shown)


hphp/runtime/vm/jit/vasm-ppc64.cpp
```diff
@@ -707,6 +712,14 @@ void Vgen::emit(const mcprep& i) {
 }
 
 void Vgen::emit(const inittc&) {
```
lbianc: Actually this VASM is handling the frame fix in the `emitEnterTCHelper`, which is architecture independent. Any suggestion if we must create a new VASM or any other way to do that? Also I had to change the place it is called, to be the first executed when entering the VM.

lbianc:
@mxw 

Some tests failed on it, is there any action that I should take?

gut:
ping

mxw:
@gut, @lbianc, etc.—I'm still not used to our new GitHub-only workflows, and as a result, I didn't even notice your pings here :\  Please feel free to ping me anytime on IRC, or ask anyone on IRC to ping me, or even shoot me emails if those don't work.

hphp/runtime/vm/jit/vasm-ppc64.cpp
```diff
@@ -474,6 +474,11 @@ struct Vgen {
 private:
   CodeBlock& frozen() { return text.frozen().code; }
 
+  void emitSaveTOC() {
+    // TOC save/restore is required by ABI for external functions.
+    a.std(rtoc(), rsfp()[AROFF(SAVED_TOC())]);
```
mxw: Does this really need its own helper?

hphp/runtime/vm/jit/vasm-ppc64.cpp
mxw: I'm pretty sure you guys added `inittc{}`, so its only real purpose is to allow the PPC64 backend to do additional work before anything else happens in enterTCHelper.  You can do whatever you want in it :]

hphp/runtime/vm/jit/vasm-ppc64.cpp
```diff
@@ -707,6 +712,14 @@ void Vgen::emit(const mcprep& i) {
 }
 
 void Vgen::emit(const inittc&) {
+  // Set info in the frame created on the caller
+  a.mflr(rfuncln());
+  a.std(rfuncln(), rsfp()[AROFF(m_savedRip)]);
+
+  // follow the ABI and create the calle frame and set its info
```
mxw: `s/calle/callee/`

hphp/runtime/vm/jit/vasm-ppc64.cpp
lbianc: Ok

hphp/runtime/vm/jit/vasm-ppc64.cpp
lbianc: This function is called in some other places, but as this change was related with the backchain, and it is used in the `inittc`, I called it only there. There will be another PR to add the other calls. I think naming a function is a better description of what the asm line is doing.

hphp/runtime/vm/jit/vasm-ppc64.cpp
lbianc: Done.

lbianc:
The additional frame created in the VASM `inittc` was useful to restore the correct value of `r31` (rvmfp), which is a non-volatile register. It was only visible when unwinding, as it uses the `r31` to find the previous frame. The `r31` value was being lost, as this register is used to manage the php stack, so the value was being overwritten.
In order to remove the additional frame, there was two possible solutions:

1 - Save the `r31` into PHP frame, which implies:
    - Change the "write_tc_cie" to look for it in the correct place, which would be different from `x64`.
    - Chance all PHP frame creation to save the `r31`, including the "enterTCHelper", which would contain a PPC64 specific code.

2 - Assign a new register to be the rvmfp, which implies:
    - Change the inittc to set the correct value in the `r26` for the stublogue.
    - Change `PPC64` abi to assign the new register.
	
The solution 2 was selected, in order to avoid `PPC64` specific code in the common unique stub code, and on the method write_tc_cie (unwind-itanium.cpp).

hphp/runtime/vm/jit/unwind-itanium.cpp
```diff
@@ -358,7 +365,24 @@ void write_tc_cie(EHFrameWriter& ehfw) {
 
   // The following calculation is related to the "top" of the record.  There is
   // an implicit -8 factor as defined by EHFrameWriter::m_cie.data_align.
+
+#if defined(__powerpc64__)
```
lbianc: @mxw This was the responsible for all the issues we had with frames. As I have already told you, a execution stores the return address in the parent's frame, so the current frame have an invalid information in the place of its m_savedRip, which will be used in the next call.
When unwinding, HHVM finds out if a frame is a VM frame analyzing the m_savedRip, which must be the one from the current context, so the following rules get it from the parent frame.
This was the reason the unwind was trying to unwind an addition PHP frame before get the last C++ frame, the m_savedRip being considered was from the next frame, i. e., still a PHP frame.

hphp/runtime/vm/jit/unique-stubs.cpp
```diff
@@ -1073,13 +1073,18 @@ TCA emitDecRefGeneric(CodeBlock& cb, DataBlock& data) {
 
 TCA emitEnterTCExit(CodeBlock& cb, DataBlock& data, UniqueStubs& us) {
   return vwrap(cb, data, [&] (Vout& v) {
+    bool saveFrame = true;
+
     // Eagerly save VM regs and realign the native stack.
     storeVMRegs(v);
 
     // Realign the native stack.
     switch (arch()) {
```
lbianc: @mxw I could not avoid this change in this unique stub, as PPC64 does not create a frame when entering the VM, there is no need to save the FP, so I added this switch.

hphp/runtime/vm/jit/unique-stubs.cpp
```diff
@@ -1098,7 +1103,7 @@ TCA emitEnterTCExit(CodeBlock& cb, DataBlock& data, UniqueStubs& us) {
     storeReturnRegs(v);
 
     // Perform a native return.
-    v << stubret{RegSet(), true};
+    v << stubret{RegSet(), saveFrame};
```
mxw: Why not just `arch() != Arch::PPC64` here instead of a variable?

hphp/runtime/vm/jit/unique-stubs.cpp
```diff
@@ -1118,12 +1123,26 @@ TCA emitEnterTCHelper(CodeBlock& cb, DataBlock& data, UniqueStubs& us) {
...
+        break;
+      case Arch::PPC64:
+        // Do no save the FP for PPC64, as no new frame is created in the
+        // stublogue
+        saveFrame = false;
+        break;
+    }
+
+    // Native func prologue.
+    v << stublogue{saveFrame};
```
mxw: Same here—the switch seems heavy-handed for just setting a single bool.

hphp/runtime/vm/jit/vasm-ppc64.cpp
```diff
@@ -484,6 +484,12 @@ struct Vgen {
     a.addi(rsfp(), rsp(), -min_frame_size);
     a.std(rvmfp(), rsfp()[AROFF(m_sfp)]);
   }
+
+  void emitSaveTOC() {
+    // TOC save/restore is required by ABI for external functions.
+    a.std(rtoc(), rsfp()[AROFF(SAVED_TOC())]);
+  }
```
mxw: Seems like this is no longer used.

hphp/runtime/vm/jit/unwind-itanium.cpp
mxw: Yes, this makes sense to me.  I think it's perfectly fine to have this conditional on architecture—but could you make this one into a switch or `if ()` instead of guarding on `#ifdef`s?

hphp/runtime/vm/jit/unique-stubs.cpp
mxw: Hmm... what's the point of having `stublogue{true}` on PPC64 in `emitFunctionEnterHelper()`, then?

lbianc: The point of have the `stublogue{true}` in this case is that there may have information pushed on the stack, so a new frame need to be created, in order to set its end. I think the same happens for `x64`, that is why the parameter is `true` for this case.
On PPC64 This is different when compared with the VM entering, because entering the VM the frame was already created and set, here the frame was allocated, but rvmfp was not set, that is why the parameter need to be `true` here.

hphp/runtime/vm/jit/unique-stubs.cpp
lbianc: Done

hphp/runtime/vm/jit/vasm-ppc64.cpp
lbianc: Removed

hphp/runtime/vm/jit/unique-stubs.cpp
```diff
@@ -1098,7 +1098,9 @@ TCA emitEnterTCExit(CodeBlock& cb, DataBlock& data, UniqueStubs& us) {
     storeReturnRegs(v);
 
     // Perform a native return.
-    v << stubret{RegSet(), true};
+    // On PPC64, as there is no new frame created when entering the VM, the FP
+    // must no be saved.
```
mxw: `s/must no/must not/`

hphp/runtime/vm/jit/unique-stubs.cpp
```diff
@@ -1098,7 +1098,9 @@ TCA emitEnterTCExit(CodeBlock& cb, DataBlock& data, UniqueStubs& us) {
     storeReturnRegs(v);
 
     // Perform a native return.
-    v << stubret{RegSet(), true};
+    // On PPC64, as there is no new frame created when entering the VM, the FP
```
mxw: +1 newline before this line

hphp/runtime/vm/jit/unwind-itanium.cpp
```diff
@@ -358,7 +369,29 @@ void write_tc_cie(EHFrameWriter& ehfw) {
...
   // an implicit -8 factor as defined by EHFrameWriter::m_cie.data_align.
-  ehfw.offset_extended_sf(dw_reg::IP, (record_size - AROFF(m_savedRip)) / 8);
+
+  switch (arch()) {
+    case Arch::ARM:
+    case Arch::X64:
+      ehfw.offset_extended_sf(dw_reg::IP, (record_size - AROFF(m_savedRip)) / 8);
+      break;
+    case Arch::PPC64:
+      // On PPC64 the return addres of a current frame is found in the parent,
```
mxw: `s/addres/address/`

hphp/runtime/vm/jit/unwind-itanium.cpp
```diff
@@ -358,7 +369,29 @@ void write_tc_cie(EHFrameWriter& ehfw) {
...
+      // so to get it, the content of the FP need to be taken to achieve the
+      // parent.
```
mxw: This sentence doesn't make sense syntactically to me.  How about:

`On PPC64, the return address for the current frame is found in the parent frame.  The following expression uses the FP to get the parent frame and recovers the return address from it.`

hphp/runtime/vm/jit/unique-stubs.cpp
lbianc: Done.

hphp/runtime/vm/jit/unwind-itanium.cpp
lbianc: Done.

PR #7328, opened 2016-09-09
[PPC64] Check if memory ref was handled

lbianc:
On PPC64 the memory ref need to be "patched" in order to eliminate the
displacement or index value, as there is no instructions with all memory
ref parameters.
These asserts ensure the instruction can be emitted with the correct
reference.


lbianc:
@mxw 

Some tests failed on it, is there any action that I should take?

aorenste:
The errors look like infra issues - I'm rerunning the build now.

PR #7329, opened 2016-09-09
Remove NO_OPT workaround from decoder

racardoso:
Summary:

This commit removes __attribute__(optimize("O0")) of decoder and
remove early-inlining-insns compilation options for ppc64.

Compilation freezes due a fine tune optimization
using early-inlining-insns and -O3 optimization level on POWER.
Before we provide a temporary solution disabling optimization for
decoder constructor (who have lots of inline and macro stuffs).

After some investigation and interlock with gcc developers we found
that's caused by early-inlining-insns due a gcc bug on ppc64el.
A bugzilla was openned on:

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=72855

Was pointed out we should be careful about those parameters as it
may have to be fine tunned for each architecture or we can find
problems like that.

### DIFF
--- a/CMake/HPHPCompiler.cmake
+++ b/CMake/HPHPCompiler.cmake
@@ -122,13 +122,21 @@ if (${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang" OR ${CMAKE_CXX_COMPILER_ID} STREQU
       "fno-canonical-system-headers"
       "Wvla"
     )
+
     list(APPEND RELEASE_CXX_OPTIONS
       "-param max-inline-insns-auto=100"
-      "-param early-inlining-insns=200"
-      "-param max-early-inliner-iterations=50"
       "-param=inline-unit-growth=200"
       "-param=large-unit-insns=10000"
     )
+
+    # The params bellow causes problem on GCC 4.8 4.8 and 5.4 on PPC64
+    # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=72855
+    if(NOT IS_PPC64)
+      list(APPEND_RELEASE_CXX_OPTIONS
+        "-param early-inlining-insns=200"
+        "-param max-early-inliner-iterations=50"
+    )
+    endif()
 
     execute_process(COMMAND ${CMAKE_CXX_COMPILER} ${CMAKE_CXX_COMPILER_ARG1} -dumpversion OUTPUT_VARIABLE GCC_VERSION)
     if(GCC_VERSION VERSION_GREATER 4.8 OR GCC_VERSION VERSION_EQUAL 4.8)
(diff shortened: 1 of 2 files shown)


aorenste:
I think the failed tests are infra issues but the cmake error seems real:
```
CMake Error at CMake/HPHPCompiler.cmake:135 (list):
list does not recognize sub-command APPEND_RELEASE_CXX_OPTIONS
Call Stack (most recent call first):
CMake/HPHPSetup.cmake:96 (include)
third-party/CMakeLists.txt:18 (include)
Generating DEBUG build
CMake Error at CMake/HPHPCompiler.cmake:135 (list):
list does not recognize sub-command APPEND_RELEASE_CXX_OPTIONS
Call Stack (most recent call first):
CMake/HPHPSetup.cmake:96 (include)
hphp/CMakeLists.txt:30 (include)
Generating DEBUG build
Not building the gmp extension.
Not building the fribidi extension.
Not building the odbc extension.
```

CMake/HPHPCompiler.cmake
```diff
@@ -122,14 +122,22 @@ if (${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang" OR ${CMAKE_CXX_COMPILER_ID} STREQU
...
-      "-param early-inlining-insns=200"
-      "-param max-early-inliner-iterations=50"
       "-param=inline-unit-growth=200"
       "-param=large-unit-insns=10000"
     )
 
+    # The params bellow causes problem on GCC 4.8 4.8 and 5.4 on PPC64
+    # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=72855
+    if(NOT IS_PPC64)
+      list(APPEND_RELEASE_CXX_OPTIONS
```
Orvid: `APPEND RELEASE_CXX_OPTIONS`

racardoso: @Orvid Fixed. Thanks for point that out

CMake/HPHPCompiler.cmake
```diff
@@ -122,14 +122,22 @@ if (${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang" OR ${CMAKE_CXX_COMPILER_ID} STREQU
...
       "-param=large-unit-insns=10000"
     )
 
+    # The params bellow causes problem on GCC 4.8 4.8 and 5.4 on PPC64
+    # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=72855
+    if(NOT IS_PPC64)
+      list(APPEND RELEASE_CXX_OPTIONS
+        "-param early-inlining-insns=200"
+        "-param max-early-inliner-iterations=50"
+    )
```
Orvid: Minor nit: The closing parenthesis here needs to be indented one more level. (so that it's at the same level as `list(APPEND`.

racardoso: @Orvid fixed

racardoso:
why there is unmerged commits?

gut:
because it didn't merge your commit, it actually squashed it and placed on top of master branch.

Don't worry, that's the usual output of a successful pull request :-)

racardoso:
@aorenste i guess merge missed that one commit ffb4b2b

Orvid:
Although compilation doesn't freeze on other platforms, this causes a significant regression in build time, nearly 7 minutes to compile this single file, so I'm reverting it as we need to keep the workaround in place for now.

racardoso:
@Orvid that's weird. In my tests here o x86 and PPC64 i do not notice this regression on build time. Although i am using `make -j16`. Btw, i guess we can reduce build time by only compile arch independent files on machines who will use it.

PR #7334, opened 2016-09-11
corrected hscan handling of output

jimyaghi:
Current release incorrectly handles HScan output and returns unexpected results.

From Redis, HScan gives similar output to ZScan, where odd numbered line represent the field name and even numbered lines represent value at the given hash key. (see http://redis.io/commands/scan for details on expected output from each).

In the current release code, HScan appears to expect a vector array of values and thus returns an incorrect, incoherent result (a 1xn serial array) when it should be returning a [ key=>value ] associative array.

I have patched the code that translates HSCAN's output so it works the exact same way that ZSCAN already does in 3.15 release. Data returned from redis is now transformed into an associative array of key=>value.

### DIFF
--- a/hphp/system/php/redis/Redis.php
+++ b/hphp/system/php/redis/Redis.php
@@ -561,7 +561,18 @@ class Redis {
   }
 
   public function hScan($key, &$cursor, $pattern = null, $count = null) {
-    return $this->scanImpl('HSCAN', $key, $cursor, $pattern, $count);
+    $flat = $this->scanImpl('HSCAN', $key, $cursor, $pattern, $count);
+    /*
+     * HScan behaves differently from the other *scan functions. The wire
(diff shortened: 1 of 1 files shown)


facebook-github-bot-6:

facebook-github-bot-9:

hphp/system/php/redis/Redis.php
```diff
@@ -561,7 +561,18 @@ public function sScan($key, &$cursor, $pattern = null, $count = null) {
   }
 
   public function hScan($key, &$cursor, $pattern = null, $count = null) {
-    return $this->scanImpl('HSCAN', $key, $cursor, $pattern, $count);
+    $flat = $this->scanImpl('HSCAN', $key, $cursor, $pattern, $count);
+    /*
```
usox: A method to format the output of hScan/zScan would avoid this code duplication

PR #7337, opened 2016-09-12
[PPC64] Replace patch Vptr and fix loadups/storeups

lbianc:
patchVptr was a function called inside many lowers to handle the memory ref. This function was removed to be a new optimization phase "fixVptrsForPPC64", that works as the lower phase, replacing and adding new VASMs, and keep it separated from lower phase.
The memory reference of loadups and storeups need to be handled on PPC64, as on this architecture there no instruction to handle a reference with base, index and displacement.

### DIFF
--- a/hphp/runtime/vm/jit/vasm-ppc64.cpp
+++ b/hphp/runtime/vm/jit/vasm-ppc64.cpp
@@ -239,6 +239,24 @@ struct Vgen {
   void emit(const landingpad&) { }
   void emit(const ldimmw& i) { a.li(Reg64(i.d), i.s); }
   void emit(const ldarx& i) { a.ldarx(i.d, i.s); }
+  void emit(const loadups& i) {
+    // A lowering would be handy here, but the Vptr's displacement is only
+    // later added at the optimizeCopies optimization
+    Vptr p(i.s);
+    if ((!p.base.isValid()) || (p.base == 0)) {
+      p.base = ppc64_asm::reg::r0; // use zero not r0
+    }
+    if (!p.index.isValid()) {
+      a.li(rAsm, p.disp);
+      p.index = rAsm;
+      p.disp = 0;
+    } else if (p.disp) {
+        a.addi(rAsm, p.index, p.disp);
+        p.index = rAsm;
+        p.disp = 0;
+    }
+    a.lxvd2x(i.d, p);
+  }
   void emit(const leap& i) { a.li64(i.d, i.s.r.disp, false); }
   void emit(const lead& i) { a.li64(i.d, (int64_t)i.s.get(), false); }
   void emit(const mfcr& i) { a.mfcr(i.d); }
@@ -290,8 +308,23 @@ struct Vgen {
   void emit(const sqrtsd& i) { a.xssqrtdp(i.d,i.s); }
 
   void emit(const stdcx& i) { a.stdcx(i.s, i.d); }
+  void emit(const storeups& i) {
+    Vptr p(i.m);
+    if ((!p.base.isValid()) || (p.base == 0)) {
+      p.base = ppc64_asm::reg::r0;
+    }
+    if (!p.index.isValid()) {
+      a.li(rAsm, p.disp);
+      p.index = rAsm;
+      p.disp = 0;
+    } else if (p.disp) {
+        a.addi(rAsm, p.index, p.disp);
+        p.index = rAsm;
+        p.disp = 0;
+    }
+    a.stxvw4x(i.s, p);
+  }
 
-  void emit(const storeups& i) { a.stxvw4x(i.s,i.m); }
   // Subtractions: d = s1 - s0
   void emit(const subq& i) {
     a.subfo(i.d, i.s0, i.s1, true);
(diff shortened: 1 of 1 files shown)


lbianc:
@mxw

hphp/runtime/vm/jit/vasm-ppc64.cpp
```diff
@@ -1564,6 +1438,207 @@ void lowerForPPC64(Vunit& unit) {
...
+    case AddressModes::Invalid:
+    default:
+      assert(false && "Invalid address mode");
+      break;
+  }
+}
+
+/* Fallback, when no Vptr is to be fixed */
+template <typename Inst>
+void fixVptrsForPPC64(Vout& v, Inst& inst) {}
```
mxw: Are there any Vinstrs with Vptr operands that you _don't_ want to be fixed up with this pass?  If not, I think writing a visitor class to use with visitOperands() would be a more robust solution here.

hphp/runtime/vm/jit/vasm-ppc64.cpp
```diff
@@ -1564,6 +1438,207 @@ void lowerForPPC64(Vunit& unit) {
...
+      Vreg index_disp_reg = v.makeReg();
+      lowerImm(Immed(p.disp), v, disp_reg);
+      v << addq{disp_reg, p.index, index_disp_reg, VregSF(RegSF{0})};
+      p.index = index_disp_reg;
+      p.disp = 0;
+      break;
+    }
+
+    case AddressModes::Invalid:
+    default:
```
mxw: This is an enum class that's only used in this function; you can nix the default case.

lbianc: Done.

hphp/runtime/vm/jit/vasm-ppc64.cpp
lbianc: Yes, good solution, thanks! Done.

hphp/runtime/vm/jit/vasm-ppc64.cpp
```diff
@@ -1564,6 +1438,142 @@ void lowerForPPC64(Vunit& unit) {
...
+      auto& inst = blocks[ib].code[ii];
+      SCOPE_EXIT {unit.freeScratchBlock(scratch);};
+      Vout v(unit, scratch, inst.irctx());
+      VptrFixVisit vptrFixVisit {v};
+
+      visitOperands(inst, vptrFixVisit);
+      if (!v.empty()) {
+        vector_splice(unit.blocks[ib].code, ii, 0,
+                      unit.blocks[scratch].code);
+      }
```
mxw: Instead of all this boilerplate, just use `vmodify()` instead.

hphp/runtime/vm/jit/vasm-ppc64.cpp
```diff
@@ -1564,6 +1438,142 @@ void lowerForPPC64(Vunit& unit) {
...
+ */
+struct VptrFixVisit {
+explicit VptrFixVisit(Vout& v) : m_v(v) {}
+  Vout& m_v;
+
+  template<class T> void def(T) {}
+  template<class T> void imm(const T&) {}
+  template<class T> void across(T) {}
+  template<class T, class H> void defHint(T, H) {}
+  template<class T, class H> void useHint(T, H) {}
```
mxw: I'm pretty sure there are not use-hints with Vptr operands, but it would be better to have

```
template<class T, class H> void useHint(T r, H) { use(r); }
```

just in case someday we decide to go wild.

hphp/runtime/vm/jit/vasm-ppc64.cpp
```diff
@@ -239,6 +239,24 @@ struct Vgen {
...
+      a.li(rAsm, p.disp);
+      p.index = rAsm;
+      p.disp = 0;
+    } else if (p.disp) {
+        a.addi(rAsm, p.index, p.disp);
+        p.index = rAsm;
+        p.disp = 0;
+    }
+    a.lxvd2x(i.d, p);
+  }
```
mxw: How does the new lowering pass interact with this and with `storeups{}`?

hphp/runtime/vm/jit/vasm-ppc64.cpp
```diff
@@ -239,6 +239,24 @@ struct Vgen {
...
+      p.base = ppc64_asm::reg::r0; // use zero not r0
+    }
+    if (!p.index.isValid()) {
+      a.li(rAsm, p.disp);
+      p.index = rAsm;
+      p.disp = 0;
+    } else if (p.disp) {
+        a.addi(rAsm, p.index, p.disp);
+        p.index = rAsm;
+        p.disp = 0;
```
mxw: nit: This is one indentation level too much.

hphp/runtime/vm/jit/vasm-ppc64.cpp
```diff
@@ -1564,6 +1438,142 @@ void lowerForPPC64(Vunit& unit) {
...
+      assert(false && "Invalid address mode");
+      break;
+  }
+}
+
+/*
+ * Every vasm that has a Vptr parameter must be handled
+ */
+struct VptrFixVisit {
+explicit VptrFixVisit(Vout& v) : m_v(v) {}
```
mxw: +1 indentation level here

hphp/runtime/vm/jit/vasm-ppc64.cpp
```diff
@@ -1564,6 +1438,142 @@ void lowerForPPC64(Vunit& unit) {
...
+  Vout& m_v;
```
mxw: Data members should go below all member functions.

hphp/runtime/vm/jit/vasm-ppc64.cpp
lbianc: Done.

hphp/runtime/vm/jit/vasm-ppc64.cpp
lbianc: The `fixVptr` changes the `Vptr` to have base and index or base and displacement. The `loadups` and `storeups` work only with base and index, so in this case we had to add this guard, as only these two instructions are different.

hphp/runtime/vm/jit/vasm-ppc64.cpp
lbianc: Done.

hphp/runtime/vm/jit/vasm-ppc64.cpp
lbianc: Done.

hphp/runtime/vm/jit/vasm-ppc64.cpp
lbianc: Done.

hphp/runtime/vm/jit/vasm-ppc64.cpp
lbianc: Done.

PR #7339, opened 2016-09-12
Fix "SYS_memfd_create" redefined warning.

racardoso:
Summary:

This commit fixes "SYS_memfd_create" redefined warning.
This warning happens on POWER because SYS_memfd_create is already
defined on: powerpc64le-linux-gnu/sys/syscall.h:31:0.

### DIFF
--- a/hphp/hack/src/heap/hh_shared.c
+++ b/hphp/hack/src/heap/hh_shared.c
@@ -142,11 +142,11 @@
 #if !defined __APPLE__ && !defined _WIN32
   // Linux version for the architecture must support syscall memfd_create
   #if defined(__x86_64__)
-    #define SYS_memfd_create 319
+    #define SYSC_memfd_create 319
   #elif defined(__powerpc64__)
-    #define SYS_memfd_create 360
+    #define SYSC_memfd_create 360
   #elif defined(__aarch64__)
-    #define SYS_memfd_create 385
+    #define SYSC_memfd_create 385
   #else
     #error "hh_shared.c requires a architecture that supports memfd_create"
   #endif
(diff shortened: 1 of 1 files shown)


hphp/hack/src/heap/hh_shared.c
```diff
@@ -142,11 +142,11 @@
 #if !defined __APPLE__ && !defined _WIN32
   // Linux version for the architecture must support syscall memfd_create
```
Orvid: Any reason not to guard this entire chain with `#ifndef SYS_memfd_create`?

racardoso: Yes. Because `SYS_memfd_create` is defined as: `#define SYS_memfd_create __NR_memfd_create`. Originally i sent a pull request that fixes a error due a hardcoded (x86_64) value: changes `return syscall(319, name, flags);` to `return syscall(__NR_memfd_create, name, flags);`. But the reviewer says that we should not rely on include files: "If we rely on the header, then if the header does NOT define __NR_memfd_create, then the binary will never use memfd_create, even if run on a machine whose kernel supports memfd_create. We need runtime checks for memfd_create support, not compile-time checks."

racardoso: @Orvid  https://reviews.facebook.net/D58995

Orvid: I'm fine with defining it when it isn't defined, but if it's already defined then we should be relying on it.

racardoso: @Orvid yes `SYS_memfd_create` is defined on  powerpc64le-linux-gnu/sys/syscall.h but defined as: 

`#define SYS_memfd_create __NR_memfd_create`

What do you suggest in this case? Originally my pull request check `#ifndef __NR_memfd_create` should i change back to it?

Orvid: `#ifndef SYS_memfd_create`

racardoso: @Orvid done.

racardoso: @Orvid done.

PR #7343, opened 2016-09-13
Add a optimized version of strcmp for POWER 8.

racardoso:
Summary:

This commit add a optimized version of strcmp for PPC64/POWER8
extracted from glibc github.

### DIFF
--- a/hphp/util/strcmp-power8.S
+++ b/hphp/util/strcmp-power8.S
@@ -1,0 +1,278 @@
+
+/* Optimized strcmp implementation for PowerPC64/POWER8.
+   Copyright (C) 2015-2016 Free Software Foundation, Inc.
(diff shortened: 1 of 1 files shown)


aorenste:
We ended up having to revert this change - there are concerns that the LGPL license is not compatible with HHVM's PHP license.

racardoso:
@aorenste ohh you're right. https://www.gnu.org/licenses/license-list.en.html:

>
 PHP License, Version 3.01 (#PHP-3.01)
This license is used by most of PHP4. It is a non-copyleft free software license. It is incompatible with the GNU GPL because it includes strong restrictions on the use of “PHP” in the name of derived products. We recommend that you not use this license for anything except PHP add-ons

PR #7347, opened 2016-09-15
[PPC64] Add a new class to support TOC

igorsnunes:
This is the first pull request of many to use a special feature for ppc:
TOC.

TOC (table of contents) is a memory region pointed by r2 that
contains immediate values to be loaded throughout the jit execution.
That is currently using hhvm's data-block as the memory region.

In this first pull request the code is not yet integrated as all our changes
in one PR would make it too big to be reviewed. For now, it just implemented
the singleton class that manages what is stored in TOC, and the "limmediate"
function that decides wether it is good or not to use TOC to load a specific
immediate value.

### DIFF
--- a/hphp/ppc64-asm/asm-ppc64.cpp
+++ b/hphp/ppc64-asm/asm-ppc64.cpp
@@ -767,8 +767,8 @@ void Assembler::loadTOC(const Reg64& rt, const Reg64& rttoc,  int64_t imm64,
 }
 
 void Assembler::limmediate(const Reg64& rt, int64_t imm64, ImmType immt) {
-  always_assert(HPHP::RuntimeOption::EvalPPC64minTOCImmSize >= 0 &&
-    HPHP::RuntimeOption::EvalPPC64minTOCImmSize <= 64);
+  always_assert(HPHP::RuntimeOption::EvalPPC64MinTOCImmSize >= 0 &&
+    HPHP::RuntimeOption::EvalPPC64MinTOCImmSize <= 64);
 
   auto fits = [](int64_t imm64, uint16_t shift_n) {
      return (static_cast<uint64_t>(imm64) >> shift_n) == 0 ? true : false;
@@ -778,7 +778,7 @@ void Assembler::limmediate(const Reg64& rt, int64_t imm64, ImmType immt) {
 #ifndef USE_TOC_ON_BRANCH
       1 ||
 #endif
-      (fits(imm64, HPHP::RuntimeOption::EvalPPC64minTOCImmSize)
+      (fits(imm64, HPHP::RuntimeOption::EvalPPC64MinTOCImmSize)
       && (immt != ImmType::TocOnly))) {
     li64(rt, imm64, immt != ImmType::AnyCompact);
     return;
--- a/hphp/ppc64-asm/asm-ppc64.cpp
+++ b/hphp/ppc64-asm/asm-ppc64.cpp
@@ -17,8 +17,16 @@
 #include "hphp/ppc64-asm/asm-ppc64.h"
 #include "hphp/ppc64-asm/decoder-ppc64.h"
 #include "hphp/runtime/base/runtime-option.h"
+#include "hphp/util/trace.h"
+
+TRACE_SET_MOD(asmppc64);
 
 namespace ppc64_asm {
+VMTOC::~VMTOC() {
+  FTRACE(1, "Number of values stored in TOC: {}\n",
+    std::to_string(m_last_elem_pos));
+}
+
 int64_t VMTOC::pushElem(int64_t elem) {
   if (m_map.find(elem) != m_map.end()) {
     return m_map[elem];
--- a/hphp/ppc64-asm/asm-ppc64.cpp
+++ b/hphp/ppc64-asm/asm-ppc64.cpp
@@ -60,7 +60,7 @@ intptr_t VMTOC::getPtrVector() {
 
 int64_t VMTOC::getValue(int64_t index, bool qword) {
   HPHP::Address addr = reinterpret_cast<HPHP::Address>(
-      reinterpret_cast<intptr_t>(index) + getPtrVector());
+      static_cast<intptr_t>(index) + getPtrVector());
   int64_t ret_val = 0;
   int max_elem = qword ? 8 : 4;
   for (int i = max_elem-1; i >= 0; i--) {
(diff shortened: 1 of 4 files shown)


hphp/runtime/base/runtime-option.h
```diff
@@ -606,6 +606,12 @@ struct RuntimeOption {
   F(int64_t, PerfWarningSampleRate, 1)                                  \
   /* Profiling flags */                                                 \
   F(bool, EnableReverseDataMap, false)                                  \
+  /* PPC64 Option: Minimum immediate size to use TOC */                 \
+  F(uint16_t, ppc64minTOCImmSize, 64)                                   \
+  /* PPC64 Option: Dump the number of elements stored in TOC */         \
+  F(bool, ppc64dumpTOCNelements, false)                                 \
+  /* PPC64 Option: Allow lwz and not only ld to be used for TOC read */ \
+  F(bool, ppc64useTOCLwz, false)                                        \
```
mxw: Why are these RuntimeOptions?

igorsnunes: We implemented these options as RuntimeOptions int order to facilitate performance tuning for each option. Do you think they should have been implemented as compilation parameters?

I think ppc64dumpTOCNelements should be kept as RuntimeOption, since it is a toggle for dumping how many values are stored in TOC.

paulbiss: ppc64dumpTOCNelements sounds like it would make sense as an admin command (though having it as a runtime option for scripts seems reasonable). The other two certainly sound like configuration settings. Please start them with a capital letter to match the rest of the file (e.g. PPC64...)

igorsnunes: @paulbiss What do you mean by configuration settings? CMake flag? preprocessor variables?

paulbiss: I mean I think they're fine as RuntimeOptions, we have a lot of perf tuning options already. But I defer to @mxw. If you do keep them though please rename them to match the rest of the options in here.

igorsnunes: @paulbiss , alright. I will rename the variables then, and wait for @mxw's feedback.

hphp/runtime/base/runtime-option.h
```diff
@@ -606,6 +606,12 @@ struct RuntimeOption {
   F(int64_t, PerfWarningSampleRate, 1)                                  \
   /* Profiling flags */                                                 \
   F(bool, EnableReverseDataMap, false)                                  \
+  /* PPC64 Option: Minimum immediate size to use TOC */                 \
+  F(uint16_t, PPC64minTOCImmSize, 64)                                   \
+  /* PPC64 Option: Dump the number of elements stored in TOC */         \
+  F(bool, PPC64dumpTOCNelements, false)                                 \
```
mxw: This should be a `Trace` module instead of a RuntimeOption, and you should use `FTRACE()` or similar for your dumping—see `util/trace.h`.

hphp/runtime/base/runtime-option.h
```diff
@@ -606,6 +606,12 @@ struct RuntimeOption {
...
+  /* PPC64 Option: Allow lwz and not only ld to be used for TOC read */ \
+  F(bool, PPC64useTOCLwz, false)                                        \
```
mxw: Capitalize the `u` in `use`, and the `m` in `min` above.

Why do these need to be configurable values?  Can you explain when and why you would ever want to change them?

mxw: e.g., Why would you want to enable or prohibit using lwz for TOC read?

igorsnunes: I think you are right. PPC64useTOCLwz should not be configurable:

Here is what it stands for: if it is set, for immediate values lower than 0xFFFFFFFF (32-bits), a lwz is emitted in spite of a ld. Which means: only 4 bytes of data-block will be used to load an immediate once lwz loads 4 bytes from memory. If PPC64useTOCLwz is not set, 8 bytes will be used (e.g. 0x00000000FFFFFFFF) and a ld will be emmited (load 64 bits from memory).

We were tweaking this parameter to make sure that no misalignment issue would happen by using lwz. And for now, this is not happening.

I will remove PPC64useTOCLwz.

hphp/runtime/base/runtime-option.h
```diff
@@ -617,6 +617,8 @@ struct RuntimeOption {
    * kept low to avoid the risk of collecting a sample while we're      \
    * processing a previous sample. */                                   \
   F(uint32_t, PerfMemEventSampleFreq, 80)                               \
+  /* PPC64 Option: Minimum immediate size to use TOC */                 \
+  F(uint16_t, PPC64MinTOCImmSize, 64)                                   \
```
mxw: Can you move this above the `Profiling flags` section divider?  This is not a profiling flag.  (You can also section of a new set of flags for PPC64, if you expect you'll have more.)

mxw:
This is failing to import to our internal Phabricator instance.  Are there any dependent diffs that also touch the same files, which might be causing a conflict?

mxw:
Okay, so it seems like the issue is that you need to rebase this diff over the other diff of yours that landed recently which touched the same files.  There's no immediate conflict, but something something source control something importing something.  Sorry for the churn—once you rebase and re-push the PR, it should be good to land.

igorsnunes:
@mxw no problem =)

hphp/ppc64-asm/asm-ppc64.cpp
```diff
@@ -16,8 +16,89 @@
...
+}
+
+intptr_t VMTOC::getPtrVector() {
+  always_assert(m_tocvector != nullptr);
+  return reinterpret_cast<intptr_t>(m_tocvector->base() + INT16_MAX + 1);
+}
+
+int64_t VMTOC::getValue(int64_t index, bool qword) {
+  HPHP::Address addr = reinterpret_cast<HPHP::Address>(
+      reinterpret_cast<intptr_t>(index) + getPtrVector());
```
mxw: This needs to be a `static_cast`, I think; it's breaking Mac builds.

igorsnunes: If I change  reinterpret_cast to static_cast my compilation starts to fail. Is that really the problem? Suggestions?

mxw: I mean specifically the `reinterpret_cast<intptr_t>`—is that what you tried changing?

igorsnunes: hmm.. My bad. I referred to the first one. I changed now and it seems to work.  I will send soon.

igorsnunes:
@mxw any idea about this one (fb/hphp_dbgo_clang-buck)?

aorenste:
@igorsnunes That looks like an unrelated infra failure.

PR #7348, opened 2016-09-16
[AArch64] Fix overflowing integer multiplication for AArch64

ptomsich:
AArch64 does not set flags on an integer-multiply and the common
way to detect an overflow on integer multiplication will only
provide EQ/NEQ flags.  To avoid taking a major penalty when
checking for overflows on integer multiplication, we make the
lowering of MulIntO architecture-specific and introduce a new
architecture-independent mulint (similarily to divint) vasm-insn.

The MulIntO node is lowered into the new 'mulint' vasm-instr, which
is a special kind of branching instruction (i.e. it takes a label
for the slow-path, in case an integer-overflow occurs).
'mulint' is marked to clobber the condition-code register.

The recommended way to detect an overflow on integer multiplication
on AArch64 is:
  mul	  calculate lower 64bit of the result
  smulh	  calculate the upper (signed/sign-extended) 64bits
  cmp	  compare if the upper 64bits are the sign-extension
	      of the lower 64bits (i.e. if all bits in the
 	      upper register are the same as the sign-bit in
	      the lower register.

To efficiently expand into such a sequence, the following vasm
instructions have been added:
  smulh	     upper 64bits of the 128bit result of a signed multiply
	       (could be named more nicely, but as only AArch64 is
	        using anyway...)
  cmpqsign   compare if one operand is the sign-extension of the
               other operand (i.e if all bits in one register are
               equivalent to the sign-bit of the other)
  mulint     integer-multiplication w/ overflow-checking

At the same time, the 'imul' vasm has been altered to _not_ set any
flags and _not_ perform any overflow-checking.  As it's only used
from the non-overflowing multiplication path in irlower-arith.c this
seems to be safe.  We an not remove 'imul', as it will still be used
to expand to lower 'MulInt'.

X-AffectedTests: quick/float-overflow.php
Signed-off-by: Philipp Tomsich <philipp.tomsich@theobroma-systems.com>

### DIFF
--- a/hphp/runtime/vm/jit/irlower-arith.cpp
+++ b/hphp/runtime/vm/jit/irlower-arith.cpp
@@ -128,7 +128,6 @@ ARITH_OPS
 
 
 void cgMulIntO(IRLS& env, const IRInstruction* inst) {
-  //  implMulIntO(vmain(env), env, inst);
   auto const d = dstLoc(env, inst, 0).reg();
   auto const s0 = srcLoc(env, inst, 0).reg();
   auto const s1 = srcLoc(env, inst, 1).reg();
--- a/hphp/runtime/vm/jit/irlower-arith.cpp
+++ b/hphp/runtime/vm/jit/irlower-arith.cpp
@@ -105,7 +105,6 @@ void implShift(Vout& v, IRLS& env, const IRInstruction* inst) {
   AO(MulInt,  BinopSF,  imul)   \
   AO(AddIntO, ArithO,   addq)   \
   AO(SubIntO, ArithO,   subq)   \
-  AO(MulIntO, ArithO,   imul)   \
   AO(AddDbl,  Binop,    addsd)  \
   AO(SubDbl,  Binop,    subsd)  \
   AO(MulDbl,  Binop,    mulsd)  \
@@ -126,6 +125,17 @@ ARITH_OPS
 #undef ARITH_OPS
 
 ///////////////////////////////////////////////////////////////////////////////
+
+
+void cgMulIntO(IRLS& env, const IRInstruction* inst) {
+  //  implMulIntO(vmain(env), env, inst);
+  auto const d = dstLoc(env, inst, 0).reg();
+  auto const s0 = srcLoc(env, inst, 0).reg();
+  auto const s1 = srcLoc(env, inst, 1).reg();
+  auto& v = vmain(env);
+
+  v << mulint{s0, s1, d, {label(env, inst->next()), label(env, inst->taken())}};
+}
 
 void cgFloor(IRLS& env, const IRInstruction* inst) {
   implRound<RoundDirection::floor>(vmain(env), env, inst);
--- a/hphp/runtime/vm/jit/irlower-arith.cpp
+++ b/hphp/runtime/vm/jit/irlower-arith.cpp
@@ -133,7 +133,7 @@ void cgMulIntO(IRLS& env, const IRInstruction* inst) {
   auto const s1 = srcLoc(env, inst, 1).reg();
   auto& v = vmain(env);
 
-  v << mulint{s0, s1, d, {label(env, inst->next()), label(env, inst->taken())}};
+  v << mulinto{s0, s1, d, {label(env, inst->next()), label(env, inst->taken())}};
 }
 
 void cgFloor(IRLS& env, const IRInstruction* inst) {
--- a/hphp/runtime/vm/jit/vasm-dead.cpp
+++ b/hphp/runtime/vm/jit/vasm-dead.cpp
@@ -144,7 +144,7 @@ bool effectful(Vinstr& inst) {
     case Vinstr::mrs:
     case Vinstr::mul:
     case Vinstr::mulsd:
-    case Vinstr::mulint:
+    case Vinstr::mulinto:
     case Vinstr::neg:
     case Vinstr::nop:
     case Vinstr::not:
--- a/hphp/runtime/vm/jit/vasm-dead.cpp
+++ b/hphp/runtime/vm/jit/vasm-dead.cpp
@@ -95,8 +95,6 @@ bool effectful(Vinstr& inst) {
     case Vinstr::fctidz:
     case Vinstr::fcvtzs:
     case Vinstr::imul:
-    case Vinstr::smulh:
-    case Vinstr::mul:
     case Vinstr::incl:
     case Vinstr::incq:
     case Vinstr::incw:
@@ -144,6 +142,7 @@ bool effectful(Vinstr& inst) {
     case Vinstr::movzbq:
     case Vinstr::movzlq:
     case Vinstr::mrs:
+    case Vinstr::mul:
     case Vinstr::mulsd:
     case Vinstr::mulint:
     case Vinstr::neg:
@@ -167,6 +166,7 @@ bool effectful(Vinstr& inst) {
     case Vinstr::shlqi:
     case Vinstr::shrli:
     case Vinstr::shrqi:
+    case Vinstr::smulh:
     case Vinstr::sqrtsd:
     case Vinstr::srem:
     case Vinstr::subbi:
--- a/hphp/runtime/vm/jit/vasm-dead.cpp
+++ b/hphp/runtime/vm/jit/vasm-dead.cpp
@@ -69,6 +69,7 @@ bool effectful(Vinstr& inst) {
     case Vinstr::cmpqi:
     case Vinstr::cmpqim:
     case Vinstr::cmpqm:
+    case Vinstr::cmpqsign:
     case Vinstr::cmpsd:
     case Vinstr::cmpsds:
     case Vinstr::cmpwim:
@@ -94,6 +95,8 @@ bool effectful(Vinstr& inst) {
     case Vinstr::fctidz:
     case Vinstr::fcvtzs:
     case Vinstr::imul:
+    case Vinstr::smulh:
+    case Vinstr::mul:
     case Vinstr::incl:
     case Vinstr::incq:
     case Vinstr::incw:
@@ -142,6 +145,7 @@ bool effectful(Vinstr& inst) {
     case Vinstr::movzlq:
     case Vinstr::mrs:
     case Vinstr::mulsd:
+    case Vinstr::mulint:
     case Vinstr::neg:
     case Vinstr::nop:
     case Vinstr::not:
--- a/hphp/runtime/vm/jit/vasm-instr.h
+++ b/hphp/runtime/vm/jit/vasm-instr.h
@@ -895,7 +895,16 @@ struct srem { Vreg s0, s1, d; };
 struct divint { Vreg s0, s1, d; };
 
 /*
- * Integer multiplication.
+ * Integer multiplication (with overflow).
+ * 
+ * Performs a 64bit integer multiplication (d := s0 * s1) and checks for a
+ * possible overflow.  Given that the method to check for the overflow will
+ * be architecture specific, we don't expose any internal flag status and
+ * mark this as killing the flag register.
+ *
+ * Targets have the same order as for branch instructions (see below):
+ *   target[0] = next (i.e. fall-through)
+ *   target[1] = taken (i.e. slow-path)
  */
 struct mulint { Vreg s0, s1, d; Vlabel targets[2]; };
 
--- a/hphp/runtime/vm/jit/vasm-instr.h
+++ b/hphp/runtime/vm/jit/vasm-instr.h
@@ -128,6 +128,7 @@ struct Vunit;
   O(absdbl, Inone, U(s), D(d))\
   O(srem, Inone, U(s0) U(s1), D(d))\
   O(divint, Inone, U(s0) U(s1), D(d))\
+  O(mulint, Inone, U(s0) U(s1), D(d))	\
   /* nop and trap */\
   O(nop, Inone, Un, Dn)\
   O(ud2, Inone, Un, Dn)\
@@ -160,6 +161,8 @@ struct Vunit;
   O(incqm, Inone, U(m), D(sf))\
   O(incqmlock, Inone, U(m), D(sf))\
   O(imul, Inone, U(s0) U(s1), D(d) D(sf))\
+  O(smulh, Inone, U(s0) U(s1), D(d)) \
+  O(mul, Inone, U(s0) U(s1), D(d)) \
   O(neg, Inone, UH(s,d), DH(d,s) D(sf))\
   O(notb, Inone, UH(s,d), DH(d,s))\
   O(not, Inone, UH(s,d), DH(d,s))\
@@ -203,6 +206,7 @@ struct Vunit;
   O(cmpqi, I(s0), U(s1), D(sf))\
   O(cmpqm, Inone, U(s0) U(s1), D(sf))\
   O(cmpqim, I(s0), U(s1), D(sf))\
+  O(cmpqsign, Inone, U(s0) U(s1), D(sf))	\
   O(cmpsd, I(pred), UA(s0) U(s1), D(d))\
   O(ucomisd, Inone, U(s0) U(s1), D(sf))\
   O(testb, Inone, U(s0) U(s1), D(sf))\
@@ -890,6 +894,11 @@ struct srem { Vreg s0, s1, d; };
  */
 struct divint { Vreg s0, s1, d; };
 
+/*
+ * Integer multiplication.
+ */
+struct mulint { Vreg s0, s1, d; Vlabel targets[2]; };
+
 ///////////////////////////////////////////////////////////////////////////////
 
 /*
@@ -939,6 +948,8 @@ struct incqm { Vptr m; VregSF sf; };
 struct incqmlock { Vptr m; VregSF sf; };
 // mul: s0 * s1 => d, sf
 struct imul { Vreg64 s0, s1, d; VregSF sf; };
+struct mul { Vreg64 s0, s1, d; };
+struct smulh { Vreg64 s0, s1, d; };
 // neg: 0 - s => d, sf
 struct neg { Vreg64 s, d; VregSF sf; };
 // not: ~s => d
@@ -992,6 +1003,7 @@ struct cmpq { Vreg64 s0; Vreg64 s1; VregSF sf; };
 struct cmpqi { Immed s0; Vreg64 s1; VregSF sf; };
 struct cmpqm { Vreg64 s0; Vptr s1; VregSF sf; };
 struct cmpqim { Immed s0; Vptr s1; VregSF sf; };
+struct cmpqsign { Vreg64 s0; Vreg64 s1; VregSF sf; };
 struct cmpsd { ComparisonPred pred; VregDbl s0, s1, d; };
 struct ucomisd { VregDbl s0, s1; VregSF sf; };
 // s1 & s0 => sf
--- a/hphp/runtime/vm/jit/vasm-instr.h
+++ b/hphp/runtime/vm/jit/vasm-instr.h
@@ -128,7 +128,7 @@ struct Vunit;
   O(absdbl, Inone, U(s), D(d))\
   O(srem, Inone, U(s0) U(s1), D(d))\
   O(divint, Inone, U(s0) U(s1), D(d))\
-  O(mulint, Inone, U(s0) U(s1), D(d))	\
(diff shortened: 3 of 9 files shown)


dave-estes:
Hey @ptomsich , did you see 0f7cc5540ce737c7a127ed68b432f7a68a44fb9b? I agree that it's pretty inefficient to do such a heavy-handed check for overflow, but the same can be said of all of the extra code that we emit for AArch64 throughout vasm-arm.cpp to synthetically set the SF in order to match the behavior of x64_64. With that said, I feel that your patch is a little too complicated and only addresses one of these cases. We have a yet-to-be-pushed patch that introduces a pass which undefs SFs for any instructions where there is no direct use. Then we've addressed each place in vasm-arm.cpp which emits code to synthetically set SF when it's not necessary. This patch did wonders to eliminate much of this extra code from being emitted.

If you think our forthcoming patch sounds promising, then perhaps we can hold off introducing this new complexity for integer multiplication here. Also, there's a good chance that I may have overlooked some other advantage of your patch. If that's the case, then I'm sorry. Can you comment?

ptomsich:
Dave,

> On 16 Sep 2016, at 19:21, Dave Estes <notifications@github.com> wrote:
> Hey @ptomsich <https://github.com/ptomsich> , did you see 0f7cc55 <https://github.com/facebook/hhvm/commit/0f7cc5540ce737c7a127ed68b432f7a68a44fb9b>? I agree that it's pretty inefficient to do such a heavy-handed check for overflow, but the same can be said of all of the extra code that we emit for AArch64 throughout vasm-arm.cpp to synthetically set the SF in order to match the behavior of x64_64. With that said, I feel that your patch is a little too complicated and only addresses one of these cases. We have a yet-to-be-pushed patch that introduces a pass which undefs SFs for any instructions where there is no direct use. Then we've addressed each place in vasm-arm.cpp which emits code to synthetically set SF when it's not necessary. This patch did wonders to eliminate much of this extra code from being emitted.
> 
Good to know, as we are also working on cleaning up the condition-code handling on AArch64.
That said, this patch has been in the making of a while (the added complexity of affecting the arch-independent lower slowed us down).
I die see 0f7cc55 earlier this week, when rebasing our change onto master.

Note that getting the condition-code handling right for AArch64 will require to model all of the subset- and inverted-subset flags.
So it will be a bit until we have our changes ready for that one.
> If you think our forthcoming patch sounds promising, then perhaps we can hold off introducing this new complexity for integer multiplication here. Also, there's a good chance that I may have overlooked some other advantage of your patch. If that's the case, then I'm sorry. Can you comment?
> 

Even with a smarter handling of the SF, we’ll need some special lowering of the integer multiplications during lowering. The idea was to lower in such a way that we get the overflowing multiplication (which really consists of an inline fast-path and a slow-path in the runtime) out of the way.

So I believe that even when you kill the unneeded SFs, you’ll be stuck with this one from MulIntO.

Cheers,
Phil.

ptomsich:
Oh by the way — this will have broken PPC64, as we don't have a machine to test in our lab.

dave-estes:
> Note that getting the condition-code handling right for AArch64 will require to model all of the subset- and inverted-subset flags.
So it will be a bit until we have our changes ready for that one.

We may be onto the same idea. My forthcoming patch simply undefs the entire SF Vreg, but that's a bit overkill. I hope to work out something later that carefully looks at the uses of SF and the exact condition codes that are examined in those uses order to determine precisely which bits of the SF need to be set. I stopped short on this idea until I can get some data that shows it is promising enough to pursue.

Is this similar to what you mean by your comments above?

ptomsich:
> On 16 Sep 2016, at 19:47, Dave Estes <notifications@github.com> wrote:
> 
> Note that getting the condition-code handling right for AArch64 will require to model all of the subset- and inverted-subset flags.
> So it will be a bit until we have our changes ready for that one.
> 
> We may be onto the same idea. My forthcoming patch simply undefs the entire SF Vreg, but that's a bit overkill. I hope to work out something later that carefully looks at the uses of SF and the exact condition codes that are examined in those uses order to determine precisely which bits of the SF need to be set. I stopped short on this idea until I can get some data that shows it is promising enough to pursue.
> 
> Is this similar to what you mean by your comments above?
> 

Yes it is… we’ve done this in GCC and it helped us quite a bit to generate more optimal code.
If you’re interested, take a look at aarch64.md and look at the various uses of the CC-register with a qualifier (e.g. CC:SWP).

ptomsich:
Independent of this, we’ll also need this to get merged, as the current approach of synthesizing a complete SF for a MulIntO is going to stand in your way.
As long as irlower-arith requests an overflow-flag for its conditional-branch, there’s little you’ll be able to do to kill the SF and derive a “best-match” flag-subset.

I had thought about having irlower-arith ask the backend what the appropriate flag is, but then that didn’t help with the MulInt expansion (which doesn’t need a condition-code at all). So the cleanest was to split the flag-generating (i.e. overflow-checking) integer-multiply (as ‘mulint’) from the low-level integer-multiply (‘imul’).

mxw:
Just to clarify—this makes code on x64 strictly worse, right?

ptomsich:
For x64, the same lowering is used as before.
It has only been moved from irlower-arith to vasm-x64.

In other words: x64 should have exactly the same performance as before for the generated code.

mxw:
Oh yeah, I see—didn't have `irlower-arith.cpp` in front of me before.  I'll take a closer look now.

hphp/runtime/vm/jit/irlower-arith.cpp
```diff
@@ -127,6 +126,17 @@ ARITH_OPS
 
 ///////////////////////////////////////////////////////////////////////////////
 
+
+void cgMulIntO(IRLS& env, const IRInstruction* inst) {
+  //  implMulIntO(vmain(env), env, inst);
```
mxw: ??

hphp/runtime/vm/jit/vasm-dead.cpp
```diff
@@ -94,6 +95,8 @@ bool effectful(Vinstr& inst) {
     case Vinstr::fctidz:
     case Vinstr::fcvtzs:
     case Vinstr::imul:
+    case Vinstr::smulh:
+    case Vinstr::mul:
```
mxw: I think everything else in this switch is sorted alphabetically; can you follow suit?

hphp/runtime/vm/jit/vasm-instr.h
```diff
@@ -939,6 +948,8 @@ struct incqm { Vptr m; VregSF sf; };
 struct incqmlock { Vptr m; VregSF sf; };
 // mul: s0 * s1 => d, sf
 struct imul { Vreg64 s0, s1, d; VregSF sf; };
+struct mul { Vreg64 s0, s1, d; };
+struct smulh { Vreg64 s0, s1, d; };
```
mxw: What are `mul{}` and `smulh{}`?  You've designated them as cross-architecture, so you will also need to implement them for x64 and ppc64 (or coordinate with the ppc64 authors, @gut and @lbianc among others, on IRC).  But as far as I can tell, they're only used for the ARM target, so I don't see why they can't be ARM-specific opcodes.

hphp/runtime/vm/jit/vasm-instr.h
```diff
@@ -992,6 +1003,7 @@ struct cmpq { Vreg64 s0; Vreg64 s1; VregSF sf; };
 struct cmpqi { Immed s0; Vreg64 s1; VregSF sf; };
 struct cmpqm { Vreg64 s0; Vptr s1; VregSF sf; };
 struct cmpqim { Immed s0; Vptr s1; VregSF sf; };
+struct cmpqsign { Vreg64 s0; Vreg64 s1; VregSF sf; };
```
mxw: Same with this.

hphp/runtime/vm/jit/vasm-instr.h
```diff
@@ -890,6 +894,11 @@ struct srem { Vreg s0, s1, d; };
  */
 struct divint { Vreg s0, s1, d; };
 
+/*
+ * Integer multiplication.
+ */
+struct mulint { Vreg s0, s1, d; Vlabel targets[2]; };
```
mxw: Document that this is multiplication with overflow detection, and what the branches mean; it's definitely not clear without any explanation.

Also, probably rename `mulint{}` to `muloint{}` or `mulinto{}` or `mulo{}` or something.

ptomsich:
I actually put something like this in earlier today, but used the CC_O flag.
Will this be ok?

Here’s the diff from my working area:

diff --git a/hphp/runtime/vm/jit/vasm-ppc64.cpp b/hphp/runtime/vm/jit/vasm-ppc64.cpp
index 93927f7..481114b 100644
--- a/hphp/runtime/vm/jit/vasm-ppc64.cpp
+++ b/hphp/runtime/vm/jit/vasm-ppc64.cpp
@@ -1469,6 +1469,13 @@ void lowerForPPC64(Vout& v, absdbl& inst) {
   v << copy{after_conv, inst.d};
 }
 
+void lowerForPPC64(Vout& v, mulint& inst) {
+  auto sf = v.makeReg();
+
+  v << imul{inst.s1, inst.s0, inst.d, sf};
+  v << jcc{CC_O, sf, inst.targets[0], inst.targets[1]};
+}
+
 void lowerForPPC64(Vout& v, decqmlock& inst) {
   Vptr p = inst.m;
   patchVptr(p, v);

ptomsich:
All requested changes are addressed with one exception: ARM64 imul continues to not set flags.
I couldn't find an instance where the flags from the imul might get used, so this feels safe for now.

gut:
@ptomsich : Naturally PPC64's `jcc` can handle the `CC_O` parameter but I just wrote `CC_NE` as it was the case for X64. Why would only PPC64 use `CC_O` then?

gut:
Oh, sorry, after your changes they are the same: PPC64 and X64 use `CC_O` as the `jcc` parameter. Ok

gut:
@ptomsich : FYI, I placed your commits on top of master and they work as expected on quick test suite. (due to lack of CI on PPC64 I did it by hand).

ptomsich:
@dave-estes: I see how your approach differs by having the ’sf.isValid()’ as a all-or-nothing decision from what my team had in mind… let me explain...

The difference in opinion seems to come from the fact that your approach still defines _all_ possible flags.
If you look at the consumer for the flag-outputs from the ‘mul’ vasm, there’s multiple (exclusive cases) for using them:
	— either you want to know if an overflow (V) has occurred (which is the most costly derived flag)
	— (or) you would be interested in NZ (for a signed multiplication)
	— (or) you would be interested in N (for an unsigned multiplication)

The NZ flags can be generated in a single instruction by using a ‘cmp’ (which is a ‘subs’ in reality).
The V flag needs to be generated by the smulh/cmp sequence we have in our mulinto lowering now.

For a flag-consumer, it doesn’t make sense to test _both_ for overflow _and_ for NZ (after all… NZ will not make sense, if an overflow has occurred).
Assuming that we have two flag-consumers, where one tests for the overflow and the other for NZ (similar to the code that is in the heavy-handed imul expansion today)… then it’s best to delay the generation of the NZ flag-generation until the basic-block that needs it.

However, when we’re in the basic block that consumes the NZ, there’s most likely a better way to express the assembly sequence anyway (i.e. rewriting in vasm-simplify): a b.eq/b.ne could be written as compare-and-branch.
So our plan had been to have two separate ‘virtual’ SF registers generated, be aggressive in the vasm-simplify phase (to get rid of any cases that don't necessarily need the NZ flags and can be expressed differently) and kill any unused ones in DCE.

One thing we want to avoid at all cost would be the MRS/MSR instructions, as I expect these to be costly on pretty much any u-arch.
All this makes the integer-multiplication different from most other cases on AArch64, which offer us the luxury of having a “and-set-flags” variant.

Let me know if this clarifies my thoughts on the CC register.

dave-estes:
Thanks a ton, @ptomsich. OK, I think I see what you're getting at. Is this a good summation?

When we finally have separate VRegs for each SF bit, then we'll have precise liveness information for each of those bits. With my approach, the forthcoming undef SF pass would then be able to selectively emit the instruction sequences to set each of those bits and would then OR them all up. With your approach we to do that much earlier using separate basic blocks (or just vasm) to set each of the bits. Your approach of using IR gives you the ability to actually move the setting of the bits closer to their uses and hopefully avoiding setting bits when a previous bit's use short-circuits the execution.

Sounds reasonable to me. My general approach to our port has been to implement as much as possible in vasm before emitting. The same patch that we will introduce the undef SF pass with also moves a lot of the implementation of the more CISC-like vasm into lowerers from their current emitters.

Still, I feel this is a very complicated change without any benefit to the codebase today. It lays the groundwork for a future optimization that relies on an uncertain multi-VReg SF feature. Furthermore, we don't even have a known workload code sequence that checks for overflow, negative, and zero flags from a single imul. If we had a critical workload or benchmark, then we'd have a much stronger motivation to design and implement the multi-VReg SF feature. And if that happened, then I'd totally be on board with your approach. Until then, I think setting flags while emitting and then landing the PR for the undef SF pass is the least disruptive path to provide the most benefit.

ptomsich:
@dave-estes: Could we then agree on either redefining the imul to only generate an overflow flag and on having a separate vasm-token for non-flag-setting integer multiplication? Generating and setting all the flags on AArch64 will introduce a very high overhead, while nobody needs any flag except the overflow.

This will leave only the small detail that AArch64 would set the NZ-flags for overflow to sort out...
When I started this change, irlower-arith assumed that an imul-overflow would signal as a CC_O (which would be the V-flag on AArch64), but in reality we’ll generate a CC_NE.  This could be done with a change of the interface between irlower-arith and the backends (either interrogating the backends to get the appropriate flag or by introducing an additional pseudo-flag CC_MULO which is then mapped onto CC_O or onto CC_NE depending on the backend).

dave-estes:
In our original port, we never actually set the NZ after multiplication. Might want to check with @lpathy, as he's the guy who added this when their port landed. In fact, that complicated my patch to set V quite a bit. I think it's safe to remove for now based on our analysis. Initially our port, we did a little trick where placed a macro into each of the vasm lowerer/emitters that we're yet setting the SF. The macro would assert if the SF Vreg was valid (i.e. wasn't undef'd by our new pass). We did the same analysis as you and found that only the V was needed for imul.

Concerning the new vasm for non-overflow, I think that sounds reasonable. My personal preference would be to leave it as is and just use the undef SF pass to catch it when emitting. But hey, who knows when that'll land. It's a pretty big change. :-) If you want to test it out, then I can probably push it to my fork soon. Just needs a final internal review.

Overall, I'm still in favor of your original approach, but I'd just prefer to see it delayed until we have multi-VReg SF. With that said, if anyone else want this PR to land now, then I certainly won't get in the way.

mxw:
Thanks for the lively discussion, folks!  I think I understand the situation better now.

Having separate Vregs for the individual condition code outputs of an instruction is not something that I see as providing much benefit.  It adds cumbersome boilerplate that is unlikely to be of much use for the mainline x64 target (especially since, in practice, I think we rarely do anything with status flags besides set it and then test it once or twice in quick succession).

I think you could achieve the same thing by simply doing liveness analysis on the existing SF Vregs, while also taking into consideration the ConditionCode immediates of the use-instrs.  As an ARM-specific analysis pass, this has the added benefit of allowing you to define which "categories" of CCs you care about.  I don't know that ARM's flags registers has the same bits that x64's does, so if what you really care about is "which of those bits needs to get set based on the CC's that we use", an arch-specific pass will probably require the least abstraction overhead.  (Even if the underlying flags are the same, this still seems simpler in a world where not all ports benefit from such analysis.)

You can think of this as (if I'm understanding correctly) an extension of the SF-unsetting pass that @dave-estes is talking about.  Rather than do liveness analysis to see if the SF has any uses at all, additionally take into account _how_ it's used.  By annotating the uses on the def-instruction, the ARM emitter should be able to emit the optimal code.

Please let me know if you think I've misunderstood.

In HHVM as it exists in master right now, `imul{}` Vinstr is required to set all applicable status flags.  If you do not either explicitly change this specification, and your port's implementation does not abide by it, then your port is not a working vasm target.  Nothing is stopping anyone from adding code to the irlower module tomorrow which does an `imul{}` and then tests its SF operand for every condition code.

What you're doing in this diff is changing the specification.  I'm fine with having two integer multiplication instructions, one which handles overflow, and the other of which sets all other flags not relating to overflow (with overflow flags set in undefined ways).  However, even though it's unused today, I don't want the "default" `imul{}` instruction to set _no_ flags.  If you want to avoid setting flags in an instruction, do it via liveness analysis.

Note that x64 is already doing SF-liveness optimizations from within `vasm-xls.cpp`.  The "right" thing to do is probably to either separate the analysis that XLS does as a separate module that is consumed both by optimization passes and the register allocator itself, or to invoke those optimization passes with the analysis results as a parameter during XLS.  The only reason this isn't already the case is basically laziness—the existing optimizations are simple enough that I didn't feel it was worth the effort to actually perform the separation.  Now that multiple backends want to do SF-liveness-based optimization, it's time to bite the bullet.  I'll put up a diff and ping you all here when it's ready.

If I haven't missed the mark completely here—and again, please let me know if I have—then I think it's best to drop this diff and wait for a more general solution.  (As an additional argument for why we should wait for a nicer solution is that this is certainly an optimization, and AFAIK, there's no data that says that overflow-checking multiplications are a performance bottleneck for any realistic workloads.  Meanwhile, optimizing flags production for _all_ instructions seems more compelling from first principles, even without data.)

mxw:
I'll close this in a few days if I hear assent with the above (or at least if I don't hear any compelling dissent).

ptomsich:
I don’t think that it will be easy to eliminate the flag-setting using the msr-insn easily, as this will require to add additional analysis to mark the specific usage as an instruction that only sets the SF… and it will make it rather hard to just kill the synthesis of individual flags.  But let’s wait for the patch and then see how to best evolve it from there.

As I’d like to get this optimised to a reasonable state until then, can we agree on the following implementation:
	(a) 'imul{}’ returns complete flags, only if the SF argument is valid => should optimize the ‘MulInt’ case
	(b) MulIntO for AArch64 expands into a non-flag-setting ‘imul{}’ (i.e. SF being invalid) and we then do our little dance with the smulh/cmp
	(c) we keep the ‘mulinto’ vasm-instr that has the explicit branch-labels and kills SF

This seems like a ‘best of both world’ solution and won’t interfere with any of the longer-term optimisation goals we discussed here.
What do you think?
;

mxw:
> I don’t think that it will be easy to eliminate the flag-setting using the msr-insn easily, as this will require to add additional analysis to mark the specific usage as an instruction that only sets the SF… and it will make it rather hard to just kill the synthesis of individual flags.

I don't know what the "msr-insn" is.  But I'm not sure I understand what you mean by the rest of this either.  We already compute the liveness intervals for all Vregs.  Supposing that you have that information, what would prevent you from doing all these optimizations "under the hood"?

If we _can_ "hide" these optimizations, then I disagree that this change gets us to the best of both worlds—it would be better to simply have one single `imul{}` instruction, whose implementation varies depending on what its flags are used for.

mxw:
I'm also not sure what the rush is.  What makes you think that HHVM on ARM is so performance-bottlenecked on multiplication?

ptomsich:
I’d like to flush this out of my backlog of changes that keep lingering in my working trees…

It started as a fix for the float-overflow issue, but was also influenced by a performance-optimisation mindset.
So I don’t expect PHP to be performance-bottlenecked by multiplies… yet, there’s no reason to make it easy for someone to come up with a microbenchmark that demonstrates how much better x64 is.

mxw:
I don't view wanting to clear your queue as a good reason to land this change—you can just strip this change locally, and pick it up again in the future if the need arises.

Your second reason is no more compelling.  If someone were to compare x64 against ARM performance on HHVM today, they would first find that the ARM port is not at 100% correctness, and would likely then see a substantive performance discrepancy, which I can guarantee is not a result of an extra branch in the instruction sequence for multiplying integers.

I understand your desire to tighten up everything you can—but there are better and more general ways to achieve this (which it sounds like both your team and @dave-estes are working on in some manner).  There's just no practical need to introduce this intermediate state; and, in fact, I would argue that changes such as this, which are better done under the covers by an optimization pass in conjunction with the ARM emitter, are actually bad, because it runs us into one of the failure modes we defined when we started admitting non-x64 ports, which was that we didn't want the ports to interfere unnecessarily with the complexity of the x64 backend that we work on internally, day-to-day.

I'm gonna close this out for now; we can revisit it if the SF liveness solutions are somehow unable to solve the problem.

PR #7359, opened 2016-09-19
[PPC64] Improve decoder and other low level asm/disasm

gut:


gut:
@mxw : Hi. Can you check if those failed tests are indeed a concern?

aorenste:
@gut - @mxw approved this internally and it's in the process of being pushed so I would guess the tests are fine.
:)

gut:
@aorenste ok, thanks.

Orvid:
The failed test was just something being flaky. It's not an issue with this diff.

PR #7360, opened 2016-09-19
Simplify some common cases of overflowing integer-multiplication

ptomsich:
While working on the overflowing integer-multiplication for AArch64, I can across the fact that some easy simplifications of integer multiplication hadn't been implemented.

The basic idea is similar to what can be done on a non-overflowing integer-multiply, but I decided to stay away from commutative transforms as these might affect precision when overflowing into floating-point.

### DIFF
--- a/hphp/runtime/vm/jit/simplify.cpp
+++ b/hphp/runtime/vm/jit/simplify.cpp
@@ -742,6 +742,27 @@ SSATmp* simplifyMulDbl(State& env, const IRInstruction* inst) {
   return constImpl(env, inst->src(0), inst->src(1), std::multiplies<double>());
 }
 
+SSATmp* simplifyMulIntOWithConstant(State& env, SSATmp* s, int64_t c)
+{
+  // X * 0 --> 0
+  if (c == 0)
+    return cns(env, 0);
+
+  // X * 1 --> X
+  if (c == 1)
+    return s;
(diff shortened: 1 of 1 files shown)


hphp/runtime/vm/jit/simplify.cpp
```diff
@@ -742,6 +742,27 @@ SSATmp* simplifyMulDbl(State& env, const IRInstruction* inst) {
...
+  if (c == 0)
+    return cns(env, 0);
+
+  // X * 1 --> X
+  if (c == 1)
+    return s;
+
+  // X * (-1) --> (0 - X)
+  if (c == -1)
+    return gen(env, SubInt, cns(env, 0), s);
```
nikic: Shouldn't this be `SubIntO` for the X == INT_MIN case?

ptomsich:
Yes, it should.

hphp/runtime/vm/jit/simplify.cpp
ptomsich: Turns out that SubIntO on AArch64 doesn't handle that right either... so I'll need to add additional test cases and fix the SubIntO while doing this one.

ptomsich:
No point in keeping this open at this time—I'll revise and reopen a pull request, once done.

PR #7361, opened 2016-09-19
[AArch64] Adds cmake option to set mcpu for GCC.

dave-estes:
GCC defaults to cortex-a53, which is inappropriate for
out-of-order machines. Additionally sets signed chars and disables
the fix for cortex-a53 errata 843419 which GCC applies even when
mcpu != cortex-a53.

### DIFF
--- a/CMake/HPHPCompiler.cmake
+++ b/CMake/HPHPCompiler.cmake
@@ -175,6 +175,28 @@ if (${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang" OR ${CMAKE_CXX_COMPILER_ID} STREQU
       list(APPEND GENERAL_CXX_OPTIONS "mcrc32")
     endif()
 
+    # ARM64
+    if(IS_AARCH64)
+      # Force char type to be signed, which is not the case on aarch64.
+      list(APPEND GENERAL_OPTIONS "fsigned-char")
+
+      # If a CPU was specified, build a -mcpu option for the compiler.
+      set(CPU "" CACHE STRING "CPU to tell gcc to optimize for (-mcpu)")
+      if(CPU)
+        list(APPEND GENERAL_OPTIONS "mcpu=${CPU}")
+      endif()
+
+      # Make sure GCC is not using the fix for errata 843419. This change
+      # interferes with the gold linker. Note that GCC applies this fix
+      # even if you specify an mcpu other than cortex-a53, which is why
+      # it's explicitly being disabled here for any cpu other than
+      # cortex-a53. If you're running a newer pass of the cortex-a53, then
+      # you can likely disable this fix with the following flag too. YMMV
+      if(NOT ${CPU} STREQUAL "cortex-a53")
+        list(APPEND GENERAL_OPTIONS "mno-fix-cortex-a53-843419")
+      endif()
+    endif()
+
     # PPC64
     if(NOT IS_PPC64)
       list(APPEND RELEASE_CXX_OPTIONS "momit-leaf-frame-pointer")


apinski-cavium:
signed chars is a bad idea.

apinski-cavium:
I think this patch should be reverted.  It is not correct at all.

apinski-cavium:
>the fix for cortex-a53 errata 843419 which GCC applies even

Only for Ubuntu and Debian.  It depends on how GCC was configured.  Again this patch is broken and should be reverted.

Orvid:
I'm curious, but what's wrong with signed chars? x86_64 defaults to signed chars, so changing that just makes it more like the original platform HHVM was written for, doesn't it?

As-for the GCC options, there is definitely value in the patch as it stands, as Ubuntu and Debian are the only distros we currently provide official packages for (admittedly not ARM packages, but still). Is there a good way to detect whether the GCC being used to compile needs to have the errata flag disabled?

CMake/HPHPCompiler.cmake
```diff
@@ -175,6 +175,28 @@ if (${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang" OR ${CMAKE_CXX_COMPILER_ID} STREQU
       list(APPEND GENERAL_CXX_OPTIONS "mcrc32")
     endif()
 
+    # ARM64
+    if(IS_AARCH64)
+      # Force char type to be signed, which is not the case on aarch64.
+      list(APPEND GENERAL_OPTIONS "fsigned-char")
+
+      # If a CPU was specified, build a -mcpu option for the compiler.
+      set(CPU "" CACHE STRING "CPU to tell gcc to optimize for (-mcpu)")
```
Orvid: As a side note, when fixing this up, use an `OPTION` instead of a normal variable. (yes, I know my code for the MSVC options doesn't use them, I should fix that >_>...)

Also, don't name it `CPU`, as it's too ambiguous as-to what the option actually does. For the MSVC section, I named the similar, but not exact same, option `MSVC_FAVORED_ARCHITECTURE`. Perhaps `AARCH64_TARGET_ARCHITECTURE`?

apinski-cavium:
>I'm curious, but what's wrong with signed chars? x86_64 defaults to signed chars, so changing that just makes it more like the original platform HHVM was written for, doesn't it?
You changed the ABI.   char is a not a standard integer type and should be treated that way. 

>Is there a good way to detect whether the GCC being used to compile needs to have the errata flag disabled?

Why disable it at all?  It might make sense to run HHVM on an cortex-a53 later on for a binary build.  There are many cortex-a53 machines which are cheap that are out there in the wild.  That can be used for testing.

dave-estes:
@apinski-cavium, Sorry I didn't respond earlier. Was on vacation last week.

Concerning char sign: we set signed chars to match the Intel default. We found a few places in HHVM that weren't cast correctly. @lpathy fixed a few as well. We were concerned about latent bugs rearing their head, so we set the default to AArch64 to match Intel. We've been running like this for ages and haven't had any problems.

Concerning the A53 errata: Keeping it on means that we have problem linking when building natively on non-A53 cores. I understand your concerns about building a single binary that will run on any core. To do this, just configure without specifying a CPU and build on you A53 machines. If you don't specify a CPU, then this errata isn't disabled and GCC uses it's default: A53. We did this on purpose, because we anticipated that distro packagers would configure HHVM without a CPU flag.

I'm not sure what your other comments about being "broken" and "not correct at all", since I couldn't determine a context for those comments. Please let me know what isn't working well for you. Or better yet, maybe you can address the shortcomings in a new PR. If you do, would you mind addressing the comment from @Orvid about using a more specific option instead of CPU?

Thanks...
-Dave

apinski-cavium:
>We were concerned about latent bugs rearing their head, so we set the default to AArch64 to match Intel. 

So what.  Again this means now you now two ABIs on ARM64, one which HHVM uses (char being signed and one for the rest of linux, unsigned).  This is not something which should be done lightly or even done at all.  This part of the patch should be reverted for sure, the other parts are questionable but less so than this part.  I am more concerned people are not writing portable C++11 code and not knowing about how char is treated is one example of people not knowing C++.

>Keeping it on means that we have problem linking when building natively on non-A53 cores.

If you reported the bug correctly to the Ubuntu/upstreams for the gold linker crash then there would be no issues.  I reported it to the list but without a testcase.  Instead I worked around it a different way of disabling gold usage for HHVM.  I still think gold is in beta for ARM64 rather than a production linker.


And really this should have been a few pull requests rather than one big pull request.

mxw:
Commenting here for posterity, since I don't think I ever saw this diff:

I think the concern about multiple ABIs on Aarch64 is compelling—we're basically trading off surprising behavior at ABI boundaries for surprising behavior within HHVM itself, one of which we have full control over.

I don't understand the resistance to the resistance to nixing the broken errata fix in gold, though.  I'm not convinced that reporting upstream will result in an instantaneous fix propagating across all machines that need to compile HHVM on Aarch64.  Moreover, we're basically forced into using gold because ld uses so much memory that it always fails on an 8GB machine, which is what I'm using through Linaro.  If I can't compile HHVM on that machine, then from FB's perspective, the port doesn't even compile, which would be an unfortunate result.

PR #7362, opened 2016-09-19
[AArch64] Optimize ldimm from ARM.

dave-estes:
Disables an x86_64-specific optmization which converted
ldimm Rd, #0 to xor (eor) Rd, Rd, Rd. It's faster for ARM to use
a simple Mov Rd, #0.


mxw:
This whole pass is supposed to be guarded on `arch() == Arch::X64`.  Did you actually witness this kicking in before making this patch?

dave-estes:
Oops. Sorry about that. This is an old patch of ours, and I think we found it through inspection. We noticed other uses of arch() == Arch::ARM and extrapolated from there. Never caught the guard for the whole pass. Gonna close.

PR #7363, opened 2016-09-19
[AArch64] Add disassembly for TRACE=ustubs.

dave-estes:


hphp/util/disasm.cpp
```diff
@@ -146,6 +148,26 @@ void Disasm::disasm(std::ostream& out, uint8_t* codeStartAddr,
...
+  auto end = reinterpret_cast<vixl::Instruction*>(codeEndAddr);
+
+  while (i <= end) {
+    decoder->Decode(i);
+    out << "  " << disasm->GetOutput() << std::endl;
+    i = i->NextInstruction();
+  }
+
+  delete decoder;
+  delete disasm;
```
mxw: Why can't you use the stack here?

hphp/util/disasm.cpp
```diff
@@ -73,7 +75,7 @@ static const xed_syntax_enum_t s_xed_syntax =
 void Disasm::disasm(std::ostream& out, uint8_t* codeStartAddr,
                     uint8_t* codeEndAddr) {
 
-#ifdef HAVE_LIBXED
+#if (defined __x86_64__) && (defined HAVE_LIBXED)
```
mxw: Prefer `defined(__x86_64__) && defined(HAVE_LIBXED)`

dave-estes: Will do.

hphp/util/disasm.cpp
dave-estes: Will do.

hphp/util/disasm.cpp
```diff
@@ -146,6 +148,23 @@ void Disasm::disasm(std::ostream& out, uint8_t* codeStartAddr,
     frontier += instrLen;
     ip       += instrLen;
   }
+
+#elif (defined __aarch64__)
```
mxw: `defined(__aarch64__)`

mxw:
The test failures are because `hphp/util/` isn't allowed to depend on `hphp/vixl/`.  `util/` is very carefully curated and kept minimal, so I think the best solution here is to add a `jit/disasm.h` module and switch on `arch()` in there, then replace uses of `Disasm` with that.  (We could rename `util/disasm.*` to `util/disasm-x64.*`, perhaps.)

dave-estes:
What do you think about moving hphp/util/asm.* and hphp/util/disasm.*  to .../jit/ and then using switch arch() for their implementations. I confirmed that the only files including disasm and asm were already in .../jit/?

dave-estes:
That was a good idea, @mxw. In fact, there was already a disassembly printer in .../jit/print.cpp, so I just changed unique-stubs.cpp to use that printer instead. As a bonus, it has support for PPC64 now too.

mxw:
You need to rebase this diff—there is no longer such a file as `mc-generator.cpp`.

dave-estes:
It's rebased now. I ended up pushing with --force. I hope that's the right way to get changes up after a rebase.

mxw:
Seems to have worked.  Tests should be good, so I'll land this once they're all clear.

PR #7364, opened 2016-09-19
[AArch64] Fixes double comparison for NAN.

dave-estes:
Fixes:
  test/slow/php7_bcbreak/7/int_semantics/nan_inf_cast.php

### DIFF
--- a/hphp/runtime/vm/jit/abi-arm.h
+++ b/hphp/runtime/vm/jit/abi-arm.h
@@ -89,6 +89,27 @@ inline vixl::VRegister x2v(PhysReg x64reg) {
   return vixl::VRegister(vixl::CPURegister(x64reg));
 }
 
+/*
+ * ARM    int meaning                   fp meaning                         JIT mapping        JIT meaning
+ * ------------------------------------------------------------------------------------------------------------------
+ * eq     equal                         equal                              CC_E/CC_Z          equal
+ * ne     not equal                     not equal or unordered             CC_NE/CC_NZ        not equal
+ * cs/hs  carry set                     greater than, equal, or unordered  CC_AE/CC_NB/CC_NC  unsigned higher or same
+ * cc/lo  carry clear                   less than                          CC_B/CC_NAE        unsigned lower
+ * mi     minus, negative               less than                          CC_S               minus (sign set)
+ * pl     plus, positive, or zero       greater than, equal, or unordered  CC_NS              plus (sign clear)
+ * vs     overflow                      unordered                          CC_O               overflow set
+ * vc     no overflow                   ordered                            CC_NO              overflow clear
+ * hi     unsigned higher               greater than, or unordered         CC_A/CC_NBE        unsigned higher
+ * ls     unsigned lower or same        less than or equal                 CC_BE/CC_NA        unsigned lower or same
+ * ge     signed greater than or equal  greater than or equal              CC_GE/CC_NL        signed greater or equal
+ * lt     signed less than              less than, or unordered            CC_L/CC_NGE        signed less than
+ * gt     signed greater than           greater than                       CC_G/CC_NLE        signed greater than
+ * le     signed less than or equal     less than, equal, or unordered     CC_LE/CC_NG        signed less or equal
+ * al     always (not used)             always (not used)                  CC_P               parity
+ * nv     not valid                                                        CC_NP              no parity
+ */
+
 inline vixl::Condition convertCC(jit::ConditionCode cc) {
   assertx(cc >= 0 && cc <= 0xF);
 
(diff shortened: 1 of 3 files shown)


mxw:
All of this needs to go under ARM-specific lowering unless there's are specific reasons why that's impossible or immensely cumbersom.

dave-estes:
I'll see what I can cook up to refactor this so that these differences when handling NaN in comparisons on ARM are lowered into vasm-*.[cpp/h] and abi-*.[cpp/h].

hphp/runtime/vm/jit/abi-arm.h
```diff
@@ -89,54 +89,80 @@ inline vixl::VRegister x2v(PhysReg x64reg) {
   return vixl::VRegister(vixl::CPURegister(x64reg));
 }
 
+/*
```
dave-estes: Is this table a valid exception case for going beyond 80 columns?

hphp/runtime/vm/jit/irlower-cmp.cpp
```diff
@@ -241,16 +241,23 @@ void cgCmpDbl(IRLS& env, const IRInstruction* inst) {
...
+    // On ARM, CC_L => lt, which for FP is true for lower or unordered
+    v << cmovq{CC_L, sf, v.cns(1), v.cns(-1), tmp};
+    v << cmovq{CC_E, sf, tmp, v.cns(0), d};
+  } else {
+    auto const tmp1 = v.makeReg();
+    auto const tmp2 = v.makeReg();
+    v << cmovq{CC_A, sf, v.cns(-1), v.cns(1), tmp1};
+    v << cmovq{CC_NE, sf, v.cns(0), tmp1, tmp2};
+    v << cmovq{CC_P, sf, tmp2, v.cns(-1), d};
+  }
```
dave-estes: I can actually implement this for ARM using the same three cmovq with the exception of the third which will use a CC_O for ARM vs. the CC_P. However, implementing it with two cmovq's leverages ARM's special unordered condition codes to eliminate the third explicit check for an unordered result. I assumed it's OK to have some lightly platform-specific code in here base another precedent earlier in the file. See 214-226 above.

With that said, I can easily revert back to using the three cmovqs for ARM too and then I'll just implement some special CC codes with switched implementations for the different platforms in their abi-xx files. Just let me know.

hphp/runtime/vm/jit/irlower-conv.cpp
```diff
@@ -198,17 +198,23 @@ void cgConvDblToInt(IRLS& env, const IRInstruction* inst) {
   v << cvttsd2siq{src, d};
   v << cmpq{indef, d, sf};
 
+  // For ARM, use the cc which ignores unordered results for the first
+  // condition. Then use the ARM equivalent cc to check for the unordered
+  // result in the second condition.
+  ConditionCode cc_nb = (arch() == Arch::ARM) ? CC_GE : CC_NB;
+  ConditionCode cc_p = (arch() == Arch::ARM) ? CC_O : CC_P;
```
dave-estes: Similar to my comment above, if we need a more portable way to represent these special condition codes for ARM, I can add a mechanism to look these up in the abi-xx files for each platform. I didn't in this patch, because I *feel* that the reduced complication of this new patch made these uses of `arch() ==` much more palatable than the previous patch. As I said earlier though, I'm happy to do it. Just let me know.

mxw: Is it just coincidental that you can just use different CC's here?  That is, does `CC_GE` on ARM not actually do the same thing as `CC_NB` on x64, but it just so happens that we can still just do two checks and get the same semantics?

dave-estes: I'd say that this new fix boils down to the inherent differences with ARM. Firstly, there is not parity bit. Instead if providing an explicit flag and then simple condition codes to test that flag, ARM provides multiple sets of condition codes that allow you to test >= without concern for unordered results and then with concern for unordered results.

Secondly, many of the ARM condition codes have multiple meanings depending on whether the context for the comparison was int or fp.

The table I added to abi-arm.h is a great reference for the ARM cc's, their meanings, and how we've mapped them to the JIT cc's and the JIT cc meanings.

With that background out of the way, the first conditional assignment of  `cc_nb` above sets it to `CC_GE` for ARM, because that maps to the ARM cc (ge) which doesn't take into account unordered results. This code here in irlower-conv.cpp doesn't consider unordered results in the first test using `cc_nb`. The second conditional assignment of `cc_p` above sets it to `CC_O` for ARM, because that it is mapped to `vs`, which is normally the overflow test in the integer context, but for the fp context it indicates an unordered result.

I tried to work this out in the actual mapping from ARM to JIT cc's, but the dual meaning of ARM cc's made it tough. I also considered introducing a new group of JIT cc's that would allow us to capture the semantics of ARM's different cc's that are concerned with unordered results too, but I thought this was way to heavy-handed.

Anyhow, my understanding of x86_64 cc's is shaky at best, so if indeed X64 is more similar to ARM than I think, then perhaps this can actually be solved in a more portable way through mapping and/or adding new JIT cc's.

And to answer your question, I wouldn't call it a coincidence, per se, but it sure is funny how such a simple change fixed the problem. Switching on Arch::ARM in the location which is producing the checks is a subtle way to take into account the context of operations being performed at that location.

mxw:
Sorry for the radio silence here; I need to sink some time into thinking about this.  I'll try to do that soon.

swalk-cavium:
Will this change fix
hphp/test/slow/php7_bcbreak/7/int_semantics/nan_inf_cast.php FAILED (4.78s)
it looks to be in the correct area.

dave-estes:
Sure will, @swalk-cavium, but as you can tell from the discussion it introduced some porting code to the irlower layer and we want to try to avoid that. @mxw is keeping this alive for reference while I work out the other solution.

That other solution is being rolled into what I was calling an Undef SF pass. As we discussed a bit on the last AArch64 call, that pass is being redesigned to be much more general. Just thought I'd mention this because I have a working patch for that pass that I intended to push this week. I'll try to make sure I mention @swalk-cavium and @ptomsich because the pass fixes the complicated mapping of condition codes between JIT (x64) and AArch64 as well as tracks exactly which status flags bits need to be synthetically set. This drastically improves the implementation of imul and many other instructions.

dave-estes:
Hey Folks. This is the PR to which I was referring: https://github.com/facebook/hhvm/pull/7481

mxw:
I'm going to close this because @dave-estes's flags-weakening pass is up, but feel free, all, to let me know if you think this should be re-opened and reconsidered.

PR #7370, opened 2016-09-20
[AArch64] Enable redundant extension elimination for AArch64

ptomsich:
AArch64 uses implicit zero-extension when performing 32bit operations.
The following rules (from the ARMv8 specification) apply for the 32bit instruction forms:
- The upper 32 bits of the source registers are ignored.
- The upper 32 bits of the destination register are set to zero.
- Right shifts and right rotates inject at bit[31], not at bit[63].
- The condition flags, where set by the instruction, are computed from the lower 32 bits.

Based on this, we can enable the same optimisation pass for eliminating redundant zero-extensions, as for x64.

### DIFF
--- a/hphp/runtime/vm/jit/vasm-simplify.cpp
+++ b/hphp/runtime/vm/jit/vasm-simplify.cpp
@@ -375,22 +375,35 @@ bool simplify(Env& env, const copyargs& inst, Vlabel b, size_t i) {
 }
 
 bool simplify(Env& env, const movzlq& inst, Vlabel b, size_t i) {
-  if (arch() != Arch::X64) return false;
-  auto const def_op = env.def_insts[inst.s];
+  switch (arch()) {
+    // Both x64 and AArch64 have implicit zero-extension, so we want to do some
+    // redundant extension elimination...
+  case Arch::X64:
+  case Arch::ARM:
+    {
+      auto const def_op = env.def_insts[inst.s];
+      
+      // Check if `inst.s' was defined by an instruction with Vreg32 operands, or
+      // movzbl{} in particular (which lowers to a movl{}).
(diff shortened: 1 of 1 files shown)


hphp/runtime/vm/jit/vasm-simplify.cpp
```diff
@@ -375,22 +375,35 @@ bool simplify(Env& env, const copyargs& inst, Vlabel b, size_t i) {
...
 bool simplify(Env& env, const movzlq& inst, Vlabel b, size_t i) {
-  if (arch() != Arch::X64) return false;
-  auto const def_op = env.def_insts[inst.s];
-
-  // Check if `inst.s' was defined by an instruction with Vreg32 operands, or
-  // movzbl{} in particular (which lowers to a movl{}).
-  if (width(def_op) != Width::Long &&
-      def_op != Vinstr::movzbl) {
-    return false;
+  switch (arch()) {
```
mxw: Can you just do `if (arch() == Arch::PPC64) return false;` instead?

Might be time for me to split up vasm-simplify per-architecture.

ptomsich: I considered the explicit list using a switch to be safer in case that another architectures comes along some day.  Note that the only thing that comes to mind is MIPS64, so this is probably a nit-pick by me.

The benefits of listing this explicitly in a switch-statement are:

- any new architectures won't pick this up automatically
- the compiler will generate a warning if not all architectures are listed

That said: if you want it changed, I'll change it...

mxw:
I'm actually just going to slide this into one of my diffs that's going to go out soon.  I'll close this once it's in.

mxw:
This should be done in master now.  Closing.

PR #7371, opened 2016-09-20
Add explicit indirect return arguments.

dave-estes:
There are stark differences between the x86_64 and AArch64 ABIs
with regards to indirect return arguments. This work is similar
to the work done for the interpreter when calling native functions.
In this case though, the ArgGroups made the port a little more
difficult. This change expands the ArgGroups to handle indirect
return arguments explicitly when the target supports them.
Otherwise, the behavior for targets without dedicated regsiters for
indirect return arguments is to pass it as the first general
purpose argument.

Fixes (AArch64):
   test/slow/php7_backported/ext_openssl/openssl_encrypt_ccm.php
   test/slow/php7_backported/ext_openssl/openssl_encrypt_gcm.php

### DIFF
--- a/hphp/runtime/vm/jit/abi-arm.cpp
+++ b/hphp/runtime/vm/jit/abi-arm.cpp
@@ -161,12 +161,19 @@ PhysReg rarg_simd(size_t i) {
   assertx(i < num_arg_regs_simd());
   return vixl::FPRegister::DRegFromCode(i);
 }
+PhysReg rarg_ind_ret(size_t i) {
+  assertx(i < num_arg_regs_ind_ret());
+  return vixl::x8;
+}
 
 RegSet arg_regs(size_t n) {
   return jit::arg_regs(n);
 }
 RegSet arg_regs_simd(size_t n) {
   return jit::arg_regs_simd(n);
+}
+RegSet arg_regs_ind_ret(size_t n) {
+  return jit::arg_regs_ind_ret(n);
 }
 
 PhysReg r_svcreq_req() { return rarg(0); }
--- a/hphp/runtime/vm/jit/abi-arm.cpp
+++ b/hphp/runtime/vm/jit/abi-arm.cpp
@@ -169,12 +169,6 @@ PhysReg rarg_ind_ret(size_t i) {
 RegSet arg_regs(size_t n) {
   return jit::arg_regs(n);
 }
-RegSet arg_regs_simd(size_t n) {
-  return jit::arg_regs_simd(n);
-}
-RegSet arg_regs_ind_ret(size_t n) {
-  return jit::arg_regs_ind_ret(n);
-}
 
 PhysReg r_svcreq_req() { return rarg(0); }
 PhysReg r_svcreq_stub() { return rarg(1); }
--- a/hphp/runtime/vm/jit/abi-ppc64.cpp
+++ b/hphp/runtime/vm/jit/abi-ppc64.cpp
@@ -172,12 +172,19 @@ PhysReg rarg_simd(size_t i) {
   assertx(i < num_arg_regs_simd());
   return simd_args[i];
 }
+PhysReg rarg_ind_ret(size_t i) {
+  assertx(i < num_arg_regs_ind_ret());
+  return PhysReg();
+}
 
 size_t num_arg_regs() {
   return sizeof(gp_args) / sizeof(PhysReg);
 }
 size_t num_arg_regs_simd() {
   return sizeof(simd_args) / sizeof(PhysReg);
+}
+size_t num_arg_regs_ind_ret() {
+  return 0;
 }
 
 
--- a/hphp/runtime/vm/jit/abi-ppc64.cpp
+++ b/hphp/runtime/vm/jit/abi-ppc64.cpp
@@ -183,9 +183,6 @@ size_t num_arg_regs() {
 size_t num_arg_regs_simd() {
   return sizeof(simd_args) / sizeof(PhysReg);
 }
-size_t num_arg_regs_ind_ret() {
-  return 0;
-}
 
 
 PhysReg r_svcreq_sf() {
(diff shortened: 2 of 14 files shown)


mxw:
What was incorrect about @lpathy's existing solution?

hphp/runtime/vm/jit/abi-arm.cpp
```diff
@@ -161,13 +161,20 @@ PhysReg rarg_simd(size_t i) {
   assertx(i < num_arg_regs_simd());
   return vixl::FPRegister::DRegFromCode(i);
 }
+PhysReg rarg_ind_ret(size_t i) {
+  assertx(i < num_arg_regs_ind_ret());
+  return vixl::x8;
```
mxw: All supported platforms on HHVM have zero or one indirect return register, so we don't need an index argument for this.

hphp/runtime/vm/jit/abi.h
```diff
@@ -78,25 +78,27 @@ PhysReg rret_type();
...
 size_t num_arg_regs();
 size_t num_arg_regs_simd();
+size_t num_arg_regs_ind_ret();
 
 /*
  * RegSet for a call with `n' arguments.
  */
 RegSet arg_regs(size_t n);
 RegSet arg_regs_simd(size_t n);
+RegSet arg_regs_ind_ret(size_t n);
```
mxw: I don't see any uses of this function.

hphp/runtime/vm/jit/arg-group.h
```diff
@@ -80,7 +83,11 @@ struct ArgDesc {
     assertx(m_kind == Kind::TypeImm);
     return m_typeImm;
   }
-  Immed disp() const { assertx(m_kind == Kind::Addr); return m_disp32; }
+  Immed disp() const {
+    assertx((m_kind == Kind::Addr) ||
+	    (m_kind == Kind::IndRet));
```
mxw: nit: Nix the parens.

hphp/runtime/vm/jit/irlower-internal.cpp
```diff
@@ -46,6 +46,13 @@ namespace {
  */
 void prepareArg(const ArgDesc& arg, Vout& v, VregList& vargs) {
   switch (arg.kind()) {
+    case ArgDesc::Kind::IndRet: {
+      auto tmp = v.makeReg();
```
mxw: `auto const`

hphp/runtime/vm/jit/abi-arm.cpp
mxw: Okay, I read further down and I'm back.  Seems okay to me.

hphp/runtime/vm/jit/arg-group.h
```diff
@@ -188,6 +199,11 @@ struct ArgGroup {
     return *this;
   }
 
+  ArgGroup& indRet(Vreg base, Immed off) {
```
mxw: Can you document here that on platforms which don't have indirect return argument registers, this behaves just like `addr`, and pushes a positional GP argument?

hphp/runtime/vm/jit/abi.h
dave-estes: Huh, I don't either. I added it to alongside `arg_regs(size_t n)` and `arg_regs_simd(size_t n)` purely for balance since it's another `RegSet`. Looks like `arg_regs_simd(size_t n)` isn't used either.

What do you think I should do?

1. Remove `arg_regs_ind_ret(size_t n)`
2. Remove `arg_regs_ind_ret(size_t n)` and `arg_regs_simd(size_t n)`
3. Leave it as

Will leave as is with the next set of changes that I push for this PR, but will push another if you prefer 1 or 2.

hphp/runtime/vm/jit/arg-group.h
dave-estes: Done.

hphp/runtime/vm/jit/arg-group.h
dave-estes: Done.

hphp/runtime/vm/jit/irlower-internal.cpp
dave-estes: Done.

hphp/runtime/vm/jit/abi-arm.cpp
dave-estes: Cool.

hphp/runtime/vm/jit/abi-ppc64.cpp
```diff
@@ -172,13 +172,20 @@ PhysReg rarg_simd(size_t i) {
...
 
 size_t num_arg_regs() {
   return sizeof(gp_args) / sizeof(PhysReg);
 }
 size_t num_arg_regs_simd() {
   return sizeof(simd_args) / sizeof(PhysReg);
 }
+size_t num_arg_regs_ind_ret() {
+  return 0;
+}
```
mxw: Oh—can you just define this in the header?

hphp/runtime/vm/jit/abi-x64.cpp
```diff
@@ -173,20 +173,30 @@ PhysReg rarg_simd(size_t i) {
...
+}
 
 size_t num_arg_regs() {
   return sizeof(gp_args) / sizeof(PhysReg);
 }
 size_t num_arg_regs_simd() {
   return sizeof(simd_args) / sizeof(PhysReg);
 }
+size_t num_arg_regs_ind_ret() {
+  return 0;
```
mxw: Also move to the header.

hphp/runtime/vm/jit/abi.h
mxw: We can kill `arg_regs_simd()` too, but for this diff, just nix `arg_regs_ind_ret()`.

hphp/runtime/vm/jit/abi-ppc64.cpp
dave-estes: Sure thing.

hphp/runtime/vm/jit/abi-x64.cpp
dave-estes: Done.

hphp/runtime/vm/jit/abi.h
dave-estes: Wend ahead and removed `args_regs_simd()` along with `args_regs_ind_ret()`.

PR #7373, opened 2016-09-20
[AArch64] Correctly calculate stub_size().

dave-estes:
Additionally refactor extract_spoff() using a lighter-weight
instruction decoding approach in keeping with the x86_64 and PPC64
implementations.

### DIFF
--- a/hphp/runtime/vm/jit/service-requests.cpp
+++ b/hphp/runtime/vm/jit/service-requests.cpp
@@ -273,11 +273,12 @@ FPInvOffset extract_spoff(TCA stub) {
         always_assert((offBytes % sizeof(Cell)) == 0);
 
         if (instr->Mask(vixl::AddSubImmediateMask) == vixl::SUB_w_imm ||
-            instr->Mask(vixl::AddSubImmediateMask) == vixl::SUB_x_imm)
+            instr->Mask(vixl::AddSubImmediateMask) == vixl::SUB_x_imm) {
           return FPInvOffset{offBytes / int32_t{sizeof(Cell)}};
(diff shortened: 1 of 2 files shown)


hphp/runtime/vm/jit/service-requests.cpp
```diff
@@ -254,42 +266,35 @@ FPInvOffset extract_spoff(TCA stub) {
...
-      return FPInvOffset{-(*decoder.offset / int32_t{sizeof(Cell)})};
+      auto instr = reinterpret_cast<vixl::Instruction*>(stub);
+
+      if (instr->IsAddSubImmediate()) {
+        auto const offBytes = safe_cast<int32_t>(instr->ImmAddSub());
+        always_assert((offBytes % sizeof(Cell)) == 0);
+
+        if (instr->Mask(vixl::AddSubImmediateMask) == vixl::SUB_w_imm ||
+            instr->Mask(vixl::AddSubImmediateMask) == vixl::SUB_x_imm)
+          return FPInvOffset{offBytes / int32_t{sizeof(Cell)}};
```
mxw: Please no goto fails in HHVM.

aorenste: Just to be a little more clear - please surround the return statement with braces so it's not a single hanging statement.

dave-estes: Thanks for that. I was gonna inquire. :-) Will post an update shortly.

dave-estes: Done.

PR #7379, opened 2016-09-23
[AArch64] Fix cdecl warnings.

dave-estes:
Similar to PPC64, the cdecl attribute isn't defined for AArch64.


brunoalr:
@dave-estes On ppc64 we also had to update `hphp/runtime/vm/jit/vtune/jitprofiling.h` 
Was the update on `hphp/runtime/vm/jit/vtune/ittnotify_config.h` enough for AArch64?

dave-estes:
Thanks for noticing, @brunoalr. We didn't need to change jitprofiling.h due to the order of includes, but I'm going to go ahead and modify it too. Thanks again.

brunoalr:
@dave-estes no problem :)

PR #7386, opened 2016-09-26
[PPC64] Improve interfaces and cleanup

gut:
On Decoder, changed some PPC64 immediate instructions that are more
readable when displayed as hex format.

On vasm-ppc64.cpp, the common functionality of creating a frame for the
following call and setting the backchain was grouped into a function.


gut:
@mxw: Hi, I'm about to send the ppc64's relocation code and I chose to send first the improvements to existing interfaces on a separated PR. After this one is merged I'll send the whole thing, ok?

gut:
btw, is there a proper tag that one could add to PR's title in order to flag it as specific change for a specific backend (architecture)? I tried **[PPC64]** but somehow it got lost.
If you have any suggestions, please feedback.

PR #7388, opened 2016-09-27
Remove NO_OPT from decoder

racardoso:
Summary:

This commit removes __attribute__(optimize("O0")) of decoder and
remove early-inlining-insns compilation options for ppc64.

After some investigation and interlock with gcc developers we found
that's caused by early-inlining-insns due a gcc bug on ppc64el.
A bugzilla was openned on:

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=72855

Was pointed out we should be careful about those parameters as it
may have to be fine tunned for each architecture to avoid problems
like that.

Removing NO_OPT increase build time on others architectures but
after pull request https://github.com/facebook/hhvm/pull/7359
we reduce the number of instructions on decoder so we can remove
NO_OPT without increase build time.

### DIFF
--- a/hphp/ppc64-asm/decoder-ppc64.h
+++ b/hphp/ppc64-asm/decoder-ppc64.h
@@ -461,13 +461,7 @@ class Decoder : private boost::noncopyable {
     }
   }
 
-  /*
-   * Disable optimizations for this constructor.  In release mode -O3 causes
-   * compilation to hang due the huge initialization list.  This is a static
-   * singleton constructor, it's only called once and, when trace is enabled so
-   * optimization here is not a big issue.
-   */
-  NO_OPT Decoder();
+  Decoder();
 
   ~Decoder() {
     for(int i = 0; i < kDecoderSize; i++)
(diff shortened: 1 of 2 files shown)


gut:
Is this test failure a concern?

aorenste:
That test failure looks like an existing infra issue.

racardoso:
ping?

hphp/ppc64-asm/decoder-ppc64.h
```diff
@@ -461,13 +461,7 @@ class Decoder : private boost::noncopyable {
...
   }
 
-  /*
-   * Disable optimizations for this constructor.  In release mode -O3 causes
-   * compilation to hang due the huge initialization list.  This is a static
-   * singleton constructor, it's only called once and, when trace is enabled so
-   * optimization here is not a big issue.
-   */
-  NO_OPT Decoder();
+  Decoder();
```
Orvid: Drop the `NO_OPT` on the implementation as well and this should be good to go.

racardoso: @Orvid  there' s no `NO_OPT` on the implementation (.cpp): https://github.com/facebook/hhvm/blob/master/hphp/ppc64-asm/decoder-ppc64.cpp#L155

Orvid: O_o... I thought there was. Oh well :D

PR #7392, opened 2016-09-28
[AArch64] Rework push/pop to avoid 8-byte operation.

dave-estes:
In addition to requiring 16 byte sp alignment at ABI boundaries,
AArch64 also requires 16 byte alignment on loads/stores. Currently,
theres a workaround in vasm-arm.cpp to temporarily use rAsm for
8 bytes pushes/pops. This change removes that workaround and instead
adds support to specify a sp operation alignment per platform which
is used wherever offsets calculations are done for pushes/pops.

### DIFF
--- a/hphp/runtime/vm/jit/phys-reg-saver.cpp
+++ b/hphp/runtime/vm/jit/phys-reg-saver.cpp
@@ -50,7 +50,7 @@ PhysRegSaver::PhysRegSaver(Vout& v, RegSet regs)
       // Push r0 twice to avoid vasm-check failures
       v << pushp{r0, r0};
     } else {
-      v << pushp{r1, r0};
+      v << pushp{r0, r1};
     }
   });
 }
--- a/hphp/runtime/vm/jit/phys-reg-saver.cpp
+++ b/hphp/runtime/vm/jit/phys-reg-saver.cpp
@@ -46,7 +46,12 @@ PhysRegSaver::PhysRegSaver(Vout& v, RegSet regs)
   }
 
   gpr.forEachPair([&] (PhysReg r0, PhysReg r1) {
-    v << pushp{r1, r0};
+    // Push r0 twice to avoid vasm-check failures
+    if (r1 == InvalidReg) {
+      v << pushp{r0, r0};
+    } else {
+      v << pushp{r1, r0};
+    }
   });
 }
 
@@ -58,7 +63,12 @@ PhysRegSaver::~PhysRegSaver() {
   auto xmm = m_regs & abi().simd();
 
   gpr.forEachPairR([&] (PhysReg r0, PhysReg r1) {
-    v << popp{r0, r1};
+    // Pop second value into new virtual to avoid vasm-check failures
+    if (r1 == InvalidReg) {
+      v << popp{r0, v.makeReg()};
+    } else {
+      v << popp{r0, r1};
+    }
   });
 
   if (!xmm.empty()) {
--- a/hphp/runtime/vm/jit/phys-reg-saver.cpp
+++ b/hphp/runtime/vm/jit/phys-reg-saver.cpp
@@ -64,8 +64,8 @@ PhysRegSaver::~PhysRegSaver() {
 
   gpr.forEachPairR([&] (PhysReg r0, PhysReg r1) {
     // Pop second value into new virtual to avoid vasm-check failures
-    if (r1 == InvalidReg) {
-      v << popp{r0, v.makeReg()};
+    if (r0 == InvalidReg) {
+        v << popp{v.makeReg(), r1};
     } else {
       v << popp{r0, r1};
     }
--- a/hphp/runtime/vm/jit/phys-reg-saver.cpp
+++ b/hphp/runtime/vm/jit/phys-reg-saver.cpp
@@ -46,8 +46,8 @@ PhysRegSaver::PhysRegSaver(Vout& v, RegSet regs)
   }
 
   gpr.forEachPair([&] (PhysReg r0, PhysReg r1) {
-    // Push r0 twice to avoid vasm-check failures
     if (r1 == InvalidReg) {
+      // Push r0 twice to avoid vasm-check failures
       v << pushp{r0, r0};
     } else {
       v << pushp{r1, r0};
@@ -63,9 +63,9 @@ PhysRegSaver::~PhysRegSaver() {
   auto xmm = m_regs & abi().simd();
 
   gpr.forEachPairR([&] (PhysReg r0, PhysReg r1) {
-    // Pop second value into new virtual to avoid vasm-check failures
     if (r0 == InvalidReg) {
-        v << popp{v.makeReg(), r1};
+      // Pop second value into new virtual to avoid vasm-check failures
+      v << popp{v.makeReg(), r1};
     } else {
       v << popp{r0, r1};
     }
--- a/hphp/runtime/vm/jit/phys-reg-saver.cpp
+++ b/hphp/runtime/vm/jit/phys-reg-saver.cpp
@@ -35,8 +35,6 @@ PhysRegSaver::PhysRegSaver(Vout& v, RegSet regs)
   auto xmm = m_regs & abi().simd();
   auto const sp = rsp();
 
-  m_adjust = gpr.size() & 0x1 ? 8 : 0;
-
   if (!xmm.empty()) {
     v << lea{sp[-16 * xmm.size()], sp};
 
@@ -47,57 +45,21 @@ PhysRegSaver::PhysRegSaver(Vout& v, RegSet regs)
     });
   }
 
-  switch (arch()) {
-    case Arch::X64:
-    case Arch::PPC64:
-      gpr.forEach([&] (PhysReg r) {
-        v << push{r};
-      });
-      break;
-    case Arch::ARM:
-      gpr.forEachPair([&] (PhysReg r0, PhysReg r1) {
-        if (r1 == InvalidReg) {
-          v << push{r0};
-        } else {
-          v << pushp{r1, r0};
-        }
-      });
-      break;
-  }
-
-  if (m_adjust) {
-    v << lea{sp[-m_adjust], sp};
-  }
+  gpr.forEachPair([&] (PhysReg r0, PhysReg r1) {
+    v << pushp{r1, r0};
+  });
 }
 
 PhysRegSaver::~PhysRegSaver() {
   auto& v = m_v;
   auto const sp = rsp();
 
-  if (m_adjust) {
-    v << lea{sp[m_adjust], sp};
-  }
-
   auto gpr = m_regs & abi().gp();
   auto xmm = m_regs & abi().simd();
 
-  switch (arch()) {
-    case Arch::X64:
-    case Arch::PPC64:
-      gpr.forEachR([&] (PhysReg r) {
-        v << pop{r};
-      });
-      break;
-    case Arch::ARM:
-      gpr.forEachPairR([&] (PhysReg r0, PhysReg r1) {
-        if (r1 == InvalidReg) {
-          v << pop{r0};
-        } else {
-          v << popp{r0, r1};
-        }
-      });
-      break;
-  }
+  gpr.forEachPairR([&] (PhysReg r0, PhysReg r1) {
+    v << popp{r0, r1};
+  });
 
   if (!xmm.empty()) {
     int offset = 0;
@@ -109,13 +71,9 @@ PhysRegSaver::~PhysRegSaver() {
   }
 }
 
-size_t PhysRegSaver::rspAdjustment() const {
-  return m_adjust;
-}
-
 size_t PhysRegSaver::dwordsPushed() const {
-  assertx((m_adjust % sizeof(int64_t)) == 0);
-  return m_regs.size() + m_adjust / sizeof(int64_t);
+  // Round up the number of regs pushed to an even number
+  return (m_regs.size() + 1) ^ 0x1;
 }
 
 ///////////////////////////////////////////////////////////////////////////////
--- a/hphp/runtime/vm/jit/vasm-arm.cpp
+++ b/hphp/runtime/vm/jit/vasm-arm.cpp
@@ -499,7 +499,7 @@ void Vgen::emit(const calls& i) {
 
 void Vgen::emit(const stublogue& i) {
   // Push FP, LR always regardless of i.saveframe (makes SP 16B aligned)
-  emit(pushp{rfp(), rlr()});
+  emit(pushp{rlr(), rfp()});
 }
 
 void Vgen::emit(const stubret& i) {
@@ -1021,7 +1021,7 @@ void Vgen::emit(const popp& i) {
 }
 
 void Vgen::emit(const pushp& i) {
-  a->Stp(X(i.s0), X(i.s1), MemOperand(sp, -16, PreIndex));
+  a->Stp(X(i.s1), X(i.s0), MemOperand(sp, -16, PreIndex));
 }
 
 ///////////////////////////////////////////////////////////////////////////////
@@ -1349,15 +1349,15 @@ void lower(Vunit& u, calltc& i, Vlabel b, size_t z) {
     // Push FP, LR for callToExit(..)
     auto r0 = v.makeReg();
     auto r1 = v.makeReg();
-    v << load{i.fp[AROFF(m_savedRip)], r0};
-    v << ldimmq{i.exittc, r1};
+    v << ldimmq{i.exittc, r0};
+    v << load{i.fp[AROFF(m_savedRip)], r1};
     v << pushp{r0, r1};
 
     // Emit call to next instruction to balance predictor's stack
     v << bln{};
 
     // Set the return address to savedRip and jump to target
-    v << copy{r0, rlr()};
+    v << copy{r1, rlr()};
     v << jmpr{i.target, i.args};
   });
 }
@@ -1367,7 +1367,7 @@ void lower(Vunit& u, resumetc& i, Vlabel b, size_t z) {
     // Push FP, LR for callToExit(..)
     auto r = v.makeReg();
     v << ldimmq{i.exittc, r};
-    v << pushp{rfp(), r};
+    v << pushp{r, rfp()};
 
     // Call the helper
     v << callr{i.target, i.args};
--- a/hphp/runtime/vm/jit/vasm-arm.cpp
+++ b/hphp/runtime/vm/jit/vasm-arm.cpp
@@ -267,9 +267,11 @@ struct Vgen {
   void emit(const orq& i);
   void emit(const orqi& i);
   void emit(const pop& i);
+  void emit(const popp& i);
   void emit(const psllq& i);
   void emit(const psrlq& i);
   void emit(const push& i);
+  void emit(const pushp& i);
   void emit(const roundsd& i);
   void emit(const sar& i);
   void emit(const setcc& i) { a->Cset(W(i.d), C(i.cc)); }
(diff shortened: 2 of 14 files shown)


mxw:
Where are we emitting raw `push{}` or `pop{}` on ARM today?  I think a better solution here is to just use `pushp{}` and `popp{}` everywhere.  I think using them everywhere cross-architecture is fine, also, since I can't think of many places where we specifically need to push and pop single qwords.

dave-estes:
I'm down for using `pushp`/`popp` **everywhere**.

There are the `pushm`'s and `pop`'s in emitFunctionEnterHelper(). You thinking I should add a `pushpm` and `poppm`?

Other than that the only `push`'s and `pop`'s are in `lower_vcall()` and `PhysRegSaver` for when we have an odd number of registers/arguments. For ARM, we can use our special zero register `xzr` in our instructions that implement `pushp`/`popp` to handle the case of the final odd register/argument. I figure we can represent that in vasm by using `InvalidReg` as the second source/destination operand? So for ARM, we'd use `xzr` for any operand with InvalidReg. And then in X64 and PPC64 we'd lower such a `pushp`/`popp` to a single `push`/`pop` and an `lea`.

Sound reasonable?

mxw:
@dave-estes—Yep, that all sounds good to me.  It should be fairly straightforward, but if for some reason it isn't, feel free to pass it back to me.  There might be some validation passes that aren't happy about an InvalidReg operand, in which case we can think of something reasonable to do in that case...

Make sure to move `push{,m}p` and `pop{,m}p` from the ARM-specific section of vasm-instr.h to live near the other pushes and pops, and also document the argument order—right now it's a mystery, which I let slide since it wasn't used by any other backends.

dave-estes:
Thanks for the feedback, @mxw. It's actually going pretty smoothly, and it's cleaning up things nicely. I'll probably push a new patch with just the ARM piece working in order to get your feedback. Then I'll modify X64 and PPC64.

hphp/runtime/vm/jit/phys-reg-saver.cpp
```diff
@@ -59,15 +51,13 @@ PhysRegSaver::PhysRegSaver(Vout& v, RegSet regs)
...
+      m_adjust = gpr.size() & 0x1 ? 8 : 0;
       break;
     case Arch::ARM:
       gpr.forEachPair([&] (PhysReg r0, PhysReg r1) {
-        if (r1 == InvalidReg) {
-          v << push{r0};
-        } else {
-          v << pushp{r1, r0};
-        }
+        v << pushp{r1, r0};
```
dave-estes: `forEachPair()` nicely returns `InvalidReg` for the second register if it's the final of an odd number of registers. `pushp`/`popp` are designed to recognize InvalidReg and to simply push `$xzr` for ARM and to simply `$sp +/- 8` for the other platforms. Note: I haven't implemented `pushp`/`popp` for X64 and PPC64 yet.

hphp/runtime/vm/jit/phys-reg-saver.cpp
```diff
@@ -96,11 +86,7 @@ PhysRegSaver::~PhysRegSaver() {
       break;
     case Arch::ARM:
       gpr.forEachPairR([&] (PhysReg r0, PhysReg r1) {
-        if (r1 == InvalidReg) {
-          v << pop{r0};
-        } else {
-          v << popp{r0, r1};
-        }
+        v << popp{r0, r1};
```
dave-estes: Similar to above, even `forEachPairR()` will use an `InvalidReg`.

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -1013,11 +1015,15 @@ Y(lslxis, X, xzr, 64)
 #undef Y
 
 void Vgen::emit(const popp& i) {
-  a->Ldp(X(i.d0), X(i.d1), MemOperand(sp, 16, PostIndex));
+  Register r0 = i.d0.isValid() ? X(i.d0) : vixl::xzr;
+  Register r1 = i.d1.isValid() ? X(i.d1) : vixl::xzr;
+  a->Ldp(r0, r1, MemOperand(sp, 16, PostIndex));
```
dave-estes: Here's the implementation of `popp` for ARM. We still use a `ldp` instruction, but we'll pass the special zero register (`$xzf`) for `InvalidReg`.

For X64 and PPC64, I planned to lower `popp` to a `pop` and `lea` of `$sp` when `InvalidReg` is used.

hphp/runtime/vm/jit/vasm-check.cpp
```diff
@@ -60,14 +60,16 @@ checkSSA(const Vunit& unit, const jit::vector<Vlabel>& blocks) {
     }
     for (auto const& inst : unit.blocks[b].code) {
       visitUses(unit, inst, [&](Vreg v) {
-        assert_flog(v.isValid(), "invalid vreg used in B{}\n{}",
+        assert_flog(v.isValid() || inst.op == Vinstr::pushp,
+		    "invalid vreg used in B{}\n{}",
```
dave-estes: Here and below are two places where I had to adjust the SSA checks to allow `InvalidReg` for uses and defs for `pushp` and `popp`.

I don't really like this, but I tried it out. Personally, I prefer to just push and pop the same registers, but I imagine that check SSA could fail too if we're def'ing the same register twice in one instruction.

hphp/runtime/vm/jit/vasm-lower.cpp
```diff
@@ -75,7 +75,9 @@ void lower_vcall(Vunit& unit, Inst& inst, Vlabel b, size_t i) {
       // there are an odd number of arguments.
       int numArgs = vargs.stkArgs.size();
       if (adjust) {
-        v << push{vargs.stkArgs[numArgs-1]};
+	// Using InvalidReg below fails SSA checks and simplify pass.
+        //v << pushp{vargs.stkArgs[numArgs-1], InvalidReg};
+        v << pushp{vargs.stkArgs[numArgs-1], vargs.stkArgs[numArgs-1]};
```
dave-estes: This is where things really started to tank. The commented line is exactly what I wanted to do when pushing the stack arguments. However, this happens early enough that check SSA started to complain of the `InvalidReg`. So for this one place, I went ahead and just pushed that odd stack argument twice.

Depending on what you think, this could easily be a simple policy for every time we have to push an odd number of things. The only down side, is that a pushp{a, a} will lower (X64 and PPC64) to push{a} and push{a} which I imagine is slower than a `push` and `lea` which we could easily do if we supported `InvalidReg`.

hphp/runtime/vm/jit/vasm-simplify.cpp
```diff
@@ -438,6 +438,7 @@ void simplify(Vunit& unit) {
 
     for (auto const& inst : blocks[b].code) {
       visitDefs(unit, inst, [&] (Vreg r) { env.def_insts[r] = inst.op; });
+      // Following line segfaults
```
dave-estes: After I doctored up the check SSA code, this started to fail. I threw my hands up and just switched the code in vasm-lower.cpp to pass the odd stack argument twice.

mxw:
The `popp{}` case is much easier to handle—you can just do `popp{r, v.makeReg()}` (or maybe the operands should be switched?  I'm still not seeing it documented anywhere—please add some!).  Then, you can add a rule in `vasm-simplify.cpp` which transforms `pop{}` into `lea{}` if its def'd Vreg has zero uses.

Handling `pushp{}` is more annoying, because we don't do any memory alias analysis for vasm.  I would recommend just adding two opcodes, `pushp1{}` and `pushp2{}` (naming suggestions welcome) for the one and two argument case.  To be honest, I don't really like this solution, but I also don't really have a better idea.

edwinsmith:
for the 1-arg pushp case, what about just pushing the same register twice? it will always be a second store on the same cache line. at least on x86, lowering to [push(r), push(r)] is close to ideal because convering one to an lea{} is a) bigger code, and b) more expensive to update rsp directly adjacently to push/pop instructions, which do it implicilty, because of having to synchronize the architectural state of rsp.

hphp/runtime/vm/jit/vasm-lower.cpp
```diff
@@ -57,12 +57,20 @@ void lower_vcall(Vunit& unit, Inst& inst, Vlabel b, size_t i) {
...
+  // Push stack arguments, in reverse order. Push in pairs without padding
+  // except for the last argument (pushed first) which should be padded if
+  // there are an odd number of arguments.
+  int numArgs = vargs.stkArgs.size();
+  int32_t const adjust = (numArgs & 0x1) ? sizeof(uintptr_t) : 0;
+  if (adjust) {
+    // Using InvalidReg below fails SSA checks and simplify pass, so just
+    // push the arg twice. It's on the same cacheline and will actually
+    // perform faster than an explicit lea.
+    v << pushp{vargs.stkArgs[numArgs-1], vargs.stkArgs[numArgs-1]};
```
dave-estes: I opted to just push the same argument twice here to avoid the vasm-check problems. Still sound reasonable to you, @edwinsmith?

dave-estes: Also, I'm a little unsure about the `rspOffset` code below (lines 108-121). I've never touched this code for AArch64, and it is working fine. Any thoughts on whether or not this is still needed?

gut:
@racardoso : Hi Rogério. Could you take a look if PPC64 has a problem with these changes and feedback here, please? Thanks.

dave-estes:
Looks like I have a problem with X64, @racardoso, @gut. It might affect PPC64 too, so I'll get it resolved and then I'll ping you again.

racardoso:
@gut yep. I'm looking it here. Quick test suite pass all tests (1101) on Release and Debug mode. I am checking all tests and oss-performance to see if there any regressions. @dave-estes ok let's us here about it.

dave-estes:
The previous patch was failing for x64, because phy-reg-saver was saving an odd number of registers when emitting code for decRefGeneric. This meant that I needed to employ the techniques that we've discussed to push the same reg twice for `pushp` and to pop to an unused virtual for `popp`. I then implemented the optimization from @mxw in vasm-simplify to change any `pop` to an unused register to an `lea`.

Additionally, I worked in a change in vasm-simplify for `popp`. The rationale for the change is that I noticed something sticky while debugging the x86 failure. x86, lowered the `popp` to two `pop`s. These were for phys-reg-saver, and so most of the registers were already assigned to physical registers, except for the last pair. The first of that pair was rax, and the second was a virtual. I then saw that the allocator assigned the final virtual to rax too. DOH. I'm not sure if this is due to weak alias analysis or if it's because the uses of the restored registers aren't known.

Anyhow this is not a problem for x86 and ppc now that I've implemented the optimization in vasm-simplify for `pop`. I was worried about this register allocator scenario for arm, so I've added this simplifier for `popp` that sets the unused register InvalidReg. I'm not sure if future ordering of passes will trigger vasm-check after this simply, but I'm happy to debug that instead of debugging a stickier issue due to the reallocating of a register for that second `popp` operand.

OK...what does everyone think? :-)

mxw:
@dave-estes—I don't agree with your justification for having InvalidReg arguments.  If the uses of `%rax` in the case you saw are unknown, that's a bug, and should be fixed.  Hiding it with an InvalidReg operand (and requiring passes, which should be order-independent, to "know" about whether it might see InvalidReg) seems harmful to me.

Was the double-pop to `%rax` actually causing incorrect behavior in the program?  If so, where was it happening?

mxw:
Also, test failures are real.  Is this passing all x64 tests in your builds?

dave-estes:
Concerning the two pops to `%rax`, it didn't cause a problem. There were an odd number of registers to save, and so phys-reg-saver pushed the final odd register twice. These two final pops were of the same value, so it worked out. Below is the vasm emitted from `emitDecRefGeneric()`. **This is before my changes to vasm-simplify to change `pop` to unused register to a `lea`**.

    --------------------------------after vasm-jumps--------------------------------
    
     B0      main
          stublogue  
          cmplim     1, [%rdi + 0xc] => %128
          jcc        E, %128, B3, else B2
    
     B3      main, B0
          pushp      %rcx, %rax
          pushp      %rsi, %rdx
          pushp      %r8, %rdi
          pushp      %r10, %r9
          pushp      %r11, %r11
          movzbl     %rsi => %130
          shrli      2, %130 => %132, %133
          movzlq     %132 => %131
          callm      [0x9ca0d40 + %131 * 8], {%rdi}
          syncpoint  pc:-1 sp:112
          popp       => %r11, %r10
          popp       => %r9, %r8
          popp       => %rdi, %rsi
          popp       => %rdx, %rcx
          popp       => %rax, %134
          jmp        B4
    
     B2      main, B0
          jcc        GE, %128, B5, else B4
    
     B5      main, B2
          declm      [%rdi + 0xc] => %129
          jcc        L, %129, B7, else B4
    
     B7      main, B5
          ud2        
    
     B4      main, B5, B2, B3
          stubret    {}
    
    3 unreachable blocks not shown. Set TRACE=vasm:6 or greater to print them.
    
    --------------------------------------------------------------------------------

Note that the final `popp` has args `%rax` and `%134`. After lowering for X64, the sequence of `popp` turns to `pop`.

      pop        => %r11
      pop        => %r10
      pop        => %r9
      pop        => %r8
      pop        => %rdi
      pop        => %rsi
      pop        => %rdx
      pop        => %rcx
      pop        => %rax
      pop        => %134

And then after register allocation, the final `pop` is assigned `%rax`.

      pop        => %r11
      pop        => %r10
      pop        => %r9
      pop        => %r8
      pop        => %rdi
      pop        => %rsi
      pop        => %rdx
      pop        => %rcx
      pop        => %rax
      pop        => %rax

I *think* this worked out by pure luck and that `%rax` was just the first unallocated register. Regardless, adding the simplifier for `pop` to `lea` side-stepped the issue altogether for X64.

The second simplifier for `popp` was a bit of a cheat. I put the check for an unused register in `popp` here because vasm-simplify already counted each registers uses. And the ARM emitter for `popp` already has the smarts to handle `InvalidReg` arguments.

I totally get your point. How does the following sound?
1. I remove the simplifier for `popp`.
2. I pull the handling of `InvalidReg` arguments from the `pushp` and `popp` emitters.

The downside of the above is that now ARM will be popping an extra value to an allocated register that is not used. I'm scratching my head to figure out how to avoid allocating that register, but I'm stumped. Using `InvaldReg` did accomplish this, but it's messy.

dave-estes:
And...will let you know about the failures ASAP.

hphp/runtime/vm/jit/phys-reg-saver.cpp
```diff
@@ -47,57 +45,31 @@ PhysRegSaver::PhysRegSaver(Vout& v, RegSet regs)
...
-        }
-      });
-      break;
-  }
-
-  if (m_adjust) {
-    v << lea{sp[-m_adjust], sp};
-  }
+  gpr.forEachPair([&] (PhysReg r0, PhysReg r1) {
+    // Push r0 twice to avoid vasm-check failures
```
mxw: Move this comment down below the `if (...) {`.

hphp/runtime/vm/jit/phys-reg-saver.cpp
```diff
@@ -47,57 +45,31 @@ PhysRegSaver::PhysRegSaver(Vout& v, RegSet regs)
...
-        if (r1 == InvalidReg) {
-          v << pop{r0};
-        } else {
-          v << popp{r0, r1};
-        }
-      });
-      break;
-  }
+  gpr.forEachPairR([&] (PhysReg r0, PhysReg r1) {
+    // Pop second value into new virtual to avoid vasm-check failures
```
mxw: Move this comment down below the `if (...) {`.

hphp/runtime/vm/jit/vasm-simplify.cpp
```diff
@@ -393,6 +393,23 @@ bool simplify(Env& env, const movzlq& inst, Vlabel b, size_t i) {
   });
 }
 
+bool simplify(Env& env, const popp& inst, Vlabel b, size_t i) {
+  if (!env.use_counts[inst.d0]) { decltype(inst.d0){Vreg::kInvalidReg}; }
+  if (!env.use_counts[inst.d1]) { decltype(inst.d1){Vreg::kInvalidReg}; }
```
mxw: What's the purpose of these two lines?  It looks to me like you're just constructing temporary Vr's on the stack which get immediately destructed... have you actually verified that this has any effect?

hphp/runtime/vm/jit/vasm-simplify.cpp
```diff
@@ -393,6 +393,23 @@ bool simplify(Env& env, const movzlq& inst, Vlabel b, size_t i) {
...
+  return false;
+}
+
+bool simplify(Env& env, const pop& inst, Vlabel b, size_t i) {
+  if (env.use_counts[inst.d]) {
+    return false;
+  } else {
```
mxw: This `else` only serves to add indentation; get rid of it.

dave-estes:
Fixed the X64 failures. The reverse iterator for pairs in phys-reg.h didn't return pairs in precisely the reverse order when there was an odd number of registers. Also the argument ordering was wrong in vasm-lower.cpp.

mxw:
@dave-estes—Sorry for the delay; our perf lockdown is winding down.

This looks great!  Could I request that you flip the arguments of `pushp*{}`, though?

Basically, I think there are two sensible ways of looking at the `pushp*{}` and `popp*{}` instructions:

1.  They're pushes and pops of two registers, in which case I think we should have `pushp{s0, s1} -> push{s0}; push{s1}` and `popp{d0, d1} -> pop{d0}; pop{d1}`.  This means that if we want to push a pair of registers and then pop them later, like we do in unique stubs, we do something like `pushp{rone, rtwo}` and then `popp{rtwo_new, rone_new}`.  I'm pretty comfortable with this antisymmetry.
2.  They're pushes and pops of a single octa-word register, i.e., `pushp{lo, hi} -> push{hi}; push{lo}` and `popp{lo, hi} -> pop{lo}; push{hi}`.  This makes pushing and popping a pair of registers symmetric in the argument order.

Both of these are valid, but I think that we rarely actually think about pushing and popping a pair of registers along the second of those two mindsets, at least not in HHVM code (though I suspect that when writing straight-up ARM assembly, this may be the prevailing mental model).  I think the way you've chosen to document the `pushp*{}` and `popp*{}` instructions, though—i.e., according to how they lower—also supports this, that we mostly think of `pushp{}` as "push one, then push the other", in which case I think `pushp{first, second} -> push{first}; push{second}` is the more intuitive operand order.

This is a little philosophical, but I do sort of do a double-take whenever I read a `pushp{}`, because I'm inclined to think of it as "push these two registers".

Let me know what you think, and whether you're willing to flip the order in this diff.  If you're amenable but working on something big elsewhere, though, I'm perfectly happy to take care of it here, so let me know.

dave-estes:
No problem, I can do this.

hphp/runtime/vm/jit/phys-reg-saver.cpp
```diff
@@ -109,13 +81,9 @@ PhysRegSaver::~PhysRegSaver() {
...
 
-size_t PhysRegSaver::rspAdjustment() const {
-  return m_adjust;
-}
-
 size_t PhysRegSaver::dwordsPushed() const {
-  assertx((m_adjust % sizeof(int64_t)) == 0);
-  return m_regs.size() + m_adjust / sizeof(int64_t);
+  // Round up the number of regs pushed to an even number
+  return (m_regs.size() + 1) ^ 0x1;
```
mxw: @dave-estes—This is my fault for not catching this, but this expression is total nonsense:

```
(2 + 1) ^ 1 == 2
(3 + 1) ^ 1 == 5
(4 + 1) ^ 1 == 4
```

and also it doesn't take into account GPR regs vs. XMM regs.  I'll have a fix upstreamed soon.

mxw: (I think your intent was probably `(m_regs.size() + 1) & ~0x1`, so I just sort of read it as that... but again, even that's not quite right because of the GP/SIMD size difference.)

dave-estes: Sure was. Sorry about that, @mxw, and thanks for upstreaming the fix.

PR #7393, opened 2016-09-28
[AArch64] Preserve fast call abi rules for dec ref.

dave-estes:
Previously the dec ref stub made use of cmplim and declm, which when
lowered on AArch64 would use virtuals. This problem was originally
addressed with special vasm variations which when lowered/emitted for
AArch64 would avoid the use of new virtuals.

The new solution introduces a flag, `fast`, in the `Vunit` which indicates to
any relevant lowerers that the platform's rAsm should be used instead of
new virtuals.

### DIFF
--- a/hphp/runtime/vm/jit/phys-reg-saver.cpp
+++ b/hphp/runtime/vm/jit/phys-reg-saver.cpp
@@ -48,6 +48,8 @@ PhysRegSaver::PhysRegSaver(Vout& v, RegSet regs)
   gpr.forEachPair([&] (PhysReg r0, PhysReg r1) {
     v << pushp{r0, r1 == InvalidReg ? r0 : r1};
   });
+
+  v << vregunrestrict{};
 }
 
 PhysRegSaver::~PhysRegSaver() {
@@ -56,6 +58,8 @@ PhysRegSaver::~PhysRegSaver() {
 
   auto gpr = m_regs & abi().gp();
   auto xmm = m_regs & abi().simd();
+
+  v << vregrestrict{};
 
   gpr.forEachPairR([&] (PhysReg r0, PhysReg r1) {
     v << popp{r0 == InvalidReg ? v.makeReg() : Vreg{r0}, r1};
--- a/hphp/runtime/vm/jit/unique-stubs.cpp
+++ b/hphp/runtime/vm/jit/unique-stubs.cpp
@@ -1022,6 +1022,7 @@ TCA emitDecRefGeneric(CodeBlock& cb, DataBlock& data) {
   CGMeta meta;
 
   auto const start = vwrap(cb, data, meta, [] (Vout& v) {
+    v << vregrestrict{};
     v << stublogue{};
 
     auto const rdata = rarg(0);
--- a/hphp/runtime/vm/jit/vasm-arm.cpp
+++ b/hphp/runtime/vm/jit/vasm-arm.cpp
@@ -337,7 +337,6 @@ struct Vgen {
   void emit(const xorqi& i);
 
   // arm intrinsics
-  void emit(const cmplims& i);
   void emit(const fcvtzs& i) { a->Fcvtzs(X(i.d), D(i.s)); }
   void emit(const mrs& i) { a->Mrs(X(i.r), vixl::SystemRegister(i.s.l())); }
   void emit(const msr& i) { a->Msr(vixl::SystemRegister(i.s.l()), X(i.r)); }
@@ -846,42 +845,6 @@ void Vgen::emit(const unpcklpd& i) {
 
 ///////////////////////////////////////////////////////////////////////////////
 
-/*
- * 'cmplims' instruction is intended for use where temporary Vregs can't be
- * allocated, E.g., in stubs called via callfaststub{}. As the memory
- * operand is not lowered, make sure that it is in one of the following
- * formats
- *  [base]
- *  [base, index]
- *  [base, #imm] where #imm is in the range [-256, 255]
- */
-void Vgen::emit(const cmplims& i) {
-  enum {
-    BASE = 1,
-    INDEX = 2,
-    DISP = 4
-  };
-
-  // TODO: Allow TLS addresses if possible
-  assertx(i.s1.seg == Vptr::DS);
-
-  uint8_t mode = (((i.s1.base.isValid()  & 0x1) << 0) |
-                  ((i.s1.index.isValid() & 0x1) << 1) |
-                  (((i.s1.disp != 0)     & 0x1) << 2));
-  switch (mode) {
-    case BASE:
-    case BASE | INDEX:
-      break;
-    case BASE | DISP:
-      assertx(i.s1.disp >= -256 && i.s1.disp <= 255);
-      break;
-    default:
-      always_assert(false);
-  }
-  a->Ldr(rAsm.W(), M(i.s1));
-  a->Cmp(rAsm.W(), i.s0.l());
-}
-
 void Vgen::emit(const cmpsd& i) {
   /*
    * cmpsd doesn't update SD, so read the flags into a temp.
@@ -1103,7 +1066,7 @@ void lower_impl(Vunit& unit, Vlabel b, size_t i, Lower lower) {
 }
 
 template<typename Inst>
-void lower(Vunit& unit, Inst& inst, Vlabel b, size_t i) {}
+void lower(const VLS& env, Inst& inst, Vlabel b, size_t i) {}
 
 ///////////////////////////////////////////////////////////////////////////////
 
@@ -1203,12 +1166,12 @@ void lowerVptr(Vptr& p, Vout& v) {
   }
 }
 
-#define Y(vasm_opc, m)                                  \
-void lower(Vunit& u, vasm_opc& i, Vlabel b, size_t z) { \
-  lower_impl(u, b, z, [&] (Vout& v) {                   \
-    lowerVptr(i.m, v);                                  \
-    v << i;                                             \
-  });                                                   \
+#define Y(vasm_opc, m)                                      \
+void lower(const VLS& e, vasm_opc& i, Vlabel b, size_t z) { \
+  lower_impl(e.unit, b, z, [&] (Vout& v) {                  \
+    lowerVptr(i.m, v);                                      \
+    v << i;                                                 \
+  });                                                       \
 }
 
 Y(decqmlock, m)
@@ -1236,18 +1199,18 @@ Y(storew, m)
 #define ISLG(w) vixl::Assembler::IsImmLogical(value, w)
 #define U2(v0, v1) v0, v1
 
-#define Y(vasm_opc, lower_opc, load_opc, chk, imm, use) \
-void lower(Vunit& u, vasm_opc& i, Vlabel b, size_t z) { \
-  lower_impl(u, b, z, [&] (Vout& v) {                   \
-    auto value = safe_cast<int64_t>(i.s0.q());          \
-    if (chk) {                                          \
-      v << i;                                           \
-    } else {                                            \
-      auto s0 = v.makeReg();                            \
-      v << load_opc{imm, s0};                           \
-      v << lower_opc{s0, use, i.sf};                    \
-    }                                                   \
-  });                                                   \
+#define Y(vasm_opc, lower_opc, load_opc, chk, imm, use)     \
+void lower(const VLS& e, vasm_opc& i, Vlabel b, size_t z) { \
+  lower_impl(e.unit, b, z, [&] (Vout& v) {                  \
+    auto value = safe_cast<int64_t>(i.s0.q());              \
+    if (chk) {                                              \
+      v << i;                                               \
+    } else {                                                \
+      auto s0 = v.makeReg();                                \
+      v << load_opc{imm, s0};                               \
+      v << lower_opc{s0, use, i.sf};                        \
+    }                                                       \
+  });                                                       \
 }
 
 Y(addli, addl, ldimml, ISAR, i.s0, U2(i.s1, i.d))
@@ -1272,15 +1235,15 @@ Y(xorqi, xorq, ldimmq, ISLG(64), Immed64(value), U2(i.s1, i.d))
 #undef ISLG
 #undef ISAR
 
-#define Y(vasm_opc, lower_opc, load_opc, store_opc, s0, m) \
-void lower(Vunit& u, vasm_opc& i, Vlabel b, size_t z) {    \
-  lower_impl(u, b, z, [&] (Vout& v) {                      \
-    lowerVptr(i.m, v);                                     \
-    auto r0 = v.makeReg(), r1 = v.makeReg();               \
-    v << load_opc{i.m, r0};                                \
-    v << lower_opc{i.s0, r0, r1, i.sf};                    \
-    v << store_opc{r1, i.m};                               \
-  });                                                      \
+#define Y(vasm_opc, lower_opc, load_opc, store_opc, s0, m)  \
+void lower(const VLS& e, vasm_opc& i, Vlabel b, size_t z) { \
+  lower_impl(e.unit, b, z, [&] (Vout& v) {                  \
+    lowerVptr(i.m, v);                                      \
+    auto r0 = v.makeReg(), r1 = v.makeReg();                \
+    v << load_opc{i.m, r0};                                 \
+    v << lower_opc{i.s0, r0, r1, i.sf};                     \
+    v << store_opc{r1, i.m};                                \
+  });                                                       \
 }
 
 Y(addlim, addli, loadl, storel, s0, m)
@@ -1293,14 +1256,14 @@ Y(orwim, orqi, loadw, storew, s0, m)
 
 #undef Y
 
-#define Y(vasm_opc, lower_opc, load_opc)                \
-void lower(Vunit& u, vasm_opc& i, Vlabel b, size_t z) { \
-  lower_impl(u, b, z, [&] (Vout& v) {                   \
-    lowerVptr(i.s1, v);                                 \
-    auto r = v.makeReg();                               \
-    v << load_opc{i.s1, r};                             \
-    v << lower_opc{i.s0, r, i.sf};                      \
-  });                                                   \
+#define Y(vasm_opc, lower_opc, load_opc)                         \
+void lower(const VLS& e, vasm_opc& i, Vlabel b, size_t z) {      \
+  lower_impl(e.unit, b, z, [&] (Vout& v) {                       \
+    lowerVptr(i.s1, v);                                          \
+    auto r = e.allow_vreg() ? v.makeReg() : Vreg(PhysReg(rAsm)); \
+    v << load_opc{i.s1, r};                                      \
+    v << lower_opc{i.s0, r, i.sf};                               \
+  });                                                            \
 }
 
 Y(cmpbim, cmpli, loadb)
@@ -1317,8 +1280,8 @@ Y(testwim, testli, loadw)
 
 #undef Y
 
-void lower(Vunit& u, cvtsi2sdm& i, Vlabel b, size_t z) {
-  lower_impl(u, b, z, [&] (Vout& v) {
+void lower(const VLS& e, cvtsi2sdm& i, Vlabel b, size_t z) {
+  lower_impl(e.unit, b, z, [&] (Vout& v) {
     lowerVptr(i.s, v);
     auto r = v.makeReg();
     v << load{i.s, r};
@@ -1326,15 +1289,16 @@ void lower(Vunit& u, cvtsi2sdm& i, Vlabel b, size_t z) {
   });
 }
 
-#define Y(vasm_opc, lower_opc, load_opc, store_opc, m)  \
-void lower(Vunit& u, vasm_opc& i, Vlabel b, size_t z) { \
-  lower_impl(u, b, z, [&] (Vout& v) {                   \
-    lowerVptr(i.m, v);                                  \
-    auto r0 = v.makeReg(), r1 = v.makeReg();            \
-    v << load_opc{i.m, r0};                             \
-    v << lower_opc{r0, r1, i.sf};                       \
-    v << store_opc{r1, i.m};                            \
-  });                                                   \
+#define Y(vasm_opc, lower_opc, load_opc, store_opc, m)            \
+void lower(const VLS& e, vasm_opc& i, Vlabel b, size_t z) {       \
+  lower_impl(e.unit, b, z, [&] (Vout& v) {                        \
+    lowerVptr(i.m, v);                                            \
+    auto r0 = e.allow_vreg() ? v.makeReg() : Vreg(PhysReg(rAsm)); \
+    auto r1 = e.allow_vreg() ? v.makeReg() : Vreg(PhysReg(rAsm)); \
+    v << load_opc{i.m, r0};                                       \
+    v << lower_opc{r0, r1, i.sf};                                 \
+    v << store_opc{r1, i.m};                                      \
+  });                                                             \
 }
 
 Y(declm, decl, loadl, storel, m)
@@ -1345,8 +1309,8 @@ Y(incwm, incw, loadw, storew, m)
 
 #undef Y
 
-void lower(Vunit& unit, cvttsd2siq& i, Vlabel b, size_t idx) {
-  lower_impl(unit, b, idx, [&] (Vout& v) {
+void lower(const VLS& e, cvttsd2siq& i, Vlabel b, size_t idx) {
+  lower_impl(e.unit, b, idx, [&] (Vout& v) {
     // Clear FPSR IOC flag.
     auto const tmp1 = v.makeReg();
     auto const tmp2 = v.makeReg();
@@ -1373,8 +1337,8 @@ void lower(Vunit& unit, cvttsd2siq& i, Vlabel b, size_t idx) {
   });
 }
 
-void lower(Vunit& u, callm& i, Vlabel b, size_t z) {
-  lower_impl(u, b, z, [&] (Vout& v) {
+void lower(const VLS& e, callm& i, Vlabel b, size_t z) {
+  lower_impl(e.unit, b, z, [&] (Vout& v) {
     lowerVptr(i.target, v);
 
     auto const scratch = v.makeReg();
@@ -1385,8 +1349,8 @@ void lower(Vunit& u, callm& i, Vlabel b, size_t z) {
   });
 }
 
-void lower(Vunit& u, jmpm& i, Vlabel b, size_t z) {
-  lower_impl(u, b, z, [&] (Vout& v) {
+void lower(const VLS& e, jmpm& i, Vlabel b, size_t z) {
+  lower_impl(e.unit, b, z, [&] (Vout& v) {
     lowerVptr(i.target, v);
 
     auto const scratch = v.makeReg();
@@ -1398,15 +1362,15 @@ void lower(Vunit& u, jmpm& i, Vlabel b, size_t z) {
 
 ///////////////////////////////////////////////////////////////////////////////
 
-void lower(Vunit& u, stublogue& i, Vlabel b, size_t z) {
-  lower_impl(u, b, z, [&] (Vout& v) {
+void lower(const VLS& e, stublogue& i, Vlabel b, size_t z) {
+  lower_impl(e.unit, b, z, [&] (Vout& v) {
     // Push both the LR and FP regardless of i.saveframe to align SP.
     v << pushp{rlr(), rvmfp()};
   });
 }
 
-void lower(Vunit& u, stubret& i, Vlabel b, size_t z) {
-  lower_impl(u, b, z, [&] (Vout& v) {
+void lower(const VLS& e, stubret& i, Vlabel b, size_t z) {
+  lower_impl(e.unit, b, z, [&] (Vout& v) {
     // Pop LR and (optionally) FP.
     if (i.saveframe) {
       v << popp{rvmfp(), rlr()};
@@ -1418,8 +1382,8 @@ void lower(Vunit& u, stubret& i, Vlabel b, size_t z) {
   });
 }
 
-void lower(Vunit& u, tailcallstub& i, Vlabel b, size_t z) {
-  lower_impl(u, b, z, [&] (Vout& v) {
+void lower(const VLS& e, tailcallstub& i, Vlabel b, size_t z) {
+  lower_impl(e.unit, b, z, [&] (Vout& v) {
     // Restore LR from native stack and adjust SP.
     v << popp{PhysReg(rAsm), rlr()};
 
@@ -1428,22 +1392,22 @@ void lower(Vunit& u, tailcallstub& i, Vlabel b, size_t z) {
   });
 }
 
-void lower(Vunit& u, stubunwind& i, Vlabel b, size_t z) {
-  lower_impl(u, b, z, [&] (Vout& v) {
+void lower(const VLS& e, stubunwind& i, Vlabel b, size_t z) {
+  lower_impl(e.unit, b, z, [&] (Vout& v) {
     // Pop the call frame.
     v << lea{rsp()[16], rsp()};
   });
 }
 
-void lower(Vunit& u, stubtophp& i, Vlabel b, size_t z) {
-  lower_impl(u, b, z, [&] (Vout& v) {
+void lower(const VLS& e, stubtophp& i, Vlabel b, size_t z) {
+  lower_impl(e.unit, b, z, [&] (Vout& v) {
     // Pop the call frame
     v << lea{rsp()[16], rsp()};
   });
 }
 
-void lower(Vunit& u, loadstubret& i, Vlabel b, size_t z) {
-  lower_impl(u, b, z, [&] (Vout& v) {
+void lower(const VLS& e, loadstubret& i, Vlabel b, size_t z) {
+  lower_impl(e.unit, b, z, [&] (Vout& v) {
     // Load the LR to the destination.
     v << load{rsp()[AROFF(m_savedRip)], i.d};
   });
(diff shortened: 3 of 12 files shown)


mxw:
I definitely don't think that we should be reifying the assembler tmp registers as part of the ABI.

We've actually talked a little bit about how to not need `cmplims{}`.  My preferred solution at the moment is to have a flag on the `Vunit`, which gets set by `Vauto` and is left unset elsewhere (or maybe is just set for Vunits created in the faststub context?), which indicates that we're not allowed to instantiate new Vregs.  Then, it's up to the backend to handle that flag correctly.  On ARM, this means that instead of lowering `cmplim{}` in the lowerARM() pass, you let the emitter handle it, and the emitter can just use ARM's assembler tmp register.

Let me know if that doesn't make sense or if you don't think it works.

dave-estes:
Yeah, I like that too, @mxw. I'll give it a shot. I'll likely have a few questions when I start in on it.

dave-estes:
Hey @mxw, this patch just sets the CodeKind to Helper for the defRecGeneric unique stub instead of using cmplims thereby ensuring that any new virtual registers introduced during lowering are allocated to only the scratch registers. I *think* this should do the trick, but I haven't been able to confirm. It does pass the same unit tests as the previous solution, but I wasn't actually able to reproduce the original problem when I tried to roll back the previous solution and this one. I did print the disassembly (TRACE=ustubs:5) and confirmed that only the scratch registers were being used (x14 and x18 on ARM) with this new solution whereas rolling back all of the solutions caused x1, x2, and x19 to be used.

Thoughts?

mxw:
@dave-estes—That's not correct because both `kGPHelperRegs` and `kGPCallerSaved` contain `x14`.  This means that `callfaststub{}` expects `x14` will be untouched, but the ABI you're specifying for the unique stub (the "helper" ABI) allows `x14` to be allocated.  Basically, your solution only works when nobody tries to use the value of `x14` after the `callfaststub{}` (given that you're seeing `x14` actually used in your `genericDecRef` implementation).

dave-estes:
Oops. I forgot to double check that kGPHelperRegs contained proper assembler and vixl macroassembler scratch registers. Don't know how x14 made it in there, but it was wrong. Note that there are two vixl macroassembler scratch registers, but I only used x16 because vasm-xls likes to use x17 as a tmp too.

mxw:
Still not a huge fan of this, because it's sort of an ARM-specific hack—using the helper ABI on x64 and ppc64 is no different from using any other ABI here, since the `kGPHelperRegs` set overlaps with the main trace ABI's GP reg set.  For ARM, there's an implicit reliance on the unreserved registers in the helper ABI not being recognized as GP registers at all elsewhere.

One solution is to add a fourth ABI kind, which on x64 and ppc64 has _no_ unreserved registers, _and_ can't spill, and which on ARM, uses the assembler tmp registers at will.  This is essentially equivalent to the solution I was proposing before, but without adding an extra flag on Vunit.

dave-estes:
OK, I think I mistook the purpose of the Helper ABI. I still like the idea of using a code kind to enforce the fast ABI, so I'm going to go ahead and make that fourth ABI for all platforms.

mxw:
Ah—"scratch register" here definitely does not mean "HHVM ABI scratch registers", which on x64 is the empty set.  I checked with @edwinsmith, and the intended meaning is "native non-callee-saved, non-reserved, non-argument registers".  That's already not the case for either ARM or ppc64, where it overlaps with argument registers on both, but fortunately, it doesn't overlap with the first several, and all the code compiled with CodeKind::Helper takes very few arguments (usually zero).

You should add a fourth ABI—maybe call it `CodeKind::FastStub`, since that makes it fairly clear that it's tightly coupled to `callfaststub{}`.  I'll update the documentation on `CodeKind::Helper` myself to make it clearer.

dave-estes:
Perfect, thanks @mxw. I'll get to work on it ASAP.

hphp/runtime/vm/jit/abi-arm.cpp
```diff
@@ -89,9 +89,15 @@ const RegSet kScratchCrossTraceRegs =
...
  * Helper code ABI registers.
  */
-const RegSet kGPHelperRegs = rAsm | vixl::x14;
+const RegSet kGPHelperRegs = vixl::x14 | rAsm;
 const RegSet kSIMDHelperRegs = vixl::d5 | vixl::d6 | vixl::d7;
 
+/*
+ * Fast Stub ABI registers.
+ */
+const RegSet kGPFastStubRegs = vixl::x16 | rAsm;
```
mxw: There's a comment in `abi-arm.cpp` that reads:

```
// x16 = used as ip0/tmp0 by MacroAssembler
```

How do you know that letting XLS use `x16` won't cause conflicts with whatever the assembler (vixl?) is doing with `x16`?

hphp/runtime/vm/jit/abi-ppc64.cpp
```diff
@@ -89,6 +89,12 @@ const RegSet kScratchCrossTraceRegs =
 const RegSet kGPHelperRegs = ppc64::rAsm | reg::r10; //TODO
 const RegSet kXMMHelperRegs; // TODO
 
+/*
+ * Fast Stub ABI registers.
+ */
+const RegSet kGPFastStubRegs = RegSet();
+const RegSet kSIMDFastStubRegs = RegSet();
```
mxw: These seem unnecessary.

hphp/runtime/vm/jit/abi-x64.cpp
```diff
@@ -78,6 +78,12 @@ const RegSet kScratchCrossTraceRegs =
 const RegSet kGPHelperRegs = x64::rAsm | reg::r11;
 const RegSet kXMMHelperRegs = reg::xmm5 | reg::xmm6 | reg::xmm7;
 
+/*
+ * Fast Stub ABI registers.
+ */
+const RegSet kGPFastStubRegs = RegSet();
+const RegSet kSIMDFastStubRegs = RegSet();
```
mxw: Kill these.

hphp/runtime/vm/jit/types.h
```diff
@@ -145,6 +145,11 @@ enum class CodeKind : uint8_t {
    * Helper code that uses scratch registers only.
    */
   Helper,
+
+  /*
+   * Fast Stub ABI can only use platform-specific "assembler" scratch registers
```
mxw: nit: Period at end of comment.  Also, can you indicate that this is intended for code called by `callfaststub{}`?

mxw:
Oh, but also all tests are failing and the failures are real.  Looks like we _were_ allocating some registers—I'm guessing the ones that `callfaststub{}`'s ABI doesn't exclude?  So x64 and ppc64 probably needs to have those available.

hphp/runtime/vm/jit/abi-arm.cpp
dave-estes: I *think* this should be fine. The VIXL macroassembler uses its temps only within the scope of one [macro-]assembly instruction. The only danger is in crafting a fast stub which is using x16 and then using a vasm instruction which is emitted with a VIXL macroassembly routine that uses x16 as a temp. Preventing this is up to the writer of the fast stub. You think that's a reasonable caveat, or should we try to come up with some more explicit way of preventing this?

hphp/runtime/vm/jit/abi-x64.cpp
dave-estes: Will remove.

hphp/runtime/vm/jit/abi-ppc64.cpp
dave-estes: Will remove.

hphp/runtime/vm/jit/types.h
dave-estes: OK

dave-estes:
Hey @mxw, this new patch implements your original idea of using a flag that rides along with the Vunit. When I originally coded it up, I really wanted to sprinkle asserts throughout vasm-arm.cpp that ensures we're not introducing new virtuals when the flag is set. I still couldn't think of a way to actuall allow it for that slow path through decRegGeneric that is protected by a PhysRegSaver. I figure it's OK as it is for now, though I'd really love to one day use the ABI and work the knowledge of PhysRegSaver to the register allocator.

hphp/runtime/vm/jit/vasm-gen-inl.h
```diff
@@ -75,13 +75,13 @@ namespace detail {
...
 TCA vwrap_impl(CodeBlock& main, CodeBlock& cold, DataBlock& data,
-               CGMeta* meta, GenFunc gen, CodeKind kind) {
+               CGMeta* meta, GenFunc gen, CodeKind kind, bool fast) {
   CGMeta dummy_meta;
 
   auto const start = main.frontier();
 
   { // Finish emitting in this scope so that we can assert about `dummy_meta'.
-    Vauto vauto { main, cold, data, meta ? *meta : dummy_meta, kind };
+    Vauto vauto { main, cold, data, meta ? *meta : dummy_meta, kind, fast };
```
mxw: This doesn't need to be an argument to the `Vauto` constructor.  Just add

```cpp
vauto.unit().fast = true;
```

after you construct it.

hphp/runtime/vm/jit/vasm-gen.h
```diff
@@ -180,7 +182,8 @@ namespace detail {
   template<class GenFunc>
   TCA vwrap_impl(CodeBlock& main, CodeBlock& cold, DataBlock& data,
                  CGMeta* meta, GenFunc gen,
-                 CodeKind kind = CodeKind::CrossTrace);
+                 CodeKind kind = CodeKind::CrossTrace,
+		 bool fast = false);
```
mxw: alignment

hphp/runtime/vm/jit/vasm-gen.h
```diff
@@ -193,6 +196,13 @@ TCA vwrap(CodeBlock& cb, DataBlock& data, CGMeta& meta, GenFunc gen,
                             [&] (Vout& v, Vout&) { gen(v); }, kind);
 }
 template<class GenFunc>
+TCA vwrap(CodeBlock& cb, DataBlock& data, CGMeta& meta, GenFunc gen,
+	  bool fast) {
```
mxw: alignment

hphp/runtime/vm/jit/vasm-gen.h
```diff
@@ -193,6 +196,13 @@ TCA vwrap(CodeBlock& cb, DataBlock& data, CGMeta& meta, GenFunc gen,
...
+  return detail::vwrap_impl(cb, cb, data, &meta,
+                            [&] (Vout& v, Vout&) { gen(v); },
+			    CodeKind::CrossTrace, fast);
```
mxw: alignment

hphp/runtime/vm/jit/vasm-unit.h
```diff
@@ -211,6 +211,7 @@ struct Vunit {
   uint16_t cur_voff{0};  // current instruction index managed by Vout
   bool padding{false};
   bool profiling{false};
+  bool avoidVreg{false}; // Indication to lowerers to avoid use of new Vregs.
```
mxw: `lowercase_with_underscores` to match the other members.

This name is also kind of misleading, because it's actually a correctness condition, not a hint.  How about `allow_implicit_vregs{true}`?

I also still think that enforcing it by means of assertions is a good idea.

dave-estes: Yeah I definitely do too, and I like you approach to implementing those assertions. I still can't think of how to get around the portion of decRefHelper that can safely use virtuals. Initially I just started adding the same support in those lowerers, but I hit a wall in some of them. In particular, I ran into some complicated Vptrs that made use of both a base and index. When lowering, two new virtuals are required and their live ranges overlap. I ran out of steam at that point. I have plans to streamline lowerVptr quite a bit, and that might accommodate lowering without using new Vregs.

With that said, I still prefer to actually figure out a way to simple relax the `allow_implicit_vregs` when emitting code protected by a PhysRegSaver because it *should* lead to better code that balances register use more on platforms that don't support register renaming (Cortex-A53 for instance).

Let me know if you think we can postpone the asserts, and I'm about to address your comment about the flag name now.

cmuellner:
I've seen that this PR removes the function emit(cmplims&), which
includes a wrong assertion. Therefore I'd like to ask about the status
of this. Otherwise I'd create a PR which removes the assertion code:

```
@@ -915,16 +915,10 @@ void Vgen::emit(const cmplims& i) {
   uint8_t mode = (((i.s1.base.isValid()  & 0x1) << 0) |
                   ((i.s1.index.isValid() & 0x1) << 1) |
                   (((i.s1.disp != 0)     & 0x1) << 2));
-  switch (mode) {
-    case BASE:
-    case BASE | INDEX:
-      break;
-    case BASE | DISP:
-      assertx(i.s1.disp >= -256 && i.s1.disp <= 255);
-      break;
-    default:
-      always_assert(false);
+  if (mode != BASE && mode != (BASE | INDEX) && mode != (BASE | DISP)) {
+    always_assert(false);
   }
```

hphp/runtime/vm/jit/vasm-instr.h
```diff
@@ -905,6 +907,13 @@ struct divint { Vreg s0, s1, d; };
 struct nop {};
 struct ud2 {};
 
+
```
mxw: nit: Spurious newline.

hphp/runtime/vm/jit/unique-stubs.cpp
```diff
@@ -1034,12 +1034,15 @@ TCA emitDecRefGeneric(CodeBlock& cb, DataBlock& data) {
...
 
   auto const start = vwrap(cb, data, meta, [] (Vout& v) {
+    v << vregrestrict{};
     v << stublogue{};
 
     auto const rdata = rarg(0);
     auto const rtype = rarg(1);
 
     auto const destroy = [&] (Vout& v) {
+      v << vregallow{};
```
mxw: This should go after we construct the `PhysRegSaver` below.

Actually, I think it would be better if we just did the `v << vregallow{}` in PhysRegSaver's constructor after the pushes, and then the `v << vregrestrict{}` in the destructor, before the pops.  Then we can make it explicit in PhysRegSaver's documentation that this is what the PhysRegSaver "buys" us.

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -1036,6 +1003,15 @@ void lower(Vunit& unit, Inst& inst, Vlabel b, size_t i) {}
 
 ///////////////////////////////////////////////////////////////////////////////
 
+bool allowVreg = true;
```
mxw: Herp derp, of course—that's the only place where these instructions are meaningful at all...

Okay, let's get the refactoring of these lower loops taken care of first.  There's a good amount of duplication among the different platforms that I've been putting off until the boilerplate started getting out of hand, which appears to be the case now.  Let me put up a diff that should give you what you need to generalize this logic and make it less globally-stateful.  I'll ping here with the diff once I've landed it.

hphp/runtime/vm/jit/unique-stubs.cpp
dave-estes: Initially I wanted to do the same thing, but it's difficult for the PhysRegSaver to know if it's currently within a restricted region in order to emit the `vregstrict` after the registers are restored. This information isn't done until the analysis to compute the dominating vreg restictions is performed while lower.

mxw: Ah, yes, I see.  Maybe we should give these intrinsics scoping semantics—i.e., each `vregrestrict{}` creates a nested "restrict" context that can be popped with a `vregallow{}`?  On the one hand, that seems more complicated—but on the other, it makes it less fragile, and means it's not dangerous to use in a context where we don't have full control over the entire Vunit.  Your dominator tracking can then just be incrementing and decrementing an "level" integer.

mxw:
@dave-estes—Okay, my refactor just sync'd.  You should be able to do everything you're doing here in `vasm-lower-inl.h` now.  (NB: Forward declare any helpers you want in a `namespace vasm_detail { ... }` and then implement them in `vasm-lower.cpp`.)  Let me know if that doesn't seem straightforward; thanks!

dave-estes:
Thanks, @mxw. That cleaned things up quite a bit. I still couldn't figure out an easy way to move the `vregConstraints` vector out of the unit since the platform-specific lowerers need access to it to determine if there is a restriction or now. Note, that I changed the meaning of that vector slightly, and so I've changed the name as well. Originally it held only the vreg constraint that dominated that block. Now it is still initialized like that, but it is also updated as that block is lowered and as we encounter `vregallow` and `vregrestrict` instead of relying on a global.

mxw:
So, I think what you actually want here is to thread more state through to the `lower_impl()` callback.  Why not define something like

```
/*
 * Vasm-to-vasm lowering state.
 */
struct VLS {
  bool allow_vreg;
};
```

in `vasm-lower.h` and change the signature of `lower_impl()` to

```
void lower_impl(const VLS& env, Vinstr& inst, Vlabel b, size_t i);
```

You'll have to update all the ARM lowerers to have a signature of `lower(const VLS& env, Vunit& unit, ...)`.

dave-estes:
Gotcha. Using `VLS env` in the ARM lowerers will certainly help the lowerers for `cmplim` and `declm`. However, while trying to add the asserts for all platforms, I ended up moving the lowerers for `vregrestrict` and `vregallow` to vasm-lower.cpp and then adding the assert in vasm-lower-inl.h. If I move the `vregConstraints` vector to VLS along with a `allowVreg` flag, then I think I'll have to make it visible in the vasm-lower* files as well. Should we change all the lowerers, or should we just stick with the `Vunit`.

hphp/runtime/vm/jit/unique-stubs.cpp
mxw: Any thoughts on this?

hphp/runtime/vm/jit/vasm-lower-inl.h
```diff
@@ -27,6 +27,12 @@ namespace HPHP { namespace jit {
 
 ///////////////////////////////////////////////////////////////////////////////
 
+namespace detail {
+jit::vector<VregConstraint> computeDominatingConstraints(Vunit& unit);
```
mxw: This needs documentation.

hphp/runtime/vm/jit/vasm-lower-inl.h
```diff
@@ -38,14 +44,23 @@ void vasm_lower(Vunit& unit, Vlower lower_impl) {
   // iterators.
   auto& blocks = unit.blocks;
 
+  VLS env{unit, true};
+  auto vregConstraints = detail::computeDominatingConstraints(unit);
```
mxw: `auto const`

hphp/runtime/vm/jit/vasm-lower-inl.h
```diff
@@ -38,14 +44,23 @@ void vasm_lower(Vunit& unit, Vlower lower_impl) {
   // iterators.
   auto& blocks = unit.blocks;
 
+  VLS env{unit, true};
```
mxw: nit: Just `VLS env { unit };`, since the `true` is always going to be overridden before anything happens.

hphp/runtime/vm/jit/vasm-lower-inl.h
```diff
@@ -38,14 +44,23 @@ void vasm_lower(Vunit& unit, Vlower lower_impl) {
...
+  auto vregConstraints = detail::computeDominatingConstraints(unit);
+
   // The vlower() implementations, and `lower', may allocate scratch blocks and
   // modify instruction streams, so we cannot use standard iterators here.
   PostorderWalker{unit}.dfs([&] (Vlabel b) {
     assertx(!blocks[b].code.empty());
 
+    env.allow_vreg = vregConstraints[b] != VregConstraint::Restrict;
```
mxw: If we scope the restrictions, this can just be `env.allow_vreg = vregConstraint[b] == 0;`

hphp/runtime/vm/jit/vasm-lower.cpp
```diff
@@ -226,28 +277,36 @@ void lower(Vunit& unit, vcallarray& inst, Vlabel b, size_t i) {
...
+}
+void lower(VLS& env, syncvmret& inst, Vlabel b, size_t i) {
+  env.unit.blocks[b].code[i] = copy2{inst.data,   inst.type,
+                                     rret_data(), rret_type()};
 }
-void lower(Vunit& unit, defvmrettype& inst, Vlabel b, size_t i) {
-  unit.blocks[b].code[i] = copy{rret_type(), inst.type};
+void lower(VLS& env, vregallow& inst, Vlabel b, size_t i) {
+  env.allow_vreg = true;
+  lower_impl(env.unit, b, i, [&] (Vout& v) {});
```
mxw: You're doing this to just kill the `vregallow{}` (and `vregrestrict{}`, below), right?  If so, you should just nop it out instead:

```
env.unit.blocks[b].code[i] = nop{};
```

It avoids a (possibly expensive) shift, which in most implementations of `std::vector` won't (I think) shrink the allocation anyway...

hphp/runtime/vm/jit/vasm-lower.h
```diff
@@ -25,14 +25,28 @@ namespace HPHP { namespace jit {
...
 
+/*
+ * Vasm-to-vasm lowering state.
+ */
+struct VLS {
+  Vunit& unit;
+  bool allow_vreg;
+};
+
+enum class VregConstraint : uint8_t {
```
mxw: If we just use integers to denote the "restriction level", we can get rid of this.

hphp/runtime/vm/jit/vasm-lower.cpp
```diff
@@ -33,6 +33,57 @@ namespace HPHP { namespace jit {
...
+
+jit::vector<VregConstraint> computeDominatingConstraints(Vunit& unit) {
+  jit::vector<VregConstraint> vregConstraints(unit.blocks.size(),
+                                              VregConstraint::Unknown);
+
+  // Build reverse postorder list of Vlabels
+  jit::vector<Vlabel> rpo;
+  PostorderWalker{unit}.dfs([&] (Vlabel ib) {
+    rpo.insert(rpo.begin(), ib);
+  });
```
mxw: Just call `sortBlocks(unit);`.

hphp/runtime/vm/jit/vasm-lower.cpp
```diff
@@ -33,6 +33,57 @@ namespace HPHP { namespace jit {
...
+
+  // Calculate each block's dominating vreg constraint. The lowerers will use
+  // this as the initial vreg constraint for each block, adjusting as they
+  // encounter vregallow/vregrestrict during lowering.
+  vregConstraints[rpo[0]] = VregConstraint::Allow;
+  for (auto b: rpo) {
```
mxw: `for (auto const b : rpo) {`

hphp/runtime/vm/jit/vasm-lower.cpp
```diff
@@ -33,6 +33,57 @@ namespace HPHP { namespace jit {
...
+        con = VregConstraint::Restrict;
+      } else if (inst.op == Vinstr::vregallow) {
+        con = VregConstraint::Allow;
+      }
+    }
+
+    // Pass the contraint to each of the successors, enforcing that a block
+    // isn't passed conflicting contraints from predecessors. i.e. A successor
+    // must be pass all restrict or all allow/none.
+    for (auto s: succs(unit.blocks[b])) {
```
mxw: `for (auto s : succs(unit.blocks[b])) {`

(More generally, spaces around operators... and colons.)

hphp/runtime/vm/jit/vasm-lower.cpp
dave-estes: Much nicer. Done.

hphp/runtime/vm/jit/vasm-lower.cpp
dave-estes: Done.

hphp/runtime/vm/jit/vasm-lower.cpp
dave-estes: Done.

hphp/runtime/vm/jit/vasm-lower.cpp
dave-estes: Nice to know. :-) Done.

hphp/runtime/vm/jit/vasm-lower-inl.h
dave-estes: Done.

hphp/runtime/vm/jit/vasm-lower-inl.h
dave-estes: Done.

hphp/runtime/vm/jit/vasm-lower-inl.h
dave-estes: Done.

hphp/runtime/vm/jit/unique-stubs.cpp
dave-estes: I really didn't like the idea. I felt like the verbose current design was a little more readable. I gave it a shot though, and the result is pretty clean and quite understandable with the addition of the comments.

hphp/runtime/vm/jit/vasm-lower.h
dave-estes: Done.

hphp/runtime/vm/jit/vasm-lower-inl.h
dave-estes: Note: I implemented this as suggested, but it drops the nesting concept a little. I use the integer to indicate the nested depth of the `vregrestrict`s when calculating the `vregConstraints`. But then when using this `env.allow_vreg` bool, we lose the nest depth. In a scenario where `vregConstraints[b] == 3`, a `vregallow` within a block will allow new vregs.

Is that you intended behavior? I think it needs to be.

hphp/runtime/vm/jit/vasm-lower-inl.h
```diff
@@ -38,14 +60,23 @@ void vasm_lower(Vunit& unit, Vlower lower_impl) {
...
   // modify instruction streams, so we cannot use standard iterators here.
   PostorderWalker{unit}.dfs([&] (Vlabel b) {
     assertx(!blocks[b].code.empty());
 
+    env.allow_vreg = vregConstraints[b] == 0;
     for (size_t i = 0; i < blocks[b].code.size(); ++i) {
-      vlower(unit, b, i);
-      lower_impl(blocks[b].code[i], b, i);
+      DEBUG_ONLY auto allow = env.allow_vreg;
+      DEBUG_ONLY auto next = unit.next_vr;
```
mxw: `auto const` for both of these.

hphp/runtime/vm/jit/vasm-lower.cpp
```diff
@@ -33,6 +33,55 @@ namespace HPHP { namespace jit {
...
+    for (size_t i = 0; i < unit.blocks[b].code.size(); ++i) {
+      auto& inst = unit.blocks[b].code[i];
+      if (inst.op == Vinstr::vregrestrict) {
+        con++;
+      } else if (inst.op == Vinstr::vregallow) {
+        con--;
+      }
+    }
+    assert_flog(con >= 0,
+		"Dominating constraint can't be unknown.");
```
mxw: If we're going to allow "nesting" of constraints, then we probably don't want this "unknown" behavior.  In particular, if we're going to have `vregallow{}` and `vregrestrict{}` be intrinsic to PhysRegSaver, then we need to let the `vregallow{}` in the constructor change the level to -1, then have the `vregrestrict{}` change the level back to 0, so that PhysRegSaver will still work without being dominated by a `vregrestrict{}` (or more generally, without being in a context where the constraint level is `>= 1`).

The check for "can I make new Vregs" will then be `vreg_level <= 0`.

Also, you commented:

> Note: I implemented this as suggested, but it drops the nesting concept a little. I use the integer to indicate the nested depth of the vregrestricts when calculating the vregConstraints. But then when using this env.allow_vreg bool, we lose the nest depth. In a scenario where vregConstraints[b] == 3, a vregallow within a block will allow new vregs.
>
> Is that you intended behavior? I think it needs to be.

Okay, so... why not change `allow_vreg` to be an int (`vreg_level` or such), and have `vregallow{}` and `vregrestrict{}` decrement and increment it as usual?

hphp/runtime/vm/jit/phys-reg-saver.cpp
```diff
@@ -31,6 +31,8 @@ PhysRegSaver::PhysRegSaver(Vout& v, RegSet regs)
   : m_v(v)
   , m_regs(regs)
 {
+  v << vregallow{};
```
mxw: This should go after the pushes.  You don't know that `push{}` and friends don't lower in a way that creates new Vregs, so it's incorrect to have a `vregallow{}` before everything is saved.

hphp/runtime/vm/jit/phys-reg-saver.cpp
```diff
@@ -69,6 +71,8 @@ PhysRegSaver::~PhysRegSaver() {
     });
     v << lea{sp[offset], sp};
   }
+
+  v << vregrestrict{};
```
mxw: Likewise, this should come before the pops.

hphp/runtime/vm/jit/vasm-lower.cpp
dave-estes: > If we're going to allow "nesting" of constraints, then we probably don't want this "unknown" behavior. In particular, if we're going to have vregallow{} and vregrestrict{} be intrinsic to PhysRegSaver, then we need to let the vregallow{} in the constructor change the level to -1, then have the vregrestrict{} change the level back to 0, so that PhysRegSaver will still work without being dominated by a vregrestrict{} (or more generally, without being in a context where the constraint level is >= 1).

Currently, `PhysRegSaver` is only used within a block and so this isn't an issue. This is because the scoped restriction level is only used when computing the dominating constraints. Once iterating over a block, then we rely on the simple bool, `allow_vreg`. I'm totally OK with allowing negative `vregConstraints` for the dominating constraints if you think it'll be an issue in the future. I'll follow up on IRC.

> Okay, so... why not change `allow_vreg` to be an int (`vreg_level` or such), and have `vregallow{}` and `vregrestrict{}` decrement and increment it as usual?

I thought of this, but I think it violates the semantic of `vregallow`. I think that nesting for the purpose of tracking to see if we're still in a `vregrestrict` scope is useful. I think that whenever we have a `vregallow`, that all of the included instructions should be allowed to use new vregs until the next `vregrestrict` closes that scope.

As an example, consider a case where the current dominating vreg constraint for a block is 3. If that blocks contains a sequence of `Vinstr`s that are enclosed by a `PhysRegSaver`, then they should be able to use new `vreg`s. But if we use the `vreg_level` approach while iterating over the `Vinstrs` in a block, then the level would only be 2 for that sequence.

hphp/runtime/vm/jit/vasm-lower-inl.h
dave-estes: Will do.

hphp/runtime/vm/jit/vasm-lower.cpp
dave-estes: After follow-up discussions: Will be changing uninitialized to INT_MAX and changing the constraint level to restrict for < 0 and allow for >=0. Will be changing boolean to also be an int level. Also changing the names of the Vinstrs to better indicate that they just adjust the vreg constraint level.

hphp/runtime/vm/jit/vasm-lower-inl.h
```diff
@@ -27,6 +27,28 @@ namespace HPHP { namespace jit {
...
+namespace detail {
+
+/*
+ * Computes the dominating vreg constraint for each block. When lowering, this
+ * will be used to determine if a lowerer can use a new vreg as a scratch, or
+ * if it must rely on platform specific assembler scratch registers.
+ * The dominating restriction is used initially for the block and then updated
+ * whenever a vregrestrict/vregunrestrict is encountered.
+ *
+ *   INT_MAX: uninitialized
```
mxw: This should not be part of the API—no consumer of `computeDominatingConstraints()`'s return should ever see any INT_MAX's.

hphp/runtime/vm/jit/vasm-lower.cpp
```diff
@@ -33,6 +33,52 @@ namespace HPHP { namespace jit {
 
 ///////////////////////////////////////////////////////////////////////////////
 
+namespace detail {
+
+jit::vector<int> computeDominatingConstraints(Vunit& unit) {
+  jit::vector<int> vregConstraints(unit.blocks.size(), INT_MAX);
```
mxw: `s/INT_MAX/std::numeric_limits<int>::max()/` everywhere.

Or better yet, something like

```
constexpr auto kInvalidConstraint = std::numeric_limits<int>::max();
```

and then use `kInvalidConstraint` everywhere.

hphp/runtime/vm/jit/vasm-lower.cpp
```diff
@@ -33,6 +33,52 @@ namespace HPHP { namespace jit {
...
+
+  // Build reverse postorder list of blocks
+  auto const rpo = sortBlocks(unit);
+
+  // Calculate each block's dominating vreg constraint. The lowerers will use
+  // this as the initial vreg constraint for each block, adjusting as they
+  // encounter vregrestrict/vregunrestrict during lowering.
+  vregConstraints[rpo[0]] = 0;
+  for (auto b : rpo) {
```
mxw: `auto const b`

hphp/runtime/vm/jit/vasm-lower.cpp
```diff
@@ -33,6 +33,52 @@ namespace HPHP { namespace jit {
...
+    // Iterate through the instructions updating the constraint accordingly.
+    auto con = vregConstraints[b];
+    for (size_t i = 0; i < unit.blocks[b].code.size(); ++i) {
+      auto& inst = unit.blocks[b].code[i];
```
mxw: Just do a range-based for:

```
for (auto const& inst : unit.blocks[b].code) {
```

The explicit index iterators you see elsewhere are because we need to pass the index to someone, or because we're manipulating the vector and can't use standard iterators.  This is just a non-mutating loop, though.

hphp/runtime/vm/jit/vasm-lower.cpp
```diff
@@ -33,6 +33,52 @@ namespace HPHP { namespace jit {
...
+        con++;
+      }
+    }
+    assert_flog(con != INT_MAX,
+                "Dominating constraint can't be uninitialized.");
+
+    // Pass the contraint to each of the successors, enforcing that a block
+    // isn't passed conflicting contraints from predecessors. i.e. A successor
+    // must be pass all restrict or all allow/none.
+    for (auto s : succs(unit.blocks[b])) {
```
mxw: `auto const s`

hphp/runtime/vm/jit/vasm-lower-inl.h
dave-estes: Done.

hphp/runtime/vm/jit/vasm-lower.cpp
dave-estes: Done.

hphp/runtime/vm/jit/vasm-lower.cpp
dave-estes: Done.

hphp/runtime/vm/jit/vasm-lower.cpp
dave-estes: Done.

hphp/runtime/vm/jit/vasm-lower.cpp
dave-estes: Done.

Issue #7320, opened 2016-09-08
Fatal error: Call to undefined function with pgsql extension


photodude:
### HHVM Version 
3.15 (and prior builds with the PocketRent/hhvm-pgsql pgsql.so extension)

### Standalone code, or other way to reproduce the problem
- `pg_set_error_verbosity()`
- `pg_set_client_encoding()`

### Expected result
no errors

### Actual result
Working on HHVM testing with pgsql I get the following errors with the PocketRent/hhvm-pgsql extension that is now included in the core as of 3.15
```php
Fatal error: Call to undefined function pg_set_error_verbosity() in /joomla-cms/libraries/joomla/database/driver/postgresql.php on line 140
Fatal error: Call to unimplemented native function pg_set_client_encoding() in /joomla-cms/libraries/joomla/database/driver/postgresql.php on line 890
```
build demonstrating the first error (the second one only shows after the first is fixed)
https://travis-ci.org/photodude/joomla-cms/jobs/132726691

I understand that the `pg_set_error_verbosity()` and ` pg_set_client_encoding()` functions are not yet implemented for various reasons...

What is the work around for this? These missing functions are causing problems for our PostgreSQL driver as we set the client encoding to `UTF8` and specify the error verbosity.

(this is a rereport of the issue from PocketRent/hhvm-pgsql as per request from @simonwelsh)

Full list of differences from the Zend pgsql module can be found at this link  https://github.com/PocketRent/hhvm-pgsql

Orvid:
If possible, could you include the link to the original issue as well?

@simonwelsh The second error there means you declared it in the the `.php` as native but you never actually implemented the function on the C++ side :P

photodude:
Unfortunately, https://github.com/PocketRent/hhvm-pgsql/issues/115 no longer exists so I can't link to it.

photodude:
@Orvid @simonwelsh I'm still learning c++ but would the fix for `pg_set_client_encoding()` be as follows? (based off the [`pg_client_encoding()` method)](https://github.com/facebook/hhvm/blob/master/hphp/runtime/ext/pgsql/pgsql.cpp#L799-L809)

```c++
static Variant HHVM_FUNCTION(pg_set_client_encoding, const Resource& connection) {
  auto pgsql = PGSQL::Get(connection);

  if (!pgsql) {
    FAIL_RETURN;
  }

  String ret(pgsql->set(clientEncoding(), CopyString);

  return ret;
}
```

[...]

PR #7396, opened 2016-09-30
Add pg_set_error_verbosity

simonwelsh:
See http://php.net/pg_set_error_verbosity for what it is. The constants
already exist, so didn't need to add them.

Part of the fix for #7320

### DIFF
--- a/hphp/runtime/ext/pgsql/ext_pgsql.php
+++ b/hphp/runtime/ext/pgsql/ext_pgsql.php
@@ -311,6 +311,9 @@ function pg_send_query(resource $connection, string $query): bool;
 function pg_set_client_encoding(resource $connection, string $encoding): int;
 
 <<__Native>>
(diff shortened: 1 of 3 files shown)


simonwelsh:
Running the example code from the docs has the same output in PHP and HHVM, so works from that perspective.

photodude:
@aorenste Will this patch be included in 3.15.2?

aorenste:
By default right now we're targeting 3.16.0.  If you really need it we could cherry-pick into 3.15.2 although we prefer not to.

photodude:
We don't support HHVM yet since we are still testing, so it's kind of relative how "needed" it is.  It would be helpful since we are likely to only support LTS versions like 3.15 and not including this fix in 3.15 would mean we would be waiting until the next LTS to add official support. Assuming we can get our test suite to pass.

aorenste:
@photodude Cherrypicked - this commit should be available in 3.15.2

photodude:
Thanks

PR #7408, opened 2016-10-07
[AArch64] Remove X64-specific cases from getEffects().

dave-estes:


mxw:
I don't understand the purpose of this diff.  Those instructions are x64-only, and are only generated by the x64 lowerer, so why is this necessary?

dave-estes:
Gotcha. We found these through inspection. I verified that `getEffects()` is executed, but didn't realize when these instructions were generated. `shlq` and `sarq` looked cross-platform to me, but I guess `shl` and `sar` are the generic versions.

I feel better with these checks just in case another platform's lowerer uses `shlq` or `sarq`, but I'm totally fine with dropping this PR altogether too if you don't think that's a likely mistake.

mxw:
Those are x64-specific instructions, and lowering to them from non-x64 architectures is a bug.  We could add a checker pass which ensures that arch-specific instructions are never produced on other architectures, I suppose, but it doesn't seem urgent.

PR #7409, opened 2016-10-07
Add support for multiple seperators in a DSN

### LINKED ISSUE #7399: connection to Postgresql fail to translate host name
### HHVM Version
3.16.0-dev (hhvm-nightly)

Joomla Postgresql driver (below is a snip from the driver that hopefully would represent Standalone code)
```php
$dsn .= "host={localhost} ";
$dsn .= "dbname={joomla_ut} user={postgres} password={}";
$this->connection = @pg_connect($dsn)
```

### Expected result
connection to Postgresql without PDOExceptions, previously worked in 3.15 and with the old 3rdparty pgsql module 

### Actual result
`PDOException: [112]: could not translate host name "localhost port=5432 dbname=joomla_ut" to address: Name or service not known`

simonwelsh:
pdo_pgsql allows seperatoring values with a space as well as a
semicolon. Add support to the DSN parser for multiple seperator
characters and set pdo_pgsql to allow ';' and ' '. pdo_mysql, currently
the only other connector using this parser, does not change.

Fixes #7399.

### DIFF
--- a/hphp/runtime/ext/pdo/pdo_driver.cpp
+++ b/hphp/runtime/ext/pdo/pdo_driver.cpp
@@ -100,8 +100,8 @@ int PDOConnection::parseDataSource(const char *data_source,
                                    int data_source_len,
                                    struct pdo_data_src_parser *parsed,
                                    int nparams,
-                                   const char *seperators/* = ";" */,
-                                   int seperatorCount/* = 1 */) {
+                                   const char *separators/* = ";" */,
+                                   int separatorCount/* = 1 */) {
   int i, j;
   int valstart = -1;
   int semi = -1;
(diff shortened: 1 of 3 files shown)


photodude:
Thanks @simonwelsh, We hope to see this fix in hhvm 3.15.2

hphp/runtime/ext/pdo/pdo_driver.cpp
```diff
@@ -98,14 +99,22 @@ bool PDOConnection::support(SupportedMethod method) {
 int PDOConnection::parseDataSource(const char *data_source,
                                    int data_source_len,
                                    struct pdo_data_src_parser *parsed,
-                                   int nparams) {
+                                   int nparams,
+                                   const char *seperators/* = ";" */,
```
Orvid: Any reason not to use `folly::StringPiece`?

photodude:
@aorenste Can this fix 714e793 be added to the 3.15 branch?

aorenste:
@photodude - I'll add it to the list

photodude:
Thanks @aorenste

Issue #7302, opened 2016-08-30
DumpTC is not dumping bytescodes


racardoso:
### HHVM Version

HipHop VM 3.15.0-dev (rel)
Compiler: heads/master-0-gd8591b1bae517929a27474fc103180d57e6328f2
Repo schema: ca03b9d0005f83442a55f54ab604919695f40474

### Standalone code, or other way to reproduce the problem

`$ nano hhvm/hphp/test/quick/echo.php`

```
<?php

function main() {
  echo false;
  echo 'x';
  echo true;
  echo 'x';
  echo null;
  echo 'x';
  echo 0;
  echo 'x';
  echo 1;
  echo 1.0;
  echo "\n";

}
main();
```
`$ hhvm -vEval.DumpTC=true --count=20 hphp/test/echo.php > /dev/null
`$ hphp/tc-print
 
### Expected result

Bytecodes dumped before disassembled code (something like):

```
ReleaseVVAndSkip
         0x6c004a1: cmpq  $0x0, 0x28(%rbp)
         0x6c004a6: jnz 0xe800856
```
### Actual result (trimmed)

```
# Read 0 symbols from file /tmp/tc_dump_helpers_addrs.txt
# Config file: 
# Repo schema: ca03b9d0005f83442a55f54ab604919695f40474

====================
Translation 0 {
  src.md5 = 61633864303138646139396434353164
  src.funcId = 6361
  src.funcName = Pseudo-main
  src.resumed = false
  src.bcStartOffset = 0
  src.guards = 0
  kind = TransLive
  hasLoop = 0
  aStart = 0x560045a
  aLen = 0x7b
  coldStart = 0xd200841
  coldLen = 0x4d
  frozenStart = 0xea000ab
  frozenLen = 0x0
}

----------
bytecode:
----------
<<< couldn't find unit 61633864303138646139396434353164 to print bytecode range [0,19) >>>
----------
x64: main
----------
<<< couldn't find unit 61633864303138646139396434353164 to print bytecode at offset 0 >>>
     0x560045a:                                                 mov $0x7f48ccd75060, %rax
     0x5600464:                                                 movq  %rax, -0x20(%rbp)
     0x5600468:                                                 movl  $0x0, -0x14(%rbp)
<<< couldn't find unit 61633864303138646139396434353164 to print bytecode at offset 6 >>>
     0x560046f:                                                 movq  %rbp, -0x30(%rbp)
[...]

PR #7415, opened 2016-10-12
Fix tc-print not mapping translations to bytecode

octmoraru:
Commit c28f559 changed MD5's constructor to take a folly::StringPiece parameter
instead of a char*. This was causing OfflineTransData::loadTCData to call the
wrong MD5 constructor; hence an incorrect function md5 was being read from the
tcdata archive.  This diff solves the problem by packing the md5 string into a
StringPiece object.
Fixex #7302.

### DIFF
--- a/hphp/runtime/vm/repo-helpers.cpp
+++ b/hphp/runtime/vm/repo-helpers.cpp
@@ -404,7 +404,7 @@ void RepoQuery::getMd5(int iCol, MD5& md5) {
       " (expected 16, got %zu) in '%s'",
       __func__, &m_stmt.repo(), iCol, size, m_stmt.sql().c_str());
   }
-  new (&md5) MD5(blob);
+  new (&md5) MD5(blob, size);
 }
 
 void RepoQuery::getTypedValue(int iCol, TypedValue& tv) {
--- a/hphp/util/md5.h
+++ b/hphp/util/md5.h
@@ -53,7 +53,8 @@ struct MD5 : private boost::totally_ordered<MD5> {
   }
 
   // Blob is assumed to be in network byte order.
-  explicit MD5(const void* blob) {
+  explicit MD5(const void* blob, size_t len) {
+    assertx(len == 16);
     q[0] = ntohq(((const uint64_t*)blob)[0]);
     q[1] = ntohq(((const uint64_t*)blob)[1]);
   }


hphp/tools/tc-print/offline-trans-data.cpp
```diff
@@ -84,7 +84,7 @@ void OfflineTransData::loadTCData(string dumpDir) {
       continue;
     }
     READ(" src.md5 = %s", md5Str);
-    tRec.md5 = MD5(md5Str);
+    tRec.md5 = MD5{folly::StringPiece{md5Str, strlen(md5Str)}};
```
markw65: Can't you omit the strlen?

octmoraru: You're right, I can.

aorenste:
Instead of casting those locations to StringPiece could you get rid of the generic `MD5(const void* blob)` constructor?  `void*` without a length seems really dangerous and there are possibly other `MD5(char*)` being missed.

octmoraru:
Would love to do that, but it seems that there are legitimate uses of the `const void*` constructor. For example, in `RepoQuery::getMd5()` [here](https://github.com/facebook/hhvm/blob/master/hphp/runtime/vm/repo-helpers.cpp#L397); even if I used `ntohq` in the caller, I would still need a `const void*` constructor for `MD5`. Thoughts?

On the other hand, it's true that I missed fixing an `MD5(char *)`; I updated the PR.

mofarrell:
I think @aorenste has a good fix there switching `MD5(const void* blob)` to `MD5(const void* blob, size_t len)` would avoid a lot of possible misuses  (Not to mention it is a kind of a misnomer because it only works with a blob of length 2 right now).  It would also make a conversion from `char*` to `StringPiece` implicit.

swtaarrs:
The `const void*` constructor isn't great but I'm not sure anything we replace uses of it with would be any better. At some point the assumption has to be made that we have 16 bytes of MD5 data to read, and that would just shift the assumption to the caller.

@octmoraru what about explicitly deleting the `const char*` constructor for `MD5`? That should make any other cases like this fail.

aorenste:
@octmoraru Sorry - I explained what I wanted poorly.  @mofarrell is right - what I intended was to replace the void* constructor with something more reasonable (just adding a length to the constructor - for now the implementation can just assert that the length is exactly 16).  Otherwise we risk an accidental construction with any pointer (not just char*) using the incorrect constructor.

mofarrell:
It looks like you didn't update this callsite. https://github.com/facebook/hhvm/blob/master/hphp/runtime/vm/repo-helpers.cpp#L397

octmoraru:
@mofarrell do you mean [this](https://github.com/facebook/hhvm/pull/7415/files#diff-0692d352d3f308d803e2f6c1861443ccR407) call site?

aorenste:
@mofarrell @octmoraru Yes - I think that's fine.  Just waiting for the internal build to run.

PR #7427, opened 2016-10-17
h2tp: Fix string conversion being lost for input "$a"

### LINKED ISSUE #7424: h2tp: "$a" is converted to $a
### HHVM Version

```
HipHop VM 3.16.0-dev (rel)
Compiler: heads/master-0-ge9eb3fc414fbae9b2c999a4f11390e23050b23b7
Repo schema: 9cc0ebe15bc9d2ba3609c5e4458deeef9ed86dcc
```

``` php
<?hh

$a = 1;
print \is_string("$a");
```

Run `h2tp` on file.
### Expected result

``` php
<?php
require_once ($GLOBALS["HACKLIB_ROOT"]);
$a = 1;
echo (\is_string("$a"));
```
### Actual result

``` php
<?php
require_once ($GLOBALS["HACKLIB_ROOT"]);
$a = 1;
echo (\is_string($a));
```

The conversion to a string via variable interpolation is lost.

jesseschalken:
Fixes #7424


aorenste:
Please add a unit test which verifies that this won't re-break in the future.

hphp/hack/src/h2tp/unparser/unparser.ml
```diff
@@ -839,6 +839,7 @@ let rec u_program v = u_of_list_spc u_def v
     | Float f -> u_pstring f
     | String (p, s) ->
       StrList [Str "\""; u_pstring (p, Php_escaping.escape s); Str "\""]
+    | String2 [elem] -> StrList [u_expr elem; Str ".\"\""]
```
Orvid: I would say either leave it like it was in this case (`"$a"`), or explicitly convert it to a string `(string)$a`.

jesseschalken:
@aorenste I tried running `make hack_test` in `hphp/hack` (assuming that runs the tests in `hphp/hack/test/h2tp`) but it fails with this:

```
Finished, 30 targets (30 cached) in 00:00:00.
Finished, 1311 targets (1311 cached) in 00:00:27.
All 7 tests passed!
All 3 tests passed!
All 1 tests passed!
All 8 tests passed!
All 2 tests passed!
All 1 tests passed!
All 21 tests passed!
All 13 tests passed!
All 2219 tests passed!
All 1664 tests passed!
All 1 tests passed!
All 58 tests passed!
All 2 tests passed!
All 1 tests passed!
All 15 tests passed!
Traceback (most recent call last):
  File "../test/integration/runner.py", line 21, in <module>
    test_save_mini.TestSaveMiniState)
AttributeError: module 'test_save_mini' has no attribute 'TestSaveMiniState'
make[5]: *** [Makefile:270: do-test] Error 1
make[4]: *** [Makefile:249: test] Error 2
make[3]: *** [hphp/hack/CMakeFiles/hack_test.dir/build.make:57: hphp/hack/CMakeFiles/hack_test] Error 2
make[2]: *** [CMakeFiles/Makefile2:1338: hphp/hack/CMakeFiles/hack_test.dir/all] Error 2
make[1]: *** [CMakeFiles/Makefile2:1345: hphp/hack/CMakeFiles/hack_test.dir/rule] Error 2
make: *** [Makefile:174: hphp/hack/CMakeFiles/hack_test.dir/rule] Error 2
```

I don't know how to fix that, though. `hphp/hack/test/integration/test_save_mini.py` just doesn't export a symbol called `TestSaveMiniState`.

aorenste:
@jesseschalken I'm not sure what's going on there - @dlreeves can you advise?

jesseschalken:
The `TestSaveMiniState` class was removed by @int3 in https://github.com/facebook/hhvm/commit/3fe6e59, but the reference [here](https://github.com/facebook/hhvm/blob/3fe6e59/hphp/hack/test/integration/runner.py#L21) was not updated. I can't see how `make hack_test` could possibly have worked since then. Does it work for anyone else?

jesseschalken:
I've split the error I get from `make hack_test` into #7465

mofarrell:
If you are still interested in landing this I would be happy to help get it landed.  I don't think `make hack_test` has worked for a little while now.  Updating the references in `test/integration/runner.py` to the new test classes should make it work.  All the tests should pass, as we are running them internally using a different 'runner'.

jesseschalken:
> If you are still interested in landing this I would be happy to help get it landed. I don't think `make hack_test` has worked for a little while now. Updating the references in `est/integration/runner.py` to the new test classes should make it work. All the tests should pass, as we are running them internally using a different 'runner'.

If you have a patch for me, great, but I really don't understand this test system enough to fix it myself.

>It would still be nice to see a test case.

Here's one, if you can get the suite to run.

```php
<?hh
echo gettype("$a");
```
```php
<?php
require_once ($GLOBALS["HACKLIB_ROOT"]);
echo (gettype((string) $a));
```

mofarrell:
:(

PR #7435, opened 2016-10-20
Fix getimagesize returning FALSE on valid jpg

MatmaRex:
getimagesize() is rather strict about the length of the marker payload data,
and fails if there are extraneous bytes before the next marker. Only a very
special case reported in bug #13213 is catered to.

libjpeg is rather resilient to such corrupted JPEG files, and raises a
recoverable error in this case. Other image processors also accept such
JPEG files, so we adapt getimagesize() to skip (but warn about) such
extraneous bytes.

(This is the same fix as PHP commit https://github.com/php/php-src/commit/82df4e263886a0da21a00c98189649287666ba5c
by Christoph M. Becker <cmbecker69@gmx.de> fixing PHP [bug 72278](https://bugs.php.net/bug.php?id=72278).)


MatmaRex:
I would appreciate if this could be backported to 3.12. Downstream issue: https://phabricator.wikimedia.org/T148606

aorenste:
It looks like a few tests are failing.

MatmaRex:
Ugh, I'm guessing the exact output of the warning differs between PHP and HHVM… I was hoping it'd just work the first time. It doesn't look like the build output here includes the expected/actual values which are failing to match, just which tests are failing? Unless someone else looks into it, I'll set up a local dev environment next week and figure this out.

MatmaRex:
Should be good now. Sorry.

MatmaRex:
Will this change be backported to 3.12? Should I submit a backport myself? (Should I submit one for 3.15 too?)

mofarrell:
We should backport to 3.12 and 3.15 at our next release.

PR #7437, opened 2016-10-23
Fix http_build_query's numeric_prefix for php7

### LINKED ISSUE #7431: http_build_query numeric_prefix php7
### HHVM Version
Compiler: tags/HHVM-3.15.2-0-g83ac3e5e3f5657be0cf4c55884044f86a7818b90
Repo schema: 608339137764e8365964a1adaa7a27d125b6076f

```php
<?php
$query = array(
	'version'            => "4.6.1",
	'php'                => "7.0.99-hhvm",
	'locale'             => "de_DE",
	'mysql'              => "5.5.5",
	'local_package'      => "de_DE",
	'blogs'              => 1,
	'users'              => 5,
	'multisite_enabled'  => 0,
	'initial_db_version' => 27916,
);
var_dump(http_build_query($query, null, "&"));
```

### Expected result
string(135) "version=4.6.1&php=7.0.99-hhvm&locale=de_DE&mysql=5.5.5&local_package=de_DE&blogs=1&users=5&multisite_enabled=0&initial_db_version=27916"

### Actual result
NULL

If the second argument of `http_build_query` is `null` it returns `null`. This happens only for php7 mode, works fine for php5. I would make a pull request, but since it is related only to php7 and  as it seems, not to the function implementation itself, I have no idea where to look for the bug.

belka-ew:
It should fix #7431.

### DIFF
--- a/hphp/runtime/ext/url/ext_url.cpp
+++ b/hphp/runtime/ext/url/ext_url.cpp
@@ -205,7 +205,7 @@ static void url_encode_array(StringBuffer &ret, const Variant& varr,
 const StaticString s_arg_separator_output("arg_separator.output");
 
 Variant HHVM_FUNCTION(http_build_query, const Variant& formdata,
-                           const String& numeric_prefix /* = null_string */,
+                           const Variant& numeric_prefix /* = null_string */,
                            const String& arg_separator /* = null_string */,
                            int enc_type /* = k_PHP_QUERY_RFC1738 */) {
   if (!formdata.isArray() && !formdata.is(KindOfObject)) {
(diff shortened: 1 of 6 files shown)


hphp/runtime/ext/url/ext_url.cpp
```diff
@@ -222,8 +222,14 @@ Variant HHVM_FUNCTION(http_build_query, const Variant& formdata,
 
   StringBuffer ret(1024);
   std::set<void*> seen_arrs;
+
+  String num_prefix;
+  if (!numeric_prefix.isNull())
+  {
```
Orvid: Minor style nit: This brace should be on the previous line.

belka-ew: I see now it was merged. Should I fix it or is it too late?

Orvid: @aorenste Do you wanna do the fixup internally?

aorenste: @Orvid @belka-ew Since I missed it on the review I'll take care of it.

PR #7449, opened 2016-10-28
Fix building on Gentoo Linux

kristapsk:
Gentoo stores `libpq-fe.h` in `/usr/include` not `/usr/include/postgresql`. Check for `pg_config.h` instead.

Maybe better way would be to check `pg_config --includedir`, but I'm not sure isn't it possible to have headers and libraries without PostgreSQL tools on other distributions / OSes.

I think this should also be merged in HHVM-3.15 tree, as currently (since 3.15.0) building on Gentoo systems is broken by default (but, as a workaround, could be patched in ebuilds).


Orvid:
cc @simonwelsh

simonwelsh:
Does this work if `libpq-fe.h` is in a separate folder from `pg_config.h` and that folder isn't already included? (e.g. `pg_config.h` is in `/usr/include` and `libpq-fe.h` is in `/usr/include/postgresql`)

`pg_config.h` isn't actually used by HHVM, so searching for it over the file that is actually included seems counter intuitive.

kristapsk:
It is used, got compile error on that one, that's why I switched to it.

https://github.com/facebook/hhvm/blob/master/hphp/runtime/ext/pgsql/pdo_pgsql_connection.cpp#L24

aorenste:
Something is strange about the update - it modifies 5000 lines which seems odd.

kristapsk:
I just updated our master branch from facebook's master, forgot about this PR being active. In reality it should not make a big difference I think.

mofarrell:
@simonwelsh Do you have any remaining issues with the pull request?

simonwelsh:
@mofarrell my concern still stands. If Gentoo can split up the include files in this way, it means we can't rely on `libpq-fe.h` and `pg_config.h` being in the same folder. Will need to change the searching to look for both files and add both resulting folders to the include dirs.

mofarrell:
@simonwelsh Seems valid.  @kristapsk Could we also put a check in for`libpq-fe.h` include dir?  I know it will probably either be with `pg_config.h` or in `/usr/include`, but to be conservative we might as well check.  Its easy to imagine another scenario where `pg_config.h` is in `/usr/include`, and `libpq-fe.h` is in `/usr/include/postgresql`.  This would break that system that is currently working.

kristapsk:
I will try to work on a different solution. Need to do some research about CMake (I'm not guru there), but have other priorities currently. Actually Gentoo does not split include files, it just creates `libpq-fe.h` symlink in `/usr/include` and because of that `PGSQL_INCLUDE_DIR` is set to `/usr/include`, which is wrong, as it have `libpg-fe.h` symlink, but does not have symlinks to other files needed, like `pg_config.h`.

mofarrell:
Im going to go ahead and close this pr for now.  @kristapsk when you get to this please comment here or open a new pull request.

PR #7455, opened 2016-10-31
Fix sign ext/mov vasms for PPC64.

racardoso:
Summary:

This commit fixes move and (signal) extensions vasms for PPC64. Some vasms
that needs to move values less than 64 bits to 64 bits registers. Those vasms were 
translated to pairs that mov and sign extend (rlwinm + exts*) that are used even 
when a sign extended are not intended like movz* or bitwise (xor*, and*) 
which can lead to a wrong behaviour or emit a extra useless sign extend/mov instruction.

This commit also fixes some lowers that uses the same VReg in instructions 
with operands that have different. Some lowered vasms are using movb and use 
the result (VReg8) in operations that use VReg64 as input operands when we should 
use movzbq in those cases. 

To summarize this fix:

- Delete some redundant/wrong move instructions on sign extension instructions.
- Fix the problem with replacement by a nop when mov{r,r}. Some lowering instructions 
like was using movb instructions, remove the instruction lead to invalid values 
on registers. Replaced by movzbq.
- Replace movb/ext by instruction versions that have operands with correctly 
width avoiding use the same Vreg in instructions that expect operands with different 
widths.
- Refactor some of X lower macros.
- Avoid lower instructions that can be emitted directly.
- Kill two ppc64 intrinsic unused vasms extrb and extrw.
- Vasm instruction extsw was changed to extsl. Using 'l' suffix to maintain name 
convention. Change first operand to VReg32 as we expect a 32 bit VReg as src 
operand.
- Remove some weird calculations for mask size when map mov* instructions to
  ppc64 asm.
- Fix extsw, extsh on ppc assembler to include Rc parameter even if we don't emit 
Rc (CR record) version of those instructions.

### DIFF
--- a/hphp/runtime/vm/jit/vasm-instr.h
+++ b/hphp/runtime/vm/jit/vasm-instr.h
@@ -316,8 +316,6 @@ struct Vunit;
   O(subsb, Inone, UA(s0) U(s1), D(d) D(sf))\
   O(uxth, Inone, U(s), D(d))\
   /* ppc64 instructions */\
-  O(extrb, Inone, UH(s,d), DH(d,s))\
-  O(extrw, Inone, UH(s,d), DH(d,s))\
   O(extsb, Inone, UH(s,d), DH(d,s))\
   O(extsw, Inone, UH(s,d), DH(d,s))\
   O(fcmpo, Inone, U(s0) U(s1), D(sf))\
@@ -1156,10 +1154,8 @@ struct uxth { Vreg16 s; Vreg32 d; };
 /*
  * ppc64 intrinsics.
  */
-struct extrb { Vreg8 s; Vreg8 d; };   // Extract and zeros the upper bits
-struct extrw { Vreg16 s; Vreg64 d; }; // Extract and zeros the upper bits
-struct extsb { Vreg64 s; Vreg64 d; }; // Extend byte sign
-struct extsw { Vreg64 s; Vreg64 d; }; // Extend word sign
+struct extsb { Vreg8 s; Vreg64 d; };  // Extend byte sign
+struct extsw { Vreg32 s; Vreg64 d; }; // Extend word sign
 struct fcmpo { VregDbl s0; VregDbl s1; VregSF sf; };
 struct fcmpu { VregDbl s0; VregDbl s1; VregSF sf; };
 struct fctidz { VregDbl s; VregDbl d; VregSF sf; };
--- a/hphp/runtime/vm/jit/vasm-instr.h
+++ b/hphp/runtime/vm/jit/vasm-instr.h
@@ -317,7 +317,7 @@ struct Vunit;
   O(uxth, Inone, U(s), D(d))\
   /* ppc64 instructions */\
   O(extsb, Inone, UH(s,d), DH(d,s))\
-  O(extsw, Inone, UH(s,d), DH(d,s))\
+  O(extsl, Inone, UH(s,d), DH(d,s))\
   O(fcmpo, Inone, U(s0) U(s1), D(sf))\
   O(fcmpu, Inone, U(s0) U(s1), D(sf))\
   O(fctidz, Inone, U(s), D(d) D(sf))\
@@ -1155,7 +1155,7 @@ struct uxth { Vreg16 s; Vreg32 d; };
  * ppc64 intrinsics.
  */
 struct extsb { Vreg8 s; Vreg64 d; };  // Extend byte sign
-struct extsw { Vreg32 s; Vreg64 d; }; // Extend word sign
+struct extsl { Vreg32 s; Vreg64 d; }; // Extend word sign
 struct fcmpo { VregDbl s0; VregDbl s1; VregSF sf; };
 struct fcmpu { VregDbl s0; VregDbl s1; VregSF sf; };
 struct fctidz { VregDbl s; VregDbl d; VregSF sf; };
--- a/hphp/runtime/vm/jit/vasm-ppc64.cpp
+++ b/hphp/runtime/vm/jit/vasm-ppc64.cpp
@@ -168,6 +168,14 @@ struct Vgen {
     copyCR0toCR1(a, rAsm);
   }
   void emit(const addsd& i) { a.fadd(i.d, i.s0, i.s1); }
+  void emit(const andb& i) {
+    a.and(Reg64(i.d), Reg64(i.s0), Reg64(i.s1), true);
+    copyCR0toCR1(a, rAsm);
+  }
+  void emit(const andl& i) {
+    a.and(Reg64(i.d), Reg64(i.s0), Reg64(i.s1), true);
+    copyCR0toCR1(a, rAsm);
+  }
   void emit(const andq& i) {
     a.and(i.d, i.s0, i.s1, true);
     copyCR0toCR1(a, rAsm);
@@ -202,18 +210,8 @@ struct Vgen {
   }
   void emit(const divint& i) { a.divd(i.d,  i.s0, i.s1, false); }
   void emit(const divsd& i) { a.fdiv(i.d, i.s1, i.s0); }
-  void emit(const extrb& i ) {
-    uint8_t sh = 8;
-    a.rlwinm(Reg64(i.d), Reg64(i.s), 0, 32-sh, 31); // extract lower byte
-    a.extsb(Reg64(i.d), Reg64(i.d));                // extend sign
-  }
-  void emit(const extrw& i ) {
-    uint8_t sh = 16;
-    a.rlwinm(i.d, Reg64(i.s), 0, 32-sh, 31); // extract lower 16 bits
-    a.extsh(i.d, i.d);                      // extend sign
-  }
-  void emit(const extsb& i) { a.extsb(i.d, i.s); }
-  void emit(const extsw& i) { a.extsw(i.d, i.s); }
+  void emit(const extsb& i) { a.extsb(i.d, Reg64(i.s)); }
+  void emit(const extsw& i) { a.extsw(i.d, Reg64(i.s)); }
   void emit(const fabs& i) { a.fabs(i.d, i.s, false); }
   void emit(const fallthru& i) {}
   void emit(const fcmpo& i) {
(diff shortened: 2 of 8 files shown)


racardoso:
Is this  `fb/hhvm-cmake — Build Failed` a infra error?

aorenste:
@ricardoso - Not infra but it looks like someone made a bad change internally so you can ignore it.

racardoso:
@aorenste ok thanks.

racardoso:
Any news on this pull request?

racardoso:
@mxw can you please review this?

hphp/runtime/vm/jit/vasm-util.cpp
```diff
@@ -29,7 +29,12 @@ bool is_nop(const copy& i) { return i.s == i.d; }
 bool is_nop(const copy2& i) { return i.s0 == i.d0 && i.s1 == i.d1; }
 
 // movb r,r is a nop, however movl is not since it zeros upper bits.
-bool is_nop(const movb& i) { return i.s == i.d; }
+bool is_nop(const movb& i) {
+  // On PPC64 movb r,r cannot be considered a trivial nop since it zeros
```
mxw: I would say that `movb{r, r}` is a trivial nop by definition (also, note that this is only possible when both src and dst regs are physical).  The destination is being interpreted as a Reg8, which means its upper bits are logically garbage after the `movb{}`.  Anywhere that you require the zeroing behavior you should do an explicit `movzbq{}` (or similar).

racardoso: The problem is ppc64 doesn't move 8 bits like x86_64. The destiny is always a 64 bit register. A `movb` is translated to a ppc64 `rlwinm` instruction who rotates and mask and it's the proper way to move 8 (or more) bits between registers on POWER. The problem is we use `movb` on some lowered vasms like `cmpbm`, `xorb`, `andb` and we expect that `movb` will not be removed even if r(dst) = r(src). We use `movb` on `mozbl` and `movzbq` as we always zero upper bits on `movb` and the destiny is always a 64 bits register anyway. Unfortunately HHVM doesn't have a `movzb` vasm i think that not consider a `movb` a trivial nop on ppc64 is easiest than create a new vasm `movzb`. What do you think? should i create a new vasm `movzb` or you have another suggestion?

mxw: Why does it matter what the destination is?  It doesn't matter what x64 or ppc64 does when we're talking about implementing a vasm instruction.  In vasm, the destination of a `movb{}` is a Vreg8, which means that a `movb{}` makes no guarantees about the upper bits of its destination (assuming the Vreg8 is backed by a regular 64-bit physical register).

If you need to copy 8 bits and zero all the upper bits, you need to lower to `movzbq{s, tmp}; movtqb{tmp, d}` instead of to `movb{s, d}`.  Is that what you want?

racardoso: > If you need to copy 8 bits and zero all the upper bits, you need to lower to movzbq{s, tmp}; movtqb{tmp, d} instead of to movb{s, d}. Is that what you want?

We are lowering those instructions to `movb` because on our implementation we zero's all upper bits on mov. I guess I can lower `movb` to `movzbq` instead.

The real problem is in some lowers we do. Let me show a example with our `xorb`:

``` 
void lowerForPPC64(Vout& v, xorb& inst) { 
  Vreg tmp1 = v.makeReg(), tmp2 = v.makeReg();
  v << movb{inst.s0, tmp1};  
  v << movb{inst.s1, tmp2}; 
  v << xorq{tmp1, tmp2, Reg64(inst.sf)}; 
}
```
The problem is if we have `r(dst) = r(src)` those `movb` will be replaced by a `nop` ending up with wrong values on registers when `xorq` is executed. I am thinking here why we really need those moves.

lbianc: I'm not sure if this is the case, but remember that in PPC64 there is no register of 8 bits, so if it's moving a value into a 8 bit register, later this register would be used as 8 bit register. The issue is that if I move a 8 bit value to a 8 bit register, and later use this register in any other instruction, on PPC64 it will consider the entire register, considering the garbage of the higher bits.
@racardoso check if this is the case.

racardoso: @lbianc I'll check if there any case when we have problem with this garbage. Thank you.

mxw: @racardoso—The output of that lowering is ill-formed vasm (we just happen not to run the `checkWidths()` pass after that point, I guess?).  By using `movb{}`, you've constrained `tmp1` and `tmp2` to be Vreg8's.  You then attempt to use them as Vreg64's, which is not allowed.

racardoso: @mxw Yes in fact I am doing some refactor here. I remove those `movb`instructions but what about `subbi` `testbi` `cmpbi` instructions I can't have garbage on the upper bits of register in this case as we are doing those operations with 64 bits. I don't know if i understand correctly but you say i can't `movb` to tmp registers and use it as Vreg64?

mxw: @racardoso—The specification for _every_ vasm instruction that has a Vreg8 src is that the upper bits of that src are considered killed when we hit the instruction (before it's executed).  For a Vreg8 dst, the upper bits are not well-defined once the instruction finished executed.

So, `movb{r, r}` takes `r := [56 bits of anything][8 defined bits]` and redefines it as `r := [56 bits of anything][the same 8 defined bits]`.  This is clearly a no-op.  It doesn't matter how you implement `movb{}` under the hood; by its vasm specification, a `movb{r, r}` is a no-op.

And you understood me correctly.  More generally, you can never use a Vreg defined with one width (e.g., Vreg8) in an instruction that expects a Vreg of a different width (e.g., Vreg64).  So, the result of any `b` instruction can only be used in instructions which take a Vreg8 (or an unconstrained Vreg).  If you want to use it for, e.g., an `addq` later or something, you need to do an intermediate `movzbq{}`.  Vreg8 and Vreg64 are not compatible types—much like how you can't use a `struct foo {};` as a `struct bar{};`, you cannot define a Vreg8 and then use it as a Vreg64.

racardoso: @mxw This specification is just for Vreg8 instructions right? Others sizes < than 8 bits like a `movl` and `movw` it zero's upper bits? Seems to me that `movl`always zero.

Yes it's a nop. Our problem before, like I show, is that we was incorrectly using it on lowered instructions result on invalid values once the instructions is changed to a nop.

I misinterpreted what you say. But it's clear now. I will update this PR. Thank you for the clarification.

racardoso: @mxw  Got it. I'll check our vasms to see if we are doing anything else like that and fix it.

racardoso:
@mxw fixed.

hphp/runtime/vm/jit/vasm-util.cpp
```diff
@@ -29,7 +29,9 @@ bool is_nop(const copy& i) { return i.s == i.d; }
 bool is_nop(const copy2& i) { return i.s0 == i.d0 && i.s1 == i.d1; }
 
 // movb r,r is a nop, however movl is not since it zeros upper bits.
-bool is_nop(const movb& i) { return i.s == i.d; }
+bool is_nop(const movb& i) {
```
mxw: Undo this change.

hphp/runtime/vm/jit/vasm-ppc64.cpp
```diff
@@ -944,213 +946,187 @@ void lowerForPPC64(Vout& v, Inst& inst) {}
...
- * "0xSSSSSSSXXXXXXXX".
- *
- * For operands the treatment is analogous, only the lower 32bits are
- * considered to perform the operation.
- *
- * PS: the core vasm of these lowerings may be further lowered, but as the
- * first one is a vasm that is not lowered (movb), there will be no issues as
- * only the first emitted vasm is not lowered on "lowerForPPC64(Vunit&)"
- */
+#undef X*/
```
mxw: Delete this.

hphp/runtime/vm/jit/vasm-instr.h
```diff
@@ -1156,10 +1154,8 @@ struct uxth { Vreg16 s; Vreg32 d; };
 /*
  * ppc64 intrinsics.
  */
-struct extrb { Vreg8 s; Vreg8 d; };   // Extract and zeros the upper bits
-struct extrw { Vreg16 s; Vreg64 d; }; // Extract and zeros the upper bits
-struct extsb { Vreg64 s; Vreg64 d; }; // Extend byte sign
-struct extsw { Vreg64 s; Vreg64 d; }; // Extend word sign
+struct extsb { Vreg8 s; Vreg64 d; };  // Extend byte sign
+struct extsw { Vreg32 s; Vreg64 d; }; // Extend word sign
```
mxw: This should either be called `extsl{}` (extend long sign), or you should change the `Vreg32` operand to `Vreg16`.

racardoso: @mxw  on ppc64 w = 32 bits not 16 as in x86_64. This is a ppc64 intrinsic. Should i maintain the suffix convention on this case?

hphp/runtime/vm/jit/vasm-ppc64.cpp
```diff
@@ -202,17 +210,8 @@ struct Vgen {
...
-    a.extsb(Reg64(i.d), Reg64(i.d));                // extend sign
-  }
-  void emit(const extrw& i ) {
-    int8_t sh = 16;
-    a.rlwinm(i.d, Reg64(i.s), 0, 32-sh, 31); // extract lower 16 bits
-  }
-  void emit(const extsb& i) { a.extsb(i.d, i.s); }
-  void emit(const extsw& i) { a.extsw(i.d, i.s); }
+  void emit(const extsb& i) { a.extsb(i.d, Reg64(i.s)); }
+  void emit(const extsl& i) { a.extsw(i.d, Reg64(i.s)); }
```
mxw: ...does PPC64 use "word" in place of "long"?

racardoso: Yes. `extsw`is a PPC64 instruction to sign extend 32 bits...

racardoso: Summary updated

PR #7457, opened 2016-10-31
tiny language fixes

rulex:


mofarrell:
@rulex All of the zend test are imported with scripts like https://github.com/facebook/hhvm/blob/10b4a1aeb38979d8632ab7496ddab390777f6b2d/hphp/test/tools/import_zend_test.py
It is probably not worth fixing the grammar in those ones.  The other changes look fine though

mofarrell:
@rules Please open a new pull request if you end up getting to this.

PR #7458, opened 2016-10-31
[Aarch64] Remove -fsigned-char flag and fix resulting failure

jim-saxman:
Remove the -fsigned-char flag for Aarch64 systems. As a result,
the unit test zend/good/ext/fileinfo/tests/finfo_file_002.php
began failing. Thus, this patch changes a cast to (signed char),
and the unit test passes again.

### DIFF
--- a/CMake/HPHPCompiler.cmake
+++ b/CMake/HPHPCompiler.cmake
@@ -183,9 +183,6 @@ if (${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang" OR ${CMAKE_CXX_COMPILER_ID} STREQU
 
     # ARM64
     if(IS_AARCH64)
-      # Force char type to be signed, which is not the case on aarch64.
-      list(APPEND GENERAL_OPTIONS "fsigned-char")
-
       # If a CPU was specified, build a -mcpu option for the compiler.
       set(CPU "" CACHE STRING "CPU to tell gcc to optimize for (-mcpu)")
       if(CPU)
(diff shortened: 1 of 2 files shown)


mxw:
What is the higher-level goal of this change?

jim-saxman:
Patch #7361 introduced the -fsigned-char flag and has the backstory for why I am removing it.

mxw:
Hmmm, I don't think I ever saw that one.  Looping in @dave-estes and @apinski-cavium.  Have you guys done any more testing w.r.t. signed chars since back then?

As I just commented on the previous PR, the concern about mismatched ABIs is pretty compelling to me, and I think we should make this change; I'm just curious if anyone else has had any problems with it since then.

dave-estes:
We've been running with and without signed chars for a while. In light of @apinski-cavium concerns, I think we should go ahead with this change to revert back to signed chars for ARM. If there are any latent bugs in HHVM, then we'll just have to find them.

dave-estes:
@mxw, is there anything else needed to land this PR?

mxw:
@dave-estes—I need to do some more testing internally.  The CMake build doesn't affect us, but our internal tests stress the system more.

Sorry for putting this off; it didn't seem to be blocking other changes, so I've been punting on it.  I'll look into it this week.

jim-saxman:
@mxw ping.... This patch still applies cleanly for me, can we get it merged?

cmuellner:
Tested here as well -> no regressions.

Orvid:
*thought this had landed months ago...*

jim-saxman:
@Orvid Me too! I can give it another round of testing on AArch64 if needed.

jim-saxman:
This patch is still clean for AArch64 tests. There are no new failures for test/all with {"","--mode interp","--mode pgo","-r --mode pgo"}.

PR #7459, opened 2016-10-31
[Aarch64] Change CPU to be more specific

jim-saxman:
Renames CPU to AARCH64_TARGET_CPU to make its usage more specific.
Additionally, move the erratum disabling inside the check. This
fixes the issue when no AARCH64_TARGET_CPU is given on the command
line, the erratum was accidentally disabled.


mxw:
Is this CPU logic not used by any other platforms?

Orvid:
It's only used by the AARCH64 config at the moment, yes.

PR #7460, opened 2016-10-31
[Aarch64] Add etch helpers to aid assembly routines

jim-saxman:
Add the etch helpers for aarch64 targets. This patch will aid in
writing assembly routines for CRC hash routines.

### DIFF
--- a/hphp/util/etch-helpers.h
+++ b/hphp/util/etch-helpers.h
@@ -68,10 +68,6 @@
 #define ETCH_ARG4       %x3
 #define ETCH_ARG5       %x4
 #define ETCH_ARG6       %x5
-#define ETCH_ARG7       %x6
-#define ETCH_ARG8       %x7
-#define ETCH_GET_ARG5   /* not used */
-#define ETCH_GET_ARG6   /* not used */
 
 #else /* Other x86 (e.g. linux) */
 #define CFI(x)            .cfi_##x
--- a/hphp/util/etch-helpers.h
+++ b/hphp/util/etch-helpers.h
@@ -49,6 +49,30 @@
 #define ETCH_GET_ARG6     /* not used */
 #define ETCH_RET1         %rax
 
+#elif defined(__aarch64__)
+#define CFI(x)          .cfi_##x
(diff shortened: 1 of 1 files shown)


hphp/util/etch-helpers.h
```diff
@@ -49,6 +49,30 @@
 #define ETCH_GET_ARG6     /* not used */
 #define ETCH_RET1         %rax
 
+#elif defined(__aarch64__)
+#define CFI(x)          .cfi_##x
+#define CFI2(x, y)      .cfi_##x y
+#define CFI3C(x, y, z)  .cfi_##x y##, z
+#define ETCH_ALIGN16    .align 4
```
apinski-cavium: I think it is better use directly balign or p2align instead of align here.  Then you don't have to worry about figuring out if the operand is a power of 2 or just the byte size.

jim-saxman:
I agree. I will update this patch to eliminate confusion.

apinski-cavium:
Looks good to me.

hphp/util/etch-helpers.h
```diff
@@ -49,6 +49,30 @@
...
+#define ETCH_ARG1       %x0
+#define ETCH_ARG2       %x1
+#define ETCH_ARG3       %x2
+#define ETCH_ARG4       %x3
+#define ETCH_ARG5       %x4
+#define ETCH_ARG6       %x5
+#define ETCH_ARG7       %x6
+#define ETCH_ARG8       %x7
+#define ETCH_GET_ARG5   /* not used */
+#define ETCH_GET_ARG6   /* not used */
```
mxw: These aren't used anywhere anymore, so can you just delete all the definitions instead of adding another dummy?

jim-saxman: Sure. I just pushed a new version that removes these #defines.

jim-saxman:
@mxw Hi, I'm getting close to pushing a new patch for some hash functions which are in Aarch64 assembly and is dependent on this patch. Can we resolve this one first, so that I don't have dependencies between the two?

PR #7480, opened 2016-11-09
remove unneeded assertions (Aarch64)

swalk-cavium:
On a debug build for an Aarch64 host  the following test was failing due to an assertion:

hphp/test/zend/good/ext/intl/tests/grapheme.php

The assertion was not needed since the call to the macro assembler will use a temporary
if the immediate is out of range.

Additionally, two other regression tests now pass.  They are:

hphp/test/slow/stack-overflow/reenter.php
hphp/test/slow/stack-overflow/catch-trace-rematerialize.php

The change was tested with 6 different option sets.

/usr/bin/php hphp/test/run -v --no-fun -m interp  
/usr/bin/php hphp/test/run -v --no-fun -m interp -r 
/usr/bin/php hphp/test/run -v --no-fun -m jit -a '-vEval.JitPGO=0' 
/usr/bin/php hphp/test/run -v --no-fun -m jit -r -a '-vEval.JitPGO=0' 
/usr/bin/php hphp/test/run -v --no-fun -m jit -a '-vEval.JitPGO=1' 
/usr/bin/php hphp/test/run -v --no-fun -m jit -r -a '-vEval.JitPGO=1'  

This was the git log for the repository before commit
swalk@1scrb-7:~/master/hhvm$ git log | head
commit 7e3addb8bc1cf1e3dbbab2de6f1690050ea33b71
Author: Michael O'Farrell <michaelofarrell@fb.com>
Date:   Mon Nov 7 07:04:47 2016 -0800

This was the repository before I pushed the change to github
swalk@1scrb-7:~/master/hhvm_110816/hhvm$ git log | head
commit 8f964f0cc101acdc8e8c5e3d43497a6d41a6b0d0
Author: Steve Walk <Steve.Walk@cavium.com>
Date:   Wed Nov 9 07:08:25 2016 -0800

A CLA was submitted last week.

### DIFF
--- a/hphp/runtime/vm/jit/vasm-arm.cpp
+++ b/hphp/runtime/vm/jit/vasm-arm.cpp
@@ -106,6 +106,7 @@ int64_t MSKTOP(int64_t value) {
 vixl::MemOperand M(Vptr p) {
   assertx(p.base.isValid());
   if (p.index.isValid()) {
+    assertx(p.disp == 0);
     return MemOperand(X(p.base), X(p.index), LSL, Log2(p.scale));
   }
   return MemOperand(X(p.base), p.disp);
--- a/hphp/runtime/vm/jit/vasm-arm.cpp
+++ b/hphp/runtime/vm/jit/vasm-arm.cpp
@@ -106,10 +106,8 @@ int64_t MSKTOP(int64_t value) {
 vixl::MemOperand M(Vptr p) {
   assertx(p.base.isValid());
   if (p.index.isValid()) {
-    assertx(p.disp == 0);
     return MemOperand(X(p.base), X(p.index), LSL, Log2(p.scale));
   }
-  assertx(p.disp >= -256 && p.disp <= 255);
   return MemOperand(X(p.base), p.disp);
 }
 
@@ -744,7 +742,6 @@ void Vgen::emit(const lea& i) {
     assertx(p.disp == 0);
     a->Add(X(i.d), X(p.base), Operand(X(p.index), LSL, Log2(p.scale)));
   } else {
-    assertx(p.disp >= -256 && p.disp <= 255);
     a->Add(X(i.d), X(p.base), p.disp);
   }
 }


swalk-cavium:
Hello facebook-github-bot,
A CLA was submitted on 11-04-2016.  It was the company version if that matters.

swalk-cavium:
How do we get information on the failing tests?

aorenste:
@swalk-cavium For the "fb" tests they're internal so we don't give information out about them - but the failures look like internal tooling issues so they can be ignored.

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -106,10 +106,8 @@ int64_t MSKTOP(int64_t value) {
 vixl::MemOperand M(Vptr p) {
   assertx(p.base.isValid());
   if (p.index.isValid()) {
-    assertx(p.disp == 0);
```
dave-estes: Definitely need this assertion. If the Vptr was lowered and it still has a disp, then we need to know if we're about to throw it away.

swalk-cavium: I'll revisit the change I was probably too quick to remove it.  I think your lowering question might be related to this pull request.
https://github.com/facebook/hhvm/pull/7058

dave-estes:
I ran test/zend/good/ext/intl/tests/grapheme.php with TRACE and saw that the offending `Vptr`s did show up after register allocation, well past the lowerer. I'd say that removing the bounds check for the disp is OK.

BTW, the other test cases that you mention were already passing for me without your changes.

swalk-cavium:
Thanks for checking.  I'm still trying to figure out the environment I was given.  The
build instructions are slightly different than the facebook instructions.  It appears they
are cherry-pick'ing some changes down in the third-party directory.

cmuellner:
I bumped into that issue as well (and was just about to make a PR
with the same changes).

The assertions are wrong. E.g. the add instruction supports an immediate,
which has 12 bits and can be shifted by 12 bits. That's not the range
[-256..+255].

Additionally vixl checks the immediate anyways later (in a correct way)
and throws an assertion if it exceeds the possible range (see IsImmAddSub()).

I was able to reproduce the problem on Ubuntu Xenial when compiling
in Debug mode:

```
FAILED: hphp/test/zend/good/ext/intl/tests/grapheme.php
[...]
14653+ Assertion failure: /mnt/hhvm-arm64/hhvm-debug/hphp/runtime/vm/jit/vasm-arm.cpp:747: void HPHP::jit::{anonymous}::Vgen::emit(const HPHP::jit::lea&): assertion `p.disp >= -256 && p.disp <= 255' failed.
```

@mxw Can we merge this?

mxw:
Should we get rid of the displacement restrictions in vasm-copy, then?  If not this, what is that pass trying to accomplish?

cmuellner:
Good catch @mxw !
I suggest to get rid of that #ifdef'ed code in vasm-copy.

If we ever happen to trigger an assertion from vixl in real life, then we can
insert code to handle too big displacement values in vasm-arm
(by building the value up in a register first using movk instructions
and using the register as operand instead of an immediate (just like
compilers are doing that)). I'm volunteering to implement that, if it is needed.

mxw:
@cmuellner—That sounds like a regression to me.  We shouldn't be making any assumptions about the magnitude of the displacement, since HHIR doesn't give any such constraints.

The purpose of the code in vasm-copy is to do exactly as you said—to build up the displaced value in a register, and then just copy it instead of doing an `lea{}` (or using a Vptr).  If +/-2^8 isn't the correct restriction, and we should be using +/-2^11, we should just do that instead, rather than making the backend less robust to displacement magnitudes.

cmuellner:
@mxw : I will prepare a Pr for that.

mxw:
I think the real issue here is that, sometime after the `lowerForARM()` pass, we're creating Vptr operands that ignore the displacement bounds (that we're apparently using the wrong bounds is an orthogonal issue)—or, we're just accidentally skipping over some Vptrs that we should be lowering.

@swalk-cavium, @dave-estes—Did either of you take a look at how these Vptrs are being created?  Or are there just Vptrs that aren't getting lowered?

swalk-cavium:
Hi Max,
No, I did not.  I'll see if I can find it now.

dave-estes:
Hey @mxw. Did you see my earlier comment?

> I ran test/zend/good/ext/intl/tests/grapheme.php with TRACE and saw that the offending Vptrs did show up after register allocation, well past the lowerer. I'd say that removing the bounds check for the disp is OK.

Also, I think the limit for a displacement should be 2^8 and not 2^11. These MemOperands are for ARM memory instructions, i.e. Loads and stores. Different loads and stores have different sized immed fields. I think @lpathy 's aim here was to support all of them with this assert. So even though a ldr (immediate) has nine bits, let's try to always lower to an eight bit displacement.

I still think it's fine to remove this assert, since indeed VIXL will handle a MemOperand using addrmode = Offset and a offset larger than 8 bit (see `CalcLSDataSize()` as it's used in `LoadStore()`). In a sense, VIXL is "lowering" the MemOperand for us, but it's always best to try to lower explicitly earlier while we're still in vasm and able to optimize.

swalk-cavium:
The unconditioned lea instruction is being emitted during vasm-xls
which is after lowerForARM().  The first offset it would have previously asserted
on was -640  or - 0x280.

1118647:--------------------------------------------------------------------------------
1151443:--------------------------------after vasm-copy---------------------------------
1151554: B881    main, B878
1173409:--------------------------------------------------------------------------------
1173411:--------------------------------after vasm-dead---------------------------------
1173522: B881    main, B878
1195293:--------------------------------------------------------------------------------
1195295:---------------------------------after vasm-xls---------------------------------
1195406: B881    main, B878
1195408:      lea        [x32 - 0x280] => x32               # (38) StStk<IRSPOff -1> t1:StkPtr, t9:Arr
1218209:--------------------------------------------------------------------------------
1218211:--------------------------------after vasm-jumps--------------------------------
1218310: B881    main, B878
1218312:      lea        [x32 - 0x280] => x32               # (38) StStk<IRSPOff -1> t1:StkPtr, t9:Arr


TRACE="vasm:6,vasm_copy:6" ../hhvm_111416/hhvm/hphp/hhvm/hhvm --file 'grapheme.php' 

egrep -n 'B881    main, B878|^----| - 0x280' /tmp/hphp.log

mxw:
Is `x32` the native stack pointer (i.e., `rsp()`)?  If so, I'm guessing this is us trying to create stack space—whatever limits we want to impose on `lea{}`s, we should probably do it there, too.

swalk-cavium:
Not sure that's right.  It's creating the space in the target independent layer.
We should be forcing target issues down, not pulling them up.  See allocateSpillSpace()
in vasm-xls.cpp.  Without the assert the assembler generates correct code.

mxw:
Right—but my point is that we only have correct code because it just so happens that the amount of spill space we're creating is less than 2^11 bytes.  Unless we want to add an additional lowering pass between the spill space creation and the register-renaming phases of XLS (which seems unnecessarily cumbersome—we'd only be doing that to support this one ARM opcode detail), it's probably easiest just to manually lower the `lea{}` that we're inserting for spill space allocation and deallocation.

mxw:
@dave-estes—But is there any limit on the size of the displacement at all?  I think on x64, it can be any valid signed 32-bit integer.  If the same is not true on ARM, then you _must_ make some explicit check, and fall back to the enregistered case accordingly.

dave-estes:
@mxw, there is a limit if the load/store is performed in a single instruction, and that depends on the type of load/store. VIXL's macro assembler will get around this if the MemOperand has a displacement that can't be represented in a single instruction by using two. From hphp/vixl/a64/macro-assembler-a64.cc:519.

    void MacroAssembler::LoadStoreMacro(const CPURegister& rt,
                                        const MemOperand& addr,
                                        LoadStoreOp op) {
      int64_t offset = addr.offset();
      LSDataSize size = CalcLSDataSize(op);
    
      // Check if an immediate offset fits in the immediate field of the
      // appropriate instruction. If not, emit two instructions to perform
      // the operation.
      if (addr.IsImmediateOffset() && !IsImmLSScaled(offset, size) &&
          !IsImmLSUnscaled(offset)) {
        // Immediate offset that can't be encoded using unsigned or unscaled
        // addressing modes.
        Register temp = AppropriateTempFor(addr.base());
        Mov(temp, addr.offset());
        LoadStore(rt, MemOperand(addr.base(), temp), op);
      } else if (addr.IsPostIndex() && !IsImmLSUnscaled(offset)) {
        // Post-index beyond unscaled addressing range.
        LoadStore(rt, MemOperand(addr.base()), op);
        Add(addr.base(), addr.base(), Operand(offset));
      } else if (addr.IsPreIndex() && !IsImmLSUnscaled(offset)) {
        // Pre-index beyond unscaled addressing range.
        Add(addr.base(), addr.base(), Operand(offset));
        LoadStore(rt, MemOperand(addr.base()), op);
      } else {
        // Encodable in one load/store instruction.
        LoadStore(rt, addr, op);
      }
    }

cmuellner:
@dave-estes, I think it's worth mentioning here, that Mov(temp, addr.offset()) is again a
MacroAssember function, which allows 64-bit immediates (for 64-bit destination registers).
It also minimises the emitted mov instructions (to build up the offset in a register).

dave-estes:
Wanted to add another point to consider. I'm currently looking at relocation for AArch64. It involves changing any PC relative instructions to update their displacements whenever A) the instruction moves or B) the target moves. It's still early days, but I think I've got a short list of AArch64 instructions that need to be considered. MOV/MOVK is not on that list currently, and I really hope it never makes the list as it is bound to be a pain to search for these MOV/MOVKs that build up displacements and to update them with new displacements. If the macroassembler starts spitting out MOV/MOVK for situations like this, then I might consider changing it to instead load literals from the instruction stream similar to smashables. It should also perform better, but I'll need input from other ARM platforms out there to confirm that.

swalk-cavium:
Is there a general place where we talk about future optims?  The facebook page perhaps?
I have a couple specializations I'd like to socialize.

Regarding above, I think building up the immediates down in the macro assembler would
be difficult to relocate.  Especially if the new address needed a 3 instruction sequence when
a 2 instruction one was generated.

cmuellner:
@dave-estes vixl uses the movk approach for non-fitting immediates for many instructions
(e.g. for the add instruction - see https://github.com/facebook/hhvm/blob/master/hphp/vixl/a64/macro-assembler-a64.cc#L427).
Calling vixl in a way to not touch this code paths might be counter-intuitive to use,
so I would prefer patching vixl as well (at least to throw assertions instead of creating MOVKs).
What would be the alternative to MOVK, using a vixl::Label (like in Vgen::emit(const jmp& i))?
Then I am concerned about cache misses when loading the offset.

@swalk-cavium I vote for email communication (@ptomsich is not on FB) and a wiki for negotiated goals/plans/responsibilities.

@ptomsich You might want to have a look at this discussion as well.

mxw:
@dave-estes—Ah, I see.  So you're saying that if the displacement is out-of-bounds, then the macro assembler will do the right thing with it?  If that's the case, then it sounds like we don't need _any_ displacement constraining or assertions—is that right?

@swalk-cavium, @cmuellner—The workflow I'd prescribe is to use the FB page for general discussion, IRC for ad hoc discussion, and GitHub issues for delegation (you can create umbrella issues as necessary).  Making an FB account is trivial, and you don't have to use it for anything but HHVM-on-Aarch64 work if you don't want :]

mxw:
Okay, it sounds like we should either repurpose this diff to delete all references to this [-256, 255] bound on displacements, or we should tell me how dense I'm being and explain to me why I'm wrong about that.

dave-estes:
Not trying to be a stick in the mud, but my preference for general discussion is also an email list. While I did create a FB account just so that I could communicate on the HHVM AArch64 group, but I don't use it often enough, so I miss things a lot. IRC and Github for the other chores suits me well.

dave-estes:
@mxw, My vote is to just remove the assert for now and to still strive to lower MemOperands so that the displacement is within a reasonable bounds (2^8 is reasonable). Lowering is my preference because it allows for later vasm-level optimizations like CSE for example.

If we really wanted to go nuts, then we could introduce special lowering that takes into account the vasm where the MemOperand is being used. We could use arch-specific knowledge of the imm width available for that instruction.

swalk-cavium:
I suspect that the various calls to lowerVptr were added to select a lowest common denominator immediate offset addressing mode.  Presumably so that any extra ldimmq's emitted (see 
lowerVptr() vasm-arm.cpp) could be used in optimization.

The root problem is that the Vptr does not know what the type of the thing it's pointing to
so it has to be careful.

The lea's in regalloc are introduced much closer to emission, so it's better for the assembler to pick up the few cases. 

I would propose just removing the assert for now and figuring out how to make a more general solution.

cmuellner:
I would also propose to remove the assertions.

Additionally I'd like to see a more instruction sensitive lowering of Vptrs,
but that is independent of this PR.

ptomsich:
Generally, a load will always be worse than a movz/movk sequence.
There's a few things to consider:
- some cores (e.g. ARM's own implementations) implement macro-fusion, which optimises part of the sequence during decoding and generates fewer u-ops from the sequence
- a load has the chance of missing through the inner cache hierarchies
- the result of loading a displacement constant will usually be used by a dependent load, which usually incurs a bigger penalty in u-arch implementation (either in terms of runtime-impact for the code or implementation-complexity in silicon) 
- the target u-archs will usually be very efficient (usually just taking a hit of a "fractional cycle" of decode-bandwidth) in getting rid of nops that might be left over, if we don't fill all 4 instruction slots in such a sequence

That said, relocation will most likely require the reservation of 4 insn slots that allow us to load up an arbitrary 64bit offset... and the code that patches things back to normal can then chose any optimal sequence that is (potentially) shorter than the 4 insns.

However, if we don't consider relocation we could always emit the optimal displacement and keep our code sequences compact. I don't agree with @cmuellner that vixl is where we want to do so—why would we want to build up the displacement as late as in the emission-stage? Unless we absolutely don't know the displacement (or an upper bound) earlier in the lowering, so we can build it up during lowering (and/or optimise during a lower optimisation pass).  I am assuming here that there will be some sort of simple instruction scheduling implemented at some point in the future... which leaves me very wary of anything that expands into multiple, sequential (i.e. no opportunity to schedule something into the gaps) instructions very late in the code-generation.

mxw:
@dave-estes—Ah, okay, I understand.  Separately: have you considered turning on email notifications from the HHVM on Aarch64 group?  That might yield some marginal friction to replying, but it's possible that's not even a bad thing (i.e., I think discussion like this one, on PRs, is probably the best vehicle for communication).  Can you try that?  If the experience persists in being terrible, I'll set up an email list (but I definitely prefer FB groups because they leave a publicly accessible paper trail).

Okay, it sounds like killing these assertions is safe because the macro assembler will handle any cases where the displacement doesn't fit into a single-instruction form, and we may want to do a more refined job about what bounds we restrict the displacement of Vptrs to, in `lea{}` and elsewhere, depending on what instruction they're being used in.

I'm going to land this diff unless anyone has any objection; sounds like it's strictly a bugfix and does not, as I thought originally, un-hide a regression.

PR #7481, opened 2016-11-09
[AArch64] Add map-flags pass.

dave-estes:
This new pass tracks every def of a VregSF and the uses. Then the
pass iterates over each use of a VregSF and updates the new field
in the instruction that defs the VregSF with the individual,
platform-specific status flags which need to be set.

The pass aims to accomplish the following:
  A) Allows platforms to map the condition codes at a use instruction
     based on the def instruction. This is especially beneficial for
     platforms that don't have a perfect 1:1 mapping of condition
     codes in terms of semantics.
  B) Allows platforms to optimize their emitters to only set the
     status flags bits that are necessry. This avoids emitting
     expesive code to synthetically set status flags unless
     absolutely required.


hphp/runtime/vm/jit/abi-ppc64.h
```diff
@@ -116,6 +116,10 @@ constexpr RegXMM rFasm = ppc64_asm::reg::f15;
 
 ///////////////////////////////////////////////////////////////////////////////
 
+inline uint8_t required_flags(jit::ConditionCode cc) { return 0xff; }
```
dave-estes: The PPC64 (and X64) implementations of this function just set the flags uint8_t to all 1's. The value set for the flags doesn't matter unless the map flags pass is used in either of those two platforms.

hphp/runtime/vm/jit/abi.cpp
```diff
@@ -77,4 +77,10 @@ RegSet arg_regs_ind_ret(size_t n) {
 
 ///////////////////////////////////////////////////////////////////////////////
 
+uint8_t required_flags(jit::ConditionCode cc) { 
+  return ARCH_SWITCH_CALL(required_flags, cc);
+}
```
dave-estes: Initially, the map flags pass was strictly for ARM. I decided to generalize it more, which led to the exposure of this function through the ABI files of the other platforms.

hphp/runtime/vm/jit/abi-arm.h
```diff
@@ -144,6 +165,46 @@ inline jit::ConditionCode convertCC(vixl::Condition cc) {
...
+    StatusFlags::V,         // vc
+    StatusFlags::CZ,        // hi
+    StatusFlags::CZ,        // ls
+    StatusFlags::NV,        // ge
+    StatusFlags::NV,        // lt
+    StatusFlags::NZV,       // gt
+    StatusFlags::NZV,       // le
+    StatusFlags::None,      // nv (invalid)
+    StatusFlags::None,      // al (invalid)
+  };
```
dave-estes: This is ARM's implementation of the platform-specific function which maps a vasm CC to the exact flags which are required to be set for the Vinstr def in order for the Vinstr use to test that condition code.

hphp/runtime/vm/jit/abi-arm.h
```diff
@@ -144,6 +165,46 @@ inline jit::ConditionCode convertCC(vixl::Condition cc) {
...
+  C    = 1 << 2,
+  Z    = 1 << 1,
+  V    = 1,
+  CZ   = C | Z,
+  NV = N | V,
+  NZV = N | Z | V,
+  NotC = N | Z | V,
+  NotV = N | C | Z,
+  All  = N | C | Z | V,
+};
```
dave-estes: These are the ARM-specific status flag combinations tested with the 16 different ARM condition codes.

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -117,6 +117,47 @@ vixl::Condition C(ConditionCode cc) {
...
+/*
+ * Uses the flags from the Vinstr which defs SF to determine
+ * whether or not the Vixl assembler should emit code which
+ * sets the status flags.
+ */
+vixl::FlagsUpdate updateFlags(uint8_t flags) {
+  return flags ? SetFlags : LeaveFlags;
+}
+
+#define UF(x) updateFlags(x)
```
dave-estes: These short macros (`UF` and `CSF`) are used for the one-line emitters to keep them under 80 columns wide.

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -274,20 +315,25 @@ struct Vgen {
   void emit(const push& i);
   void emit(const roundsd& i);
   void emit(const sar& i);
+  void emit(const sarqi& i);
```
dave-estes: This and the other shift vasm below are now implemented with an emitter instead of lowering them to the variations that do and don't set status flags. Those have been removed. They were initially created in an attempt to avoid expensive synthetic status flags setting. It's no longer needed since we're tracking precisely which status flags need to be set when emitting an instruction that defs SF.

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -595,11 +630,11 @@ void Vgen::emit(const unwind& i) {
 ///////////////////////////////////////////////////////////////////////////////
 
 void Vgen::emit(const andbi& i) {
-  a->And(W(i.d), W(i.s1), MSKTOP(i.s0.l()), SetFlags);
+  a->And(W(i.d), W(i.s1), MSKTOP(i.s0.l()), updateFlags(i.fl));
```
dave-estes: Note: The shortened `UF` macro isn't used and below here since this is not one of these single-line emitters.

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -620,46 +655,54 @@ void Vgen::emit(const cloadq& i) {
...
-  a->Eor(rAsm, rAsm, X(i.d));
-  a->Tbnz(rAsm, 63, &Overflow);
+  // If we have to set any flags, then always set N and Z since it's cheap.
+  // Only set V when absolutely necessary. C is not supported.
+  if (i.fl) {
+    vixl::Label after;
+
+    checkSF(i, StatusFlags::NotC);
+
+    if (flagRequired(i.fl, StatusFlags::V)) {
```
dave-estes: Avoids the expensive overflow checking unless the V bit is required by a use.

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -620,46 +655,54 @@ void Vgen::emit(const cloadq& i) {
...
-  a->Msr(NZCV, rAsm);
-  a->B(&after);
-
-  // Check the signs of hi and lo.
-  a->Bind(&checkSign);
-  a->Eor(rAsm, rAsm, X(i.d));
-  a->Tbnz(rAsm, 63, &Overflow);
+  // If we have to set any flags, then always set N and Z since it's cheap.
+  // Only set V when absolutely necessary. C is not supported.
+  if (i.fl) {
```
dave-estes: Avoids all synthetic status flag setting if there are no SF bits required. If even one is required, then the relatively cheap Z and N are checked.

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -620,46 +655,54 @@ void Vgen::emit(const cloadq& i) {
...
+    vixl::Label after;
+
+    checkSF(i, StatusFlags::NotC);
```
dave-estes: This is the first use of checkSF with a specific flag. In this case we want to assert of  any flags outside of NotC are required by a use. i.e. We don't support synthetically setting C.

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -117,6 +117,47 @@ vixl::Condition C(ConditionCode cc) {
...
+  uint8_t required = i.fl;
+  uint8_t set = static_cast<uint8_t>(s);
+  always_assert_flog((required & set) == required,
+                     "should def SF but does not: {}\n",
+                     vinst_names[Vinstr(i).op]);
+}
+
+template<class Inst> void checkSF(const Inst& i) {
+  checkSF(i, StatusFlags::None);
+}
```
dave-estes: Originally, I used this function to protect instructions which didn't support setting status flags. With all of the SF setting tidied up, we can probably remove it.

hphp/runtime/vm/jit/vasm-instr.h
```diff
@@ -131,90 +131,90 @@ struct Vunit;
   O(nop, Inone, Un, Dn)\
   O(ud2, Inone, Un, Dn)\
   /* arithmetic instructions */\
```
dave-estes: I considered many different ways to annotate a Vinstr with information about the Vinstrs that use the VregSF that it defs. Storing it in the Vunit was problematic if the blocks were mutated at all, changing their index in the block. I also used the pointers to the Vinstrs, but those also change often with copy constructors. Storing it in the Vinstr itself made the most sense. Vinstrs have restrictions on their size, so I opted to put required flags only in the Vinstrs that def a VregSF. Furthermore, I used a uint8_t since no platforms should have more than 8 status flags.

mxw: Vinstr has an annotated two-byte hole.  Why not put it there?  It also avoids the churn of adding flag bits to all the sub-structs, which is nice since it's not used at all on x64 (or on ppc64, at least yet).

hphp/runtime/vm/jit/abi-arm.h
```diff
@@ -144,13 +165,57 @@ inline jit::ConditionCode convertCC(vixl::Condition cc) {
   return mapping[cc];
 }
 
+enum class StatusFlags : uint8_t {
+  None = 0,
+  N    = 1 << 3,
+  C    = 1 << 2,
+  Z    = 1 << 1,
+  V    = 1,
```
mxw: nit: It's a little weird seeing this in reverse bit order—maybe flip it?

hphp/runtime/vm/jit/abi.h
```diff
@@ -159,6 +159,10 @@ inline RegSet fcall_array_regs() { return vm_regs_with_sp(); }
 
 ///////////////////////////////////////////////////////////////////////////////
 
+uint8_t required_flags(jit::ConditionCode cc);
```
mxw: "Please document me." —`required_flags()`

Also, you don't need the `jit::`.

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -202,17 +257,17 @@ struct Vgen {
   void emit(const unwind& i);
 
   // instructions
-  void emit(const addl& i) { a->Add(W(i.d), W(i.s1), W(i.s0), SetFlags); }
-  void emit(const addli& i) { a->Add(W(i.d), W(i.s1), i.s0.l(), SetFlags); }
-  void emit(const addq& i) { a->Add(X(i.d), X(i.s1), X(i.s0), SetFlags); }
-  void emit(const addqi& i) { a->Add(X(i.d), X(i.s1), i.s0.q(), SetFlags); }
+  void emit(const addl& i) { a->Add(W(i.d), W(i.s1), W(i.s0), UF(i.fl)); }
```
mxw: Oh, I guess this sort of demands that the flags bits be in the instruction structs...

hphp/runtime/vm/jit/vasm-arm.cpp
mxw: I don't see the value in this discrepancy.  Why not just name the function `UF()`?

hphp/runtime/vm/jit/vasm.h
```diff
@@ -107,6 +107,7 @@ bool checkBlockEnd(const Vunit& v, Vlabel b);
  */
 void allocateRegisters(Vunit&, const Abi&);
 void fuseBranches(Vunit&);
+void mapFlags(Vunit&);
```
mxw: I'm not sure what "map" refers to.  I would call this something like `annotateSFUses()`, but any self-explanatory name is good.  (You can just name the file `vasm-flags.cpp`.)

hphp/runtime/vm/jit/vasm-instr.h
```diff
@@ -908,105 +897,105 @@ struct ud2 {};
...
  */
 // add: s0 + {s1|m} => {d|m}, sf
-struct addl   { Vreg32 s0, s1, d; VregSF sf; };
-struct addli  { Immed s0; Vreg32 s1, d; VregSF sf; };
-struct addlm  { Vreg32 s0; Vptr m; VregSF sf; };
-struct addlim { Immed s0; Vptr m; VregSF sf; };
-struct addq  { Vreg64 s0, s1, d; VregSF sf; };
-struct addqi { Immed s0; Vreg64 s1, d; VregSF sf; };
-struct addqim { Immed s0; Vptr m; VregSF sf; };
+struct addl   { Vreg32 s0, s1, d; VregSF sf; uint8_t fl; };
```
mxw: I would abstract away the underlying representation here with `using Vflags = uint8_t`, if the idea is to make it more generally usable (we might want to make it a union of per-arch information, etc.)

hphp/runtime/vm/jit/vasm-map-flags.cpp
```diff
@@ -0,0 +1,204 @@
...
+///////////////////////////////////////////////////////////////////////////////
+
+struct Env {
+  Vunit& unit;
+
+  // Map of VregSF to the inst that defs it
+  jit::hash_map<size_t, Vinstr*> sf_defs;
+
+  // Map of VregSF to list of insts that use it
+  jit::hash_map< size_t, jit::vector<Vinstr*> > sf_uses;
```
mxw: nit: Nix extra spaces.

mxw: Also, maybe combine these maps into a single map from `size_t` to `struct { Vinstr* def; jit::vector<Vinstr*> uses; };` to avoid some lookup overhead.

hphp/runtime/vm/jit/vasm-map-flags.cpp
```diff
@@ -0,0 +1,204 @@
...
+  switch (inst->op) {
+#define O(name, imms, uses, defs) \
+    case Vinstr::name: { \
+      auto& i = inst->name##_; (void)i; \
+      imms \
+      break; \
+    }
+#define Inone
+#define fl fl; return i.fl
+#define I(x) (void)i.x;
```
mxw: This is... pretty weird.  Do we have macro dependencies on the names of immediates elsewhere in the codebase already?

hphp/runtime/vm/jit/vasm-map-flags.cpp
```diff
@@ -0,0 +1,204 @@
...
+///////////////////////////////////////////////////////////////////////////////
+
+/*
+ * For each use of an SF, return the StatusFlags bits needed by the use's CC.
+ *
+ * Specializations are below, which can additionally change the CC if the
+ * platform doesn't have a simple CC mapping for the given def and use.
+ */
+template<typename Inst>
+uint8_t mapFlags(Env& env, Inst& inst) {
```
mxw: Should these functions take a `const Env& env`?

hphp/runtime/vm/jit/vasm-map-flags.cpp
```diff
@@ -0,0 +1,204 @@
...
+/*
+ * Map SF pass for a Vunit.
+ */
+void mapFlags(Vunit& unit) {
+  assertx(check(unit));
+  auto& blocks = unit.blocks;
+
+  Env env { unit };
+  env.sf_defs.clear();
+  env.sf_uses.clear();
```
mxw: Unnecessary; these are getting default initialized already.

hphp/runtime/vm/jit/vasm-map-flags.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+  // Set up Env, only visiting reachable blocks.
+  // Tracks each VregSF to the instruction which def'd it.
+  // Tracks each VregSF to a list of each instruction which uses it.
+  for (auto const b : labels) {
+    assertx(!blocks[b].code.empty());
+    for (auto const& inst : blocks[b].code) {
+      visitDefs(unit, inst, [&] (Vreg r, Width w) {
+        if (w == Width::Flags) {
+          env.sf_defs[r] = (Vinstr*)(&inst);
```
mxw: watduck

If you want a non-const pointer, just do `auto& inst` in the loop header instead of `auto const& inst`.

hphp/runtime/vm/jit/vasm-map-flags.cpp
```diff
@@ -0,0 +1,204 @@
...
+      visitUses(unit, inst, [&] (Vreg r, Width w) {
+        if (w == Width::Flags) {
+          env.sf_uses[r].push_back((Vinstr*)(&inst));
+        }
+      });
+    }
+  };
+
+  // Calls mapFlags on each 
+  for (auto it = env.sf_defs.begin(); it != env.sf_defs.end(); it++) {
```
mxw: Use range-based for.

hphp/runtime/vm/jit/vasm-map-flags.cpp
```diff
@@ -0,0 +1,204 @@
...
+    auto r = it->first;
+    auto d = it->second;
+
+    // For each instruction that defs an sf, initially clear the flag bits.
+    uint8_t flags = 0x00;
+
+    // Call mapFlags for each use. This will a) update the instruction's
+    // condition code (cc) based on any platform specific mapping b) update
+    // the flag bits (c) for this VregSF def based on the instruction's cc.
+    for (auto const u : env.sf_uses[r]) {
```
mxw: It's a little wasteful to construct `sf_uses` when you're just going to iterate through it a single time.  Can you just do this work in place of where you're currently adding a new mapping to `sf_uses`?  If you iterate the blocks in RPO order, you'll always hit the def first, so you'll have that available at every use.

hphp/runtime/vm/jit/vasm-map-flags.cpp
mxw: Oh, how about this—can you go through the opcode table and replace all the `I(cc)`s with `IC(cc)` and the `I(fl)`s with `IF(fl)`?  That way we can collect them explicitly using some `#define`s.  You'll just need to make sure that everywhere where we're currently defining `I()` and expecting it to hit all the immediates, we instead define `IF()` and `IC()` as well to be the same thing.

mxw: @dave-estes—Oof, sorry for the churn, but I just realized (because of your modification of `vasm-visit.h`) that `visitOperands()` can already visit immediates by type...

You can just define anonymous visitor classes like:

```
struct Visitor {
  void imm(ConditionCode& cc) {
    assertx(!found);
    do_your_stuff();
    found = true;
  }
  template<class I> void imm(I&) {}

  /* Define all the other necessary visitors; see vasm-xls.cpp for some examples. */

  bool found{false};
};
```

That way we don't need to rely on macro table manipulations to get what we want.

hphp/runtime/vm/jit/vasm-map-flags.cpp
mxw: Just use `.at()`?

hphp/runtime/vm/jit/vasm-map-flags.cpp
mxw: @dave-estes—I don't see how that could fit into the scope of this pass.  That sounds like a peephole optimization you could do as part of `vasm-simplify.cpp` after running `annotateSFUses()`, because it also depends on the defs of those `cmovq{}`s being functionally chained, and not used as srcs by any later instructions.  This pass should be restricted to annotating flags and conditions.

hphp/runtime/vm/jit/vasm-flags.cpp
```diff
@@ -0,0 +1,215 @@
...
+Vflags annotateSFUses(Env& env, jcc& inst) {
+  if (env.sf_def_uses[inst.sf].def->op == Vinstr::ucomisd) {
+    switch (inst.cc) {
+    case CC_AE: // CC_NB
+      inst.cc = CC_GE;
+      break;
+    default:
+      break;
+    }
+  }
```
mxw: Can you do an arch check for this also?

hphp/runtime/vm/jit/vasm-flags.cpp
```diff
@@ -0,0 +1,215 @@
...
+  return static_cast<Vflags>(required_flags(inst.cc));
```
mxw: Would be nice to structure this so that the specializations don't have to repeat the boilerplate `return static_cast<Vflags>(required_flags(inst.cc));`

hphp/runtime/vm/jit/vasm-flags.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+///////////////////////////////////////////////////////////////////////////////
+
+}
+
+/*
+ * Map SF pass for a Vunit.
+ */
+void annotateSFUses(Vunit& unit) {
+
```
mxw: nit: Delete extraneous newline.

hphp/runtime/vm/jit/vasm-flags.cpp
```diff
@@ -0,0 +1,215 @@
...
+  if (!arch_any(Arch::ARM)) return;
```
mxw: This is unnecessary—we can just not invoke this pass for the other platforms.

Also: Don't you want to call this somewhere in `emitARM()` or `optimizeARM()`?

hphp/runtime/vm/jit/vasm-flags.cpp
```diff
@@ -0,0 +1,215 @@
...
+      visitUses(unit, inst, [&] (Vreg r, Width w) {
+        if (w == Width::Flags) {
+          env.sf_def_uses[r].uses.push_back(&inst);
+        }
+      });
+    }
+  };
+
+  // Calls annotateSFUses on each
+  for (auto&& kv : env.sf_def_uses) {
```
mxw: Uh... why `auto&&`?

hphp/runtime/vm/jit/vasm-flags.cpp
```diff
@@ -0,0 +1,215 @@
...
+    DEBUG_ONLY auto r = kv.first;
```
mxw: Just use `kv.first` directly in the assertion?

hphp/runtime/vm/jit/abi-arm.h
```diff
@@ -144,13 +165,57 @@ inline jit::ConditionCode convertCC(vixl::Condition cc) {
   return mapping[cc];
 }
 
+enum class StatusFlags : Vflags {
```
mxw: `enum class StatusFlags : uint8_t {`

hphp/runtime/vm/jit/vasm-flags.cpp
```diff
@@ -0,0 +1,215 @@
...
+    auto du = kv.second;
+
+    // For each instruction that defs an sf, initially clear the flag bits.
+    Vflags flags = 0x00;
+
+    // Call annotateSFUses for each use. This will a) update the instruction's
+    // condition code (cc) based on any platform specific mapping b) update
+    // the flag bits (c) for this VregSF def based on the instruction's cc.
+    for (auto const u : du.uses) {
+      flags |= annotateSFUses(env, u);
```
mxw: You've defined `Vflags` as a sort of generic type that different platforms can use as they wish—but there is an implicit requirement here that it be coherently `|`-able as an integer.  I don't think it's worth generalizing until there are more users, but at least you should document this behavior when defining `Vflags`.

hphp/runtime/vm/jit/vasm-flags.cpp
```diff
@@ -0,0 +1,211 @@
...
+  FlagVisitor fv {};
+  visitOperands(*inst, fv);
+  return fv.getFlags();
+}
+
+void setFlags(Vinstr* inst, Vflags flags) {
+  FlagVisitor fv {};
+  visitOperands(*inst, fv);
+  fv.setFlags(flags);
+}
```
mxw: One of these is unused, and the other is used in one place—can you delete both?

hphp/runtime/vm/jit/vasm-flags.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+///////////////////////////////////////////////////////////////////////////////
+
+/*
+ * Check the particular SF uses and change the CC if the platform doesn't have
+ * a simple CC mapping for the given def and use. This is to be done in the
+ * specializations below as general mappings are usually correct.
+ */
+template<typename Inst>
+void mapSFUses(const Env& env, Inst& inst) {}
```
mxw: This name doesn't tell me anything about what this pass is doing :(

mxw:
@dave-estes—This looks great!  Thanks for the work.  I need to poke at our internal build files a bit before landing this.  It looks like there are some conflicts, though—can you rebase and re-push first?  Also: is this fixing any unit tests?

cc/ @swalk-cavium, @apinski-cavium, @cmuellnerTheobromaSystems, @jim-saxman for any further comments, suggestions, or ideas.  Further applications of this pass can be left to later PRs.

dave-estes:
It fixes test/slow/php7_bcbreak/7/int_semantics/nan_inf_cast.php.

Will start the rebase now.

mxw:
Quick update—Right now, GitHub sync from internal master and PR sync to internal Phabricator are both nonfunctional.  @dave-estes, once they're both working again, I'll ping here and ask you to do another rebase (since I think the last one was a no-op, since we haven't been syncing to GitHub for five days, apparently...).

mxw:
@dave-estes—Okay, things are back to normal... can you rebase again?  Sorry for the churn :(

dave-estes:
No worries. I've just pushed the newly rebased patch.

PR #7486, opened 2016-11-11
Disallow static variables

### LINKED ISSUE #7453: Hack: Static variables are permitted, but they are not type safe
### HHVM Version

```
HipHop VM 3.16.0-dev (rel)
Compiler: heads/master-0-ge9eb3fc414fbae9b2c999a4f11390e23050b23b7
Repo schema: 9cc0ebe15bc9d2ba3609c5e4458deeef9ed86dcc
```

``` hack
<?hh // strict

class Foo {
  public function fooMethod(): void {}
}

function foo(bool $b = false): void {
  static $x;
  if ($b) {
    $x = 1;
  } else {
    $x = new Foo();
    foo(true);
    $x->fooMethod();
  }
}

/* HH_IGNORE_ERROR[1002] */
foo();
```

### Expected result

A type error from `hh_client` that static variables are disallowed.
### Actual result

No error from `hh_client`, but this error is produced at runtime:

```
Uncaught Error: Call to a member function fooMethod() on integer
```
### Discussion

I expected static variables to be disallowed by Hack, in the same way that global variables and references generally are disallowed, but they are permitted, and this creates a hole in the type system.

jesseschalken:
The Hack typechecker generally bans syntax that introduces a reference, such as `&` or `global` (see https://docs.hhvm.com/hack/unsupported/references), and does flow analysis assuming that references don't exist and that local variables are in fact local.

Static variables are also references and therefore break the assumptions of the type checker, exposing a hole, as in the function below:

```hack
function foo(bool $b = false): void {
  static $x;
  if ($b) {
    $x = 1;
  } else {
    $x = new Foo();
    foo(true); // <= will assign $x = 1
    $x->fooMethod(); // <= Call to fooMethod() on int, but allowed by Hack
  }
}
```

Since static variables offer little above static properties besides some conciseness, and also hamper readability because the entire function must be searched for `static $f` before knowing if `$f = ...;` modifies shared state or not, just disallow them and require the developer to use a static property instead.

Fixes #7453


dlreeves:
In Hack we take the philosophy of eliminating type holes in strict mode, but leaving them present in partial/decl mode. This is what we do for references, where we ban them in strict mode but allow them in partial mode even though they are unsound.

In order to allow this distinction we cannot remove support for this at the parser level, instead it needs to be done at a later pass in the type checker. You will want to look in [typing/typing.ml](https://github.com/facebook/hhvm/blob/master/hphp/hack/src/typing/typing.ml#L591) and emit an error in strict mode if we see a static variable, otherwise we leave the behavior as is.

simonwelsh:
I'm against this change. It's taking a rather extreme approach to something that could be solved by allowing types on static variables, as you have the exact same problem with untyped static members on classes.

```hack
class Baz {
  private static $x;

  public static function foo(bool $b = false): void {
    if ($b) {
      self::$x = 1;
    } else {
      self::$x = new Foo();
      self::foo(true); // <= will assign self::$x = 1
      self::$x->fooMethod(); // <= Call to fooMethod() on int, but allowed by Hack
    }
  }
}
```

Removing static variables will also force a fair amount of perfectly type safe code into static classes.

jesseschalken:
@dlreeves

I took the same approach as it currently does for `global $foo;` which is to reject it at the parser level. I'm happy to make it just rejected by the typechecker in strict mode like `&$foo` but, as a separate issue, shouldn't the same apply to `global $foo;`?

@simonwelsh

My only problem with a type annotation on the static variable is that static variables still hamper readability because I can no longer assume that a simple variable assignment is only local in effect, which was one of the useful side effects of Hack banning references. If static variables are permitted then every variable assignment could potentially be modifying shared state and I have to scan the function looking for `static $foo;` to find out.

Besides that, what's most important is that the hole is plugged, and without doing a great deal of work in the parsing/inference/typechecking code, banning static vars, at least in strict mode, is the shortest path to that end.

andrewjkennedy:
Here's an idea for eliminating this via typing.
1. Conservatively assume that any function containing a static variable can be re-entered.
2. For such functions, apply a fixed-point iteration on the entire function in order to obtain types for the statics. (Just as we do for loops.) For the counterexample, we would deduce that `$x` has type `int|Foo` and thereby give an error to the call to `fooMethod`.

elgenie:
`global $foo`; is rejected because it makes latter `$foo`s in the function body unnecessarily difficult for both humans and the typechecker to reason about. Forcing `$GLOBALS['foo']` allows for the code to function the same as before, but makes the dubiousness explicit.

     global <typehint> $foo = 10; // where does the typehint go? $foo could be declared in dozens of files
     some_function(); // hahaha, this could change the value of GLOBALS['foo'] ...
     takes_int($foo); // ... and your type inference is worthless

jesseschalken:
@elgenie

> `global $foo;` is rejected because it makes latter `$foo`s in the function body unnecessarily difficult for both humans and the typechecker to reason about. 

The same applies to `static $foo;`, at least partially. Looking at this code in isolation

```php
$x = new Foo();
foo(true);
$x->fooMethod();
```

it appears that there shouldn't be any way `$x` could be anything besides a `Foo`, but if static variables are permitted, you have to scan the entire function body (or at least all the code that this is reachable from) looking for `static $x;` before you can even know whether or not the standard reasoning about variables applies.

I would be happy if the assumption that local variables are local was broken in favour of general support for references, but breaking that assumption only to support such a rarely used feature as static variables seems to me a net loss.

jesseschalken:
@andrewjkennedy 

Also consider the location of the `static $x;` line, since it's a normal variable before that point and a reference thereafter. If the `static $x;` line is inside a conditional block or loop it may get complicated.

For that reason it may be necessary to more directly model what happens at runtime with a type internal to the typechecker like this:

```hack
class ref<T> {
  public function __construct(public T $val) {}
}
```

and the addition of these rules:

1. `static $x = 1;` is typed as `$x = new ref(1);`
2. `$x = ...` when `$x` is a `ref<T>` is typed as `$x->val = ...`
3. `$x` when `$x` is a `ref<T>` is typed as `$x->val`

Then inferring `T` follows the normal generic inference rules.

Support for references generally would also be possible, by typing `$a =& $b` as `$a = ($b = new ref($b))`.

I would still prefer static variables be banned in strict mode (I am yet to update this PR as such), but if you can make them rigorously type safe some other way, great. If you can get references type-checked at the same time, even better.

elgenie:
@jesseschalken

Since  function `static`s are not accessible from outside function scope (with the possible exception of very esoteric and obviously non-type-safe reflection stuff), it should still be _very_ possible for the type-checker to reason about it in terms of the scope of the function that it's looking at. 

For that reason, I don't see this as a typechecker hole. Probably the correct-but-annoying solution would be, in strict mode, to require type annotations on `static` inside functions and methods when a recursive assignment cannot be provably ruled out. Some patterns should probably be detected.

e.g. 

   function foo() {
       // if inference is possible ...
       static $do_this_once = null;
       // if inference is not possible ...
       static ?Foo $do_this_once = null;
       if (null === $do_this_once) {
          // ... a bunch of work
         $do_this_once = new Foo() // ... 
       }
       // ... should $do_this_once be considered a `?Foo` or a `Foo`? the latter should be provable!
   }

Orvid:
As this isn't likely to make it in without significant changes, I'm going to close this for now.

PR #7490, opened 2016-11-15
[Aarch64] Fix decqmlock with acquire and release semantics

jim-saxman:
The VASM decqmlock and incqmlock have no ordering semantics when
using ldxr and stxr Aarch64 instructions. The current code is
broken and could be fixed by adding DMB barrier instructions,
however the cleaner approach taken here is to use the load-acquire
and store-release, which is equal to x86 including ordering
characteristics. This is also the same as PPC64's decqmlock.

### DIFF
--- a/hphp/runtime/vm/jit/vasm-arm.cpp
+++ b/hphp/runtime/vm/jit/vasm-arm.cpp
@@ -667,9 +667,9 @@ void Vgen::emit(const vasm_opc& i) {   \
   auto adr = M(i.m);                   \
   vixl::Label again;                   \
   a->bind(&again);                     \
-  a->ldxr(rAsm, adr);                  \
+  a->ldaxr(rAsm, adr);                 \
   a->arm_opc(rAsm, rAsm, 1, SetFlags); \
-  a->stxr(rAsm.W(), rAsm, adr);        \
+  a->stlxr(rAsm.W(), rAsm, adr);       \
   a->Cbnz(rAsm.W(), &again);           \
 }
 
--- a/hphp/vixl/a64/assembler-a64.cc
+++ b/hphp/vixl/a64/assembler-a64.cc
@@ -1135,6 +1135,23 @@ void Assembler::ldr(const FPRegister& ft, double imm) {
 }
 
 
+void Assembler::ldaxr(const Register& rt,
+                      const MemOperand& src) {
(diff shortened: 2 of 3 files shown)


apinski-cavium:
>which is equal to x86 including ordering
characteristics. 

I would be careful what you say here because lock inc or lock dec does not have any memory model associated with it.

apinski-cavium:
Really decqmlock/incqmlock should have order specified with them instead of doing this kind of patch.

apinski-cavium:
So I looked further into these usage and I say this should be written such that we actually have the memory model added.
For an example the usage in cgCheckCold:
  v << decqmlock{v.cns(counterAddr)[0], sf};
is an acquire barrier (well it is really consume but nobody should be using that C11 memory model).

But the usage cgIncProfCounter is relax as no stores or loads around it matter to it.

And incqmlock is not used at all.

Also it might be a good idea to also emit ARMv8.1 LSE instructions where LSE is available.

mxw:
@swtaarrs, @edwinsmith, @ottoni—What do you all think the design intent is for these instructions?

edwinsmith:
IIRC, they're for independent profiling counters, so any memory model that doesn't silently drop increments is probably fine.

swtaarrs:
I'm with @edwinsmith: I wouldn't expect our usage of `decqmlock` for profile counters to require anything other than atomicity, even in `cgCheckCold`. There's a bunch of translation metadata that needs to be in place for retranslation to work but that should all be guarded by other locks and atomics.

@jim-saxman and @apinski-cavium: what other memory accesses are you worried about this being reordered with? Were you seeing actual failures caused by the lack of acquire/release semantics?

It looks like `incqmlock` is unused since 4ca340e1d185ef91f1e4d31248f83d986ebc90f2 so we can probably delete it, especially if we end up doing something nontrivial to `decqmlock` that should be mirrored in any related instructions.

jim-saxman:
The current implementation will perform an atomic update without the acquire-release. Based on everyone's responses, I am going to drop this patch. @apinski-cavium I agree, we should adopt the 8.1 LSE instructions for CPU's that support them. Let's give some time to ARM Ltd. to add them to the VIXL assembler.

mxw:
Okay, I'm going to close without landing.  Ping here if it should be reopened.

PR #7495, opened 2016-11-16
[AArch64] Cleanup and fix abi-related vasm.

dave-estes:
Moved several abi-related vasm implementations from emitters to
lowerers. Change calltc to push FP/LR pair instead of the saved
RIP from the activation record and the i.exittc. Remove extra
pop of FP/LR in the emulated return to enterTCExit. While x86_64
actually pops the LR when a return occurs AArch64 does not.
Lastly, add syncpoint after calling destructors in emitDecRefHelper.

### DIFF
--- a/hphp/runtime/vm/jit/vasm-arm.cpp
+++ b/hphp/runtime/vm/jit/vasm-arm.cpp
@@ -171,29 +171,23 @@ struct Vgen {
 
   // native function abi
   void emit(const call& i);
-  void emit(const callm& i);
   void emit(const callr& i) { a->Blr(X(i.target)); }
   void emit(const calls& i);
   void emit(const ret& i) { a->Ret(); }
 
   // stub function abi
-  void emit(const stublogue& i);
-  void emit(const stubret& i);
-  void emit(const callstub& i);
   void emit(const callfaststub& i);
-  void emit(const tailcallstub& i);
+  void emit(const callstub& i);
 
   // php function abi
-  void emit(const phplogue& i);
-  void emit(const phpret& i);
   void emit(const callphp& i);
-  void emit(const tailcallphp& i);
   void emit(const callarray& i);
   void emit(const contenter& i);
 
   // vm entry abi
+  void emit(const calltc& i);
   void emit(const inittc& i) {}
-  void emit(const leavetc&);
+  void emit(const leavetc& i);
 
   // exceptions
   void emit(const landingpad& i) {}
@@ -235,7 +229,6 @@ struct Vgen {
   void emit(const jcci& i);
   void emit(const jmp& i);
   void emit(const jmpi& i);
-  void emit(const jmpm& i);
   void emit(const jmpr& i) { a->Br(X(i.target)); }
   void emit(const lea& i);
   void emit(const leap& i) { a->Mov(X(i.d), i.s.r.disp); }
@@ -308,7 +301,6 @@ struct Vgen {
   void emit(const addxi& i) { a->Add(X(i.d), X(i.s1), i.s0.q()); }
   void emit(const asrxi& i);
   void emit(const asrxis& i);
-  void emit(const bln& i);
   void emit(const cmplims& i);
   void emit(const cmpsds& i);
   void emit(const fabs& i) { a->Fabs(D(i.d), D(i.s)); }
@@ -488,71 +480,26 @@ void Vgen::emit(const call& i) {
   }
 }
 
-void Vgen::emit(const callm& i) {
-  a->Ldr(rAsm, M(i.target));
-  a->Blr(rAsm);
-}
-
 void Vgen::emit(const calls& i) {
   emitSmashableCall(a->code(), env.meta, i.target);
 }
 
 ///////////////////////////////////////////////////////////////////////////////
 
-void Vgen::emit(const stublogue& i) {
-  // Push FP, LR always regardless of i.saveframe (makes SP 16B aligned)
-  emit(pushp{rlr(), rfp()});
-}
-
-void Vgen::emit(const stubret& i) {
-  if (i.saveframe) {
-    emit(popp{rfp(), rlr()});
-  } else {
-    emit(popp{PhysReg(rAsm), rlr()});
-  }
-  emit(ret{});
+void Vgen::emit(const callfaststub& i) {
+  emit(call{i.target, i.args});
+  emit(syncpoint{i.fix});
 }
 
 void Vgen::emit(const callstub& i) {
   emit(call{i.target, i.args});
 }
 
-void Vgen::emit(const callfaststub& i) {
-  emit(call{i.target, i.args});
-  emit(syncpoint{i.fix});
-}
-
-void Vgen::emit(const tailcallstub& i) {
-  // Pop off FP/LR pair and jump to target
-  emit(popp{rfp(), rlr()});
-  emit(jmpi{i.target, i.args});
-}
-
 ///////////////////////////////////////////////////////////////////////////////
-
-void Vgen::emit(const phplogue& i) {
-  // Save LR in m_savedRip on the current VM frame pointed by 'i.fp'
-  a->Str(X(rlr()), X(i.fp)[AROFF(m_savedRip)]);
-}
-
-void Vgen::emit(const phpret& i) {
-  a->Ldr(X(rlr()), X(i.fp)[AROFF(m_savedRip)]);
-  if (!i.noframe) {
-    a->Ldr(X(i.d), X(i.fp)[AROFF(m_sfp)]);
-  }
-  emit(ret{});
-}
 
 void Vgen::emit(const callphp& i) {
   emitSmashableCall(a->code(), env.meta, i.stub);
   emit(unwind{{i.targets[0], i.targets[1]}});
-}
-
-void Vgen::emit(const tailcallphp& i) {
-  // To make callee's return as caller's return, load the return address at
-  // i.fp[AROFF(m_savedRip)] into LR and jmp to target
-  a->Ldr(X(rlr()), X(i.fp)[AROFF(m_savedRip)]);
-  emit(jmpr{i.target, i.args});
 }
 
 void Vgen::emit(const callarray& i) {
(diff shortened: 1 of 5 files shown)


hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -1328,55 +1275,145 @@ void lower(Vunit& unit, cvttsd2siq& i, Vlabel b, size_t idx) {
...
+    // Then directly jump to the target
+    v << jmpi{i.target, i.args};
+  });
+}
+
+void lower(Vunit& u, stubret& i, Vlabel b, size_t z) {
+  lower_impl(u, b, z, [&] (Vout& v) {
+    // Pop LR and (optionally) FP
+    if (i.saveframe) {
+      v << popp{rfp(), rlr()};
```
mxw: Why `rfp()` instead of `rvmfp()`?

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -1328,55 +1275,145 @@ void lower(Vunit& unit, cvttsd2siq& i, Vlabel b, size_t idx) {
...
   lower_impl(u, b, z, [&] (Vout& v) {
+    // Pop the call frame
     v << lea{rsp()[16], rsp()};
   });
 }
 
 void lower(Vunit& u, loadstubret& i, Vlabel b, size_t z) {
   lower_impl(u, b, z, [&] (Vout& v) {
+    // Load the LR to the destination
     v << load{rsp()[8], i.d};
```
mxw: Use `AROFF(m_savedRip)` instead of `8`.

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -1328,55 +1275,145 @@ void lower(Vunit& unit, cvttsd2siq& i, Vlabel b, size_t idx) {
...
+  lower_impl(u, b, z, [&] (Vout& v) {
+    // Restore the saved RIP to LR
+    v << load{i.fp[AROFF(m_savedRip)], rlr()};
+
+    // If there's no frame, then load fp
+    if (!i.noframe) {
+      v << load{i.fp[AROFF(m_sfp)], i.d};
+    }
+
+    // Return
```
mxw: These inline comments all seem a little redundant.  Try to avoid comments of the form

```
// Do the thing.
do_the_thing();
```

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -1328,55 +1275,145 @@ void lower(Vunit& unit, cvttsd2siq& i, Vlabel b, size_t idx) {
...
+    v << ret{i.args};
+  });
+}
+
+void lower(Vunit& u, stublogue& i, Vlabel b, size_t z) {
+  lower_impl(u, b, z, [&] (Vout& v) {
+    // Push both the LR and FP regardless of i.saveframe to align SP
+    v << pushp{rfp(), rlr()};
```
mxw: Same question.

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -1328,55 +1275,145 @@ void lower(Vunit& unit, cvttsd2siq& i, Vlabel b, size_t idx) {
...
+
+///////////////////////////////////////////////////////////////////////////////
+
+void lower(Vunit& u, callstub& i, Vlabel b, size_t z) {
+  lower_impl(u, b, z, [&] (Vout& v) {
+    v << call{i.target, i.args};
+  });
+}
+
+void lower(Vunit& u, callfaststub& i, Vlabel b, size_t z) {
```
mxw: (I guess the heart of the issue is that if we ever changed the implicit defs of `callstub{}` and `callfaststub{}` to be anything but a subset of `call{}`, this would be a bug.  I think `callm{}` and `call{}` can be assumed to have the same implicit defs, though, so that one is fine.)

hphp/runtime/vm/jit/vasm-arm.cpp
dave-estes: Done.

hphp/runtime/vm/jit/vasm-arm.cpp
dave-estes: Done.

hphp/runtime/vm/jit/vasm-arm.cpp
dave-estes: Changed.

hphp/runtime/vm/jit/vasm-arm.cpp
dave-estes: Changed.

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -1328,55 +1275,145 @@ void lower(Vunit& unit, cvttsd2siq& i, Vlabel b, size_t idx) {
...
+  v << load{i.target, scratch};
+  v << callr{scratch, i.args};
+  });
+}
+
+void lower(Vunit& u, jmpm& i, Vlabel b, size_t z) {
+  lower_impl(u, b, z, [&] (Vout& v) {
+  lowerVptr(i.target, v);
+
+  auto scratch = v.makeReg();
```
dave-estes: Done.

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -1328,55 +1275,145 @@ void lower(Vunit& unit, cvttsd2siq& i, Vlabel b, size_t idx) {
   });
 }
 
+void lower(Vunit& u, callm& i, Vlabel b, size_t z) {
+  lower_impl(u, b, z, [&] (Vout& v) {
+  lowerVptr(i.target, v);
```
dave-estes: Done.

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -1328,55 +1275,145 @@ void lower(Vunit& unit, cvttsd2siq& i, Vlabel b, size_t idx) {
...
+
+  // Load the target from memory and then call it
+  v << load{i.target, scratch};
+  v << callr{scratch, i.args};
+  });
+}
+
+void lower(Vunit& u, jmpm& i, Vlabel b, size_t z) {
+  lower_impl(u, b, z, [&] (Vout& v) {
+  lowerVptr(i.target, v);
```
dave-estes: Done.

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -488,91 +478,62 @@ void Vgen::emit(const call& i) {
...
+  a->Mov(rAsm, i.exittc);
+  a->Stp(X(rvmfp()), rAsm, MemOperand(sp, -16, PreIndex));
 
-  a->Adr(rAsm, &End);
-  a->Str(rAsm, X(i.fp)[AROFF(m_savedRip)]);
-  a->Blr(X(i.target));
-  a->bind(&End);
-  // m_savedRip will point here.
-  emit(unwind{{i.targets[0], i.targets[1]}});
+  // Branch and link to nowhere to balance the LR stack
```
dave-estes: Done.

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -488,91 +478,62 @@ void Vgen::emit(const call& i) {
...
 
-void Vgen::emit(const callarray& i) {
-  emit(call{i.target, i.args});
-}
+void Vgen::emit(const calltc& i) {
+  vixl::Label stub;
 
-void Vgen::emit(const contenter& i) {
-  vixl::Label End;
+  // Preserve the FP and LR (exittc) on the stack.
```
dave-estes: May not need to, but we did it since we have to push in pairs. ARM has to push it's LR before doing any branch with link (bl/blr), and so it's typically done with FP in that pair making a new frame. `leavetc` below restores it...again although it might not need to. The previous implementation for `calltc` has the same comment, but it was pretty wrong as it pushed savedRIP along with LR.

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -488,91 +478,62 @@ void Vgen::emit(const call& i) {
...
+  a->B(&end);
 
-void Vgen::emit(const phpret& i) {
-  a->Ldr(X(rlr()), X(i.fp)[AROFF(m_savedRip)]);
-  if (!i.noframe) {
-    a->Ldr(X(i.d), X(i.fp)[AROFF(m_sfp)]);
-  }
-  emit(ret{});
-}
+  // We call into this stub from the end below. Take that
```
dave-estes: Done.

hphp/runtime/vm/jit/vasm-arm.cpp
dave-estes: Gotcha. Didn't think of that. Changed.

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -488,91 +480,72 @@ void Vgen::emit(const call& i) {
...
+  a->Bl(&stub);
   emit(unwind{{i.targets[0], i.targets[1]}});
 }
 
+///////////////////////////////////////////////////////////////////////////////
+
+void Vgen::emit(const calltc& i) {
+  vixl::Label stub;
+
+  // Preserve the FP and LR (exittc) on the stack.
```
mxw: I think it'd be better to push and pop `rAsm` instead, then.  We should only ever call `leavetc{}` from the top frame in a VM reentry, which means that `rvmfp()` should be the same as it was at the `calltc{}` anyway.  Doing this save/restore could hide incorrect uses of `leavetc{}` (and also I think it's just less ambiguous if we avoid doing unnecessary work, for later code archaeology purposes).

dave-estes: Sure thing. I'll try it out.

mxw:
Thanks!  I'll land as soon as the last tests finish up.

mxw:
@dave-estes—Oh, I realized just before I was about to land—this needs to be rebased on top of the `pushp{}/popp{}` change, which I just upstreamed (and I think some of the operands might need to be flipped, to match the new ordering).  The "Files changed" tab seems to be confused for this reason.  Can you give it a poke?

dave-estes:
@mxw: Done.

PR #7496, opened 2016-11-16
[AArch64] Remove remaining extraneous ARM vasm.

dave-estes:
Several ARM-specific vasm instructions were added for lowering, but
they are no longer needed. Additional cleanup includes:
  - Change complex emitters to lowerers.
  - Flag untested implementations with notYetTested assertions.

### DIFF
--- a/hphp/runtime/vm/jit/vasm-arm.cpp
+++ b/hphp/runtime/vm/jit/vasm-arm.cpp
@@ -195,6 +195,7 @@ struct Vgen {
   void emit(const unwind& i);
 
   // instructions
+  void emit(const absdbl& i) { a->Fabs(D(i.d), D(i.s)); }
   void emit(const addl& i) { a->Add(W(i.d), W(i.s1), W(i.s0), SetFlags); }
   void emit(const addli& i) { a->Add(W(i.d), W(i.s1), i.s0.l(), SetFlags); }
   void emit(const addq& i) { a->Add(X(i.d), X(i.s1), X(i.s0), SetFlags); }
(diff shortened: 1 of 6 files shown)


hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -1504,6 +1476,35 @@ Y(storewi, storew, ldimmw, i.s)
 
 #undef Y
 
+void lower(Vunit& u, cloadq& i, Vlabel b, size_t z) {
+  lower_impl(u, b, z, [&] (Vout& v) {
+    auto scratch = v.makeReg();
```
mxw: `auto const`

dave-estes:
@mxw -- There are several emitters/lowerers implemented in the tip that are never actually used. I felt a little uncomfortable, so by using the "not tested" macros and commenting out some of the lowerers, we'll now see assertions pop if these vasm are ever used in the future. That'll trigger some maintenance to go remove the assertions, but at least we know then that the implementation for that vasm is now "tested" ... albeit far from extensively.

mxw:
@dave-estes—I prefer either just implementing them in an obviously correct way by inspection, or just deleting them.  I have a mild preference for the former, but I understand your preference for defensiveness, too.  But leaving these implementations around only to assert seems unnecessary in the presence of source control.

Interestingly, I think some of these Vinstrs are never used by anyone, so I'm going to delete them in a separate diff...

dave-estes:
Sounds good, @mxw. I just axed the `notTested()` code with the new change. I haven't merged up, because I wanted to preserve the diff history, but once this change get's approved, I can rebase and push --force.

mxw:
Looping in the ppc64 folks: @gut, @lbianc, @racardoso.  Can you guys kill your `fabs{}` also?  It doesn't seem to serve any purpose except to obfuscate the implementation of `absdbl{}`.

racardoso:
@mxw Doing a rapid inspection seems we are using this vasm only on `absdbl{}` since I am re-factoring some vasms I will check if we can kill this vasm.

gut:
@mxw: we didn't notice the 05608721cd308b3a89aaa1be6ca17d154ccb209f change. Now the absdbl can be emited directly as ppc64's fabs as the register type matches now. 
Please organize who would make such a short change.

racardoso:
@mxw I check https://github.com/facebook/hhvm/commit/05608721cd308b3a89aaa1be6ca17d154ccb209f we can kill fabs. I already did it on `hphp/runtime/vm/jit/vasm-ppc64.cpp` and test works ok.

mxw:
@dave-estes—Can you rebase this diff?  It's good to go aside from the merge conflicts.

dave-estes:
Thanks, @mxw. It's done.

racardoso:
@dave-estes @mxw about ppc64 we can kill `fabs{}`should I open another pull request with the changes or paste de diff here and someone update this pull request?

mxw:
@racardoso—A new PR would be great.  I'm about to upstream this one—once you see it close here on GitHub, you can rebase onto master and then make your changes.

racardoso:
@mxw Ok great. Once you close this one I will rebase and send a PR.

mxw:
@racardoso—While you're there, please try to nix any other PPC64 vasm instructions that are not used, or only used for trivial lowering.  I just sampled three of them—`ldarx{}, `xxlxor{}`, and `xxpermdi{}`—and all three are unnecessary; there are probably many others that can be deleted.

racardoso:
@mxw Ok I'll check others vasms that can be killed and send it on the PR.

PR #7498, opened 2016-11-17
Fix TOC register restore due to new backchain fix on PPC64

gut:
The TOC register restoration was being handled in 2 different ways by 2
pull requests (7389 and 7327) but together there were 7 tests from the
quick test suite that started failing.

When exiting the VM, the r1 register is read in order to reload the r2
(toc register). Other than that, only the r31 is used as a frame pointer
in order to not only read from stack but only to pop the frame.

The TOC register had the same data before toc/relocation PR (7389) so
this bug was not affecting when fix backchain PR (7327) was integrated.

### DIFF
--- a/hphp/runtime/vm/jit/vasm-ppc64.cpp
+++ b/hphp/runtime/vm/jit/vasm-ppc64.cpp
@@ -865,10 +865,10 @@ void Vgen::emit(const stublogue& i) {
 
 void Vgen::emit(const stubret& i) {
   // rvmfp, if necessary.
-  if (i.saveframe) {
-    a.ld(rvmfp(), rsp()[AROFF(m_sfp)]);
-    a.mr(rsfp(), rvmfp());
-  }
+  if (i.saveframe) a.ld(rvmfp(), rsp()[AROFF(m_sfp)]);
+
+  // restore r1 appropriately
+  a.mr(ppc64_asm::reg::r1, rvmfp());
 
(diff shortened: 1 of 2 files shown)


gut:
I just noticed that despite solving the tests that were broken, the backchain is corrupted and that was the reason for PR 7327 ... hang on

gut:
PR text was updated.
@mxw : Please review.

hphp/runtime/vm/jit/vasm-ppc64.cpp
```diff
@@ -865,10 +865,10 @@ void Vgen::emit(const stublogue& i) {
...
 void Vgen::emit(const stubret& i) {
   // rvmfp, if necessary.
-  if (i.saveframe) {
-    a.ld(rvmfp(), rsp()[AROFF(m_sfp)]);
-    a.mr(rsfp(), rvmfp());
-  }
+  if (i.saveframe) a.ld(rvmfp(), rsp()[AROFF(m_sfp)]);
+
+  // restore r1 appropriately
+  a.mr(ppc64_asm::reg::r1, rvmfp());
```
mxw: Use `rsfp()` instead, so that this is more self-documenting.

gut: damn, sorry.

gut:
@mxw: updated. Thanks

Issue #32, opened 2010-02-22
Clustering not compiling right


chrisgraham:
I enabled cpp clustering on my build but it did not include references to the cpp directory in my cmake stuff, so I had to add it in manually.
I am compiling on latest Kubuntu.

Issue #16, opened 2010-02-21
compile error HPHP::Variant::Variant(ssize_t) cannot be overloaded


danielrmz:
I ran `make` and get the following error. 

[ 17%] Building CXX object src/CMakeFiles/hphp_runtime_static.dir/cpp/base/string_data.cpp.o
In file included from /home/daniel/workspace/hiphop-php/src/cpp/base/zend/zend_string.h:22,
                 from /home/daniel/workspace/hiphop-php/src/cpp/base/string_data.cpp:22:
/home/daniel/workspace/hiphop-php/src/cpp/base/type_variant.h:95: error: ‘HPHP::Variant::Variant(ssize_t)’ cannot be overloaded
/home/daniel/workspace/hiphop-php/src/cpp/base/type_variant.h:92: error: with ‘HPHP::Variant::Variant(int)’
make[2]: **\* [src/CMakeFiles/hphp_runtime_static.dir/cpp/base/string_data.cpp.o] Error 1
make[1]: **\* [src/CMakeFiles/hphp_runtime_static.dir/all] Error 2

PR #7499, opened 2016-11-18
Correction for load/store exclusive disassembly

swalk-cavium:
This brings forward some of the changes in vixl which has to do with
the disassembly of the loadStoreExclusive Aarch64 instructions.  It was
noticed that some instructions where not disassembled correctly when
examining the output from TRACE=printir:9

Before
====
    478     (03) IncProfCounter<1>
    479         Main:
    480               0x11a00030  d2800100              movz x0, #0x8
    481               0x11a00034  f2b4f000              movk x0, #0xa780, lsl #16
    482               0x11a00038  f2dfffe0              movk x0, #0xffff, lsl #32
    483               0x11a0003c  c85f7c12              unimplemented (Unimplemented)
    484               0x11a00040  f1000652              subs x18, x18, #0x1 (1)
    485               0x11a00044  c8127c12              unimplemented (Unimplemented)
    486               0x11a00048  35ffffb2              cbnz w18, #-0xc
    487

After
===
  (comments added manually)

    511     (03) IncProfCounter<1>
    512         Main:
    513               0x2aa00030  d2800100              movz x0, #0x8
    514               0x2aa00034  f2ae0800              movk x0, #0x7040, lsl #16 
    515               0x2aa00038  f2dfffe0              movk x0, #0xffff, lsl #32 
    516               0x2aa0003c  c85f7c12              ldxr x18, [x0]          // <<--- fixed
    517               0x2aa00040  f1000652              subs x18, x18, #0x1 (1) 
    518               0x2aa00044  c8127c12              stxr w18, x18, [x0]     // <<--- fixed
    519               0x2aa00048  35ffffb2              cbnz w18, #-0xc
    520 

Since code generation only generates these two instructions currently, the
others are not yet tested.  The remaining instructions will require an update to
the assembler.

The visitor for the Instrument class will count this as an "other".
The visitor for the Simulator class will report an unimplemented instruction (no change).

### DIFF
--- a/hphp/runtime/vm/jit/vasm-arm.cpp
+++ b/hphp/runtime/vm/jit/vasm-arm.cpp
@@ -109,6 +109,7 @@ vixl::MemOperand M(Vptr p) {
     assertx(p.disp == 0);
     return MemOperand(X(p.base), X(p.index), LSL, Log2(p.scale));
   }
+  assertx(p.disp >= -256 && p.disp <= 255);
   return MemOperand(X(p.base), p.disp);
 }
 
@@ -743,6 +744,7 @@ void Vgen::emit(const lea& i) {
     assertx(p.disp == 0);
     a->Add(X(i.d), X(p.base), Operand(X(p.index), LSL, Log2(p.scale)));
   } else {
+    assertx(p.disp >= -256 && p.disp <= 255);
     a->Add(X(i.d), X(p.base), p.disp);
   }
 }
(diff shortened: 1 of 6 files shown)


fredemmott:
It'd be good to make the pull request title more descriptive - it's what ends up being the git commit title after import.

swalk-cavium:
You're right of course.  I'll be more careful.

fredemmott:
You should be able to edit it with a button in the top right; if not, comment with what it should be and we'll change it :)

mxw:
Looks like this has some artifacts from another diff—can you rebase onto current master?  Looks good to me otherwise.

swalk-cavium:
Hi Max, 
Which is the artifact?  There was another specifier 's' in the disassembler.  Rather
than introduce extra diffs when/if we bring the latest vixl forward I added the specifier.

If that's not the issue, could you add more words?

mxw:
You're removing some assertions that you removed in a separate PR.

swalk-cavium:
My bad.  I'll fix.

swalk-cavium:
I think I've put the two lines back now.

PR #7502, opened 2016-11-19
[arm64] Fix geneneric emit() function signature.

cmuellner:
A recent change in the hhvm source tree broke the vasm module for aarch64.
This patch fixes the regression.

The problem manifests in assertion failures of the following form:

Assertion Failure: /mnt/hhvm-arm64/hhvm/hphp/runtime/vm/jit/vasm-arm.cpp:155: void HPHP::jit::{anonymous}::Vgen::emit(Inst&) [with Inst = HPHP::jit::store]: assertion `false' failed.

### DIFF
--- a/hphp/runtime/vm/jit/vasm-arm.cpp
+++ b/hphp/runtime/vm/jit/vasm-arm.cpp
@@ -150,7 +150,7 @@ struct Vgen {
 
   /////////////////////////////////////////////////////////////////////////////
 
-  template<class Inst> void emit(Inst& i) {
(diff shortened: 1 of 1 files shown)


jim-saxman:
Thanks @cmuellner. I confirmed this patch fixes Aarch64 JIT.

ptomsich:
@mxw : any chance we could get this merged, so the various AArch64 teams can all pick up this fix?

aorenste:
This looks pretty reasonable to me - I'll merge it.

PR #7516, opened 2016-11-30
[arm64] Const reference patches

cmuellner:
Two patches, which changes read-only arguments to be const references instead of copies
in the arm64 JIT code path.


mxw:
The `jit::` namespace structs like `Vreg`, `ConditionCode`, and `Vptr` should be passed by value and not by const ref—they're not going to grow more than a few registers in size.  Not sure how big/extensible the ARM structs are, but I suspect it's unlikely that passing by const ref will be better for them either.

mxw:
Closing without upstreaming since I don't think we want this, but feel free to ping here more if you feel otherwise.

PR #7522, opened 2016-12-02
specialize testb lowering to avoid extra zero extension

swalk-cavium:
Comparing side-by-side log files between x86_64 and Aarch64 shows there are
some inefficiencies.  For example,

/data/swalk/try007$ TRACE='printir:9,vasm:6' ../hhvm/hphp/hhvm/hhvm --file 'Spyc.php'

X86 produces two instructions.
X86_64
======
15572 B9: [profCount=1] (preds B0) 
15573 --- bc 14, fp 0, spOff 4 ()
15574 14: JmpNZ 7 (21)
15575 (40) JmpNZero t11:Bool -> B10<Unlikely>
15576 Main:
15577 0x1760089c: 84 c0 test %al, %al 
15578 0x1760089e: 0f 85 47 f8 5f 01 jnz 0x18c000eb
Aarch64 needs improvement even after the conditional branch is cleaned up.

Aarch64
======
102 15016 B9: [profCount=1] (preds B0) 
103 15017 --- bc 14, fp 0, spOff 4 ()
104 15018 14: JmpNZ 7 (21)
105 15019 (40) JmpNZero t11:Bool -> B10<Unlikely>
106 15020 Main:
107 15021 0x2fa008dc 53001c01 uxtb w1, w0
108 15022 0x2fa008e0 53001c00 uxtb w0, w0 // <<--
109 15023 0x2fa008e4 6a01001f tst w0, w1
110 15024 0x2fa008e8 540000a0 b.eq #+0x14
111 15025 0x2fa008ec 58000052 ldr x18, pc+8 (addr 0x2fa008f4)
112 15026 0x2fa008f0 d61f0240 br x18 
113 15027 0x2fa008f4 38e00128 unallocated (Unallocated)
114 15028 0x2fa008f8 00000000 unallocated (Unallocated)

After (hand edited)
===
  14982  B9: [profCount=1] (preds B0)
  14983     --- bc 14, fp 0, spOff 4 ()
  14984   14: JmpNZ 7 (21)
  14985     (40) JmpNZero t11:Bool -> B10<Unlikely>
  14986         Main:
  14987               0x258008dc  72001c1f              tst w0, #0xff   // <<---
  14988               0x258008e0  540000a0              b.eq #+0x14
  14989               0x258008e4  58000052              ldr x18, pc+8 (addr 0x258008ec)
  14990               0x258008e8  d61f0240              br x18
  14991               0x258008ec  2ec00128              unimplemented (Unimplemented)
  14992               0x258008f0  00000000              unallocated (Unallocated)
  14993 

I ran regression tests with 6 option sets, before/after.  No issues found.


mxw:
lgtm and all tests have passed.  cc/ @dave-estes, @cmuellnerTheobromaSystems—Speak up if you see any problem with this.

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -602,6 +603,10 @@ void Vgen::emit(const andli& i) {
   a->And(W(i.d), W(i.s1), MSKTOP(i.s0.l()), SetFlags);
 }
 
+void Vgen::emit(const testbi& i) {
+  a->And(vixl::wzr, W(i.s1), MSKTOP(0xff), SetFlags);
```
dave-estes: I think `testbi` is supposed to test an immediate `s0` against a register `s1`. `MSKTOP` only masks off the top 32 bits. I think you'll need to mask off the the upper 48 bits. Also for the sake of consistency maybe use `Tst` instead of `And` like `testli` below.

mxw:
Back to you to address @dave-estes's comments.

dave-estes:
Also, this change feels a bit like a peephole optimization placed in the lowerer. The right place might be to add it to vasm-simplify or another pass before lowerForARM. What do you think, @mxw?

mxw:
@dave-estes—Disagreed.  There's nothing peephole about this lowering; we're looking at the single instructing and nothing else: no other instructions, no use/def information.  Things like this, especially when platform-specific, should definitely be kept to architecture-specific lowerers.

Agreed that this change is currently incorrect, though :]

hphp/runtime/vm/jit/vasm-arm.cpp
swalk-cavium: What's the approved way for creating 8-bit constants?

swalk-cavium:
I think it would be more work to detect this further downstream.  This is the point
where the inefficiency is introduced.

hphp/runtime/vm/jit/vasm-arm.cpp
swalk-cavium: I think you're right about the Tst.  We could emit it here instead of waiting for the
assembler to recognize the alias.

dave-estes:
@mxw, yeah maybe not peephole, but it certainly smells of paltform-specific algebraic simplification. I'm OK with it staying in the lowerer, but I am a little concerned about too may more optimizations like this muddying up the lowerers/emitters. However, if it ever does get out of hand, then we can just refactor and work in a new pass.

hphp/runtime/vm/jit/vasm-arm.cpp
dave-estes: lpathy added `MSKTOP` originally. I'd recommend just adding a new function. While your at it, maybe we can rename these function from ALL-CAPS-MACRO-STYLE to something more verbose and camelCased. `Mask32` and `Mask8` maybe? What do you think?

mxw:
@dave-estes—Current precedent has all instruction-local optimizations handled in arch-dependent lower or emit routines (e.g., `x64::Vgen::emit(const testqi&)`), so I think we should stick with that at least for now.  It seems like a pretty arbitrary line, though.  What makes lowering `pushp{}` to two pushes different from lowering `testqi{}` depending on its immediate?  If it has something to do with there being a conditional: what about `stublogue{}`, whose implementation depends on the `saveframe` operand?

The boundary between the vimplifier and the vemitters/vlowerers is pretty clear to me: If it requires more than instruction-local knowledge, it goes in the former; else it's in the latter.  But I'm skeptical that there's much principled justification for splitting the emitters/lowerers further.  Not married to the status quo organization—I just want to make sure that if we ever change it, it's not just churn.

swalk-cavium:
Apparently I don't know how to fly git yet.  Looks like more changed than I intended.

swalk-cavium:
I agree with @dave-estes in that the shouting function name is odd.
We should clean those up all at once in some future change.

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -602,6 +603,10 @@ void Vgen::emit(const andli& i) {
   a->And(W(i.d), W(i.s1), MSKTOP(i.s0.l()), SetFlags);
 }
 
+void Vgen::emit(const testbi& i) {
+  a->Tst(W(i.s1), MSKTOP(0xff));
```
mxw: This is still wrong, in the way that @dave-estes mentioned.  As you have it, regardless of what you set `s0` to, it'll be essentially replaced with `0xff`.  You're basically making all of the following transformations:

```
testbi{0x0, r} -> testbi{0xff, r}
testbi{0x1, r} -> testbi{0xff, r}
testbi{0x2, r} -> testbi{0xff, r}
...
```

You're also still only masking the top 32 bits, instead of the top 56 bits as is necessary for correctness, in case there's junk in the upper 56 bits of `i.s1`.

swalk-cavium: This is the first use of the testbi.  Part of this issue is mapping the semantics
of the Vinstrs which were based off of x86 onto ARM.    You're right.  It's ignoring
s0.  

Previously, the assembler would eventually recognize the tst alias.  I thought the
previous comment was to emit the tst directly and fix the upper-case function names
later.

swalk-cavium:
Ok, I see.  Let me re-test.

hphp/runtime/vm/jit/vasm-arm.cpp
mxw: I'm not really sure why you guys need this MSKTOP thing—I assume it's because you need 32-bit negative values to zero-extend instead of sign-extend when widened to `int64_t`.  Is that correct, @dave-estes?

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -602,6 +603,10 @@ void Vgen::emit(const andli& i) {
   a->And(W(i.d), W(i.s1), MSKTOP(i.s0.l()), SetFlags);
 }
 
+void Vgen::emit(const testbi& i) {
+  a->Tst(W(i.s1), MSKTOP(i.s0.l()));
+}
```
mxw: Why not implement this inline with the other `test*{}` instructions?

I'm also still not sure I understand the semantics of `MSKTOP()` and why you're using `i.s0.l()` for byte instructions when you could be using `i.s0.b()`—but this is a more general question, not specific to this instruction in particular.  @dave-estes, can you explain in more detail?

swalk-cavium: It looks like the ones calling MSKTOP are not declared inline.

I'll change the method to .b() as you suggest and push again.

swalk-cavium: Looks like the other 'b' instructions use the .l() method.  See

    420 #define Y(vasm_opc, simd_w, vr_w, gpr_w, imm) \
    421 void Vgen::emit(const vasm_opc& i) {          \
    422   if (i.d.isSIMD()) {                         \
    423     emitSimdImmInt(a, i.s.simd_w(), i.d);     \
    424   } else {                                    \
    425     Vreg##vr_w d = i.d;                       \
    426     a->Mov(gpr_w(d), imm);                    \
    427   }                                           \
    428 }
    429 
    430 Y(ldimmb, ub, 8, W, MSKTOP(i.s.l())) // <<---
runtime/vm/jit/vasm-arm.cpp

mxw: Right, hence why I said

> but this is a more general question, not specific to this instruction in particular

I don't understand why the byte instructions use `l()` in general instead of `b()`.  I realize that they, in fact, do so.

swalk-cavium: I was wondering if it has something to do with the safe_cast business.
I didn't want to generate a sign extension incorrectly.

     83   int8_t  b() const { return safe_cast<int8_t>(m_int); }
     84   uint8_t ub() const { return safe_cast<uint8_t>(m_int); }
     85  
hhvm/hphp/util/immed.h

dave-estes: I think we _should_ use `.b()` here, and that we still need a new MSKTOP24. Using `b()` will ensure that the new value can be cast as a byte, but we still need to get rid of the upper 24 bits before testing it. `tst` is an alias of `ands`. We are using W registers here, but if there are sign bits in that second byte of the immediate, then `ands` will use them.

I think as it coded now, the implementation of `testbi` is incorrect.

dave-estes: Oh yeah, I missed `ub()`. That'll do the trick. Now lets get rid of MSKTOP altogether...in another PR of course. :-)

swalk-cavium: Replying to @mxw above, I think the whole point of MSKTOP as a function is to have an assertx.  The right  shifting is just replicating the sign bit.  The conditional is just checking
to make sure there wasn't anything specified in the upper bits.

The usual immediate operand to an AND instruction is unsigned.  So, in this case we could
do the same kind of function and check a different set of bits.  I'm not convinced it would
buy all that much.

I'll push the i.s0.bu() change so we can declare victory and move on to the next one.

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -602,6 +603,10 @@ void Vgen::emit(const andli& i) {
   a->And(W(i.d), W(i.s1), MSKTOP(i.s0.l()), SetFlags);
 }
 
+void Vgen::emit(const testbi& i) {
+  a->Tst(W(i.s1), i.s0.ub());
+}
+
 void Vgen::emit(const testli& i) {
   a->Tst(W(i.s1), MSKTOP(i.s0.l()));
 }
```
mxw: Can you move these guys to be inline in the struct definition, like the other `test` instructions?

dave-estes:
Looks like your last change may have picked up some additional work on accident.

swalk-cavium:
I'm not sure what you mean.  I moved it to inline per Max's suggestion. Added, committed,  then pushed the same branch.

https://github.com/facebook/hhvm/pull/7522/commits/24cc43f450fea03530ffca3a27d217b7191b7929

dave-estes:
I'm not 100% what I'm looking at either. I clicked the link from the email notification to see changes since last import and I saw some LoadStoreExclusive work. https://github.com/facebook/hhvm/pull/7522/files/592a8d356b0b97b6c53156aadbbd2a0140829770..24cc43f450fea03530ffca3a27d217b7191b7929

swalk-cavium:
OK, I guess I'm going to learn more git (the hard way).

swalk-cavium:
Ok, there was a merge I didn't expect when I pulled down my repo from github and started again.

mxw:
@swalk-cavium—If you go to the "Files changed" tab from this page, you'll see if you scroll down a bit that a lot of the LoadStoreExclusive code that you submitted for a separate PR has also been included in this PR.  Could you remove that code from this PR?

swalk-cavium:
Yes, working on it.  I was making sure I understood the sequence which got me
here so that I don't do it again.  I think the remaining clearcase ideology has been
removed.

swalk-cavium:
Is there anything else I need to do for this PR?

PR #7534, opened 2016-12-07
[test]: Fixing expected string of empty_filename.php.

cmuellner:
I've observed that the expected strings for empty_filename.php
seem to be wrong in the same way on the following machines:

 * x64: CentOS 7
 * aarch64: Ubuntu Xenial
 * aarch64: CentOS 7
 * aarch64: OpenSuse 42.2

This patch changes the expected string to match the output
on the named machines.


mxw:
This works fine for me on CentOS 7.  I'm guessing it's a permissions issue on your machine.  @paulbiss, @Orvid, @mofarrell, any thoughts?

mofarrell:
https://3v4l.org/qcBeM  I think we have it right.  If its not working properly on your machine, I think the right answer is to fix it not change the test here.

cmuellner:
I just tested on a fresh setup x64 Fedora 25 machine:

```
cm@cm:~$ php foo.php # PHP 7.0.13
PHP Warning:  chmod(): No such file or directory in /home/cm/foo.php on line 3
bool(false)
bool(false)
PHP Warning:  disk_free_space(): No such file or directory in /home/cm/foo.php on line 5
bool(false)
PHP Warning:  chmod(): No such file or directory in /home/cm/foo.php on line 6
bool(false)
bool(false)
PHP Warning:  disk_free_space(): No such file or directory in /home/cm/foo.php on line 8
bool(false)
PHP Warning:  chmod(): No such file or directory in /home/cm/foo.php on line 9
bool(false)
bool(false)
PHP Warning:  disk_free_space(): No such file or directory in /home/cm/foo.php on line 11
bool(false)

```

```
cm@cm:~$ ./src/hhvm/hphp/hhvm/hhvm foo.php 

Warning: chmod(): No such file or directory in /home/cm/foo.php on line 3
bool(false)
bool(false)

Warning: disk_free_space(): Operation not permitted in /home/cm/foo.php on line 5
bool(false)

Warning: chmod(): No such file or directory in /home/cm/foo.php on line 6
bool(false)
bool(false)

Warning: disk_free_space(): Operation not permitted in /home/cm/foo.php on line 8
bool(false)

Warning: chmod(): No such file or directory in /home/cm/foo.php on line 9
bool(false)
bool(false)

Warning: disk_free_space(): Operation not permitted in /home/cm/foo.php on line 11
bool(false)

```

cmuellner:
@mxw, could you please provide the git commit ID of your HHVM build, where this works for you?

swalk-cavium:
I'm seeing the same issue as @cmuellner on an Aarch64 platform.
The hphp/test/slow/ext_file/empty_filename.php test fails.

I have a build from earlier this week on x86_64 and it passes.

Aarch64
=====
  29698 read(7, "\0\0\365?\0\2\237\300\0\0\0\0\0\0\0\0", 16) = 16
  29699 fstat(7, {st_mode=S_IFREG|0644, st_size=176095232, ...}) = 0
  29700 newfstatat(AT_FDCWD, "/home/swalk/master/hhvm_111416/hhvm/hphp/hhvm/verify.hhbc-wal", 0xffffe803a768, 0) = -1 ENOENT (No such file or director        y)
  29701 fstat(7, {st_mode=S_IFREG|0644, st_size=176095232, ...}) = 0
  29702 fcntl(7, F_SETLK, {l_type=F_UNLCK, l_whence=SEEK_SET, l_start=0, l_len=0}) = 0
  29703 write(1, "\nWarning: disk_free_space(): Ope"..., 145
  29704 Warning: disk_free_space(): Operation not permitted in /home/swalk/master/hhvm_111416/hhvm/hphp/test/slow/ext_file/empty_filename.php on line         4

jjones-cavium:
(from an email to Steve Walk)
I'm currently looking at:

https://github.com/facebook/hhvm/blob/267e4ebb386f022bc135866afc5e1bd9a03f7caa/hphp/runtime/ext/std/ext_std_file.cpp

which is where disk_free_space is implemented. I note the following things:

- The return type is "Variant". That supports my supposition that the return type is either string or boolean.
- Examining File::TranslatePath indicates that an empty string is returned 
- The macro CHECK_BOOST returns a string or false. The string is derived from a BOOST error type: boost::system::error_code

My guess is that BOOST is generating different error messages between x86 and AArch64. In my opinion, if PHP really needs to report the same error messages across platforms, it should do one of several things:

- Do the translation from the pair {function being called, errno value} to string without relying on an external library
- Have an early check (during build setup?) that checks whether or not boost returns the expected strings
- Configure boost to "do the right thing", i.e. return error messages in a consistent format. Maybe the method message() of boost::system::error_code is the wrong one to use?

mxw:
But surely "No such file or directory" and "Operation not permitted" are semantically different errors?

@cmuellner—This works for me on ~all commit hashes.  It's definitely a local issue.

cmuellner:
@jjones-cavium , I've reproduced that on x64 (short for x86_64 or amd64).
As I don't have a single machine (arm64 or x86_64) producing the correct behavior, could you or @swalk-cavium please provide the strace output of a correct run?

@mxw, I guess you mean environment issue, because I would not call it a local issue, when I have observed it on three different arm64 Linux distributions, two x86_64 Linux distributions and people from Cavium observe this as well.

jjones-cavium:
The problem is that error_code, according to the BOOST source at:

https://github.com/boostorg/system/blob/develop/include/boost/system/error_code.hpp#L227

are system dependent, whereas error_condition is portable. CHECK_BOOST should be changed to use error_condition instead of error_code.

swalk-cavium:
For what it's worth, it looks like the current version may depend on the version of the boost
library.  The actual 'Operation not permitted' is from an EPERM down in libc.

152 Breakpoint 3, __GI___strerror_r (errnum=1, buf=0xffffffffd8a8 "", buflen=64) at _strerror.c:38
153 38      in _strerror.c
154 1: x/i $pc
155 => 0xffffb6574730 <__GI___strerror_r>:  stp     x29, x30, [sp,#-112]!
156 #0  __GI___strerror_r (errnum=1, buf=0xffffffffd8a8 "", buflen=64) at _strerror.c:38
157 #1  0x0000ffffb7eee398 in ?? () from /usr/lib/aarch64-linux-gnu/libboost_system.so.1.58.0
158 #2  0x0000ffffb7eeeda4 in ?? () from /usr/lib/aarch64-linux-gnu/libboost_system.so.1.58.0
159 #3  0x0000000004980290 in boost::system::error_code::message[abi:cxx11]() const (this=0xffffffffd9e0) at /usr/include/boost/system/error_code.hpp:357
160 #4  0x00000000071a0504 in HPHP::f_disk_free_space (directory=...) at /home/swalk/master/hhvm_111416/hhvm/hphp/runtime/ext/std/ext_std_file.cpp:1374
161 #5  0x000000000d8008f4 in ?? ()

swalk-cavium:
@cmuellner - How would I provide the strace output.  It's kind of large.

swalk-cavium:
Reduced test case
<?php
var_dump(disk_free_space(''));
?>

cmuellner:
@swalk-cavium, you could copy'n'paste it to pastebin.com and post the link here.

swalk-cavium:
Ok.  I though you might want the whole file.  Here's what I think is what the 
test case expects

http://pastebin.com/qez2ug2K

cmuellner:
Ok, it seems to be a behavior change in boost.
The problem can be reduced to the following lines:

```
//g++ mytest.cpp -lboost_filesystem -lboost_system
#include <iostream>
#include <string>
#include <boost/filesystem.hpp>
namespace fs = boost::filesystem;
int main()
{
	boost::system::error_code ec;
	fs::space_info sb = fs::space("", ec);
	if (ec) {
		std::cerr << ec.message() << std::endl;
		std::cerr << "value: " << ec.value() << std::endl;
	}
}
```

On Linux distros with a recent boost version (e.g. Fedora 25 with boost 1.60) this gives:

```
cm@cm:~$ ./a.out 
Operation not permitted
value: 1
```

On Linux distros with older boost versions (e.g. Debian 8 with boost 1.49) this gives:

```
cmuellner@purcell:~$ ./a.out 
No such file or directory
value: 2
```

I've tested also the new C++ filesystem API:

```
#include <iostream>
#include <string>
#include <experimental/filesystem>
namespace fs = std::experimental::filesystem;
int main()
{
	std::error_code ec;
	fs::space_info sb = fs::space("", ec);
	if (ec) {
		std::cerr << ec.message() << std::endl;
		std::cerr << "value: " << ec.value() << std::endl;
	}
}

```

...which outputs:
```
No such file or directory
value: 2
```

This is most likely a bug in the boost function `space_info space(const path& p, error_code* ec)`.
I've created a PR in the boostorg/filesystem repo (https://github.com/boostorg/filesystem/pull/39),
but I guess that will not propagate so quickly (if at all).

cmuellner:
I've update the PR to accept both error strings.

Although there is a PR for the bug in boost and C++'s,
it might need quite some time until the bugfix is
propagated down to all distros (if ever).

I personally don't care if the HHVM team is interested in merging such changes or not
(it boils down to the question how to deal with known environment bugs
affecting the own project and which environments are considered at all).
But I wanted to provide a solution for people which stumble over that in
the near future. And in the far future I guess this either gets
resolved by boost being fixed or HHVM using C++'s filesystem API.

Affected boost versions: at least 1.60
    
Affected Linux distributions: at least Ubuntu Xenial 16.04,
Fedora 25, and OpenSuse LEAP 42.2.

swalk-cavium:
Hi @cmuellner, 
I thought the bug was in the HHVM extension source as Joel mentioned earlier.
Not specficially in the boost source.  I haven't looked at how big a change it would
be.  See the links below for info.

http://stackoverflow.com/questions/16389931/how-to-use-error-condition-in-boostsystem-library
http://blog.think-async.com/2010/04/system-error-support-in-c0x-part-2.html

cmuellner:
@swalk-cavium, the boost::filesystem API offers two versions for each call. One of the form foo(a1, ...) and one of the form foo(a1, ..., e), where e is an error_code object.
The second form does not throw an exception in case of an error, but provides the specific error via the error_code object.

In the specific case the system call is vfstat() and the boost call is is space().

HHVM needs the specific error value of the call (i.e. ENOENT as provided by errno).

The problem now is that space() initializes the error_code object wrong
(in any error case the value will be 1 indicating EPERM).
And later the HHVM extension cannot check if the specific error was ENOENT,
as all information left indicates EPERM.

http://www.boost.org/doc/libs/1_60_0/libs/filesystem/doc/reference.html#space

cmuellner:
Just wanted to mention, that boost accepted my PR an merged it
(https://github.com/boostorg/filesystem/pull/39).
I don't know of any plans of backporting (neither to boost stable releases
nor to Linux distributions).

hphp/test/slow/ext_file/empty_filename.php.expectregex
```diff
@@ -0,0 +1,20 @@
...
+bool\(false\)
+
+Warning\: disk_free_space\(\)\: (No such file or directory|Operation not permitted) in [^\r\n]+\/test\/slow\/ext_file\/empty_filename.php on line 7
+bool\(false\)
+
+Warning\: chmod\(\)\: No such file or directory in [^\r\n]+\/test\/slow\/ext_file\/empty_filename.php on line 8
+bool\(false\)
+bool\(false\)
+
+Warning\: disk_free_space\(\)\: Operation not permitted in [^\r\n]+\/test\/slow\/ext_file\/empty_filename.php on line 10
```
mofarrell: @cmuellner If this was optional, like line 12, I think this would be good to land into master.

cmuellner:
@mofarrell thank's for the review!

PR #7538, opened 2016-12-08
Change return from a nullptr to bool to remove narrowing error.

dave-estes:
For AArch64, sizeof(bool) = 1 and sizeof(nullptr) = 8.


dave-estes:
BTW, @aorenste, someone mentioned that this might already be fixed internally to FB. I just couldn't verify, and so I didn't close this PR.

apinski-cavium:
I am running into this also ...

aorenste:
@dave-estes Yes - it's been fixed internally but we're having temporary trouble with our system which pushes changes back to github.

PR #7556, opened 2016-12-16
[Aarch64] Enable CRC32 instructions for hashing

jim-saxman:
This patch enables CRC32 instructions for the Aarch64 architecture.
It adds new assembly routines to hash-crc-arm.S which utilize the
CRC32 instructions to accelerate string hashing.

It adds a new cmake flag, enabled with -DENABLE_AARCH64_CRC=1, in the
same fashion as the existing ENABLE_AVX flag. This is required since
there are not compile time flags which indicate whether or not the
given CPU truly supports the CRC32 instructions, as they optional.

It renames NO_SSECRC to NO_HWCRC to make the code processor agnostic.

It moves the x64 assembly from the hash-crc.S file to hash-crc-x64.S.

Lastly, it implements hash_int64 and hash_int64_pair using the CRC
instructions.

### DIFF
--- a/hphp/tools/oss_hot_section_ordering
+++ b/hphp/tools/oss_hot_section_ordering
@@ -818,7 +818,7 @@
 .text.hash_string_i_unaligned_crc*
 .text._ZN4HPHP9ArrayData10IsValidKeyERKNS_7VariantE*
 .text._ZN4HPHP21tvMatchesRepoAuthTypeENS_10TypedValueENS_12RepoAuthTypeE*
-.text._ZN4HPHP18IsSSEHashSupportedEv*
+.text._ZN4HPHP17IsHwHashSupportedEv*
 .text._ZNK4HPHP6String5toKeyEv*
 .text._ZN4HPHP5VarNRC1EPNS_10StringDataE*
 .text._ZN4HPHP5VarNRC2EPNS_10StringDataE*
--- a/hphp/util/hash-crc-arm.S
+++ b/hphp/util/hash-crc-arm.S
@@ -1,0 +1,189 @@
+#include "hphp/util/etch-helpers.h"
+#include "hphp/util/hphp-config.h"
+
+#if defined(ENABLE_AARCH64_CRC) && !defined(NO_HWCRC) && \
+    !defined(__CYGWIN__) && !defined(__MINGW__) && !defined(_MSC_VER)
+
+.file   "hphp/util/hash-crc-arm.S"
+
+#define HASH_FUNC_NAME _ZN4HPHP20hash_string_i_unsafeEPKcj
+
+ETCH_SECTION(HASH_FUNC_NAME)
+.globl    HASH_FUNC_NAME
+ETCH_TYPE(HASH_FUNC_NAME, @function)
+ETCH_ALIGN16
+ETCH_NAME(HASH_FUNC_NAME):
+        CFI(startproc)
+        negs    w1, w1
+        mov     w9, #0xffffffff
+        beq     ETCH_LABEL(iend)
+        mov     w3, w1
+        mov     x4, #0xdfdfdfdfdfdfdfdf
+        b       ETCH_LABEL(iheader)
+
+ETCH_ALIGN16
+ETCH_LABEL(iloop):
+        crc32cx w9, w9, x2
+ETCH_LABEL(iheader):
+        ldr     x10, [x0], #8
+        adds    w3, w3, #8
+        mov     x2, x4
+        and     x2, x2, x10
+        bcc     ETCH_LABEL(iloop)
+        lsl     w3, w3, #3
+        lsl     x2, x2, x3
+        crc32cx w9, w9, x2
+
+ETCH_LABEL(iend):
+        lsr     w0, w9, #1
+        ret
+        CFI(endproc)
+
+ETCH_SIZE(HASH_FUNC_NAME)
+#undef HASH_FUNC_NAME
+
+
+#define HASH_FUNC_NAME _ZN4HPHP13hash_string_iEPKcj
+
+ETCH_SECTION(HASH_FUNC_NAME)
+.globl    HASH_FUNC_NAME
+ETCH_TYPE(HASH_FUNC_NAME, @function)
+ETCH_ALIGN16
+ETCH_NAME(HASH_FUNC_NAME):
+        CFI(startproc)
+        mov     w9, #0xffffffff
+        subs    w3, wzr, w1
+        beq     ETCH_LABEL(icend)
+
+        tst     w0, #7
+        mov     x4, #0xdfdfdfdfdfdfdfdf
+        bne     ETCH_LABEL(iuheader)
+        b       ETCH_LABEL(icheader)
+
+ETCH_ALIGN16
+ETCH_LABEL(icloop):
+        crc32cx w9, w9, x2
+ETCH_LABEL(icheader):
+        ldr     x10, [x0], #8
+        adds    w3, w3, #8
+        mov     x2, x4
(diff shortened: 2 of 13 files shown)


jim-saxman:
@swalk-cavium , Dave said you were interested in the CRC instructions and hashes. Can you please review and comment on this patch?

@binliu19 , I see you improved the generic hashing routines last month. Can you please review and comment on this patch?

hphp/tools/oss_hot_section_ordering
```diff
@@ -818,7 +818,7 @@
 .text.hash_string_i_unaligned_crc*
 .text._ZN4HPHP9ArrayData10IsValidKeyERKNS_7VariantE*
 .text._ZN4HPHP21tvMatchesRepoAuthTypeENS_10TypedValueENS_12RepoAuthTypeE*
-.text._ZN4HPHP18IsSSEHashSupportedEv*
+.text._ZN4HPHP17IsHwHashSupportedEv*
```
jim-saxman: How is this file generated? @binliu19 The hash-crc-x74.S assembly has the macro IF_SSE42() to define the function names, but those names didn't make it into this file? Should I add them? How?

hphp/util/hash.h
```diff
@@ -59,23 +64,33 @@ inline size_t hash_int64_fallback(int64_t key) {
...
+#if defined(USE_HWCRC) && defined(__SSE4_2__)
   // crc32 is commutative, so we need to perturb k1 so that (k1, k2) hashes
   // differently from (k2, k1).
   k1 += k1;
   __asm("crc32q %1, %0\n" : "+r" (k1) : "rm"(k2));
   return k1;
+#elif defined(USE_HWCRC) && defined(ENABLE_AARCH64_CRC)
+  size_t res;
+  k1 += k1;
+  __asm("crc32cx %w0, %w1, %x2\n" : "=r"(res) : "r"(k2), "r"(k1));
```
jim-saxman: I swapped k1 and k2 since k1 is now possibly 33 bits and the crc32cx instruction uses a 64-bit reg. for %x2.

jim-saxman:
@swalk-cavium Yes, I did have a runtime check in IsHwHashSupported that checked HWCAP, but the CRC bit was only just recently added to the ARM kernel. And we need to know at compile time, since the hash_int64* live in a header file.

jim-saxman:
This patch is dependent on #7460. I can't find a way to chain them together in GitHub, so you'll need to pull in #7460 to build this patch.

swalk-cavium:
Are there regression tests?  Or is the oss_hot_section_ordering the only check.
I seem to recall a 1-char difference in the polynomial described by the x86 instruction
and the Aarch64.

jim-saxman:
@swalk-cavium There are no regressions in quick, slow, zend/good, or the OSS performance test suite. 

ARM specifies two different polynomials. The instructions with 'c' use the same polynomial as x86. 

For hash_int64 and hash_int64_pair, I believe I can use either polynomial as long as they are the same between the two. They don't necessarily have to match crc32q x86, they just need to produce a good hash.

swalk-cavium:
My concern was since the hash is exposed at the php language level that someone
will put the hash in a database.  I only see one flavor of crc32 in 
hhvm/hphp/test/slow/ext_hash/ext_hash.php

Do you know which one that uses?

jim-saxman:
I think I understand where you are going. I didn't modify any hashes in hphp/runtime/ext/hash/. There are two #ifdef FACEBOOK's in ext_hash.cpp but not for crc32 nor crc32b. It used to be a few months ago that the functions now in hash-crc-x86.S were also under a FACEBOOK define, but that's not the case now.

I believe from inspection that the test/slow/ext_hash/ext_hash.php test uses a big lookup-table driven version of CRC, and not the hash-crc*.S versions, so I'm not sure which polynomial that unit test uses. In any case, I will put this on a debugger and verify.

hphp/util/hash-crc-arm.S
```diff
@@ -0,0 +1,189 @@
...
+ETCH_NAME(HASH_FUNC_NAME):
+        CFI(startproc)
+        negs    w1, w1
+        mov     w9, #0xffffffff
+        beq     ETCH_LABEL(iend)
+        mov     w3, w1
+        mov     x4, #0xdfdfdfdfdfdfdfdf
+        b       ETCH_LABEL(iheader)
+
+ETCH_ALIGN16
```
apinski-cavium: We should have the alignment there.  That is the inner loop should be aligned so the start of the loop is aligned for the fetch units of the common processors.  The ones Cavium has loads 16 byte aligned from the data cache IIRC.  So this is a good thing for our processors.    Now I noticed this is a straight copy of the x86 code which might not be a good thing.

jim-saxman: It works slightly to our advantage as well which is why I replicated it to the other inner loops in this file.  But I want to make sure it doesn't seriously disadvantage someone's core.

apinski-cavium: Looking into GCC's tuning flags, I notice the following:
Some loop tunings are 16 byte aligned and others are 8 byte aligned.  I know there was some discussion on the GCC mailing list about this too.  For ThunderX even though it is 8 byte aligned for gcc, the fetcher happens on a 16 byte boundary.  What I found for ThunderX was for some benchmarks 16 byte was better than 8 bytes but I did not dig far enough into why.  I think 16 byte is reasonable here and might only cause a few cycle hit on some cores.

jim-saxman: Hmm, the Cortex-A72 and A73 appear to have fetchers which are 3 instructions wide [A72](http://www.anandtech.com/show/10347/arm-cortex-a73-artemis-unveiled/2) [A73](http://techreport.com/review/28189/inside-arm-cortex-a72-microarchitecture).

swalk-cavium:
I looked at the source a bit more.  I don't see any place where this hash function would be directly visible in PHP/HACK.  I think we would be consistent with the x64 hash_string_i_unsafe
call.  I don't think I have any other comments.

mxw:
@binliu19—Could you take a look at this, and share any thoughts you might have?

Everyone else—Sorry for the delayed response; I thought I'd already pinged @binliu19, who is definitely the right person to look at this change.

binliu19:
Sorry for dropping this.. Somehow I didn't see an email notification on it.  Thanks @mxw for pinging me on IRC.  I will take a look.

hphp/tools/oss_hot_section_ordering
binliu19: I don't think we often write to this file directly.  The file should be generated from profiling after building.  Since the function is hot, it should be picked up next time we update this list.  On the other hand, it is OK to edit here directly.

jim-saxman:
@binliu19 Thanks for the comment. I was pretty sure it was an auto-generated file.

mxw:
@binliu19 has given this the vocal stamp of approval.  I'm just waiting for a clean internal test run, but there's no more action to be taken here (except for nagging me if it doesn't get landed fairly soon).

jim-saxman:
@mxw I have a second version of this patch which incorporates @apinski-cavium review comment. I can't push it yet, I still need to verify its performance. Shall we wait a day until get the results and then @binliu19 can review the second version too?

jim-saxman:
@mxw @apinki-cavium @binliu19. Sorry, Max, I thought you were commenting on #7602. Please ignore my last comment -It's totally irrelevant to this patch - and I'll nag you if it doesn't get landed fairly soon. Thanks.

mxw:
Sorry for the delay—there were some internal build issues to slough through.  This has now landed.  Thanks for your patience, all!

PR #7562, opened 2016-12-20
Aarch64 - Add a few more folder methods

swalk-cavium:
There were a few opportunities for additional folder methods in Aarch64.

I ran the regression suite with six option sets.  No new regressions detected.

echo "Test Started - `date`"

HPHP_TEST=all
PATH=`pwd`/hphp/hhvm:${PATH}
 
echo "/usr/bin/php hphp/test/run -v -m interp ${HPHP_TEST}"
      /usr/bin/php hphp/test/run -v -m interp ${HPHP_TEST} || true

echo "/usr/bin/php hphp/test/run -v -m interp -r ${HPHP_TEST}"
      /usr/bin/php hphp/test/run -v -m interp -r ${HPHP_TEST} || true

echo "/usr/bin/php hphp/test/run -v -m jit -a '-vEval.JitPGO=0' ${HPHP_TEST}"
      /usr/bin/php hphp/test/run -v -m jit -a '-vEval.JitPGO=0' ${HPHP_TEST} || true

echo "/usr/bin/php hphp/test/run -v -m jit -r -a '-vEval.JitPGO=0' ${HPHP_TEST}"
      /usr/bin/php hphp/test/run -v -m jit -r -a '-vEval.JitPGO=0' ${HPHP_TEST} || true

echo "/usr/bin/php hphp/test/run -v -m jit -a '-vEval.JitPGO=1' ${HPHP_TEST}"
      /usr/bin/php hphp/test/run -v -m jit -a '-vEval.JitPGO=1' ${HPHP_TEST} || true

echo "/usr/bin/php hphp/test/run -v -m jit -r -a '-vEval.JitPGO=1' ${HPHP_TEST}"
      /usr/bin/php hphp/test/run -v -m jit -r -a '-vEval.JitPGO=1' ${HPHP_TEST} || true

echo "Test Ended - `date`"

PR #7563, opened 2016-12-21
Correct invalid test case assumption

swalk-cavium:
This is a similar issue to 
https://github.com/facebook/hhvm/pull/7562

strncmp() returns less than, equal to, or greater than zero.  This test case assumes
it is implemented identically across various hosts.  That is not a safe assumption.

Before change (Aarch64)
===============

swalk@1scrb-7:~/master/hhvm_120516/hhvm$ /usr/bin/php hphp/test/run -v hphp/test/zend/good/ext/standard/tests/strings/strncmp_variation6.php
Warning: $_SERVER/$_ENV variables not available, please check
your ini setting: variables_order, it should have both 'E' and 'S'
You are using the binary located at: ./hphp/hhvm/hhvm
Running 1 tests in 1 threads (0 in serial)
hphp/test/zend/good/ext/standard/tests/strings/strncmp_variation6.php FAILED (3.80s)

After change (Aarch64)
==============

swalk@1scrb-7:~/master/hhvm_120516/hhvm$ /usr/bin/php hphp/test/run -v hphp/test/zend/good/ext/standard/tests/strings/strncmp_variation6.php
Warning: $_SERVER/$_ENV variables not available, please check 
your ini setting: variables_order, it should have both 'E' and 'S' 
You are using the binary located at: ./hphp/hhvm/hhvm
Running 1 tests in 1 threads (0 in serial)
hphp/test/zend/good/ext/standard/tests/strings/strncmp_variation6.php passed (3.78s)

All tests passed.

After change (x86_64)
==============

swalk@xxx-lnx1:/data/swalk/hhvm$ hphp/test/run --verbose --threads 1 'hphp/test/zend/good/ext/standard/tests/strings/strncmp_variation6.php'
Warning: $_SERVER/$_ENV variables not available, please check
your ini setting: variables_order, it should have both 'E' and 'S'
You are using the binary located at: ./hphp/hhvm/hhvm
Running 1 tests in 1 threads (0 in serial)
hphp/test/zend/good/ext/standard/tests/strings/strncmp_variation6.php passed (1.25s)

All tests passed.


swalk-cavium:
@mxw - new pull request

hphp/test/zend/good/ext/standard/tests/strings/strncmp_variation6.php
```diff
@@ -25,7 +25,7 @@
 /* Checking with the out of range ASCII value(given in binary format) */ 
 $str1 = decbin(256);
 $str2 = decbin( ord( chr(256) ));
-var_dump( strncmp($str1, $str2, 8) );  //comparing all the 8-bits; expected: int(1)
+var_dump( strncmp($str1, $str2, 8) !== 0 );  //comparing all the 8-bits; expected: int(1)
```
mxw: How about `> 0` instead?

mxw:
Makes sense.  I'm not sure how much we try to keep these zend tests in sync with upstream(?) though.  @paulbiss, @Orvid, @mofarrell—Do you guys know?

mofarrell:
@mxw @swalk-cavium As far as I know we keep all the changes in a script https://github.com/facebook/hhvm/blob/master/hphp/test/tools/import_zend_test.py
This seems kind of like an artifact of the way things were done.  I don't know that anyone regularly reimports the zend tests.

aorenste:
@mxw - Oops - I saw Aarch64 and assigned you blindly.  Correct - we generally try to keep those tests in sync with zend.  A recent version of php7 still has the test the same way - but looking at the code they return the result of C's memcmp() so this test certainly seems bogus.

swalk-cavium:
@aorenste - I'm not sure what you mean.  Am I putting the changes in the wrong place?
I'm trying to make them all green for Aarch64 and it looks like we're down to the system configuration/bad test level.  I'll have another one shortly.

mofarrell:
@swalk-cavium I think the change is fine.  We just would like it reflected somewhere in our [import script](https://github.com/facebook/hhvm/blob/master/hphp/test/tools/import_zend_test.py).  This test is a zend test, and is imported.  We maintain a script that can import all of zend's tests, and apply the reasonable transformations we need to make them pass.

mxw:
@aorenste—Nah, that's cool; I just happened not to know how zend test import works here.

@swalk-cavium—Give a shout if what @mofarrell is saying is still unclear to you.

aorenste:
@swalk-cavium You're not putting the changes in the wrong place.  The issue is that the tests in the zend directory are directly imported from the zend php source so we try to avoid modifying them - however this test is clearly wrong - your fix is correct.  The thing we need to figure out is do we accept your change as is or do we disable the test.  I'm leaning toward accepting the change as is.

mxw:
@aorenste—Why not do what @mofarrell suggested, and update the importer script so that it patches in the change?

swalk-cavium:
Any other magic scripts I need to know about?

swalk-cavium:
Hi,  I took a stab at modifying the import script.  It's not clear to me if the script will
recreate the corresponding .expectf file.  After much installing of packages I'm not sure
I have the corresponding zend installed so I was not actually able to run the modified script.

hphp/test/tools/import_zend_test.py
```diff
@@ -1366,6 +1366,8 @@ def parse_headers(zend):
         test = test.replace('EmptyFile.txt', 'md5_EmptyFile.txt')
     if 'ext/standard/tests/strings/lcfirst.php' in full_dest_filename:
         test = test.replace('dummy.txt', 'dummy-lcfirst.txt')
+    if 'ext/standard/tests/strings/strncmp_variation6.php' in full_dest_filename:
+        test = re.sub(r"^var_dump\( strncmp\(\$str1, \$str2, 8\) \)', 'var_dump( strncmp($str1, $str2, 8) > 0)', test)
```
mofarrell: This doesn't have matched quotes.  Can you also confirm this functions properly?

hphp/test/zend/good/ext/standard/tests/strings/strncmp_variation6.php.expectf
```diff
@@ -771,6 +771,6 @@ int(0)
 int(0)
 
 -- Checking with out of character's range, given in binary format --
-int(1)
+bool(true)
```
mofarrell: This also needs to be handled in the script.  Sorry didn't notice this last time.  See something like https://github.com/swalk-cavium/hhvm/blob/1b3401bb9ccbbd2dd5f2c61059270ea54ece9c95/hphp/test/tools/import_zend_test.py#L1381

hphp/test/zend/good/ext/standard/tests/strings/strncmp_variation6.php.expectf
```diff
@@ -771,6 +771,6 @@ int(0)
...
 
-*** Done ***
```
mofarrell: If we could not add a newline at the end of the file that would be nice, but not a big deal.

hphp/test/zend/good/ext/standard/tests/strings/strncmp_variation6.php.expectf
swalk-cavium: Ok.  I think the RE will have to depend on the preceding comment.  Lots of 'int(0)' in
the file.

hphp/test/zend/good/ext/standard/tests/strings/strncmp_variation6.php.expectf
swalk-cavium: This issue should go away if .expectf is regenerated each time, right?

swalk-cavium:
Ok, newbie question.  Where exactly are these tests coming from?  I downloaded the Zend framework but I don't see any of these files.  Even before any suffix changes.

mofarrell:
https://github.com/php/php-src/blob/6053987bc27e8dede37f437193a5cad448f99bce/ext/standard/tests/strings/strncmp_variation6.phpt
They are all in the php source tree.

hphp/test/tools/import_zend_test.py
```diff
@@ -1366,6 +1366,10 @@ def parse_headers(zend):
         test = test.replace('EmptyFile.txt', 'md5_EmptyFile.txt')
     if 'ext/standard/tests/strings/lcfirst.php' in full_dest_filename:
         test = test.replace('dummy.txt', 'dummy-lcfirst.txt')
+    if 'ext/standard/tests/strings/strncmp_variation6.php' in full_dest_filename:
+        test = re.sub(r'^var_dump\( strncmp\(\$str1, \$str2, 8\) \)', 'var_dump( strncmp($str1, $str2, 8) > 0)', test)
+        exp  = re.sub(r'range, given in binary format --\nint\(1\)', 'range, given in binary format --\nbool(true)', exp)
```
mofarrell: In general we try to keep lines under 80 chars in all files.  All of the style guidelines are [here](https://github.com/facebook/hhvm/blob/master/hphp/doc/coding-conventions.md), and mostly aimed towards C++ development.  Not that this is actually in that guide, but our linter also yells about the 2 spaces before the assignment operator.

hphp/test/tools/import_zend_test.py
```diff
@@ -1366,6 +1366,10 @@ def parse_headers(zend):
         test = test.replace('EmptyFile.txt', 'md5_EmptyFile.txt')
     if 'ext/standard/tests/strings/lcfirst.php' in full_dest_filename:
         test = test.replace('dummy.txt', 'dummy-lcfirst.txt')
+    if 'ext/standard/tests/strings/strncmp_variation6.php' in full_dest_filename:
+        test = re.sub(r'^var_dump\( strncmp\(\$str1, \$str2, 8\) \)', 'var_dump( strncmp($str1, $str2, 8) > 0)', test)
+        exp = re.sub(r'range, given in binary format --\nint\(1\)', 'range, given in binary format --\nbool(true)', exp)
```
mofarrell: The line length is still longer than 80 chars.

aorenste:
@swalk-cavium The tests are named .phpt in the original zend source - for example this file is ["ext/standard/tests/strings/strncmp_variation6.phpt"](https://github.com/php/php-src/blob/master/ext/standard/tests/strings/strncmp_variation6.phpt).  Our script splits the zend file out into separate files for each section.

PR #7569, opened 2016-12-28
bad assumption about proc_nice results

swalk-cavium:
proc_nice_basic.php is the only test which depends on the relative priority difference
before and after a call to proc_nice().  The other tests in the suite make sure the
function call accepts various parameters.

The test was passing on the CI @ http://hhvm-arm64.theobroma-systems.com/index-fb.html,
4 core machine, but consistently failing in our internal runs on 48-core machines.
It would pass when run by hand, so I suspected it had something to do with system load.

After instrumenting a version of the test to print out more information it became clear
that the test was assuming that a call to proc_nice would always change priority by a known
delta.  This is not the case if this test or another one that calls proc_nice has already been
run.  It just happened in our environment that one of the hhvm servers already run a test
with a call to proc_nice, so the starting priority was lower.

I was able to demonstrate this failure with the following command which runs the test
on a single thread multiple times.

/usr/bin/php ./run --relocate 3 zend/good/ext/standard/tests/general_functions/proc_nice_basic.php

Running the replay in the report shows that only the first one passes.

../hhvm/hhvm -c /home/swalk/master/hhvm_120516/hhvm/hphp/test/zend/config.ini -vEval.EnableArgsInBacktraces=true -vEval.EnableIntrinsicsExtension=true -vRepo.Local.Mode=-- -vRepo.Central.Path=/home/swalk/master/hhvm_120516/hhvm/hphp/hhvm/verify.hhbc -vEval.Jit=true   -vEval.ProfileHWEnable=false  --count=6 -vEval.JitAHotSize=6000000 -vEval.HotFuncCount=0 -vEval.PerfRelocate=3 -vResourceLimit.CoreFileSize=0    --file 'zend/good/ext/standard/tests/general_functions/proc_nice_basic.php'
bool(true)
bool(false)
bool(false)
bool(false)
bool(false)
bool(false)

I don't think this test is adding anything that is not tested elsewhere.


hphp/test/tools/import_zend_test.py
```diff
@@ -81,6 +81,9 @@
     # Zend's "run-tests" uses the cgi-fcgi sapi to run tests
     # our implementation is simply different by design
     '/ext/standard/versioning/php_sapi_name_variation001.phpt',
+
+    # bad assumption about system call effect
+    'zend/good/ext/standard/tests/general_functions/proc_nice_basic.php',
```
mofarrell: I feel like this would be better off as a flaky test.  In any case the actual file location should be changed or deleted as well (depending on whether the test is marked `flaky`, or `no_import`).

Also I believe the path here is in the PHP repo.  So in this case:
`/ext/standard/tests/general_functions/proc_nice_basic.phpt`  (note the `t` at the end)

swalk-cavium:
I can move the entry to the flaky table and adjust the file name.  Not sure
what you mean by deleting the actual file location.  Remove the file from
where it was put in the hhvm repo?  And the generated .expectf / .noserver files?

hphp/test/tools/import_zend_test.py
```diff
@@ -81,6 +81,7 @@
     # Zend's "run-tests" uses the cgi-fcgi sapi to run tests
     # our implementation is simply different by design
     '/ext/standard/versioning/php_sapi_name_variation001.phpt',
+
```
mofarrell: Don't add a line here if there is no change.

hphp/test/tools/import_zend_test.py
```diff
@@ -334,6 +335,9 @@
     '/ext/standard/tests/general_functions/getservbyname_variation9.php',
     '/ext/standard/tests/general_functions/getservbyname_variation10.php',
     '/ext/standard/tests/general_functions/getservbyport_variation1.php',
+
+    # bad assumption about system call effect
+    '/ext/standard/tests/general_functions/proc_nice_basic.php',
```
mofarrell: The actual file needs to be moved from `good/` to `flaky/` as well, otherwise this will not take effect until the next time the tests are imported.

swalk-cavium: Does this include the .expectf , .skipif, and .noserver files too?

mofarrell: Yes.

PR #7576, opened 2017-01-04
Fix PIC register clobbered by r30 error on gcc.

racardoso:
Summary:

On gcc 5.4.1 and above we cannot clobber r30 since it's used
as PIC register on PPC64. We still need to save it on
CALLEE_SAVED_BARRIER on gcc versions bellow 5.4.1 since it's
a callee saved register. On gcc 5.4.1 and above we don't need
to explicit save this register on CALLEE_SAVED_BARRIER since
this register is reserved by gcc abi.

Note that we can still clobber r30 if we pass -fno-pic option
to gcc.


brunoalr:
Note that this is a PPC64 issue.

PR #7586, opened 2017-01-07
Adding common interface implemented by Map and Vector that is mutable and iterable

acrylic-origami:
`MutableMap` and `MutableVector` both already implement `KeyedContainer` and `IndexAccess`. The extends `KeyedTraversable`, while the latter gives these classes basic mutability through `set[All]` and `removeKey`.

Currently, the only way to admit both `Map` and `Vector` to an immutable wrapper with mutable descendants is to upper-bound the collection generic type by `ConstIndexAccess` and refine to a mutable `IndexWrapper` in the descendants. For example:

```php
<?hh // strict
class ImmutableWrapper<Tk, +Tv, +TCollection as \ConstIndexAccess<Tk, Tv>> {
	public function __construct(private TCollection $collection) {}
	public function get_collection(): TCollection {
		return $this->collection;
	}
}
class MutableWrapper<Tk, Tv, +TCollection as \IndexAccess<Tk, Tv>> extends ImmutableWrapper<Tk, Tv, TCollection> {
	public function mutate(Tk $k, Tv $v): void {
		$this->get_collection()->set($k, $v); // for example
	}
}
```

However, without iterability, working with the `IndexAccess` family is cumbersome, and attempts to add iterability get messy. Keeping track of keys in a separate iterable collection, for example, leads quickly to generic hell, especially when the mutability of this collection is considered.

Of the `Traversable` descendants, I'm proposing `KeyedContainer` because it and `Indexish` are the only common ancestors of `Map` and `Vector` to define an indexing method — in this case, the square bracket syntax. `Indexish` is also a good candidate, but I figure that in the future, some core Hack classes might extend `KeyedContainer` but not `Indexish`, judging from Typing[4005] (the [illegal-use-of-square-bracket-indexing error](https://github.com/facebook/hhvm/blob/35c6d1ef834bf2cb81c4ab206c786e475968db0d/hphp/hack/src/utils/errors.ml#L1669)).

### DIFF
--- a/hphp/hack/hhi/collections/interfaces.hhi
+++ b/hphp/hack/hhi/collections/interfaces.hhi
@@ -270,6 +270,19 @@ interface MapAccess<Tk, Tv> extends ConstMapAccess<Tk, Tv>,
 }
 
 /**
+ * Represents a write-enabled (mutable) sequence of key/value pairs
+ * that can be both indexed using square-bracket syntax (e.g.
+ * `$keyed_container[key]`) and mutated.
+ *
+ * This interface provides no new methods, but it combines the
+ * iterability and mutability of some Hack collections into a common
+ * superclass.
+ */
+interface MutableKeyedContainer<Tk, Tv> extends IndexAccess<Tk, Tv>,
+                                                KeyedContainer<Tk, Tv> {
+}
+
+/**
  * Represents a read-only (immutable) sequence of values, indexed by integers
  * (i.e., a vector).
  *
@@ -530,8 +543,8 @@ interface ConstVector<+Tv> extends ConstCollection<Tv>,
  * @guide /hack/collections/interfaces
  */
 interface MutableVector<Tv> extends ConstVector<Tv>,
-                                    Collection<Tv>,
-                                    IndexAccess<int, Tv> {
+                                    MutableKeyedContainer<int, Tv>,
+                                    Collection<Tv> {
   /**
    * Returns a `MutableVector` containing the values of the current
    * `MutableVector`. Essentially a copy of the current `MutableVector`.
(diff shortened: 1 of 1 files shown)


dlreeves:
KeyedContainer was introduced to be a union of Hack Collections and PHP arrays. This requires adding special support in the runtime to have arrays behave as instances of KeyedContainer. Do you imagine arrays also being instances of MutableKeyContainer?

acrylic-origami:
@dlreeves That's a good point. I don't include them because they don't implement `IndexAccess`. Functionally, they definitely fit the concept behind `MutableKeyedContainer`. However, there are no interfaces for square-bracket syntax for mutation.

`MutableKeyedContainer` would actually be in the right place to be that interface, if it ditched the extension of `IndexAccess` and allowed setting via square brackets behind the scenes.

For now, I figured `Map` and `Vector` was a good start for `MutableKeyedContainer.

fredemmott:
@sgolemon was thinking of this a long while back - any comments, and still have a copy of the diagram?

dlreeves:
If it doesn't support arrays then I wonder if a mutable version of Indexish would be better. At one point we allowed Indexish to be modified with `[]` in the type checker, but removed it since the immutable collections also implemented this interface.

acrylic-origami:
@dlreeves Are you proposing reviving this historic mutable `Indexish` implementation? I'm not sure I follow your rationale that `Indexish` is more appropriate because of arrays, when arrays implement `Indexish` as well. Could you elaborate?

Also, did this mutable `Indexish` include a mechanism to remove elements? That's the one functional thing `IndexAccess` would implement over a mutable `Indexish` as it stands.

dlreeves:
@acrylic-origami - Ok took a look at the type hierarchy again. Would having `IndexAccess` extend `Indexish` solve this particularly issue for you? That way you have the mutable methods available via `IndexAccess`, but can access values using square-brackets. You still wouldn't be able to write using square-brackets though.

acrylic-origami:
@dlreeves Well, `Indexish` can't be usefully extended by custom interfaces, but `IndexAccess` definitely can.  So, I suggested they stay separate and are combined by a common descendant to avoid misleading upcasts of custom `IndexAccess`es to `KeyedContainer`. That way, the only and more correct option would be to upcast to `ConstIndexAccess` to achieve that functionality.

mofarrell:
https://pbs.twimg.com/media/C_pNhu3XgAAibNw.jpg

PR #7587, opened 2017-01-07
Ignore .hhconfig for Hack typechecker

vjon:
Update .gitignore to include .hhconfig file, necessary for using the Hack typechecker (which is in turn necessary for several of the built-in tests to pass).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.hhbc
 .mkdir
 .DS_Store
+.hhconfig
 hphp.log
 /build
 /deps


mofarrell:
I'm a little confused why we need to ignore hhconfig files.  Are tests failing with the config files?  We currently have several hhconfig files throughout the project.  Do you need to delete those to make tests pass?

vjon:
Sorry, my wording the first time was a little vague, let me try again.

When building the master branch, I encountered multiple errors stating that the Hack typechecker was required for a number of tests in the `slow` test suite. Thus, following the documentation to enable the typechecker requires the developer to create a `.hhconfig` file in the project root directory.

However, as the `.gitignore` in the root directory doesn't currently ignore this new file, the file will continue to appear in the git staging area when the developer begins working in other parts of the code base. Simply adding `.hhconfig` to the top-level `.gitignore` fixes this.

mofarrell:
The tests that require the typechecker should all have an hhconfig setup by the test suite.

From https://github.com/facebook/hhvm/blob/master/hphp/test/README.md:
> .php.hhconfig - A blank or syntactically valid Hack typechecker configuration file if you want the test to be able to be run in typechecker mode.

I'm curious what the errors said exactly.  I don't know anyone else who requires this to be able to run the tests.

vjon:
Hmm, I've been unable to reproduce this with a fresh copy of the repository.

I'll close this Issue now; hopefully, if this happens again, I'll have more specific information regarding the cause.

PR #7607, opened 2017-01-18
test: Cleanup after running 1775.php.

cmuellner:
After creating the file /tmp/lock.txt it should be deleted.


mofarrell:
Could you make this use a random generated filename?  Like here:
https://github.com/facebook/hhvm/blob/master/hphp/test/slow/ext_file/fputs.php

In theory we want the tests to be able to run in parallel with each other.

PR #7608, opened 2017-01-18
test: Cleanup after running eval.php.

cmuellner:
After creating the file /tmp/remote_eval.test it should be deleted.


mofarrell:
Same as https://github.com/facebook/hhvm/pull/7607
A random filename if possible.

mofarrell:
It looks like changing this file will require modification of [eval.php.test.in](https://github.com/facebook/hhvm/tree/master/hphp/test/slow/ext_xdebug/remote/eval.php.test.in)
It adds a breakpoint on the last line of the file (which has changed).

cmuellner:
@mofarrell : Sorry for not properly testing before pushing.

I've changed the PR to set the breakpoint to the fclose() line (like it was before), i.e. line 46. If you prefer the breakpoint to be on the last line (the unlink()), then I will update the PR again.

PR #7619, opened 2017-01-24
Fix failing test cases, triggered by missing optional member-reflection data.

cmuellner:
There are three failing test cases (all archs, all build types), which are caused by
a warning message of the member-reflection module (in case the member-reflection
data is not embedded in HHVM):

* config_overrides.php
* tier_overrides_default.php
* tier_overrides_tier.php

This PR fixes those tests.


mofarrell:
`(init_member_reflection: Unable to find embedded data)?` does not make the newline optional.  This means these tests will fail when the data is embedded.

cmuellner:
@mofarrell that is true, but the first commit removes the (IMHO) unneeded additional newline, when calling logger->verbose(). So when applying both commits the test does not fail (this is of course tested).

Issue #7627, opened 2017-01-27
MySQL poll() syscall can hang request indefinitely


RyanGordon:
### HHVM Version

HHVM 3.17.1

### Standalone code, or other way to reproduce the problem
1. Start a SELECT query on a MySQL table that takes a long time to run. Use `strace -e poll -f` to watch for poll() syscalls
2. sudo kill -SIGSTOP `MYSQL PID` (close to simulating a MySQL that is frozen or under very high load)
3. Observe that the `poll` syscall hangs there indefinitely.

Simple example observed output:
```
...
poll([{fd=9, events=POLLIN|POLLPRI}], 1, 4294967295) = 1 ([{fd=9, revents=POLLIN}])
poll([{fd=9, events=POLLIN|POLLPRI}], 1, 4294967295) = 1 ([{fd=9, revents=POLLIN}])
poll([{fd=9, events=POLLIN|POLLPRI}], 1, 4294967295
# --> sudo kill -SIGSTOP 5424
...
# poll indefinitely hangs due to high default timeout
```

### Expected result

While this is not buggy behavior, it is undesirable because there is currently no way to configure the timeout on `poll` through the PDO extension to enable clients to be more resilient to failure edge cases.

### Actual result

The client code hangs indefinitely.

mofarrell:
cb8c85679d4e800b86bfd77a618cf19a69878a67 should mitigate this issue

PR #7628, opened 2017-01-27
Add MySQL PDO Net Read/Write Timeout Config

RyanGordon:
This patch address issue #7627

These settings were previously available in the deprecated MySQL extension, but was never carried over into the PDO MySQL extension. There is a way to set the connect timeout through PDO, but not the read or write timeout. This PR addresses that by adding two new settings that finish wiring through the write_timeout and read_timeout exposed in sql-common/client.c


mofarrell:
Are these standard?
I can't find any documentation for them in PHP
http://php.net/manual/en/pdo.constants.php

RyanGordon:
No, these aren't standard (They're not in upstream PHP PDO). These are completely new parameters that allow us to set MYSQL_OPT_READ_TIMEOUT and PDO_ATTR_WRITE_TIMEOUT as implemented in the MySQL client library: https://github.com/mysql/mysql-server/blob/71f48ab393bce80a59e5a2e498cd1f46f6b43f9a/sql-common/client.c#L5699 (docs here: https://dev.mysql.com/doc/refman/5.7/en/mysql-options.html)

mofarrell:
@wez I believe you implemented most of pdo.  What do you think about this?  Since we don't have mysqli we are leaning towards accepting this.

Orvid:
If they aren't standard they should be placed in the `HH\` namespace.

wez:
If no other database drivers have a concept of a separated read vs. write timeout (it's been a while since I looked, and they didn't way back then), this would be motivation for making these PDO_MYSQL specific.  The pre-namespaces way of dealing with a driver specific attribute in PDO is to use `PDO_MYSQL_` as the prefix for the global constant that is being defined.

It should be a similar trivial patch to the upstream PHP to make whatever naming we settle on here work with both implementations.

RyanGordon:
That sounds good, should I move this to both `HH\` namespace and rename to be `PDO_MYSQL_` or just rename to `PDO_MYSQL_`?

Orvid:
Sorry it's taken so long to reply, I missed your response :( `HH_PDO_MYSQL` is reasonable. As these are class constants, moving them into the `HH` namespace is a bit more difficult that it would otherwise be.

RyanGordon:
Sounds good! I will change them to HH_PDO_MYSQL and test it out and push it asap

PR #7631, opened 2017-01-29
Add a DebugOpt build type

swtaarrs:
This builds with -DDEBUG like a Debug build but adds -O2, resulting in a binary
with assertions enabled that isn't slower than molasses.


swtaarrs:
This inherits `-DDEBUG` from [here](https://github.com/facebook/hhvm/blob/master/CMake/HPHPSetup.cmake#L188-L190).

Does anyone know why we set `-DNDEBUG` in both `HPHPSetup.cmake` and `HPHPCompiler.cmake` for the Release build types?

PR #7632, opened 2017-01-29
Allow classname<T> in instanceof test

### LINKED ISSUE #7617: instanceof on generic overeager
### HHVM Version
3.18-dev
hh_client: 441002bba095256f8f01dda97189cd7164a029f7 Tue Jan 24 08:34:40 2017

```hack
<?hh // strict

function isInstanceOf<T>(classname<T> $type, mixed $what): T {
  if ($what instanceof $type) {
    return $what;
  }
  throw new Exception();
}
```

### Expected result
```
No errors!
```
### Actual result

```
test.php:4:7,28: 'instanceof' cannot be used on a generic classname because types are erased at runtime (Typing[4162])
```

Given that `$type` is a `classname`, it's not erased at runtime. Just the `T` is.

jesseschalken:
As far as I can tell, the commit https://github.com/facebook/hhvm/commit/e5e54bb597bc55249403011399eadc85772763cb is in error.

If `$y` is `classname<T>` (name of a class that is compatible with `T`) and `$x instanceof $y` is `true`, `$x` is necessarily compatible with `T`. I don't see the relevance to type erasure mentioned in the error message.

Fixes #7617

### DIFF
--- a/hphp/hack/src/typing/typing.ml
+++ b/hphp/hack/src/typing/typing.ml
@@ -4162,7 +4162,6 @@ and condition env tparamet =
         let env, obj_ty = Env.expand_type env obj_ty in
         match obj_ty with
         | _, Tabstract (AKgeneric _, _) ->
-          Errors.instanceof_generic_classname p;
           env, obj_ty
         | _, Tabstract (AKdependent (`this, []), Some (_, Tclass _)) ->
           let obj_ty =
--- a/hphp/hack/src/utils/errors.ml
+++ b/hphp/hack/src/utils/errors.ml
@@ -667,7 +667,6 @@ module Typing                               = struct
   let instanceof_always_false               = 4159 (* DONT MODIFY!!!! *)
   let instanceof_always_true                = 4160 (* DONT MODIFY!!!! *)
   let ambiguous_member                      = 4161 (* DONT MODIFY!!!! *)
-  let instanceof_generic_classname          = 4162 (* DONT MODIFY!!!! *)
   (* EXTEND HERE WITH NEW VALUES IF NEEDED *)
 end
 
@@ -2014,10 +2013,6 @@ let instanceof_always_false pos =
 let instanceof_always_true pos =
   add Typing.instanceof_always_true pos
     "This 'instanceof' test will always succeed"
-
-let instanceof_generic_classname pos =
-  add Typing.instanceof_generic_classname pos
-    "'instanceof' cannot be used on a generic classname because types are erased at runtime"
 
 
 (*****************************************************************************)
--- a/hphp/hack/src/utils/errors_sig.ml
+++ b/hphp/hack/src/utils/errors_sig.ml
@@ -321,8 +321,6 @@ module type S = sig
     -> (Pos.t * string) list -> unit
   val instanceof_always_false : Pos.t -> unit
   val instanceof_always_true : Pos.t -> unit
-  val instanceof_generic_classname : Pos.t -> unit
-
 
   val to_json : Pos.absolute error_ -> Hh_json.json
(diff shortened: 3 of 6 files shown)


andrewjkennedy:
The problem with this construct is that it leads to unsoundness. Consider, for example:
```
function mycast<T>(mixed $x, classname<T> $t) : T {
  if ($x instanceof $t) { return $x; }
  else throw new Exception("Type didn't match");
}
function expectListOfString(List<string> $x) { ... }
```
Now if I write `expectListOfString(mycast($y, List::class))` but with `$y` having type `List<int>` then no exception will get thrown at runtime (because of generics erasure).

jesseschalken:
And yet this code passes but it has the exact same problem:

```php
class List<T> {
  public function add(T $x): void {}
}

function main(): void {
  $y = new List();
  $y->add(1);
  foo($y);
}

function foo(mixed $x): void {
  if ($x instanceof List) {
    expectListOfString($x);
  }
}

function expectListOfString(List<string> $x): void {
  // ..
}

```

andrewjkennedy:
With 
```
enable_experimental_tc_features = instanceof
```
set in the .hhconfig file, or by running `hh_single_type_check`, that code does get rejected (in `//strict` mode). With the following message:
```
File "jesse.php", line 16, characters 24-25:
Invalid argument (Typing[4110])
File "jesse.php", line 20, characters 34-39:
This is a string
File "jesse.php", line 15, characters 7-8:
It is incompatible with a value of generic type T#1 from this instanceof test matching \List<T#1>
File "jesse.php", line 20, characters 34-39:
Considering that this type argument is invariant with respect to List
```

But I had forgotten that this was still marked as an "experimental" feature, whereas the `classname` restriction is turned on regardless. Thanks for spotting this!

It's probably time to turn on the "safe instanceof" feature permanently. (See https://github.com/facebook/hhvm/commit/4c2a6097643f33b4da8c297e52363c6a2f23f4f4 and related commits.) In general, we're trying to close all known strict-mode type soundness issues.

jesseschalken:
@andrewjkennedy Awesome, I didn't know about `enable_experimental_tc_features = instanceof`. According to the commit, it looks like it already does what I suggested in my comment [here](https://github.com/facebook/hhvm/issues/7617#issuecomment-276058377). 

I agree `enable_experimental_tc_features = instanceof` should be turned on by default, or at least move the banning of `instanceof (classname<T>)` underneath that option so it is consistent.

Also I think the message

> 'instanceof' cannot be used on a generic classname because types are erased at runtime

should be more explanatory, like

> 'instanceof' cannot be used on a classname of a type parameter because that type parameter may itself be generic but 'instanceof' cannot check generics because they are erased at runtime

If you agree, I'll open a new PR to update the message.

andrewjkennedy:
Thanks @jesseschalken that's a great suggestion for improving the error message. Perhaps "may itself be an instantiated generic type". The word "generic" is overloaded - I often see people talk about "the generic" when they mean a type parameter such as "T", and other times "generic" means a constructed type such as "C<int>".

andrewjkennedy:
Sorry screwy markup - that should be `C<int>` above.

fredemmott:
@jesseschalken if I'm understanding https://github.com/facebook/hhvm/commit/603cbd7cbf79f059365a1ff7787fef7b193e7ec3 correctly, this can be closed: while the safe behavior is still experimental, if the experimental flag isn't set, the old unsafe behavior is back - the latest build of the typechecker is now allowing https://github.com/facebook/fbshipit/blob/master/src/shipit/repo/ShipItRepo.php#L96-L98 to be removed without the experimental option

fredemmott:
whoops, that should be on the issue, not here

PR #7633, opened 2017-01-29
build: apply x86 flags on x86 only

compnerd:
`-msse4.3` is a x86 only flag.  We would previously apply it
unconditionally.  The result is both architecture and OS dependent
unfortunately.  Windows uses x86 and AMD64, others use i[3-6]86 and
x86_64.  Check all the possible casses.

### DIFF
--- a/CMake/HPHPCompiler.cmake
+++ b/CMake/HPHPCompiler.cmake
@@ -73,15 +73,20 @@ if (${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang" OR ${CMAKE_CXX_COMPILER_ID} STREQU
   endif()
 
   if (${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang") # using Clang
-    list(APPEND GENERAL_OPTIONS
+    if (CMAKE_SYSTEM_PROCESSOR MATCHES x86 OR
+        CMAKE_SYSTEM_PROCESSOR MATCHES "i[3456]86" OR
+        CMAKE_SYSTEM_PROCESSOR MATCHES AMD64 OR
+        CMAKE_SYSTEM_PROCESSOR MATCHES x86_64)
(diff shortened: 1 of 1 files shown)


compnerd:
BTW, is there an older version of a differential that exposes the clang issue the build system mentions?  Id be interested in that as a clang developer.

mofarrell:
This seems to be omitting the flag on our mac build, breaking the build.
```
In file included from /temp/hhvm-mac-build/third-party/folly/src/folly/Checksum.cpp:24:
/opt/homebrew/Cellar/llvm/3.6.2/bin/../lib/clang/3.6.2/include/nmmintrin.h:28:2: error: "SSE4.2 instruction set not enabled"
#error "SSE4.2 instruction set not enabled"
 ^
[  8%] Building C object third-party/libmbfl/mbfl/CMakeFiles/mbfl.dir/__/filters/mbfilter_iso8859_7.c.o
-- Check size of ulong - failed
/temp/hhvm-mac-build/third-party/folly/src/folly/Checksum.cpp:47:23: error: use of undeclared identifier '_mm_crc32_u8'
      sum = (uint32_t)_mm_crc32_u8(sum, data[offset]);
                      ^
-- Check size of int8
/temp/hhvm-mac-build/third-party/folly/src/folly/Checksum.cpp:55:20: error: use of undeclared identifier '_mm_crc32_u64'
    sum = uint32_t(_mm_crc32_u64(sum, *src));
                   ^
/temp/hhvm-mac-build/third-party/folly/src/folly/Checksum.cpp:61:21: error: use of undeclared identifier '_mm_crc32_u8'
    sum = (uint32_t)_mm_crc32_u8(sum, data[offset]);
                    ^
4 errors generated.
```

mofarrell:
This could use a rebase as well.

PR #7637, opened 2017-02-01
Pass to reuse ldimmq immediate values

swalk-cavium:
This adds a quick pass for ARM targets which attempts to reuse immediate values
generated by the ldimmq Vinstr.  This pass only operates on single blocks.  The span
of this optimization is tunable so that it does not increase register pressure.  The
default for the new option is 8,  (-vEval.JitLdimmqSpan=8).

Here is an example of the pass firing on test/zend/good/ext/intl/tests/grapheme.php

Without
=====

    36918     (38) DecRef<-> t7:Str
    36919         Main:
    36920               0x468005f4  d29ffe40              movz x0, #0xfff2
    36921               0x468005f8  f2a05fe0              movk x0, #0x2ff, lsl #16 
    36922               0x468005fc  78e06b61              ldrsh w1, [x27, x0] 
    36923               0x46800600  11000421              add w1, w1, #0x1 (1) 
    36924               0x46800604  78206b61              strh w1, [x27, x0] 
    36925               0x46800608  d29ffe80              movz x0, #0xfff4         //<<--+
    36926               0x4680060c  f2a05fe0              movk x0, #0x2ff, lsl #16 //<<--+
    36927               0x46800610  78e06b61              ldrsh w1, [x27, x0] 
    36928               0x46800614  11000421              add w1, w1, #0x1 (1) 
    36929               0x46800618  78206b61              strh w1, [x27, x0] 

With
===

    36929     (38) DecRef<-> t7:Str
    36930         Main:
    36931               0x200005f0  d29ffe40              movz x0, #0xfff2
    36932               0x200005f4  f2a05fe0              movk x0, #0x2ff, lsl #16 
    36933               0x200005f8  78e06b61              ldrsh w1, [x27, x0] 
    36934               0x200005fc  11000421              add w1, w1, #0x1 (1) 
    36935               0x20000600  78206b61              strh w1, [x27, x0] 
    36936               0x20000604  91000800              add x0, x0, #0x2 (2)  //<<---
    36937               0x20000608  78e06b61              ldrsh w1, [x27, x0] 
    36938               0x2000060c  11000421              add w1, w1, #0x1 (1) 
    36939               0x20000610  78206b61              strh w1, [x27, x0] 

This was tested by running the regression tests with 6 different option sets.
No regressions were introduced.

Performance testing is pending.

### DIFF
--- a/hphp/runtime/vm/jit/vasm-arm.cpp
+++ b/hphp/runtime/vm/jit/vasm-arm.cpp
@@ -1641,12 +1641,13 @@ void optimizeARM(Vunit& unit, const Abi& abi, bool regalloc) {
   simplify(unit);
 
   lowerForARM(unit);
-  reuseImmq(unit);
+
   simplify(unit);
 
   if (!unit.constToReg.empty()) {
     foldImms<arm::ImmFolder>(unit);
   }
+  reuseImmq(unit);
 
   optimizeCopies(unit, abi);
 
--- a/hphp/runtime/vm/jit/vasm-arm.cpp
+++ b/hphp/runtime/vm/jit/vasm-arm.cpp
@@ -1641,7 +1641,7 @@ void optimizeARM(Vunit& unit, const Abi& abi, bool regalloc) {
   simplify(unit);
 
   lowerForARM(unit);
-
+  reuseImmq(unit);
   simplify(unit);
 
   if (!unit.constToReg.empty()) {
--- a/hphp/runtime/vm/jit/vasm-reuseimm.cpp
+++ b/hphp/runtime/vm/jit/vasm-reuseimm.cpp
@@ -14,7 +14,6 @@
    +----------------------------------------------------------------------+
 */
 
-
 #include "hphp/runtime/vm/jit/vasm.h"
 
 #include "hphp/runtime/vm/jit/vasm-gen.h"
@@ -50,7 +49,6 @@ struct ImmState {
 
 struct Env {
   Vunit& unit;
-  int max_span;
   std::vector<ImmState> immStateVec;
 };
 
@@ -78,8 +76,7 @@ folly::Optional<int> reuseCandidate(Env& env, int64_t p, Vreg& reg) {
     }
     if ((p < q) && (q < (p + 4095))) {
       reg = elem.base;
-//    return folly::Optional<int>(safe_cast<int>(-safe_cast<int>(q - p)));
-      return folly::Optional<int>(safe_cast<int>(safe_cast<int>(p - q)));
+      return folly::Optional<int>(safe_cast<int>(p - q));
     }
   }
   return folly::none;
@@ -88,7 +85,7 @@ folly::Optional<int> reuseCandidate(Env& env, int64_t p, Vreg& reg) {
 template <typename Inst>
 void reuseImmq(Env& env, const Inst& inst, Vlabel b, size_t i) {
   // leaky bucket
-  env.immStateVec[i % env.max_span] = ImmState{};
+  env.immStateVec[i % RuntimeOption::EvalJitLdimmqSpan] = ImmState{};
 }
 
 template<typename ReuseImm>
@@ -99,36 +96,36 @@ void reuseimm_impl(Vunit& unit, Vlabel b, size_t i, ReuseImm reuse) {
 void reuseImmq(Env& env, const ldimmq& ld, Vlabel b, size_t i) {
   if (isMultiword(ld.s.q())) {
     Vreg base;
-    folly::Optional<int> off = reuseCandidate(env, ld.s.q(), base);
+    auto const off = reuseCandidate(env, ld.s.q(), base);
 
     if (off.hasValue()) {
-//    if (off.value() >= 0 ) {
-        reuseimm_impl(env.unit, b, i, [&] (Vout& v) {
-          v << addqi{off.value(), base, ld.d, v.makeReg()};
-        });
-//    } else {
-//      reuseimm_impl(env.unit, b, i, [&] (Vout& v) {
-//        v << subqi{-off.value(), base, ld.d, v.makeReg()};
-//      });
-//    }
-    return;
+      reuseimm_impl(env.unit, b, i, [&] (Vout& v) {
+        v << addqi{off.value(), base, ld.d, v.makeReg()};
+      });
+      return;
     }
   }
-  env.immStateVec[i % env.max_span] = ImmState{ld.s, ld.d};
+  env.immStateVec[i % RuntimeOption::EvalJitLdimmqSpan] = ImmState{ld.s, ld.d};
 }
 
 void reuseImmq(Env& env, Vlabel b, size_t i) {
   assertx(i <= env.unit.blocks[b].code.size());
-  auto const& inst = env.unit.blocks[b].code[i];
+  auto& inst = env.unit.blocks[b].code[i];
 
   if (isBlockEnd(inst) || isCall(inst)) {
     for (auto &elem : env.immStateVec) elem.reset();
-  } else if (inst.op == Vinstr::ldimmq) {
-    reuseImmq(env, inst.ldimmq_, b, i);
-  } else {
-    reuseImmq(env, inst, b, i);
+    return;
+  } 
+
+  switch (inst.op) {
+#define O(name, ...)                      \
+    case Vinstr::name:                    \
+      reuseImmq(env, inst.name##_, b, i); \
+      break;
+
+    VASM_OPCODES
+#undef O
   }
-  return;
 }
 
 }
@@ -140,17 +137,18 @@ void reuseImmq(Vunit& unit) {
 
   if (RuntimeOption::EvalJitLdimmqSpan <= 0) return;
 
-  Env env { unit, RuntimeOption::EvalJitLdimmqSpan };
-  env.immStateVec.reserve(env.max_span);
+  Env env {unit};
+  env.immStateVec.resize(RuntimeOption::EvalJitLdimmqSpan);
 
   auto const labels = sortBlocks(unit);
 
-  for (auto const b : labels) {
-    // required for initial size of vector
-    for (auto i = 0; i < env.max_span; i++) {
-      env.immStateVec.push_back(ImmState{});
-    }
+  // required for initial size of vector
+  for (auto i = 0; i < RuntimeOption::EvalJitLdimmqSpan; i++) {
+    env.immStateVec.push_back(ImmState{});
+  }
 
+  for (auto const b : labels) {
+    for (auto &elem : env.immStateVec) elem.reset();
     for (size_t i = 0; i < blocks[b].code.size(); ++i) {
       reuseImmq(env, b, i);
     }
--- a/hphp/runtime/vm/jit/vasm-reuseimm.cpp
+++ b/hphp/runtime/vm/jit/vasm-reuseimm.cpp
@@ -27,6 +27,8 @@
 #include "hphp/util/arch.h"
 #include "hphp/util/assertions.h"
 
+#include <folly/Optional.h>
+
 #include <cstdlib>
 
 TRACE_SET_MOD(vasm);
@@ -40,28 +42,16 @@ struct ImmState {
   ImmState() : val{0}, base{} {}
   ImmState(Immed64 a, Vreg b) : val{a}, base{b} {}
 
+  void reset() { base = Vreg{}; }
+
   Immed64 val;
   Vreg base;
 };
 
-struct ImmStateAry {
-  void clear();
-  ImmState& operator[](size_t i) { return immState_ary[i % size]; }
-
-  int size;
-  enum { max_history = 16};
-  ImmState immState_ary[max_history];
-};
-
-void ImmStateAry::clear() {
-  for (int i=0; i < size; i++) {
-    immState_ary[i].base = Vreg{};
-  }
-}
-
 struct Env {
   Vunit& unit;
-  ImmStateAry& immStateAry;
+  int max_span;
+  std::vector<ImmState> immStateVec;
 };
 
 bool isMultiword(int64_t imm) {
@@ -78,65 +68,28 @@ bool isMultiword(int64_t imm) {
 }
 
 // candidate around +/- uimm12
-bool reuseCandidate(Env& env, int64_t p, Vreg& reg, int& delta) {
-  delta = 0;
-  for (int i = 0; i < env.immStateAry.size; i++) {
-    ImmState elem = env.immStateAry[i];
+folly::Optional<int> reuseCandidate(Env& env, int64_t p, Vreg& reg) {
+  for (auto const& elem : env.immStateVec) {
     if (!elem.base.isValid()) continue;
     int64_t q = elem.val.q();
     if ((p >= q) && (p < (q + 4095))) {
-      delta = safe_cast<int>(p - q);
       reg = elem.base;
-      return true;
+      return folly::Optional<int>(safe_cast<int>(p - q));
     }
     if ((p < q) && (q < (p + 4095))) {
-      delta = -safe_cast<int>(q - p);
       reg = elem.base;
-      return true;
+//    return folly::Optional<int>(safe_cast<int>(-safe_cast<int>(q - p)));
+      return folly::Optional<int>(safe_cast<int>(safe_cast<int>(p - q)));
     }
   }
-  return false;
+  return folly::none;
 }
 
 template <typename Inst>
-void reuseImmq(Env& env, const Inst& inst, Vlabel b, size_t i) { 
-  env.immStateAry[i] = ImmState{};
-  return;
+void reuseImmq(Env& env, const Inst& inst, Vlabel b, size_t i) {
+  // leaky bucket
+  env.immStateVec[i % env.max_span] = ImmState{};
 }
-
-// purge table across any instruction with RegSet or control flow change
-#define Y(vasm_opc) \
-void reuseImmq(Env& env, const vasm_opc& c, Vlabel b, size_t i) { \
-  env.immStateAry.clear(); \
-  return; \
-} 
-
-Y(fallthru)
-Y(call)
-Y(callarray)
-Y(callfaststub)
-Y(callm)
-Y(callr)
-Y(calls)
-Y(callstub)
-Y(calltc)
-Y(contenter)
-Y(jcc)
-Y(jcci)
-Y(jmpi)
-Y(jmpm)
-Y(jmpr)
-Y(landingpad)
-Y(leavetc)
-Y(nothrow)
-Y(phpret)
-Y(ret)
-Y(syncpoint)
-Y(ud2)
-Y(unwind)
-
-#undef Y
-
 
 template<typename ReuseImm>
 void reuseimm_impl(Vunit& unit, Vlabel b, size_t i, ReuseImm reuse) {
@@ -145,38 +98,37 @@ void reuseimm_impl(Vunit& unit, Vlabel b, size_t i, ReuseImm reuse) {
 
 void reuseImmq(Env& env, const ldimmq& ld, Vlabel b, size_t i) {
   if (isMultiword(ld.s.q())) {
-    int off;
     Vreg base;
-    if (reuseCandidate(env, ld.s.q(), base, off)) {
-     if (off >= 0 ) {
-       reuseimm_impl(env.unit, b, i, [&] (Vout& v) {
-         v << addqi{off, base, ld.d, v.makeReg()};
-       });
-     } else {
-       reuseimm_impl(env.unit, b, i, [&] (Vout& v) {
-         v << subqi{-off, base, ld.d, v.makeReg()};
-       });
-      }
+    folly::Optional<int> off = reuseCandidate(env, ld.s.q(), base);
+
+    if (off.hasValue()) {
+//    if (off.value() >= 0 ) {
+        reuseimm_impl(env.unit, b, i, [&] (Vout& v) {
+          v << addqi{off.value(), base, ld.d, v.makeReg()};
+        });
+//    } else {
+//      reuseimm_impl(env.unit, b, i, [&] (Vout& v) {
+//        v << subqi{-off.value(), base, ld.d, v.makeReg()};
+//      });
+//    }
     return;
     }
   }
-  env.immStateAry[i] = ImmState{ld.s, ld.d};
+  env.immStateVec[i % env.max_span] = ImmState{ld.s, ld.d};
 }
 
 void reuseImmq(Env& env, Vlabel b, size_t i) {
   assertx(i <= env.unit.blocks[b].code.size());
   auto const& inst = env.unit.blocks[b].code[i];
 
-  switch (inst.op) {
-#define O(name, ...)                      \
-    case Vinstr::name:                    \
-      reuseImmq(env, inst.name##_, b, i); \
-      return;
-
-  VASM_OPCODES
-#undef O
+  if (isBlockEnd(inst) || isCall(inst)) {
+    for (auto &elem : env.immStateVec) elem.reset();
+  } else if (inst.op == Vinstr::ldimmq) {
+    reuseImmq(env, inst.ldimmq_, b, i);
+  } else {
+    reuseImmq(env, inst, b, i);
   }
-  not_reached();
+  return;
 }
 
 }
@@ -186,19 +138,18 @@ void reuseImmq(Vunit& unit) {
   assertx(check(unit));
   auto& blocks = unit.blocks;
 
-  if (RuntimeOption::EvalJitLdimmqSpan <= 0) {
-    return;
-  }
+  if (RuntimeOption::EvalJitLdimmqSpan <= 0) return;
 
-  ImmStateAry immStateAry;
-  Env env { unit, immStateAry };
+  Env env { unit, RuntimeOption::EvalJitLdimmqSpan };
+  env.immStateVec.reserve(env.max_span);
 
   auto const labels = sortBlocks(unit);
-  env.immStateAry.size =
-    std::min<int>(RuntimeOption::EvalJitLdimmqSpan, ImmStateAry::max_history);
 
   for (auto const b : labels) {
-    env.immStateAry.clear();
+    // required for initial size of vector
+    for (auto i = 0; i < env.max_span; i++) {
+      env.immStateVec.push_back(ImmState{});
+    }
 
     for (size_t i = 0; i < blocks[b].code.size(); ++i) {
       reuseImmq(env, b, i);
--- a/hphp/runtime/vm/jit/vasm-reuseimm.cpp
+++ b/hphp/runtime/vm/jit/vasm-reuseimm.cpp
@@ -14,7 +14,6 @@
    +----------------------------------------------------------------------+
 */
 
-#include <cstdlib>
 
 #include "hphp/runtime/vm/jit/vasm.h"
 
@@ -27,14 +26,19 @@
 
 #include "hphp/util/assertions.h"
 
+#include <cstdlib>
+
 TRACE_SET_MOD(vasm);
 
 namespace HPHP { namespace jit {
 
+namespace {
+
+// track ldimmq values
 struct qTrack {
   Immed64 val;
   Vreg base;
-  int  ttl;
+  int ttl;
 };
 
 struct Env {
@@ -43,26 +47,32 @@ struct Env {
   std::deque<qTrack> qdeque; // leaky bucket of recent immediates
 };
 
-static bool isMultiword(int64_t imm) {
-  uint64_t val = abs(imm);
-  if (val > (1 << 16)) return true;
+bool isMultiword(int64_t imm) {
+  switch(arch()) {
+  case Arch::X64:
+    break;
+  case Arch::ARM:
+    uint64_t val = std::abs(imm);
+    if (val > (1 << 16)) return true;
+    break;
+  case Arch::PPC64:
+    break;
+  }
   return false;
 }
 
 // candidate around +/- uimm12
-static bool reuseCandidate(Env& env, int64_t p, Vreg& reg, int& delta)
-{
+bool reuseCandidate(Env& env, int64_t p, Vreg& reg, int& delta) {
   delta = 0;
-  for (const auto& itr : env.qdeque) {
+  for (auto const& itr : env.qdeque) {
     int64_t q = itr.val.q();
-    if ((p >= q) && (p < (q+4095))) {
+    if ((p >= q) && (p < (q + 4095))) {
       delta = safe_cast<int>(p-q);
       reg = itr.base;
       return true;
     }
-    if ((p < q) && (q < (p+4095))) {
-      delta = safe_cast<int>(q-p);
-      delta = -delta;
+    if ((p < q) && (q < (p + 4095))) {
+      delta = -safe_cast<int>(q-p);
       reg = itr.base;
       return true;
     }
@@ -73,7 +83,7 @@ static bool reuseCandidate(Env& env, int64_t p, Vreg& reg, int& delta)
 template <typename Inst>
 void reuseImmq(Env& env, const Inst& inst, Vlabel b, size_t i) { return; }
 
-// purge table across any instruction with RegSet or flow change
+// purge table across any instruction with RegSet or control flow change
 #define Y(vasm_opc) \
 void reuseImmq(Env& env, const vasm_opc& c, Vlabel b, size_t i) { \
   env.qdeque.clear(); \
@@ -120,11 +130,11 @@ void reuseImmq(Env& env, const ldimmq& ld, Vlabel b, size_t i) {
      if (off >= 0 ) {
        reuseImm_impl(env.unit, b, i, [&] (Vout& v) {
          v << addqi{off, base, ld.d, v.makeReg()};
-        });
+       });
      } else {
        reuseImm_impl(env.unit, b, i, [&] (Vout& v) {
          v << subqi{-off, base, ld.d, v.makeReg()};
-        });
+       });
       }
     return;
     }
@@ -150,12 +160,13 @@ void reuseImmq(Env& env, Vlabel b, size_t i) {
   not_reached();
 }
 
+}
 // Opportunistically reuse immediate q values
 void reuseImmq(Vunit& unit) {
   assertx(check(unit));
   auto& blocks = unit.blocks;
 
-  Env env {unit};
+  Env env { unit };
 
   auto const labels = sortBlocks(unit);
 
--- a/hphp/runtime/vm/jit/vasm-reuseimm.cpp
+++ b/hphp/runtime/vm/jit/vasm-reuseimm.cpp
@@ -24,6 +24,7 @@
 #include "hphp/runtime/vm/jit/vasm-unit.h"
 #include "hphp/runtime/vm/jit/vasm-util.h"
 
+#include "hphp/util/arch.h"
 #include "hphp/util/assertions.h"
 
 #include <cstdlib>
@@ -50,12 +51,11 @@ struct Env {
 bool isMultiword(int64_t imm) {
   switch(arch()) {
   case Arch::X64:
+  case Arch::PPC64:
     break;
   case Arch::ARM:
     uint64_t val = std::abs(imm);
     if (val > (1 << 16)) return true;
-    break;
-  case Arch::PPC64:
     break;
   }
   return false;
--- a/hphp/runtime/vm/jit/vasm-reuseimm.cpp
+++ b/hphp/runtime/vm/jit/vasm-reuseimm.cpp
@@ -139,11 +139,6 @@ void reuseImmq(Vunit& unit) {
 
   auto const labels = sortBlocks(unit);
 
-  // required for initial size of vector
-  for (auto i = 0; i < RuntimeOption::EvalJitLdimmqSpan; i++) {
-    env.immStateVec.push_back(ImmState{});
-  }
-
   for (auto const b : labels) {
     for (auto& elem : env.immStateVec) elem.reset();
     for (size_t i = 0; i < blocks[b].code.size(); ++i) {
--- a/hphp/runtime/vm/jit/vasm-reuseimm.cpp
+++ b/hphp/runtime/vm/jit/vasm-reuseimm.cpp
@@ -137,7 +137,7 @@ void reuseImmq(Vunit& unit) {
 
   if (RuntimeOption::EvalJitLdimmqSpan <= 0) return;
 
-  Env env {unit};
+  Env env { unit };
   env.immStateVec.resize(RuntimeOption::EvalJitLdimmqSpan);
 
   auto const labels = sortBlocks(unit);
--- a/hphp/runtime/vm/jit/vasm-reuseimm.cpp
+++ b/hphp/runtime/vm/jit/vasm-reuseimm.cpp
@@ -1,0 +1,182 @@
+/*
+   +----------------------------------------------------------------------+
+   | HipHop for PHP                                                       |
+   +----------------------------------------------------------------------+
+   | Copyright (c) 2010-present Facebook, Inc. (http://www.facebook.com)  |
+   +----------------------------------------------------------------------+
+   | This source file is subject to version 3.01 of the PHP license,      |
+   | that is bundled with this package in the file LICENSE, and is        |
+   | available through the world-wide-web at the following url:           |
+   | http://www.php.net/license/3_01.txt                                  |
+   | If you did not receive a copy of the PHP license and are unable to   |
+   | obtain it through the world-wide-web, please send a note to          |
+   | license@php.net so we can mail you a copy immediately.               |
+   +----------------------------------------------------------------------+
+*/
+
+#include <cstdlib>
+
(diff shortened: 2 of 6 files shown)


mxw:
Sorry @swalk-cavium—I wasn't assigned this diff, so I didn't see it until now.  Will review soon.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,182 @@
...
+   | that is bundled with this package in the file LICENSE, and is        |
+   | available through the world-wide-web at the following url:           |
+   | http://www.php.net/license/3_01.txt                                  |
+   | If you did not receive a copy of the PHP license and are unable to   |
+   | obtain it through the world-wide-web, please send a note to          |
+   | license@php.net so we can mail you a copy immediately.               |
+   +----------------------------------------------------------------------+
+*/
+
+#include <cstdlib>
```
mxw: Always include the header that contains the forward declarations implemented in the source file first—so just move this to below `#include "hphp/util/assertions.h"`.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,182 @@
...
+#include "hphp/util/assertions.h"
+
+TRACE_SET_MOD(vasm);
+
+namespace HPHP { namespace jit {
+
+struct qTrack {
+  Immed64 val;
+  Vreg base;
+  int  ttl;
```
mxw: nit: Spurious space.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,182 @@
...
+};
```
mxw: Can you document this struct?  As a name, `qTrack` is not particularly self-documenting...

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+static bool isMultiword(int64_t imm) {
+  uint64_t val = abs(imm);
+  if (val > (1 << 16)) return true;
+  return false;
+}
+
+// candidate around +/- uimm12
+static bool reuseCandidate(Env& env, int64_t p, Vreg& reg, int& delta)
+{
```
mxw: oneline

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,182 @@
...
+  }
+  not_reached();
+}
+
+// Opportunistically reuse immediate q values
+void reuseImmq(Vunit& unit) {
+  assertx(check(unit));
+  auto& blocks = unit.blocks;
+
+  Env env {unit};
```
mxw: nit: `env{unit}` or `env { unit }`.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,182 @@
...
+Y(ret)
+Y(syncpoint)
+Y(ud2)
+Y(unwind)
+
+#undef Y
+
+
+template<typename ReuseImm>
+void reuseImm_impl(Vunit& unit, Vlabel b, size_t i, ReuseImm reuse) {
```
mxw: Avoid mixing naming styles within a single name.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,182 @@
...
+      return true;
+    }
+  }
+  return false;
+}
+
+template <typename Inst>
+void reuseImmq(Env& env, const Inst& inst, Vlabel b, size_t i) { return; }
+
+// purge table across any instruction with RegSet or flow change
```
mxw: What does "flow change" mean?

Also, this blacklist is pretty fragile to future changes.  Either make it a whitelist, or use introspection to determine what sort of instruction we have—e.g., using the Vinstr visitor machinery.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,182 @@
...
+  if (val > (1 << 16)) return true;
+  return false;
+}
+
+// candidate around +/- uimm12
+static bool reuseCandidate(Env& env, int64_t p, Vreg& reg, int& delta)
+{
+  delta = 0;
+  for (const auto& itr : env.qdeque) {
+    int64_t q = itr.val.q();
```
mxw: `auto const`

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,182 @@
...
+};
+
+struct Env {
+  Vunit& unit;
+  int tick;
+  std::deque<qTrack> qdeque; // leaky bucket of recent immediates
+};
+
+static bool isMultiword(int64_t imm) {
+  uint64_t val = abs(imm);
```
mxw: I don't see `abs()` defined anywhere.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,182 @@
...
+  int  ttl;
+};
+
+struct Env {
+  Vunit& unit;
+  int tick;
+  std::deque<qTrack> qdeque; // leaky bucket of recent immediates
+};
+
+static bool isMultiword(int64_t imm) {
```
mxw: Everything in this file (notably including the class definitions) except the one exposed function should be enclosed in an anonymous namespace.  Then you can strip all these `static`s.

Also: This `isMultiword()` predicate seems fairly ARM-specific, but this file is purportedly architecture-independent.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,182 @@
...
+  return false;
+}
+
+// candidate around +/- uimm12
+static bool reuseCandidate(Env& env, int64_t p, Vreg& reg, int& delta)
+{
+  delta = 0;
+  for (const auto& itr : env.qdeque) {
+    int64_t q = itr.val.q();
+    if ((p >= q) && (p < (q+4095))) {
```
mxw: Space between operands: `q + 4095`.  Here and everywhere below.

Also, why this magic number?

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,182 @@
...
+      delta = safe_cast<int>(p-q);
+      reg = itr.base;
+      return true;
+    }
+    if ((p < q) && (q < (p+4095))) {
+      delta = safe_cast<int>(q-p);
+      delta = -delta;
```
mxw: Just `delta = -safe_cast<int>(q - p)`.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,182 @@
...
+    env.qdeque.clear();
+
+    for (size_t i = 0; i < blocks[b].code.size(); ++i) {
+      reuseImmq(env, b, i);
+      env.tick++;
+      if (!env.qdeque.empty()) {
+        auto& p = env.qdeque.back();
+        if (p.ttl < env.tick) {
+          env.qdeque.pop_back(); 
+        }
```
mxw: This tick machinery seems a bit obfuscated and also threads around a lot of state—plus, there's a nontrivial cost to using a deque here.

Instead, how about using a fixed-sized circular array?  You can move the cursor and clear whatever's at it on each iteration, and you can pass that cursor into the `reuseImmq()` call, with the API being that it should store its analysis state at that location.  That way, only the toplevel loop needs to know about the sliding window.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,182 @@
...
+    if (reuseCandidate(env, ld.s.q(), base, off)) {
+     if (off >= 0 ) {
+       reuseImm_impl(env.unit, b, i, [&] (Vout& v) {
+         v << addqi{off, base, ld.d, v.makeReg()};
+        });
+     } else {
+       reuseImm_impl(env.unit, b, i, [&] (Vout& v) {
+         v << subqi{-off, base, ld.d, v.makeReg()};
+        });
+      }
```
mxw: Why not just always use `addqi{}`?

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+void reuseImmq(Env& env, const ldimmq& ld, Vlabel b, size_t i) {
+  if (isMultiword(ld.s.q())) {
+    int off;
+    Vreg base;
+    if (reuseCandidate(env, ld.s.q(), base, off)) {
+     if (off >= 0 ) {
+       reuseImm_impl(env.unit, b, i, [&] (Vout& v) {
+         v << addqi{off, base, ld.d, v.makeReg()};
+        });
```
mxw: You have a single extra space (here and below).

hphp/runtime/vm/jit/vasm-reuseimm.cpp
swalk-cavium: Getting the name right is always a pain.  It's intended to track the ldimmQ values
until the time-to-live value expires.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
swalk-cavium: abs() is the absolute value of an integer.  Defined in cstdlib.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
swalk-cavium: "flow change"  control flow change.

I'm not sure about your whitelist comment.  This is the same technique used in
other places in the compiler, just using the X-macro pattern.  Can you elaborate?

hphp/runtime/vm/jit/vasm-reuseimm.cpp
swalk-cavium: I'm not sure a circular buffer would be a win in this case.  Looking at some of the hphp.log files it shows that for the default EvalJitLdimmqSpan of 8  there might be 2 elements in the
dequeue.  peeking at the front/back of a dequeue is supposed to be fast.  
If we wanted to use a circular buffer wouldn't you have to put empty marker entries
in the buffer to get the same span?   With the current implementation each instruction in
the block is a tick and we're only tracking the ones we care about.

swalk-cavium:
Hi @mxw,
I think I've fixed most of these.  We can talk about the queue stuff more if you'd like.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,193 @@
...
+};
+
+struct Env {
+  Vunit& unit;
+  int tick;
+  std::deque<qTrack> qdeque; // leaky bucket of recent immediates
+};
+
+bool isMultiword(int64_t imm) {
+  switch(arch()) {
```
mxw: `switch (arch()) {`

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,193 @@
...
+  return false;
+}
+
+// candidate around +/- uimm12
+bool reuseCandidate(Env& env, int64_t p, Vreg& reg, int& delta) {
+  delta = 0;
+  for (auto const& itr : env.qdeque) {
+    int64_t q = itr.val.q();
+    if ((p >= q) && (p < (q + 4095))) {
+      delta = safe_cast<int>(p-q);
```
mxw: `p - q`

Consider skimming through https://github.com/facebook/hhvm/blob/master/hphp/doc/coding-conventions.md.  It's pretty short.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,193 @@
...
+      reg = itr.base;
+      return true;
+    }
+    if ((p < q) && (q < (p + 4095))) {
+      delta = -safe_cast<int>(q-p);
```
mxw: `q - p`

Except also, why can't this just be `delta = safe_cast<int>(p - q)`?  Both of them are signed integers within a certain small delta from one another, so that should be fine, no?

hphp/runtime/vm/jit/vasm-reuseimm.cpp
mxw: poke

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,193 @@
...
+#include <cstdlib>
+
+TRACE_SET_MOD(vasm);
+
+namespace HPHP { namespace jit {
+
+namespace {
+
+// track ldimmq values
+struct qTrack {
```
mxw: Can you call this something like `ImmState` or `RegConst`, or something similarly expressive?  `qTrack` seems really oblique to me.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
mxw: A circular buffer containing eight 16-byte entries will only occupy 2 cache lines—it'll certainly be cheaper than a deque.  The cost of this pass probably isn't a dominating factor right now, but doing this the cheaper way will save us from having to speed it up in the future when we start tightening the screws on JIT time.  (Plus, it's pretty straightforward.)

You can just use the default `Vreg{}` as a sentinel invalid entry in your state struct.

swalk-cavium:
@mxw - I think I've updated the code to address your concerns.  You were right, I didn't need the tick to implement the leaky bucket.  The offset into the basic block worked just as well.  I did add
a kill flag, so if we set the span option to zero it will skip this pass.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,212 @@
...
+
+void ImmStateAry::clear() {
+  for (int i=0; i < size; i++) {
+    immState_ary[i].base = Vreg{};
+  }
+}
+
+struct Env {
+  Vunit& unit;
+  ImmStateAry& immStateAry;
```
mxw: This is pretty weird.  The Env is local to this pass; why not just have it contain the `ImmStateAry`, instead of a reference to one you construct on the stack just to pass it to the Env?

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,212 @@
...
+}
+
+// Opportunistically reuse immediate q values
+void reuseImmq(Vunit& unit) {
+  assertx(check(unit));
+  auto& blocks = unit.blocks;
+
+  if (RuntimeOption::EvalJitLdimmqSpan <= 0) {
+    return;
+  }
```
mxw: nit: This can be oneline (`if (RuntimeOption::EvalJitLdimmqSpan <= 0) return;`)

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,212 @@
...
+  Immed64 val;
+  Vreg base;
+};
+
+struct ImmStateAry {
+  void clear();
+  ImmState& operator[](size_t i) { return immState_ary[i % size]; }
+
+  int size;
+  enum { max_history = 16};
```
mxw: You're not actually using this as an enum.  If you want a constant, use a constant.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,212 @@
...
+  if (RuntimeOption::EvalJitLdimmqSpan <= 0) {
+    return;
+  }
+
+  ImmStateAry immStateAry;
+  Env env { unit, immStateAry };
+
+  auto const labels = sortBlocks(unit);
+  env.immStateAry.size =
+    std::min<int>(RuntimeOption::EvalJitLdimmqSpan, ImmStateAry::max_history);
```
mxw: Why not just let the RuntimeOption dictate everything?  You can just use a `std::vector` instead of a C-style array.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,212 @@
...
+};
+
+struct ImmStateAry {
+  void clear();
+  ImmState& operator[](size_t i) { return immState_ary[i % size]; }
+
+  int size;
+  enum { max_history = 16};
+  ImmState immState_ary[max_history];
+};
```
mxw: This struct seems unnecessary.  Just use a `std::vector<ImmState>` and you'll be good to go.  (You can just do the mod operation inline.)

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,212 @@
...
+    if (val > (1 << 16)) return true;
+    break;
+  }
+  return false;
+}
+
+// candidate around +/- uimm12
+bool reuseCandidate(Env& env, int64_t p, Vreg& reg, int& delta) {
+  delta = 0;
+  for (int i = 0; i < env.immStateAry.size; i++) {
```
mxw: You can use a range-based for here, once you're using a `std::vector`.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,212 @@
...
+  case Arch::ARM:
+    uint64_t val = std::abs(imm);
+    if (val > (1 << 16)) return true;
+    break;
+  }
+  return false;
+}
+
+// candidate around +/- uimm12
+bool reuseCandidate(Env& env, int64_t p, Vreg& reg, int& delta) {
```
mxw: Instead of having an out-parameter, return a `folly::Optional<int>`.  You can return `folly::none` when we can't reuse, or whatever the delta is when we can.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,212 @@
...
+  return false;
+}
+
+template <typename Inst>
+void reuseImmq(Env& env, const Inst& inst, Vlabel b, size_t i) { 
+  env.immStateAry[i] = ImmState{};
+  return;
+}
+
+// purge table across any instruction with RegSet or control flow change
```
mxw: I wasn't commenting about this macro pattern—I was commenting about having a hardcoded set of vasm operations that are "control flow".

I actually think you can just use the `isBlockEnd()` predicate defined in `vasm-instr.h`.  That doesn't include `landingpad{}` or `unwind{}`, but then again, neither of those instructions is actually control flow; they're just marker instructions.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,212 @@
...
+void reuseimm_impl(Vunit& unit, Vlabel b, size_t i, ReuseImm reuse) {
+  vmodify(unit, b, i, [&] (Vout& v) { reuse(v); return 1; });
+}
+
+void reuseImmq(Env& env, const ldimmq& ld, Vlabel b, size_t i) {
+  if (isMultiword(ld.s.q())) {
+    int off;
+    Vreg base;
+    if (reuseCandidate(env, ld.s.q(), base, off)) {
+     if (off >= 0 ) {
```
mxw: Each indentation level should be two spaces.  Can you just go through this function and fix it everywhere?

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,212 @@
...
+       reuseimm_impl(env.unit, b, i, [&] (Vout& v) {
+         v << addqi{off, base, ld.d, v.makeReg()};
+       });
+     } else {
+       reuseimm_impl(env.unit, b, i, [&] (Vout& v) {
+         v << subqi{-off, base, ld.d, v.makeReg()};
```
mxw: You didn't answer my question about why we can't just use `addqi{}`.  The immediate is signed; there's no problem with adding negative numbers.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,212 @@
...
+    ImmState elem = env.immStateAry[i];
+    if (!elem.base.isValid()) continue;
+    int64_t q = elem.val.q();
+    if ((p >= q) && (p < (q + 4095))) {
+      delta = safe_cast<int>(p - q);
+      reg = elem.base;
+      return true;
+    }
+    if ((p < q) && (q < (p + 4095))) {
+      delta = -safe_cast<int>(q - p);
```
mxw: You didn't answer my question about why we can't just use `delta = safe_cast<int>(p - q)` here (and thus squash these two conditional blocks together).

hphp/runtime/vm/jit/vasm-reuseimm.cpp
swalk-cavium: I made it an array because of your previous comment about deque.  I'll change to vector.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
swalk-cavium: I thought it easier  to debug.  I can squash.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
swalk-cavium: isBlockEnd() is not sufficient.  It does not include call, callr, callm, calls, and syncpoint.
I would expect this table to become invalid crossing those instructions.

swalk-cavium:
@mxw - Some comments added.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
mxw: `syncpoint{}` is also a marker instruction, so it should be fine.  As for calls—the register allocator will ensure that any Vregs that are live across a call will be preserved, either by being in a callee-saved register, or by spilling it to the stack.  If you don't special-case calls, correctness won't be affected, but since you're trying to avoid added register pressure, I think it makes sense to handle them in this way.

I would just use `isBlockEnd() || /* is one of the call instructions */` in that case.

swalk-cavium:
@mxw - Hi Max, I think I've updated per all your comments.

mxw:
@swalk-cavium—Whoops, I must have missed the notification for you last comment.  I'll take a look and hopefully land.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,163 @@
...
+    if (!elem.base.isValid()) continue;
+    int64_t q = elem.val.q();
+    if ((p >= q) && (p < (q + 4095))) {
+      reg = elem.base;
+      return folly::Optional<int>(safe_cast<int>(p - q));
+    }
+    if ((p < q) && (q < (p + 4095))) {
+      reg = elem.base;
+//    return folly::Optional<int>(safe_cast<int>(-safe_cast<int>(q - p)));
+      return folly::Optional<int>(safe_cast<int>(safe_cast<int>(p - q)));
```
mxw: This amount of casting seems excessive.  I'm still not sure why this is a separate block from the above, since they're identical.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,163 @@
...
+    if (off.hasValue()) {
+//    if (off.value() >= 0 ) {
+        reuseimm_impl(env.unit, b, i, [&] (Vout& v) {
+          v << addqi{off.value(), base, ld.d, v.makeReg()};
+        });
+//    } else {
+//      reuseimm_impl(env.unit, b, i, [&] (Vout& v) {
+//        v << subqi{-off.value(), base, ld.d, v.makeReg()};
+//      });
+//    }
```
mxw: Nix this commented-out code.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+template<typename ReuseImm>
+void reuseimm_impl(Vunit& unit, Vlabel b, size_t i, ReuseImm reuse) {
+  vmodify(unit, b, i, [&] (Vout& v) { reuse(v); return 1; });
+}
+
+void reuseImmq(Env& env, const ldimmq& ld, Vlabel b, size_t i) {
+  if (isMultiword(ld.s.q())) {
+    Vreg base;
+    folly::Optional<int> off = reuseCandidate(env, ld.s.q(), base);
```
mxw: nit: `auto const`

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,163 @@
...
+//    if (off.value() >= 0 ) {
+        reuseimm_impl(env.unit, b, i, [&] (Vout& v) {
+          v << addqi{off.value(), base, ld.d, v.makeReg()};
+        });
+//    } else {
+//      reuseimm_impl(env.unit, b, i, [&] (Vout& v) {
+//        v << subqi{-off.value(), base, ld.d, v.makeReg()};
+//      });
+//    }
+    return;
```
mxw: nit: Indentation.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,163 @@
...
+void reuseImmq(Env& env, Vlabel b, size_t i) {
+  assertx(i <= env.unit.blocks[b].code.size());
+  auto const& inst = env.unit.blocks[b].code[i];
+
+  if (isBlockEnd(inst) || isCall(inst)) {
+    for (auto &elem : env.immStateVec) elem.reset();
+  } else if (inst.op == Vinstr::ldimmq) {
+    reuseImmq(env, inst.ldimmq_, b, i);
+  } else {
+    reuseImmq(env, inst, b, i);
```
mxw: You're not using the `inst` argument in this overload.  Either don't pass it, or use the macro-generated switch pattern that we use elsewhere in vasm code, in case we extend this pass later (seems possible, so I'd go with this route—it's also very familiar).

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+  Env env { unit, RuntimeOption::EvalJitLdimmqSpan };
+  env.immStateVec.reserve(env.max_span);
+
+  auto const labels = sortBlocks(unit);
+
+  for (auto const b : labels) {
+    // required for initial size of vector
+    for (auto i = 0; i < env.max_span; i++) {
+      env.immStateVec.push_back(ImmState{});
```
mxw: It looks like you're adding `EvalJitLdimmqSpan` entries to this vector for every block that you process, which makes the vector `O(sb)` in memory instead of `O(1)`...

You should do `resize()` above instead of `reserve()`, and then clear the vector at the start of every block like you do above.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,163 @@
...
+  }
+  env.immStateVec[i % env.max_span] = ImmState{ld.s, ld.d};
+}
+
+void reuseImmq(Env& env, Vlabel b, size_t i) {
+  assertx(i <= env.unit.blocks[b].code.size());
+  auto const& inst = env.unit.blocks[b].code[i];
+
+  if (isBlockEnd(inst) || isCall(inst)) {
+    for (auto &elem : env.immStateVec) elem.reset();
```
mxw: `auto&`

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,163 @@
...
+}
+
+// Opportunistically reuse immediate q values
+void reuseImmq(Vunit& unit) {
+  assertx(check(unit));
+  auto& blocks = unit.blocks;
+
+  if (RuntimeOption::EvalJitLdimmqSpan <= 0) return;
+
+  Env env { unit, RuntimeOption::EvalJitLdimmqSpan };
```
mxw: Why thread `max_span` around as state?  Just use `EvalJitLdimmqSpan` everywhere.

swalk-cavium:
@mxw - Hi Max,  I fixed the rebase and what I think was your one comment.

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -1641,7 +1641,7 @@ void optimizeARM(Vunit& unit, const Abi& abi, bool regalloc) {
   simplify(unit);
 
   lowerForARM(unit);
-
+  reuseImmq(unit);
```
cmuellner: I think it would be better to call this after foldImms(). foldImms() eliminates immediates a bit more efficient than this pass.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+    VASM_OPCODES
+#undef O
+  }
+}
+
+}
+
+// Opportunistically reuse immediate q values
+void reuseImmq(Vunit& unit) {
```
cmuellner: Why is this pass restricted to `ldimmq{}`? It could operate over all `ldimmN{}`s. Note that a ldimmq{} exits to create a `Vreg64`, so it is also used for values which fit in the 8-bit range. E.g. the immediate 0x7f` might be shared with a `ldimmq{}` and a `ldimmb{}`.

swalk-cavium: @cmuellner - Hi Christoph, The intent of this pass was to reduce code size on ARMv8.  
Since we have to recreate some  immediates with multi-instruction sequences (movz/movk), it's checking a small leaky window to see if a close value is already in a register.  The window
size is optionable in case this increases register pressure.  Values which can be done in
one instruction are not considered.

cmuellner: True, I agree that a `ldimmb` will not have anything to share. But a `movz` instruction is limited to 16-bit values (shifted by 0/16/32/48). So a reuse of a constant from `ldimml` and `ldimmq` could be possible.

swalk-cavium: @cmuellner - I agree, it's possible, but it currently doesn't seem likely.  There's only a couple of places in the code where ldimml's are generated.  ldimmq is the  one that sticks out in the address calculation.  I was kind of hoping to land this one after almost 4 months.

hphp/runtime/vm/jit/vasm-arm.cpp
mxw: @swalk-cavium—Do you have any stats comparing placement of this pass before/after `foldImms()`?

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+TRACE_SET_MOD(vasm);
+
+namespace HPHP { namespace jit {
+
+namespace {
+
+// track ldimmq values
+struct ImmState {
+  ImmState() : val{0}, base{} {}
```
mxw: You don't need `, base{}`—also, I think it's a bit more self-documenting for the default constructor to be empty, and to have `Immed64 val{0}` below.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,161 @@
...
+// candidate around +/- uimm12
+folly::Optional<int> reuseCandidate(Env& env, int64_t p, Vreg& reg) {
+  for (auto const& elem : env.immStateVec) {
+    if (!elem.base.isValid()) continue;
+    int64_t q = elem.val.q();
+    if ((p >= q) && (p < (q + 4095))) {
+      reg = elem.base;
+      return folly::Optional<int>(safe_cast<int>(p - q));
+    }
+    if ((p < q) && (q < (p + 4095))) {
```
mxw: Combine this condition with the above, to avoid duplicating the bodies.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,161 @@
...
+}
+
+// candidate around +/- uimm12
+folly::Optional<int> reuseCandidate(Env& env, int64_t p, Vreg& reg) {
+  for (auto const& elem : env.immStateVec) {
+    if (!elem.base.isValid()) continue;
+    int64_t q = elem.val.q();
+    if ((p >= q) && (p < (q + 4095))) {
+      reg = elem.base;
+      return folly::Optional<int>(safe_cast<int>(p - q));
```
mxw: Use `folly::make_optional` instead.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,161 @@
...
+    }
+  }
+  return folly::none;
+}
+
+template <typename Inst>
+void reuseImmq(Env& env, const Inst& inst, Vlabel b, size_t i) {
+  // leaky bucket
+  env.immStateVec[i % RuntimeOption::EvalJitLdimmqSpan] = ImmState{};
```
mxw: Why mix this idiom with `reset()`?  Personally, I think this is clearer, so just kill `reset()`?

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,161 @@
...
+  }
+  env.immStateVec[i % RuntimeOption::EvalJitLdimmqSpan] = ImmState{ld.s, ld.d};
+}
+
+void reuseImmq(Env& env, Vlabel b, size_t i) {
+  assertx(i <= env.unit.blocks[b].code.size());
+  auto& inst = env.unit.blocks[b].code[i];
+
+  if (isBlockEnd(inst) || isCall(inst)) {
+    for (auto &elem : env.immStateVec) elem.reset();
```
mxw: `auto& elem`

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,161 @@
...
+      });
+      return;
+    }
+  }
+  env.immStateVec[i % RuntimeOption::EvalJitLdimmqSpan] = ImmState{ld.s, ld.d};
+}
+
+void reuseImmq(Env& env, Vlabel b, size_t i) {
+  assertx(i <= env.unit.blocks[b].code.size());
+  auto& inst = env.unit.blocks[b].code[i];
```
mxw: `auto const&`

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+  Env env { unit };
+  env.immStateVec.resize(RuntimeOption::EvalJitLdimmqSpan);
+
+  auto const labels = sortBlocks(unit);
+
+  // required for initial size of vector
+  for (auto i = 0; i < RuntimeOption::EvalJitLdimmqSpan; i++) {
+    env.immStateVec.push_back(ImmState{});
+  }
```
mxw: You already did a `resize()` above, which will default initialize `EvalJitLdimmqSpan` elements.  Now you're appending that many more elements, which will never be touched again.

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+  for (auto const b : labels) {
+    for (auto &elem : env.immStateVec) elem.reset();
```
mxw: `auto& elem` (here and elsewhere)

hphp/runtime/vm/jit/vasm-reuseimm.cpp
```diff
@@ -0,0 +1,161 @@
...
+    }
+  }
+  env.immStateVec[i % RuntimeOption::EvalJitLdimmqSpan] = ImmState{ld.s, ld.d};
+}
+
+void reuseImmq(Env& env, Vlabel b, size_t i) {
+  assertx(i <= env.unit.blocks[b].code.size());
+  auto& inst = env.unit.blocks[b].code[i];
+
+  if (isBlockEnd(inst) || isCall(inst)) {
```
mxw: Why do you need the `isBlockEnd()` check here if you're clearing the vector whenever we start processing a new block below?

hphp/runtime/vm/jit/vasm-arm.cpp
swalk-cavium: @mxw - No.

swalk-cavium: Actually, I'm not convinced it matters one way or the other.  The foldIms() pass is detecting the case
where the immediate can be pushed into the instruction.  This pass is just checking for immediates built up 
from multiple instructions.  @mxw, @cmuellner

cmuellner: Ok I understand, just three remarks:

1) with your approach you don't get immediates, that are shared, cannot be embedded into an (e.g. logical) instruction and build up within one mov* instruction (e.g. `ldimmq{0x2300, %0} ; andqi{%0, %1, %2}` -> `andqi{0x2300}` -> `movz x1, 0x2300 ;  and x0, x0, x1`). There would be optimizing potential for them as well.

2) your implementation of isMultiword() is too eager, as it does not consider all variants of mov* (movz 16, movz immr/imms, movk, movn). So you would reuse immediates, that would be embedded in an (e.g. logical) instruction anyways (e.g. `ldimmq{0xbfffffffffffffff, %0} ; andqi{%0, %1, %2}` -> `andqi{0xbfffffffffffffff}` -> `and x0, x0, 0xbfffffffffffffff`). This might end up in more code being generated (in the case when you emit an additional `add` to calculate the target immediate). When you put your pass after foldImms(), then you would not see that effect.

3) Regarding the `ldimml{}` (from the comments below): I agree that it is not used very often directly (only irlower-intrinsic.cpp uses that instruction directly). However in vasm-arm.cpp we are lowering `addli{}`, `andli{}`, `cmpli{}`, `subli{}`, `testli{}` and `storeli{}` to `ldimml{}`s. And these are emitted more often. So there would be optimizing potential for them as well (e.g. andli{} could share a non-embeddable mask (e.g. 0x2300)).

swalk-cavium: @cmuellner - I pretty much agree with all these points.  The initial intent was to have something that could start cleaning up the extra instructions around the addressing.
Ideally, there'd be a vixl callout to address \#2. above.   I suggest getting something in that's
correct and then improving forever.

Since I have to change the code again anyway, I can move that one call down.

swalk-cavium: Looks like there's something in the vixl repo that can be embraced.

    369 int MacroAssembler::MoveImmediateHelper(MacroAssembler* masm,
    370                                         const Register& rd,
    371                                         uint64_t imm) {
    372   bool emit_code = (masm != NULL);
src/vixl/a64/macro-assembler-a64.cc

cmuellner: You are right. I am obviously trying to force my ideas into your PR... Sorry for that.
So please put this pass after `foldImms()`, then I am fine.
I'll come up with a PR, which implements my ideas, once this PR lands (and when I'm sure that my idea improves performance in at least once benchmark).

Just one question regarding the naming: why not calling your pass `reuseImms()` (similar to `foldImms()` and `optimizeCopies()` and `annotateSFUses()` and `optimizeJmps()`)?

hphp/runtime/vm/jit/vasm-reuseimm.cpp
swalk-cavium: @mxw - Hi Max, without this code the transformation never fires since the part that does
the scanning things the vector doesn't have any elements.  We need the elements for the
leaky bucket.

swalk-cavium:
@mxw - Hi Max, Updates per last inspection.  Checked to make sure it's firing on grapheme.php,
ran regression tests with 6 option sets.  No new issues discovered.  Can you inspect?

hphp/runtime/vm/jit/vasm-reuseimm.cpp
mxw: I'm quite confident that this is not the case...

You're doing `env.immStateVec.resize(RuntimeOption::EvalJitLdimmqSpan);` above, after which the vector will contain a number of default-initialized ImmState elements equal to `EvalJitLdimmqSpan`.  But then you append _another_ set of ImmStates, and your vector ends up having twice as many as you use...

cf. http://en.cppreference.com/w/cpp/container/vector/resize

PR #7639, opened 2017-02-01
tests: Increase VMStackElms in stack-overflow tests.

cmuellner:
Fedora on aarch64 uses a default page size of 64k.
This setting breaks two stack-overflow tests, which set
the VMStackElms variable to 512 (0x200).

The error message, which comes from hphp/runtime/vm/bytecode.cpp, is:
Uncaught exception: VM stack size of 0x200 is below the minimum of 0x2000

This patch fixes the two stack-overflow tests by setting the value
of VMStackElms to 8192 (0x2000).

### DIFF
--- a/hphp/test/slow/stack-overflow/catch-trace-rematerialize.php.opts
+++ b/hphp/test/slow/stack-overflow/catch-trace-rematerialize.php.opts
@@ -1,1 +1,1 @@
--vEval.VMStackElms=512
+-vEval.VMStackElms=8192
--- a/hphp/test/slow/stack-overflow/reenter.php.opts
+++ b/hphp/test/slow/stack-overflow/reenter.php.opts
@@ -1,1 +1,1 @@
--vEval.VMStackElms=512
+-vEval.VMStackElms=8192


swalk-cavium:
@cmuellner - I've also seen these two tests fail on one of our Aarch64 Ubuntu machines.

cmuellner:
@swalk-cavium I've never seen this on Ubuntu (14.04). Do you have a diff of the output? Or does this patch fix your problem as well? Are you using a page size of 4k or another?

swalk-cavium:
@cmuellner - I think they changed one of the boot parameters so the page size was
65Kb.  It's one of the lab machines I sneak onto to do testing.  If you send me your e-mail
address I can share our my latest test results.  steve Dot walk aT cavium dOt com.  I don't
have a reasonable way to publish them yet.

apinski-cavium:
So internally at Cavium we sometimes use 4k page size kernel while other times we use 64k page size kernel even with userland of Ubuntu.   Ubuntu's default kernel page size is 4k while CentOS/RHEL/Fedora is 64k.  I can't remember what SUSE's default page size is (it might be 64k).

cmuellner:
@apinski-cavium OpenSUSE uses 4k pages as well.

PR #7641, opened 2017-02-01
Update third-party to make individual subprojects more modular.

aorenste:


aorenste:
Note that this will need https://github.com/hhvm/hhvm-third-party/pull/110 to land at the same time to remain consistent.

CMake/HPHPFunctions.cmake
```diff
@@ -1,3 +1,5 @@
+cmake_policy(SET CMP0026 OLD)
```
Orvid: What is this needed for?

CMakeLists.txt
```diff
@@ -1,5 +1,11 @@
 CMAKE_MINIMUM_REQUIRED(VERSION 2.8.7 FATAL_ERROR)
 
+if (NOT ${CMAKE_BINARY_DIR} STREQUAL ${CMAKE_SOURCE_DIR})
+  message(FATAL_ERROR
+    "HHVM requires you to build in-source.  Please run 'cmake .' "
```
Orvid: We don't, not entirely anyways.
I had to make a lot of changes to make it work for Windows, but it does work. We have conflicting `config.h` headers in our dependencies, but your work with setting up the individual projects correctly in third-party should be able to solve that by handling the include directory order correctly.

We don't respect it everywhere either, but it's better than it would be otherwise.

All of my work on Windows has been with out-of-tree builds.

hphp/CMakeLists.txt
```diff
@@ -28,6 +30,13 @@ if (EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/facebook")
                         ${CMAKE_MODULE_PATH})
 endif()
 
+find_package(PostgreSQL)
```
Orvid: Why? This doesn't belong here, it belongs in the extension config mechanism. (where it already is)

hphp/runtime/base/req-containers.h
```diff
@@ -344,7 +345,7 @@ template <typename T,
           typename Compare = std::less<T>>
 using priority_queue = std::priority_queue<T, Container, Compare>;
 
-#ifdef HAVE_BOOST1_49
+#if BOOST_VERSION > 104900
```
Orvid: We require 1.51 minimum, this is always true. Kill the macro entirely.

CMake/HPHPFunctions.cmake
aorenste: Without it newer versions of cmake complains about the use of `get_target_property(... LOCATION)`.  I couldn't figure out an easy way to avoid using it for now so this policy setting tells cmake to use the old behavior.
https://cmake.org/cmake/help/v3.0/policy/CMP0026.html

CMakeLists.txt
aorenste: Ah - I wasn't aware of that - I'll pull that out.

hphp/CMakeLists.txt
aorenste: I don't see any use of `find_package(PostgreSQL)` in any existing `CMakeLists.txt` or `*.cmake`.  I'll move it to HHVMExtensionConfig.cmake.

hphp/runtime/base/req-containers.h
aorenste: seems reasonable

aorenste:
@Orvid Okay - PR updated

CMake/HHVMExtensionConfig.cmake
```diff
@@ -72,6 +72,12 @@ include(CheckFunctionExists)
 include(HPHPFunctions)
 include(Options)
 
+find_package(PostgreSQL)
```
aorenste: I think it's available around 2.8.2+

aorenste:
@Orvid Okay - I think I moved the PostgreSQL stuff into the right place.

CMake/HPHPFunctions.cmake
Orvid: The problem is that this call to `cmake_policy` is an error on versions of CMake without that policy -_-...

That message is a warning, not an error.

Once we require CMake >= 3.0 we can get rid of it via generator expressions, but we've not tried to do that yet.

CMake/HHVMExtensionConfig.cmake
```diff
@@ -951,6 +951,13 @@ function (HHVM_EXTENSION_INTERNAL_HANDLE_LIBRARY_DEPENDENCY extensionID dependen
       HHVM_EXTENSION_INTERNAL_ADD_LINK_LIBRARIES(${MYSQL_CLIENT_LIBS})
     endif()
   elseif (${libraryName} STREQUAL "pgsql")
+    FIND_PACKAGE(PostgreSQL)
+    if (${PostgreSQL_VERSION_STRING} VERSION_LESS "9.1.0")
+      message(STATUS "System PostgreSQL version is too old: Disabling extension.")
+      set(HHVM_EXTENSION_${extensionID}_ENABLED_STATE 2 CACHE INTERNAL "" FORCE)
```
Orvid: Why are you doing this rather than `HHVM_EXTENSION_INTERNAL_SET_FAILED_DEPENDENCY(${extensionID} ${dependencyName})`?

This should also be completely replacing the FIND_PATH stuff below.

I'll put up a diff in a moment to switch over to that.

Orvid: Why are you doing this rather than `HHVM_EXTENSION_INTERNAL_SET_FAILED_DEPENDENCY(${extensionID} ${dependencyName})`?

This should also be completely replacing the FIND_PATH stuff below.

I'll put up a diff in a moment to switch over to that.

CMake/HHVMExtensionConfig.cmake
```diff
@@ -951,6 +951,13 @@ function (HHVM_EXTENSION_INTERNAL_HANDLE_LIBRARY_DEPENDENCY extensionID dependen
       HHVM_EXTENSION_INTERNAL_ADD_LINK_LIBRARIES(${MYSQL_CLIENT_LIBS})
     endif()
   elseif (${libraryName} STREQUAL "pgsql")
+    FIND_PACKAGE(PostgreSQL)
+    if (${PostgreSQL_VERSION_STRING} VERSION_LESS "9.1.0")
+      message(STATUS "System PostgreSQL version is too old: Disabling extension.")
+      HHVM_EXTENSION_INTERNAL_SET_FAILED_DEPENDENCY(${extensionID} ${dependencyName})
+      return()
+    endif()
+
```
Orvid: Drop this change entirely for now, there's work that still needs to be done before this can be switched.

PR #7644, opened 2017-02-02
Throw an exception when Redis returns an OOM error or aborts a transaction

yadavsaroj:
## Summary

When Redis is out of memory, the hhvm Redis client blocks until the socket times out while executing a transaction.

The problem is that `exec` calls `flushCallbacks` which ignores the errors coming back from Redis and is stuck waiting to get the responses for the commands in the transaction, which are never going to come.

More specifically, when Redis is OOM:
1. When the client queues a command in a transaction, Redis returns - `OOM command not allowed when used memory > 'maxmemory'`.
2. When the client tries to execute this transaction, Redis returns - `Transaction discarded because of previous errors`. 

In this PR, both of these errors are thrown as exceptions so the caller knows that there is a problem with Redis.

## Notes 
1. I have used Redis errors as is when throwing exceptions. `Transaction discarded because of previous errors` does not look very intuitive as a standalone exception. Is there is a better way here? 

1. I could not find a test file for existing exception cases in this function. What's the best place to add tests for such exceptions? I will be happy to add more tests.  

1. Finally, this is my first submission to hhvm. If I am off somewhere, happy to change things.

### DIFF
--- a/hphp/system/php/redis/Redis.php
+++ b/hphp/system/php/redis/Redis.php
@@ -1097,7 +1097,8 @@ class Redis {
             throw new RedisException("Sync with master in progress");
           }
           if (!strncmp($line, 'OOM ', 4)) {
-            throw new RedisException("OOM command not allowed when used memory > 'maxmemory'");
+            throw new RedisException(
+              "OOM command not allowed when used memory > 'maxmemory'");
           }
           if (!strncmp($line, 'EXECABORT ', 10)) {
             throw new RedisException("Transaction discarded because of previous errors");
--- a/hphp/system/php/redis/Redis.php
+++ b/hphp/system/php/redis/Redis.php
@@ -1096,6 +1096,12 @@ class Redis {
           if (!strncmp($line, '-ERR SYNC ', 10)) {
             throw new RedisException("Sync with master in progress");
           }
+          if (!strncmp($line, 'OOM ', 4)) {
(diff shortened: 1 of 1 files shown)


hphp/system/php/redis/Redis.php
```diff
@@ -1096,6 +1096,12 @@ protected function sockReadData(&$type) {
           if (!strncmp($line, '-ERR SYNC ', 10)) {
             throw new RedisException("Sync with master in progress");
           }
+          if (!strncmp($line, 'OOM ', 4)) {
+            throw new RedisException("OOM command not allowed when used memory > 'maxmemory'");
```
mofarrell: Max 80 characters per line please.
https://github.com/facebook/hhvm/blob/master/hphp/doc/coding-conventions.md

yadavsaroj: @mofarrell sorry about the delay. For some reason, I did not get notification from github about your feedback. Just pushed a change for it.

yadavsaroj:
@mofarrell Just wondering why this PR got closed.

mofarrell:
The diff was landed.  It is in master now.  See the latest commit for the file: https://github.com/facebook/hhvm/commits/master/hphp/system/php/redis/Redis.php

yadavsaroj:
:)

PR #7646, opened 2017-02-05
Fixing casing and missing getSortKey on Collator signatures

asuth:
See http://php.net/manual/en/class.collator.php

### DIFF
--- a/hphp/hack/hhi/stdlib/builtins_intl.hhi
+++ b/hphp/hack/hhi/stdlib/builtins_intl.hhi
@@ -390,14 +390,14 @@ class Collator {
   public function asort(&$arr, $sort_flag = null) {}
   public function compare($str1, $str2) {}
   static public function create($locale) {}
-  public function getattribute($attr) {}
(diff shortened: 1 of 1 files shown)


hphp/hack/hhi/stdlib/builtins_intl.hhi
```diff
@@ -390,14 +390,15 @@ class Collator {
...
-  public function getlocale($type = 0) {}
-  public function getstrength() {}
-  public function setattribute($attr, $val) {}
-  public function setstrength($strength) {}
-  public function sortwithsortkeys(&$arr) {}
+  public function getAttribute($attr) {}
+  public function getErrorCode() {}
+  public function getErrorMessage() {}
+  public function getLocale($type = 0) {}
+  public function getSortKey($str) {}
```
mofarrell: Could you add the corresponding `collator_get_sort_key($coll, $str)`

asuth: good catch. Done!

PR #7661, opened 2017-02-10
[AArch64] Adds relocation support.

dave-estes:


dave-estes:
Hey @mxw. Mentioning you here because I still can't find the option to add reviewers.

Also maybe you folks are interested in taking a look too, @swalk-cavium @cmuellnerTheobromaSystems ? Anyone else?

swalk-cavium:
How did you test?

dave-estes:
Existing unit tests and OSS

swalk-cavium:
I should have added more words.  I saw a command in the administrative server
that might be useful to exercise the code.  See /relocate, there's a random option.

dave-estes:
Oh yeah, thanks for the reminder. I stress tested using that feature quite a bit while flushing out some early bugs. It was pretty helpful. I ran the server and used `?relocate=random` as well as `?relocate=all`. Then I built a file of URLs using the test/slow unit tests. Then I ran siege with that URLs file.

hphp/runtime/vm/jit/vasm-emit.cpp
```diff
@@ -134,7 +134,7 @@ void emitVunit(Vunit& vunit, const IRUnit& unit,
   CodeBlock cold;
   CodeBlock* frozen = &code.frozen();
 
-  auto const do_relocate = arch_any(Arch::X64, Arch::PPC64) &&
+  auto const do_relocate = arch_any(Arch::X64, Arch::PPC64, Arch::ARM) &&
```
mxw: Just delete this first condition—now it's just checking that we're running on any architecture, which is trivially true...

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -481,6 +481,7 @@ void Vgen::emit(const mcprep& i) {
   auto const mov_addr = emitSmashableMovq(a->code(), env.meta, 0, r64(i.d));
   auto const imm = reinterpret_cast<uint64_t>(mov_addr);
   smashMovq(mov_addr, (imm << 1) | 1);
+  //smashMovq(a->code().toDestAddress(mov_addr), (imm << 1) | 1);
```
mxw: ?

hphp/runtime/vm/jit/tc-relocate.cpp
```diff
@@ -590,7 +590,9 @@ void liveRelocate(int time) {
   case Arch::PPC64:
     break;
   case Arch::ARM:
-    // Relocation is not supported on arm.
+  default:
```
mxw: Why did you add this?

hphp/runtime/vm/jit/tc-recycle.cpp
```diff
@@ -137,20 +139,33 @@ void clearProfCaller(TCA toSmash, const Func* func, int numArgs,
...
+    isBranch = instr->IsCondBranchImm() || instr->IsUncondBranchImm() ||
+               instr->IsUncondBranchReg() || instr->IsCompareBranch() ||
+               instr->IsTestBranch();
+    isNop = instr->Mask(SystemHintFMask) == SystemHintFixed &&
+            instr->ImmHint() == NOP;
+    isCall = instr->Mask(UnconditionalBranchMask) == BL ||
+             instr->Mask(UnconditionalBranchToRegisterMask) == BLR;
+#elif defined(__powerpc64__)
     ppc64_asm::DecodedInstruction di(start);
+    isBranch = di.isBranch(); isNop = di.isNop(); isCall =  di.isCall();
```
mxw: Each statement should have its own line.

hphp/runtime/vm/jit/tc-recycle.cpp
```diff
@@ -137,20 +139,33 @@ void clearProfCaller(TCA toSmash, const Func* func, int numArgs,
...
 #else
     x64::DecodedInstruction di(start);
+    isBranch = di.isBranch(); isNop = di.isNop(); isCall =  di.isCall();
```
mxw: Here also.

hphp/runtime/vm/jit/tc-recycle.cpp
```diff
@@ -137,20 +139,33 @@ void clearProfCaller(TCA toSmash, const Func* func, int numArgs,
 void clearTCMaps(TCA start, TCA end) {
   auto const profData = jit::profData();
   while (start < end) {
-#if defined(__powerpc64__)
+    bool isBranch, isNop, isCall;
+#if defined(__aarch64__)
```
mxw: Since you're adding these bools, you should just convert these `#if`s to be regular `if ()`s instead, switching on architecture.

hphp/runtime/vm/jit/tc-recycle.cpp
```diff
@@ -165,7 +180,11 @@ void clearTCMaps(TCA start, TCA end) {
         s_smashedCalls.erase(it);
       }
     }
+#if defined(__aarch64__)
+    start += vixl::kInstructionSize;
+#else
     start += di.size();
+#endif
```
mxw: Just add an `instSz` parameter above with the bools.

hphp/runtime/vm/jit/smashable-instr-arm.h
```diff
@@ -45,13 +44,13 @@ constexpr size_t smashableCallLen() { return 4 + 8 + 2 * 4; }
...
                       PhysReg d);
-TCA emitSmashableCmpq(CodeBlock& cb, CGMeta& fixups, int32_t imm,
+TCA emitSmashableCmpq(CodeBlock& cb, CGMeta& meta, int32_t imm,
                       PhysReg r, int8_t disp);
-TCA emitSmashableCall(CodeBlock& cb, CGMeta& fixups, TCA target);
-TCA emitSmashableJmp(CodeBlock& cb, CGMeta& fixups, TCA target);
-TCA emitSmashableJcc(CodeBlock& cb, CGMeta& fixups, TCA target,
+TCA emitSmashableCall(CodeBlock& cb, CGMeta& meta, TCA target);
+TCA emitSmashableJmp(CodeBlock& cb, CGMeta& meta, TCA target);
+TCA emitSmashableJcc(CodeBlock& cb, CGMeta& meta, TCA target,
```
mxw: \o/

hphp/runtime/vm/jit/smashable-instr-arm.h
```diff
@@ -17,6 +17,7 @@
 #ifndef incl_HPHP_JIT_SMASHABLE_INSTR_ARM_H_
 #define incl_HPHP_JIT_SMASHABLE_INSTR_ARM_H_
 
+#include "hphp/runtime/vm/jit/cg-meta.h"
```
mxw: Why did you add this?  This header doesn't seem to require the definition of CGMeta.

hphp/runtime/vm/jit/smashable-instr-arm.cpp
```diff
@@ -208,60 +231,85 @@ uint32_t smashableCmpqImm(TCA inst) {
...
+  Instruction* after = target->NextInstruction()->NextInstruction();
+  const auto rd = ldr->Rd();
+
+  assertx(b->IsCondBranchImm() &&
+          b->ImmPCOffsetTarget() == after &&
+          ldr->IsLoadLiteral() &&
+          ldr->Mask(LoadLiteralMask) == LDR_x_lit &&
+          ldr->ImmPCOffsetTarget() == target &&
+          br->Mask(UnconditionalBranchToRegisterMask) == BR &&
+          br->Rn() == rd);
```
mxw: Can all of these changes be pulled into a separate PR?

The more you can split this up, the better.

dave-estes:
Thanks, @mxw. Leaving the office for two days of PTO soon. Will work on review changes first thing on Monday.

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -0,0 +1,993 @@
...
+   | that is bundled with this package in the file LICENSE, and is        |
+   | available through the world-wide-web at the following url:           |
+   | http://www.php.net/license/3_01.txt                                  |
+   | If you did not receive a copy of the PHP license and are unable to   |
+   | obtain it through the world-wide-web, please send a note to          |
+   | license@php.net so we can mail you a copy immediately.               |
+   +----------------------------------------------------------------------+
+*/
+
+#include "hphp/runtime/vm/jit/relocation.h"
```
mxw: The first include should always be the corresponding header—in this case, `relocation-arm.h`.

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -0,0 +1,993 @@
...
+ * in the instruction stream atomically. We can't, however, update
+ * multiple instructions atomically since there may be a thread
+ * executing instructions within that range. Therefore, MOV/MOVK
+ * instructions can't be adjusted for live ranges of instructions.
+ * Any emitted code that uses MOV/MOVK and which must be adjusted
+ * while live, should instead use LDR (literal). Smashables use this
+ * approach; they embed a literal in the instruction stream which
+ * can then be updated atomically. At the time of this writing, only
+ * the vasm opcode (leap) produces a live region that must be updated,
+ * and so it uses an approach with LDR (literal) instead of MOV/MOVK.
```
mxw: Thanks for documenting all of this!  A lot of it is well-suited to accompany the APIs that are laid out in `relocation.h`.  Can you move the non-ARM-specific bits there, and leave the rest here?

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -0,0 +1,993 @@
...
+      src_count = smashableCallLen() >> vixl::kInstructionSizeLog2; dest_count = 1;
+      for (auto i = Instruction::Cast(src);
+           i < Instruction::Cast(src + src_count * kInstructionSize);
+           i = i->NextInstruction()) {
+        rewrites.insert(i);
+      }
+      return true;
+    }
+  }
+#endif
```
mxw: Delete this.

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -0,0 +1,993 @@
...
+    return TcaRange(s, e);
+  }
+  if (s && !e) {
+    return TcaRange(s, s + rng.size());
+  }
+  if (!s && e) {
+    return TcaRange(e - rng.size(), e);
+  }
+  return rng;
+}
```
mxw: This looks like it's duplicated for every architecture.  Can you just add it to `relocation.h` and implement it a single time in `relocation.cpp`?

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -0,0 +1,993 @@
...
+
+void fixupRanges(AsmInfo* asmInfo, AreaIndex area, RelocationInfo& rel) {
+  asmInfo->clearBlockRangesForArea(area);
+  for (auto& ii : asmInfo->instRangesForArea(area)) {
+    ii.second = fixupRange(rel, ii.second);
+    asmInfo->updateForBlock(area, ii.first, ii.second);
+  }
+}
```
mxw: Same with this.

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -0,0 +1,993 @@
...
+             * Calculate the new offset and update. We're just adjusting, and
+             * not relocating. Therefore we can't change relative instructions
+             * to absolute. Our only recourse is to mark it as far and then
+             * retry the entire relocation again.
+             */
+            int imm = Instruction::Cast(new_target) - instr2;
+            if ((instr->IsPCRelAddressing() && !is_int21(imm)) ||
+                (instr->IsLoadLiteral() && !is_int19(imm)) ||
+                (instr->IsCondBranchImm() && !is_int19(imm >> vixl::kInstructionSizeLog2)) ||
+                (instr->IsUncondBranchImm() && !is_int26(imm >> vixl::kInstructionSizeLog2))) {
```
mxw: I can't tell for sure on GitHub, but this looks way longer than 80 chars to me.

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -0,0 +1,993 @@
...
+        (instr->Mask(UnconditionalBranchToRegisterMask) == BLR)) {
+      if (auto fixup = FixupMap::findFixup(start)) {
+        meta.fixups.emplace_back(start, *fixup);
+      }
+      if (auto ct = getCatchTrace(start)) {
+        meta.catches.emplace_back(start, *ct);
+      }
+    }
+  }
+}
```
mxw: Looks like this could be largely shared between the three architectures also, with maybe just a small handful of arch-specific metadata adjustments.  Can you nix this from the individual `relocation-xxx.h` APIs, and just have one shared implementation in `relocation.cpp`?

It should be easy to diff the existing individual implementations; they're largely identical—I think it's just that x64 does literals and code pointers, ARM does smashables, and PPC does neither.

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -0,0 +1,993 @@
...
+
+  if (target) {
+    TCA adjusted = rel.adjustedAddressAfter((TCA)target);
+    if (!adjusted) { adjusted = (TCA)target; }
+    int imm = (adjusted - dest) >> vixl::kInstructionSizeLog2;
+    if (is_int26(imm) && !far.count(instr)) {
+      vixl::MacroAssembler a { dest_block };
+      dest_block.setFrontier(dest);
+      a.b(imm);
+      src_count = smashableJmpLen() >> vixl::kInstructionSizeLog2; dest_count = 1;
```
mxw: Please don't write more than one statement per line—it impedes readability.

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -0,0 +1,993 @@
...
+      instr->IsLoadLiteral() ||
+      instr->IsCondBranchImm() ||
+      instr->IsUncondBranchImm()) {
+
+    TCA target = (TCA)instr->ImmPCOffsetTarget();
+    size_t range = end - start;
+
+    // If the target is outside of the range of this relocation,
+    // then update it.
+    if (size_t(target - start) >= range) {
```
mxw: Also, is it possible for `target` to be before `start`?

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -0,0 +1,993 @@
...
+    }
+
+    if (instr->IsCondBranchImm() ||
+        instr->IsUncondBranchImm()) {
+      jmp_dest = target;
+    }
+    return true;
+  } else {
+    return false;
+  }
```
mxw: Unnecessary `else`; just `return false;` is sufficient.

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -0,0 +1,993 @@
...
+     */
+    if (next->Mask(LoadStoreUnsignedOffsetMask) == LDR_x_unsigned
+        && next->Rt() == rd) {
+      // Only transform if the MOV/MOVK sequence def'd a vixl scratch,
+      // otherwise we run the risk of not def'ing a live register.
+      auto const dst = vixl::Register(rd, 64);
+      if (dst.Is(rVixlScratch0) || dst.Is(rVixlScratch1)) {
+        if (is_int19(imm)) {
+          // TODO(CDE): Support MOV/MOVK, LDR to LDR literal.
+          not_implemented();
```
mxw: We should probably fail more gracefully...

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -0,0 +1,993 @@
...
+    /*
+     * If the next instruction after a sequence of MOV/MOVK uses the
+     * target, see if the entire sequence can be converted to a
+     * single PC relative instruction. Supported transformations include:
+     *   MOV/MOVK, LDR to LDR literal - Not yet implemented
+     *   MOV/MOVK, BR<L> to B<L>
+     *   MOV/MOVK absolute to ADR
+     */
+    if (next->Mask(LoadStoreUnsignedOffsetMask) == LDR_x_unsigned
+        && next->Rt() == rd) {
```
mxw: Trailing `&&` preferred over leading `&&`.

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -0,0 +1,993 @@
...
+
+  return false;
+}
+
+/*
+ * Returns true if the source is a PC Relative instruction.
+ *   ADR/ADRP
+ *   LDR (literal)
+ *   B.<cc> (immediate) - Not yet supported as it's not used.
+ *   B (immediate)
```
mxw: It also seems to do a bunch of other stuff (like potentially overwriting an instruction in `dest_block`) and it takes a bunch of "out" arguments by reference.  Can you document the behavior more thoroughly?

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -0,0 +1,993 @@
...
+  } else {
+    return false;
+  }
+}
+
+/*
+ * Returns true if the instruction(s) contain an immediates
+ *   MOV/MOVK
+ *   LDR (literal) - Not supported as it's highly unlikely to have a
+ *                   19 bit address
```
mxw: This also needs more documentation.

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -0,0 +1,993 @@
...
+            a.bl(imm);
+          }
+
+          dest_count += (dest_block.frontier() - dest) / kInstructionSize;
+          src_count = (next - instr) / kInstructionSize + 1;
+          is_absolute = false;
+
+          // Add range of source instructions to rewrites (MOV/MOVKs/B<L>)
+          for (auto i = instr;
+               i < next->NextInstruction();
```
mxw: Why `next->NextInstruction()` here...

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -0,0 +1,993 @@
...
+        vixl::MacroAssembler a { dest_block };
+        auto const dst = vixl::Register(rd, 64);
+        a.adr(dst, imm);
+
+        dest_count += (dest_block.frontier() - dest) / kInstructionSize;
+        src_count = (next - instr) / kInstructionSize;
+        is_absolute = false;
+
+        // Add range of source instructions to rewrites (MOV/MOVKs)
+        for (auto i = instr; i < next; i = i->NextInstruction()) {
```
mxw: ...but not here?

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -0,0 +1,993 @@
...
+            TCA target = *addr;
+            TCA adjusted = rel.adjustedAddressAfter(target);
+
+            if (!adjusted) {
+              // Consider the case of a non-initialized mcprep smashableMovq
+              target = reinterpret_cast<TCA>((uint64_t(target) >> 1));
+              adjusted = rel.adjustedAddressAfter(target);
+              if (adjusted) {
+                adjusted = reinterpret_cast<TCA>((uint64_t(adjusted) << 1) | 1);
+              }
```
mxw: Just to make sure I'm not totally off-the-mark—the reason you have to deal with this case, and literals in general, is because unlike on x64, you can't just embed your immediates within instructions, and thus the memory location in the instruction stream where the immediate is found might require position adjustment.  Is that correct?

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -0,0 +1,993 @@
...
+              instr->IsLoadLiteral() ||
+              instr->IsCondBranchImm() ||
+              instr->IsUncondBranchImm()) {
+            TCA old_target = (TCA)instr->ImmPCOffsetTarget();
+            TCA adjusted_target = rel.adjustedAddressAfter(old_target);
+            TCA new_target = (adjusted_target) ? adjusted_target : old_target;
+
+            /*
+             * Calculate the new offset and update. We're just adjusting, and
+             * not relocating. Therefore we can't change relative instructions
```
mxw: What do you mean by "We're just adjusting, and not relocation"?

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -0,0 +1,993 @@
...
+    if (internal_refs_need_update) {
+      src = start;
+      bool ok = true;
+      while (src != end) {
+        Instruction* instr = Instruction::Cast(src);
+        TCA dest = rel.adjustedAddressAfter(src);
+        Instruction* instr2 = Instruction::Cast(dest);
+
+        // Adjust this instruction if A) it wasn't written from a pc relative
+        // instruction to an aboslute (or vice-versa) and B) it isn't a literal.
```
mxw: `s/aboslute/absolute/`

hphp/runtime/vm/jit/relocation-arm.cpp
mxw: You probably need to include both, still... but I guess it's fine to do what the other files are doing for now.

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -0,0 +1,993 @@
...
+
+          dest_count += (dest_block.frontier() - dest) / kInstructionSize;
+          is_relative = false;
+        }
+      } else if (instr->IsCondBranchImm()) {
+        imm >>= vixl::kInstructionSizeLog2;
+        // TODO(CDE): Suppport relative-to-absolute relocation for
+        //            conditional branches.
+        assert_flog(is_int19(imm),
+                    "Can't yet relocate B.<cc>, imm won't fit in 19 bits.\n");
```
mxw: I'm a little skeptical of committing code that ignores a case just because our tests don't trigger it.  You could try writing a C++ unit test (there should be some in runtime/vm/jit/tests or runtime/tests, etc.) that creates a massive Vunit and then runs relocation on it.

hphp/runtime/vm/jit/relocation-arm.cpp
mxw: The optimal optimizations we can definitely skip; but this pass should be correct, even on workloads not exercised by our unit tests.

hphp/runtime/vm/jit/tc-relocate.cpp
```diff
@@ -773,6 +777,12 @@ void relocateTranslation(
   }
   memset(main.base(), 0xcc, main.frontier() - main.base());
   memset(cold.base(), 0xcc, cold.frontier() - cold.base());
+#ifdef __aarch64__
+  __builtin___clear_cache(reinterpret_cast<char*>(main.base()),
+                          reinterpret_cast<char*>(main.frontier()));
+  __builtin___clear_cache(reinterpret_cast<char*>(cold.base()),
+                          reinterpret_cast<char*>(cold.frontier()));
+#endif
```
mxw: Use `if (arch() == Arch::ARM) { ...` instead.

hphp/tools/hfsort/jitsort.cpp
```diff
@@ -218,7 +218,7 @@ int jitsort(int pid, int time, FILE* perfSymFile, FILE* relocResultsFile) {
 
     auto perfCmd =
         "perf record -g -p {0} -e instructions -o {1} -- sleep {2} && "
-        "perf script -i {1} -f comm,ip > {3}";
+        "perf script -i {1} -F comm,ip > {3}";
```
mxw: This doesn't seem like it belongs in this diff.

hphp/runtime/vm/jit/tc-recycle.cpp
```diff
@@ -137,20 +139,44 @@ void clearProfCaller(TCA toSmash, const Func* func, int numArgs,
...
+      instSz = di.size();
+    } else if (arch() == Arch::X64) {
+      x64::DecodedInstruction di(start);
+      isBranch = di.isBranch();
+      isNop = di.isNop();
+      isCall = di.isCall();
+      instSz = di.size();
+    } else {
+      not_reached();
+    }
```
mxw: This should probably be a switch statement.

hphp/runtime/vm/jit/smashable-instr-arm.cpp
```diff
@@ -261,7 +273,25 @@ ConditionCode smashableJccCond(TCA inst) {
   if (br->Bits(31, 10) != 0x3587C0 || br->Bits(4, 0) != 0) return CC_NP;
 
   return arm::convertCC(InvertCondition(static_cast<Condition>(b->Bits(3, 0))));
+}
+
+///////////////////////////////////////////////////////////////////////////////
+
+TCA getSmashableFromTargetAddr(TCA addr) {
```
mxw: Can you document this function (in the header file)?

I'm not totally clear on what it does... it looks like if you pass it an address which is the target of an immediately preceding smashable jmp, you get that smashable jmp returned.  That's bizarrely specific for such a general-sounding function name, but maybe I'm misunderstanding something.

hphp/runtime/vm/jit/relocation-arm.cpp
mxw: It's possible it wouldn't even be a benefit, since right now the literals have good locality with the instruction we're executing.

hphp/runtime/vm/jit/relocation-arm.cpp
mxw: poke

hphp/tools/hfsort/jitsort.cpp
dave-estes: Moved to #7781.

mxw:
This is finally ready for a rebase over the landed PRs :D

mxw:
I'll try to take a look at this before Monday.  It looks like there are some conflicts to rebase over, but I think they're relatively trivial.

jim-saxman:
@dave-estes @mxw  I rebased this patch this morning, and we have many quick failures. reverting fe3dfa7  "Add ir and vasm for func guard emission" results in quick passing.

mxw:
@jim-saxman—Failures with or without this PR patched?  I'm not seeing any regression on @cmuellner's CI setup

mxw:
If it's just with this PR—I think relocation will now get run on the func guard code where it wasn't before, so it seems possibly that your relocator is messing with the func guard instruction sequence when it shouldn't be, or perhaps not setting the watchpoint correctly.

I'll start taking a look here concurrently, in any case.

mxw:
Oh—maybe the func guard isn't being marked smashable?  That would explain why x64 isn't affected, since we don't use that machinery, I'm pretty sure.

Sorry for the rapid-fire comments!

dave-estes:
I'm back from holiday, so I'll take a look too.

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -0,0 +1,951 @@
...
+ *       sequence can be optimized during static relocation, but that an
+ *       actual smashable can't be optimized until live relocation and then
+ *       only if it has been smashed. Currently there is no means to know
+ *       if a smashable has been smashed, and so this optimization of
+ *       smashables during live relocation is not yet performed at all.
+ *
+ * dest_count, src_count and rewrites are updated to reflect when an
+ * instruction(s) is rewritten to a different instruction sequence.
+ */
+bool relocateSmashableHelper(RelocationInfo& rel, CGMeta& meta,
```
mxw: nit: Why `Helper` in all these names?

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -0,0 +1,951 @@
...
+
+    /*
+     * Finally update any internal refs if needed. This indicates that the
+     * range of instructions grew/shrank and therefore the internal refs
+     * may be off.
+     */
+    if (internal_refs_need_update) {
+      src = start;
+      bool ok = true;
+      while (src != end) {
```
mxw: Can you make into for loops all the while loops you've written whose last line is the increment statement?  It makes it a lot easier to see the loop at a high level without reading for detail.

(Also, this loop should probably be over Instruction*, rather than raw addresses.)

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -0,0 +1,951 @@
...
+           *   LDR (literal)
+           *   B[.<cc>] (immediate)
+           */
+          if (instr->IsPCRelAddressing() ||
+              instr->IsLoadLiteral() ||
+              instr->IsCondBranchImm() ||
+              instr->IsUncondBranchImm()) {
+            auto old_target = reinterpret_cast<TCA>(instr->ImmPCOffsetTarget());
+            auto adjusted_target = rel.adjustedAddressAfter(old_target);
+            auto new_target = (adjusted_target) ? adjusted_target : old_target;
```
mxw: Unnecessary parens.

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -0,0 +1,951 @@
...
+
+            /*
+             * Calculate the new offset and update. At this stage, we've already
+             * relocated and now We're just adjusting an internal reference.
```
mxw: `s/We're/we're/`

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -0,0 +1,951 @@
...
+
+              // Write the new mov/movk sequence.
+              vixl::MacroAssembler a { dest_block };
+              auto const dst = vixl::Register(rd, 64);
+              a.Mov(dst, adjusted);
+
+              // If the new sequence is longer than the original, then we must
+              // gracefully fail.
+              length -= (dest_block.frontier() - dest) / kInstructionSize;
+              if (length < 0) {
```
mxw: This is never true, since `length` is unsigned.

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -0,0 +1,951 @@
...
+            auto next = instr->NextInstruction();
+            while (next->IsMovk()) {
+              if (next->Rd() == rd) {
+                target |= next->ImmMoveWide() << (16 * next->ShiftMoveWide());
+                length++;
+              }
+              next = next->NextInstruction();
+            }
+            // Adjust the mov/movk sequence if necessary
+            auto adjusted = rel.adjustedAddressAfter((TCA)target);
```
mxw: Prefer C++-style casts in general.

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -0,0 +1,951 @@
...
+                   Assembler::ShiftMoveWide(adjLength));
+          next->SetInstructionBits(bits);
+          adjLength++;
+        }
+        always_assert_flog(length >= adjLength,
+                           "Can't adjust MOV/MOVK, new sequence is longer.\n");
+        shift++;
+        next = next->NextInstruction();
+        tmp >>= 16;
+      }        while (tmp);
```
mxw: nit: whitespace

hphp/runtime/vm/jit/relocation-arm.cpp
dave-estes: Done.

swalk-cavium:
@dave-estes - Once your rebase is done, would you like me to test some?

dave-estes:
Sure, @swalk-cavium. That'd be great.

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -126,367 +163,342 @@ InstrSet findLiterals(Instruction* start, Instruction* end) {
...
+  srcCount = smashableJmpLen() >> vixl::kInstructionSizeLog2;
+  destCount = 1;
+  for (auto i = src;
+       i < src + srcCount * kInstructionSize;
+       i = i->NextInstruction()) {
+    env.rewrites.insert(i);
+  }
+  env.updateInternalRefs = true;
+  FTRACE(3,
+         "Relocated smashable at src 0x{:08x} with target 0x{:08x} to 0x{:08x} (0x{:08x})\n",
```
mxw: nit: Line length.

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -126,367 +163,342 @@ InstrSet findLiterals(Instruction* start, Instruction* end) {
...
+    /*
+     * Calculate the new offset and determine if it can be encoded
+     * in a PC relative instruciton or if it needs to be converted
+     * to make use of an absolute target.
+     * Note: Use the VIXL scratch registers when transforming. Their
+     *       scope is just a single macroassembler directive, whereas
+     *       the scope of rAsm is an entire vasm instruction.
+     */
+    int imm = src->ImmPCOffsetTarget() - dest;
+    bool isRelative = true;
```
mxw: This is a note rather than a nit, but: you shouldn't feel compelled to adhere to a particular case convention when writing new code.  Be consistent within a single file or subsystem (e.g., in `vasm-*` code, we tend towards `lower_case_with_underscores` everywhere), but outside of that, just do what's most natural for you.

(I only mention this because I see that you camelCase'd everything between your previous revision and this one.)

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -126,367 +163,342 @@ InstrSet findLiterals(Instruction* start, Instruction* end) {
...
-                    CGMeta& meta,
-                    TCA* exit_addr,
-                    InstrSet& far) {
-  auto src = start;
-  bool internal_refs_need_update = false;
-  auto dest_start = dest_block.frontier();
-  size_t asm_count{0};
-  TCA jmp_dest = nullptr;
+size_t relocateImpl(Env& env) {
+  auto destStart = env.destBlock.frontier();
```
mxw: nit: `auto const`

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -497,120 +509,108 @@ size_t relocateImpl(RelocationInfo& rel,
...
       bool ok = true;
-      while (src != end) {
-        auto instr = Instruction::Cast(src);
-        auto dest = rel.adjustedAddressAfter(src);
-        auto instr2 = Instruction::Cast(dest);
+      for (auto src = Instruction::Cast(env.start);
+           src < Instruction::Cast(env.end);
+           src = src->NextInstruction()) {
+        auto destAddr = env.rel.adjustedAddressAfter(reinterpret_cast<TCA>(src));
+        auto dest = Instruction::Cast(destAddr);
```
mxw: nit: `auto const` on these lines.

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -497,120 +509,108 @@ size_t relocateImpl(RelocationInfo& rel,
...
+      // Update the destAddr and reset the frontier
+      destAddr += destCount * kInstructionSize;
+      assertx(destAddr <= env.destBlock.frontier());
+      env.destBlock.setFrontier(destAddr);
 
-    if (exit_addr) {
-      *exit_addr = jmp_dest;
-    }
+      asmCount += destCount;
+    } // while (src != env.end)
```
mxw: This comment is a lie now, and also most editors make it pretty easy to chase the start of a scope.

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -779,7 +778,7 @@ void adjustInstruction(RelocationInfo& rel, Instruction* instr,
       }
       next = next->NextInstruction();
     }
-    auto adjusted = (uint64_t)rel.adjustedAddressAfter((TCA)target);
+    auto adjusted = (uint64_t)rel.adjustedAddressAfter(reinterpret_cast<TCA>(target));
```
mxw: nit: Line length.

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -318,6 +320,55 @@ bool relocatePCRelative(Env& env, Instruction* src, Instruction* dest,
...
+      if (!is_int19(imm) || env.far.count(src)) {
+        env.destBlock.setFrontier(destAddr);
+        destCount--;
+
+        vixl::MacroAssembler a { env.destBlock };
+        vixl::Label end;
+        auto const rt = vixl::Register(src->Rt(), 64);
+        auto const tmp = rVixlScratch0;
+        a.SetScratchRegisters(vixl::NoReg, vixl::NoReg);
+	if (src->Mask(CompareBranchMask) == CBZ_x) {
```
mxw: indentation

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -318,6 +320,55 @@ bool relocatePCRelative(Env& env, Instruction* src, Instruction* dest,
...
+        env.destBlock.setFrontier(destAddr);
+        destCount--;
+
+        vixl::MacroAssembler a { env.destBlock };
+        vixl::Label end;
+        auto const bit_pos = src->ImmTestBranchBit40();
+        auto const rt = vixl::Register(src->Rt(), 64);
+        auto const tmp = rVixlScratch0;
+        a.SetScratchRegisters(vixl::NoReg, vixl::NoReg);
+	if (src->Mask(TestBranchMask) == TBZ) {
```
mxw: indentation

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -579,11 +632,15 @@ size_t relocateImpl(Env& env) {
...
            *   B[.<cc>] (immediate)
+	   *   CB[N]Z
+	   *   TB[N]Z
            */
           if (src->IsPCRelAddressing() ||
               src->IsLoadLiteral() ||
               src->IsCondBranchImm() ||
-              src->IsUncondBranchImm()) {
+              src->IsUncondBranchImm() ||
+	      src->IsCompareBranch() ||
```
mxw: indentation

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -318,6 +320,55 @@ bool relocatePCRelative(Env& env, Instruction* src, Instruction* dest,
         destCount += (env.destBlock.frontier() - destAddr) / kInstructionSize;
         isRelative = false;
       }
+    } else if (src->IsCompareBranch()) {
+      imm >>= vixl::kInstructionSizeLog2;
+      if (!is_int19(imm) || env.far.count(src)) {
```
mxw: Consider writing a helper `imm_fits_instr(int imm, const Instruction* instr)` which does the work of switching on the instruction kind, shifting the immediate, and dispatching to the appropriate size check.  There are a few places where you're hand-rolling this dispatch that could be papered over, such as...

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -601,7 +658,11 @@ size_t relocateImpl(Env& env) {
                 (src->IsCondBranchImm() &&
                  !is_int19(imm >> vixl::kInstructionSizeLog2)) ||
                 (src->IsUncondBranchImm() &&
-                 !is_int26(imm >> vixl::kInstructionSizeLog2))) {
+                 !is_int26(imm >> vixl::kInstructionSizeLog2)) ||
```
mxw: ...here

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -740,6 +805,14 @@ void adjustInstruction(RelocationInfo& rel, Instruction* instr,
         imm >>= vixl::kInstructionSizeLog2;
         always_assert_flog(is_int26(imm),
           "Can't adjust B, imm won't fit in 26 bits.\n");
+      } else if (instr->IsCompareBranch()) {
```
mxw: ...also here

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -0,0 +1,951 @@
...
+      return true;
+    }
+  }
+
+  /*
+   * Smashable calls are not yet supported, because we don't have any
+   * reloc debug info logged when smashing a call.
+   */
+
+  return false;
```
mxw: Yep, that makes sense—I mostly just meant that you didn't need to leave an opening for handling calls, since you can just do it in a separate helper.

hphp/runtime/vm/jit/relocation-arm.cpp
mxw: I personally use `lower_case_with_underscores` in all new code I write, so if that's your inclination, I'll never complain :]

hphp/runtime/vm/jit/relocation-arm.cpp
mxw: It would definitely be good to get this refactored a bit (and I suspect an extra decode won't be a big deal), but doing it as a separate PR seems perfectly fine to me.  This has been in rebase hell long enough (which is my fault)—let's get it landed :]

mxw:
Looks good to me.  I'm going to ping some other reviewers internally and see if they have major comments.  Otherwise, I'm hoping to land this soon.  Thanks for nursing this for so long—you've been more patient than should be necessary, and I'll try to speed up my response times in the future.

PR #7701, opened 2017-02-28
[AArch64] Pick a different VA due to CentOS kernel

jim-saxman:
The Aarch64 CentOS 7.3.1611 kernel doesn't support a full 48-bit
VA space, so a new location for the FIXED address is required.


jim-saxman:
@mxw @apinski-cavium @swalk-cavium @cestes @cmuellnerTheobromaSystems  Please review.

cmuellner:
Successfully tested with Linux version 4.3.0-gryphon_evt1_sw_1.02.03+ with 4 kB pages (on an APM Gryphon node (Merlin equivalent)), by running "hhvm hphp/test/run all" and observing no additional fails.

PR #7704, opened 2017-03-01
Make build use ocaml from third-party with not installed on the host …

aloiret:
…system

Summary:
Whether ocaml was built from third-party or not, gen_build_id.ml was always
called using the host ocaml binary. This led to a build failure when no OCaml
interpreter was installed on the host system.

Signed-off-by: Arthur Loiret <arthur.loiret@adeline.mobi>


aloiret:
Also see https://github.com/hhvm/hhvm-third-party/pull/113

mofarrell:
This is waiting on a third-party update.  Right now there the `OCAML_EXECUTABLE` var is still not defined.

mofarrell:
This was landed as a part of another change.

PR #7707, opened 2017-03-02
[Aarch64] setcc/movzbl -> setcc peephole

swalk-cavium:
This improves JIT code generation for Aarch64 be removing unneeded zero extend
instructions.

Before
=====
    cset w0,rel
    uxtb x0,w0

After
=====
    cset W0,rel

The write to a 32-bit register clears the upper portion anyway, so the instruction is not needed.
This saved over 100 instructions in the vm-perf/center-of-mass.php benchmark.

The regression suite was run with 6 different option sets.  No new regressions were noted.

### DIFF
--- a/hphp/runtime/vm/jit/vasm-simplify.cpp
+++ b/hphp/runtime/vm/jit/vasm-simplify.cpp
@@ -522,18 +522,31 @@ bool simplify(Env& env, const pop& inst, Vlabel b, size_t i) {
   });
 }
 
-bool simplify(Env& env, const movzbq& inst, Vlabel b, size_t i) {
+template<typename Movzb>
+bool simplify_movzb(Env& env, const Movzb& inst, Vlabel b, size_t i) {
   auto const def_op = env.def_insts[inst.s];
 
   if (def_op != Vinstr::setcc) return false;
   if (!(arch() == Arch::ARM &&
     width(def_op) != Width::Long)) return false;
 
-  // movzbq{} is redundant--operation on 32-bit registers clear the upper bits.
+  // movzbX{} is redundant--operation on 32-bit registers clear the upper bits.
   return simplify_impl(env, b, i, [&] (Vout& v) {
     v << copy{inst.s, inst.d};
     return 1;
   });
+}
+
+bool simplify(Env& env, const movzbl& inst, Vlabel b, size_t i) {
+  return simplify_movzb<movzbl>(env, inst, b, i);
+}
+
+bool simplify(Env& env, const movzbw& inst, Vlabel b, size_t i) {
+  return simplify_movzb<movzbw>(env, inst, b, i);
+}
+
+bool simplify(Env& env, const movzbq& inst, Vlabel b, size_t i) {
+  return simplify_movzb<movzbq>(env, inst, b, i);
 }
 
 ///////////////////////////////////////////////////////////////////////////////
--- a/hphp/runtime/vm/jit/vasm-simplify.cpp
+++ b/hphp/runtime/vm/jit/vasm-simplify.cpp
@@ -522,6 +522,20 @@ bool simplify(Env& env, const pop& inst, Vlabel b, size_t i) {
   });
 }
 
+bool simplify(Env& env, const movzbq& inst, Vlabel b, size_t i) {
+  auto const def_op = env.def_insts[inst.s];
+
+  if (def_op != Vinstr::setcc) return false;
+  if (!(arch() == Arch::ARM &&
+    width(def_op) != Width::Long)) return false;
+
+  // movzbq{} is redundant--operation on 32-bit registers clear the upper bits.
+  return simplify_impl(env, b, i, [&] (Vout& v) {
+    v << copy{inst.s, inst.d};
+    return 1;
+  });
+}
+
 ///////////////////////////////////////////////////////////////////////////////
 
 /*


ptomsich:
uxtb is a 8-bit zero-extend... so it it won't matter that there's an implicit zero-extend on results being moved to w-form registers.

In fact, this could be generalised to eliminate any extension-op operating on the result of a cset (as the result of the cset is always a zero-extende 1 or 0).

swalk-cavium:
@ptomsich  - Ah, you're right of course.  The uxtb ones jumped out when I was examining the code.  I can add the others in another pull request.

cmuellner:
@swalk-cavium can you add a bit more context to the "before" (best would be the whole block with the IR). I'd like to find out, why this uxtb is created at all. It's not the emit code, since we only do setcc/Cset in a single line without a `uxtb`:

`void emit(const setcc& i) { a->Cset(W(i.d), C(i.cc)); }`

swalk-cavium:
123723     --- bc 326, fp 0, spOff 6 (center)
 123724  326: CGetL2 L:2
 123725     (90) StStk<IRSPOff 0> t1:StkPtr, t26:Int
 123726         Main:
 123727               0x24401c7c  381a83a1              sturb w1, [x29, #-88]
 123728               0x24401c80  f81a03a0              stur x0, [x29, #-96]
 123729 
 123730     (91) StStk<IRSPOff -1> t1:StkPtr, t27:Int
 123731         Main:
 123732               0x24401c84  381983a1              sturb w1, [x29, #-104]
 123733               0x24401c88  f81903a2              stur x2, [x29, #-112]
 123734 
 123735 
 123736     --- bc 328, fp 0, spOff 7 (center)
 123737  328: Lt
 123738     (94) t28:Bool = LtInt t26:Int, t27:Int
 123739         Main:
 123740               0x24401c8c  eb02001f              cmp x0, x2
 123741               0x24401c90  1a9fa7e0              cset w0, lt
 123742 
 123743     (95) StStk<IRSPOff 0> t1:StkPtr, t28:Bool
 123744         Main:
 123745               0x24401c94  52800421              movz w1, #0x21
 123746               0x24401c98  381a83a1              sturb w1, [x29, #-88]
 123747               0x24401c9c  d3401c01              uxtb x1, w0
 123748               0x24401ca0  f81a03a1              stur x1, [x29, #-96]
 123749 
 123750 
 123751     --- bc 329, fp 0, spOff 6 (center)
 123752  329: JmpNZ -32 (297)

TRACE='printir:9,vasm:6' /home/swalk/master/hhvm/hphp/hhvm/hhvm -vEval.Jit=1 center-of-mass.php

swalk-cavium:
See
```
     71 void implCmpEqDbl(IRLS& env, const IRInstruction* inst, ComparisonPred pred) {
     72   auto const s0 = srcLoc(env, inst, 0).reg();
     73   auto const s1 = srcLoc(env, inst, 1).reg();
     74   auto const d  = dstLoc(env, inst, 0).reg();
     75 
     76   auto& v = vmain(env);
     77   auto const cmpsd_res = v.makeReg();
     78   auto const movtdb_res = v.makeReg();
     79 
     80   v << cmpsd{pred, s0, s1, cmpsd_res};
     81   v << movtdb{cmpsd_res, movtdb_res};
     82   v << andbi{1, movtdb_res, d, v.makeReg()};
     83 }
```
runtime/vm/jit/irlower-cmp.cpp

swalk-cavium:
I went back and checked my notes.  There all over the cgInstr lowering.

```
    124 void cgCmpBool(IRLS& env, const IRInstruction* inst) {
    125   auto const s0 = srcLoc(env, inst, 0).reg();
    126   auto const s1 = srcLoc(env, inst, 1).reg();
    127   auto const d  = dstLoc(env, inst, 0).reg();
    128 
    129   auto& v = vmain(env);
    130   auto const sf = v.makeReg();
    131   auto const extended0 = v.makeReg();
    132   auto const extended1 = v.makeReg();
    133 
    134   assert(inst->src(0)->type() <= TBool);
    135   assert(inst->src(1)->type() <= TBool);
    136 
    137   v << movzbq{s0, extended0};
    138   v << movzbq{s1, extended1};
    139   v << subq{extended1, extended0, d, sf};
    140 }
```
runtime/vm/jit/irlower-cmp.cpp

ptomsich:
Isn’t a TBool defined as equivalent to a C++ bool?
If that’s the case it is
	(a) wider than 8bits
	(b )true whenever not-equals zero ( != 0 )

I.e. the movzbq (byte -> quad) extension seems odd (e.g. 0x100 would be true by definition, but after the extension).

Also, the subq looks odd (given this definition), as both 0x1 and 0x2 are “true”, but will not compare to equal using subq.
The appropriate comparison seem to be checking whether both values are 0 (or not equals zero)...

swalk-cavium:
@ptomsich - My intent was to show that the zero extensions were generated all over that
file.  The code I was trying to improve was probably coming from a relop.

I found a picture of the type system over in hhvm/hphp/hhbbc/type-system.h around
line 40.  It looks like the TBool type is defined with the x-macros trick around line 610.

The basic types have their own bits apparently around line 200.
I don't think it's safe to assume false == 0, true != 0.

mxw:
The hhbbc type system and the jit type system are not the same.  Additionally, TBool is a type, not a value.

mxw:
The hhbbc type system and the jit type system are not the same.  Additionally, TBool is a type, not a value.

mxw:
@ptomsich—Our type system represents PHP values, so a runtime value which has type TBool (i.e., has DataType KindOfBoolean) is a PHP boolean.  PHP booleans are always either identically 0 or identically 1, which is why subtracting them is non-degenerate.

The reason we're doing zero-extension is because the result of the CmpBool HHIR instruction is must have type TInt—i.e., a PHP integer, which is by definition an `int64_t`.  That said, I don't see any reason why we zero-extend both operands and then do a `subq{}`.  We could just as easily do a `subb{}` and then zero-extend the result, which is probably better.  I might put up a diff myself that does that.

cmuellner:
When BoolT required values to be either 0 (false) or 1 (true),
then the result ("d") needs to be of this range, right?

But what if:

```
s0 = 0 (false)
s1 = 1 (true)
d = 0-1 = -1 = 0xff (for subb) or 0xffffffffffffffff (for subq).
```

Isn't that violating the spec?

ptomsich:
@mxw — If I understand this correctly, we are looking for an implementation of
	CmpBool(uint1_t a, uint1_t b) -> int64_t		[with uint1_t being a 1-bit integer (i.e. 0 or 1)]
defined as
	result := (a ^ b)

So the optimal expansion for AArch64 seems to be a single eor-instruction.

mxw:
@cmuellner—Er, yes; I meant sign-extend.  Vasm doesn't have any arch-independent sign-extension instructions, so I guess that could be one reason we did it this way.  (Also, maybe two zero-extends are cheaper than a sign-extend on x64, since the zero-extends are just `mov %r, %r`; not sure.)

@ptomsich—So, the `CmpBool` HHIR instruction is actually performing the work of the PHP spaceship operator `<=>`, which returns -1, 0, or 1 depending on which operand compares larger (or if they're equal).  So subtraction is actually what we want here.

cmuellner:
@mxw still `sub` is incorrect:

If the spaceship operator ('<=>') returns: 
Return 0 if values on either side are equal
Return 1 if value on the left is greater
Return -1 if the value on the right is greater

Then what if:
`s0=4` and `s1=2` => `d=2`
`s0=2` and `s1=4` => `d=-2`

ptomsich:
> On 11 Apr 2017, at 19:24, Max Wang <notifications@github.com> wrote:
> 
> @cmuellner <https://github.com/cmuellner>—Er, yes; I meant sign-extend. Vasm doesn't have any arch-independent sign-extension instructions, so I guess that could be one reason we did it this way. (Also, maybe two zero-extends are cheaper than a sign-extend on x64, since the zero-extends are just mov %r, %r; not sure.)
> 
> @ptomsich <https://github.com/ptomsich>—So, the CmpBool HHIR instruction is actually performing the work of the PHP spaceship operator <=>, which returns -1, 0, or 1 depending on which operand compares larger (or if they're equal). So subtraction is actually what we want here.
> 

Ah, I see. Then it’s best modelled as a pure subtraction (and not a cmp/subs).
I wasn’t aware (from the HHIR spec) that this should map onto { -1, 0, 1 }.

If you submit the change, you might want to update the HHIR spec as well, so nobody repeats my mistake.

Thanks,
Phil.

cmuellner:
Reading the source helps...
There is cgCmpInt and cgCmpBool. The first for integers the second for booleans.

```
cgCmpInt:
  v << cmpq{s1, s0, sf};
  v << setcc{CC_G, sf, tmp1};
  v << movzbq{tmp1, tmp2};
  v << cmovq{CC_L, sf, tmp2, v.cns(-1), d};
```

```
cgCmpBool:
  v << movzbq{s0, extended0};
  v << movzbq{s1, extended1};
  v << subq{extended1, extended0, d, sf};
```

Nevermind.

mxw:
Yeah, our documentation in this area is lacking.  I think we adopted `Cmp` as meaning "spaceship", which I guess makes sense if you're very familiar with recent developments in PHP, but completely opaque otherwise.

jim-saxman:
OSS performance test suite ran fine for us. LGTM.

mxw:
I think we can close out this diff without landing, per the work that @cmuellner is doing around sign- and zero-extension assumptions on ARM.  Does that sound correct to all?

swalk-cavium:
As long as the original issue is resolved, I have no objection.  See center-of-mass.php above.

cmuellner:
@mxw - I would say this is a perfectly valid peephole optimization for arm64 and I would like to see this getting merged.

The vasm sequence is emitted (e.g.) here:

```
cgCmpInt():
  v << cmpq{s1, s0, sf};
  v << setcc{CC_G, sf, tmp1};
  v << movzbq{tmp1, tmp2};
  v << cmovq{CC_L, sf, tmp2, v.cns(-1), d};
```

`setcc{}` is defined as follows:

`struct setcc { ConditionCode cc; VregSF sf; Vreg8 d; };`

In x64 setcc only sets the lower byte register, but on aarch64 it sets the whole register. Therefore a zero-extension is not necessary on aarch64.

@swalk-cavium - Would it be possible to extend this PR to also optimize for other movzbN{}s as well.

swalk-cavium:
@cmuellner - Yes, it's trivial.  Testing started.

swalk-cavium:
@mxw - Hi Max, It may very well be that this change is subsumed by the flurry of changes
from @cmuellner.  Once his sequence of changes land I can retest.  I don't think it makes 
sense to land something that will never fire.

swalk-cavium:
@mxw - Hi Max, I went back and checked this one since #cmuellner's changes have landed.
There are still opportunities for this to fire in center-of-mass.php and quick/all_type_comparison_test.php.

Can you inspect?

cmuellner:
@swalk-cavium, did you test on top of the not yet merged PR #7811?

swalk-cavium:
@cmuellner - Ah, no.  I looked a the git log.  I thought I saw all your changes.  Never mind.  Back to waiting.

mxw:
@swalk-cavium, @cmuellner—#7811 should have landed.  Really sorry for the delay; I was working out of a different office and then out on PTO last week.  I'm back now and will be responsive to updates.

swalk-cavium:
@mxw , @cmuellner - I just tested on a fresh build. This particular pattern is no longer generated, so this pull request isn't needed anymore.

Issue #7709, opened 2017-03-03
hphp_set_property() Fatal error if php7 support is enabled


bartaakos:
I'm working on a mixed php/hack codebase where I'd like to enable php7 mode - 7.1 actually :) - for the php5 part moving towards php7/hack. Currently I face a reflection issue described below.

The code works both with "<?php" and "<?hh" open tags if php7 mode is disabled but fails if enabled.

### HHVM Version

HipHop VM 3.18.0 (rel)
Compiler: tags/HHVM-3.18.0-0-g9b285191feb2bb1558bb6682da135263bd2a9e60
Repo schema: cc987335f4691fa6ed98336c2378d8e71a751b22

### Standalone code, or other way to reproduce the problem

```php
<?php

class Dto
{
    public $prop;
}

class DtoSerializer
{
    public function toDto($destinationObject, $data)
    {
        $destinationReflection = new \ReflectionObject($destinationObject);
        $destinationProperties = $destinationReflection->getProperties();
        foreach ($destinationProperties as $property) {
            if (!$property->isPublic() || $property->isStatic()) {
                continue;
            }

            $name = $property->getName();
            $value = isset($data[$name]) ? $data[$name] : null;
            $property->setValue($destinationObject, $value);
        }

        return $destinationObject;
    }
}

function main() {
    $serializer = new DtoSerializer();

    $dto = $serializer->toDto(new Dto(), ['prop' => 'Some data']);

    var_dump($dto);
}

main();
```

https://3v4l.org/iJBSg

### Expected result

```
object(Dto)#2 (1) {
  ["prop"]=>
  string(9) "some data"
}
```

### Actual result

```
Fatal error: Uncaught TypeError: Argument 2 passed to hphp_set_property() must be an instance of string, null given in /in/iJBSg:21
Stack trace:
#0 /in/iJBSg(21): ReflectionProperty->setValue()
#1 /in/iJBSg(31): DtoSerializer->toDto()
#2 /in/iJBSg(36): main()
#3 {main}

Process exited with code 255.
```

mofarrell:
It looks like https://github.com/facebook/hhvm/blob/master/hphp/runtime/ext/reflection/ext_reflection-internals-functions.php#L90 needs to be `?string` type.  Want to submit a pull request?

bartaakos:
Sure :)

I've just realised that the same applies to getProperty() so made the same change there as well.

I added a test case - does it look ok? I mean could we state somehow that php7 mode should be turned on?

[...]

PR #7710, opened 2017-03-03
Reflection get/set property's $cls should be nullable

bartaakos:
Related issue: https://github.com/facebook/hhvm/issues/7709


mofarrell:
Could you move the test to https://github.com/facebook/hhvm/tree/master/hphp/test/slow/reflection

All the tests under the hphp/test/zend directory were imported from zends test suite.

mofarrell:
It seems like this is going to be slightly more involved.  By making them nullable, the C++ types used in the native definition have to change.  Now instead of having the C++ type String, it most likely needs to use a Variant, and do a check for null and other variant types.  See the [type correspondence table](https://github.com/facebook/hhvm/wiki/Extension-API#php-and-c-extensions).

mofarrell:
These are internal functions, so the typehints are fine being used to trigger coercion now that we don't enforce strict type hints in systemlib.

PR #7712, opened 2017-03-05
Fix handling of b" in token_get_all()

nikic:
In the best PHP tradition, "single-char" tokens can actually have
more than one character... In particular b" is considered a
single-char token and token_get_all() returns string "b\"" for it.
This change brings HHVM in line with PHP behavior. Previously
HHVM returned just "\"" in this case, dropping the "b" entirely.

### DIFF
--- a/hphp/parser/scanner.h
+++ b/hphp/parser/scanner.h
@@ -263,7 +263,7 @@ public:
     incLoc(rawText, rawLeng, type);
   }
   void stepPos(const char *rawText, int rawLeng, int type = -1) {
-    if (shortTags()) {
+    if (full()) {
       m_token->setText(rawText, rawLeng);
     }
     incLoc(rawText, rawLeng, type);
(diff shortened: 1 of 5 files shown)


Orvid:
There still seems to be conflicts with master, could you rebase this?

hphp/parser/scanner.h
```diff
@@ -263,7 +263,7 @@ struct Scanner {
     incLoc(rawText, rawLeng, type);
   }
   void stepPos(const char *rawText, int rawLeng, int type = -1) {
-    if (shortTags()) {
+    if (full()) {
```
nikic: I was very confused about the purpose of `stepPos()` until I realized that the reference to short tags was a bug introduced in https://github.com/facebook/hhvm/commit/7119bff22c524cfe66dfbd8ba5a9d1781381980f#diff-868580dfadd010b428015b3fc416a1acL247.

Issue #7716, opened 2017-03-06
Memory leak in pgsql driver


Norbyte:
### HHVM Version
HipHop VM 3.18.1 (rel)
Compiler: tags/HHVM-3.18.1-0-g4f792a827d9b6fcc4bcd7940a1b92280f3fdedc3
Repo schema: b9aebb86bbda024eeeed6f608c9a49076563f9c7

### Standalone code, or other way to reproduce the problem
```php
<?php
$pdo = new PDO('pgsql:host=(host);dbname=(dbname)', '(user)', '(password)');
$stmt = $pdo->prepare('select 1');
for ($i = 0; $i < 100; $i++)
{
    $stmt->execute();
}
```

During testing with `ab -c 4 -n 1000000 http://<url>`, HHVM started leaking memory at a rate of about ~40 MB/sec. (The OOM killer killed it after a few minutes).

This issue does not occur if another driver (eg. mysql, sqlite) is selected. Directly calling `$pdo->exec()`, instead of preparing a statement with `$pdo->prepare()` also seems to work without leaks. The leak only occurs when executing a pgsql prepared statement,

photodude:
@simonwelsh any thoughts on this?

Norbyte:
Not sure if it helps, but I created a core dump of a running HHVM process after triggering this leak, and almost 80% of the dump is this exact sequence repeating endlessly (with some noise inbetween):

```
00 00 00 00 01 00 00 00 00 45 4C 45 43 54 20 31   .........ELECT 1
00 45 00 00 00 00 00 00 00 00 00 00 00 00 00 00   .E..............
00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00   ................
00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00   ................
00 00 00 00 00 00 00 00 00 00 00 00 F0 63 1C 03   ............đc..
00 00 00 00 00 00 00 00 00 00 00 00 70 63 1C 03   ............pc..
00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00   ................
00 00 00 00 00 00 00 00 06 00 00 00 00 00 00 00   ................
00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00   ................
00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00   ................
00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00   ................
[...]

PR #7718, opened 2017-03-07
Fix pgsql prepared statement memory leak

Norbyte:
The pgsql Result struct mishandled PGresult resources when the
move assignment operator was used, resulting in PGresult leakage.

Fixes Issue #7716.

### DIFF
--- a/hphp/runtime/ext/pgsql/pq.h
+++ b/hphp/runtime/ext/pgsql/pq.h
@@ -39,9 +39,11 @@ struct Result {
   }
 
   Result& operator=(Result&& other) {
-    clear();
-    m_res = other.m_res;
-    other.m_res = nullptr;
+    if (this != &other) {
+      clear();
+      m_res = other.m_res;
+      other.m_res = nullptr;
+    }
     return *this;
   }
 
--- a/hphp/runtime/ext/pgsql/pq.h
+++ b/hphp/runtime/ext/pgsql/pq.h
@@ -39,6 +39,7 @@ struct Result {
   }
 
   Result& operator=(Result&& other) {
+    clear();
     m_res = other.m_res;
     other.m_res = nullptr;
     return *this;


simonwelsh:
Not being that good at this sort of C++, what'll happen if `other == this`?

Norbyte:
Hmm, thats a tricky question. The only relevant answer I found was this SO question: http://stackoverflow.com/questions/13127455/what-does-the-standard-library-guarantee-about-self-move-assignment

... which quotes the standard like so:
> If a function argument binds to an rvalue reference parameter, the implementation may assume that this parameter is a unique reference to this argument.

and later on:
> The latest working draft, N4618 has been modified to clearly state that in the MoveAssignable requirements the expression:
>> `t = rv`
> (where `rv` is an rvalue), `t` need only be the equivalent value of `rv` prior to the assignment if `t` and `rv` do not reference the same object. And regardless, `rv`'s state is unspecified after the assignment.
> There is an additional note for further clarification:
> `rv` must still meet the requirements of the library component that is using it, whether or not `t` and `rv` refer to the same object.

Norbyte:
After testing this for a while, I could only trigger self-move assignment calls when I explicitly used `x = std::move(x)`; I couldn't make it self-assign with normal assignment operator use at all.

Orvid:
In general, standard library implementations work correctly with self-assignment, so we should do the same.

Norbyte:
They work correctly with self-assignment, but generally don't with self-move assignment, which is what we're discussing here.
For example, this is the std::basic_string move assignment of libstdc++ as of 4.8.5:

```C++
      /**
       *  @brief  Move construct string.
       *  @param  __str  Source string.
       *
       *  The newly-created string contains the exact contents of @a __str.
       *  @a __str is a valid, but unspecified string.
       **/
      basic_string(basic_string&& __str) noexcept
      : _M_dataplus(__str._M_dataplus)
      {
#if _GLIBCXX_FULLY_DYNAMIC_STRING == 0
        __str._M_data(_S_empty_rep()._M_refdata());
#else
        __str._M_data(_S_construct(size_type(), _CharT(), get_allocator()));
#endif
      }
```

(Note how it overwrites `__str` with an empty string even if `this == &__str`; also the `__str is a valid, but unspecified string.` postcondition comment. The same applies for many other std containers as well.)

Anyhow, I don't think this is really worth arguing about, so I added a self-assignment check.

PR #7729, opened 2017-03-09
[AArch64] Disable retranslateAll for ARM architectures

jim-saxman:
The retranslateAll feature is currently broken on ARM
architectures. The OSS Performance Test Suite benchmarks segfault
after the retranslateAll code runs. Fixing it will require some
more substantial work, so temporarily disable it for now.

### DIFF
--- a/hphp/runtime/base/runtime-option.cpp
+++ b/hphp/runtime/base/runtime-option.cpp
@@ -30,6 +30,7 @@
 #include <folly/portability/SysTime.h>
 #include <folly/portability/Unistd.h>
 
+#include "hphp/util/arch.h"
 #include "hphp/util/atomic-vector.h"
 #include "hphp/util/build-info.h"
 #include "hphp/util/hdf.h"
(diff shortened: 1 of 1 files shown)


jim-saxman:
@dave-estes thanks for the quick review. @mxw @swalk-cavium please review and we can discuss what needs to be done to enable retranslateAll for ARM.

Orvid:
Is there a specific place where it's hitting an issue? We know we've had a few issues with retranslateAll internally, so it's possible that you're hitting an issue we just haven't encountered yet.

cc. @paulbiss for context on the known issues that we've hit.

jim-saxman:
It is causing a SEGV in __builtin___clear_cache() during vasm-arm-cpp’s ~Vgen.
This function is only used for ARM.

We need to port some of @pualbiss’s changes to x64 over to ARM from his https://github.com/facebook/hhvm/commit/b01abd12d38ae754025c22d1662b73edc3a9145f , specifically the toDestAddr().

Our hope is that we can do this along with @dave-estes relocation work for ARM.

-Jim

swalk-cavium:
@jim-saxman - The temporary change looks fine to me.  I hadn't seen this in regular testing so I forgot you mentioned something.

Issue #7434, opened 2016-10-20
Missing PDO::MYSQL_ATTR_MULTI_STATEMENTS constant


caioiglesias:
HipHop VM 3.16.0-dev (rel)

step to reproduce: `print_r(PDO::MYSQL_ATTR_MULTI_STATEMENTS);`

expected: `1013`

actual: `\nFatal error: Couldn't find constant PDO::MYSQL_ATTR_MULTI_STATEMENTS in /.../Mysql.php on line 342`

This error prevents Magento 2.1.2 from completing the installation.

Orvid:
This should be fairly trivial to implement, https://github.com/php/php-src/blob/6053987bc27e8dede37f437193a5cad448f99bce/ext/pdo_mysql/mysql_driver.c#L568

It's just a flag being passed to MySQL itself.

aftabnaveed:
Same error on HHVM 3.18

caioiglesias:
well @aftabnaveed ... it hasn't been implemented yet.

halid:
I'm getting the same error.

version: HHVM 3.18

racardoso:
I made a PR to fix this issue https://github.com/facebook/hhvm/pull/7737

jesseschalken:
This was fixed in https://github.com/facebook/hhvm/commit/b07bca0f325155a435f05329b2660711557af065

https://3v4l.org/6VS1h (not sure why Zend versions don't have that constant defined on 3v4l.org)

PR #7737, opened 2017-03-20
Include missing constant on ext/pdo_mysql.

racardoso:
Summary:

This commit implements MYSQL_ATTR_MULTI_STATEMENT constant.
Solves issue https://github.com/facebook/hhvm/issues/7434


hphp/runtime/ext/pdo_mysql/pdo_mysql.h
```diff
@@ -21,7 +21,7 @@
 #include "hphp/runtime/ext/pdo/pdo_driver.h"
 
 namespace HPHP {
-///////////////////////////////////////////////////////////////////////////////
+/////////////////////////////////////////////////////////////////////////////////
```
Orvid: Drop this change.

hphp/runtime/ext/pdo_mysql/pdo_mysql.cpp
```diff
@@ -203,11 +203,16 @@ bool PDOMySqlConnection::create(const Array& options) {
 #ifdef CLIENT_MULTI_RESULTS
     |CLIENT_MULTI_RESULTS
 #endif
-#ifdef CLIENT_MULTI_STATEMENTS
-    |CLIENT_MULTI_STATEMENTS
-#endif
     ;
 
+  #ifdef CLIENT_MULTI_STATEMENTS
+    if (!options.empty()) {
```
Orvid: This logic is the inverse of what is intended.
The way you've written it here, it will default to no multi-statement, but enable multi-statements if you've provided any options.

This needs to be `if (options.empty())` instead.

hphp/runtime/ext/pdo_mysql/pdo_mysql.h
racardoso: dropped.

hphp/runtime/ext/pdo_mysql/pdo_mysql.cpp
racardoso: Done.

Orvid:
We just merged #7628 so you'll need to rebase before we can import it. The changes look good to me though.

racardoso:
@Orvid rebase done.

PR #7738, opened 2017-03-20
stream_socket_get_name function add ":" for unix domain socket

s2x:


Orvid:
We actually have a fix that landed ~2 days ago, but I had our Github sync disabled over the weekend as I had been trying to debug an issue. I've re-enabled the sync so it should land properly in ~45 minutes. 

Sorry about the duplicated effort :(

Orvid:
Sync has caught up, and https://github.com/facebook/hhvm/commit/1787575c7a1554808336413abc9c9468f220ba4d is the commit that @mofarrell fixed it in.

PR #7593, opened 2017-01-12
Aarch64 - loadb/movxbl  -> loadzbl peephole

swalk-cavium:
On Aarch64 the loadb/movxbl sequence would generate ldrsb/uxtb.
This occurred around the  Vinstr cmpbm instructions.  It is equivalent to just use 
the ldrb instruction and save one instruction per instance.

Before
====
    14854    0: FPushFuncD 0 "ut_main"
    14855     (03) EndGuards
    14856     (07) t3:Func = LdFuncCached<ut_main> -> B2<Catch>
    14857         Main:   
    14858               0x296008b0  d28217e0              movz x0, #0x10bf 
    14859               0x296008b4  38e06b60              ldrsb w0, [x27, x0]     // <<--+
    14860               0x296008b8  53001c00              uxtb w0, w0             // <<--+
    14861               0x296008bc  d29fffe1              movz x1, #0xffff 
    14862               0x296008c0  f2a05fe1              movk x1, #0x2ff, lsl #16 
    14863               0x296008c4  38616b61              ldrb w1, [x27, x1] 
    14864               0x296008c8  6b00003f              cmp w1, w0

After
=====
     320    0: FPushFuncD 0 "ut_main"
     321     (03) EndGuards
     322     (07) t3:Func = LdFuncCached<ut_main> -> B2<Catch>
     323         Main:
     324               0x1d40082c  d28217e0              movz x0, #0x10bf
     325               0x1d400830  38606b60              ldrb w0, [x27, x0]  //<<---
     326               0x1d400834  d29fffe1              movz x1, #0xffff
     327               0x1d400838  f2a05fe1              movk x1, #0x2ff, lsl #16
     328               0x1d40083c  38616b61              ldrb w1, [x27, x1]
     329               0x1d400840  6b00003f              cmp w1, w0

This was tested by running the regression tests with six different option sets.
No new issues were found.

PR #7744, opened 2017-03-23
jit: arm: Use unsigned loads instead of signed loads.

cmuellner:
It seems like the arm64 tree prefers to load bytes
in signed form, which is neither a requirement nor
a necessity. And additionally it makes things slower
on certain ARM64 cores.

This patch changes the emitted loads from signed to unsigned.
No regressions have been observed in the tests.

### DIFF
--- a/hphp/runtime/vm/jit/vasm-arm.cpp
+++ b/hphp/runtime/vm/jit/vasm-arm.cpp
@@ -271,19 +271,19 @@ struct Vgen {
   void emit(const lea& i);
   void emit(const leap& i) { a->Mov(X(i.d), i.s.r.disp); }
   void emit(const lead& i) { a->Mov(X(i.d), i.s.get()); }
-  void emit(const loadb& i) { a->Ldrsb(W(i.d), M(i.s)); }
+  void emit(const loadb& i) { a->Ldrb(W(i.d), M(i.s)); }
   void emit(const loadl& i) { a->Ldr(W(i.d), M(i.s)); }
   void emit(const loadsd& i) { a->Ldr(D(i.d), M(i.s)); }
-  void emit(const loadtqb& i) { a->Ldrsb(W(i.d), M(i.s)); }
+  void emit(const loadtqb& i) { a->Ldrb(W(i.d), M(i.s)); }
   void emit(const loadtql& i) { a->Ldr(W(i.d), M(i.s)); }
   void emit(const loadups& i);
-  void emit(const loadw& i) { a->Ldrsh(W(i.d), M(i.s)); }
+  void emit(const loadw& i) { a->Ldrh(W(i.d), M(i.s)); }
   void emit(const loadzbl& i) { a->Ldrb(W(i.d), M(i.s)); }
   void emit(const loadzbq& i) { a->Ldrb(W(i.d), M(i.s)); }
   void emit(const loadzlq& i) { a->Ldr(W(i.d), M(i.s)); }
   void emit(const movb& i) { a->Mov(W(i.d), W(i.s)); }
   void emit(const movl& i) { a->Mov(W(i.d), W(i.s)); }
-  void emit(const movtqb& i) { a->Sxtb(W(i.d), W(i.s)); }
+  void emit(const movtqb& i) { a->Uxtb(W(i.d), W(i.s)); }
   void emit(const movtql& i) { a->Mov(W(i.d), W(i.s)); }
   void emit(const movzbl& i) { a->Uxtb(W(i.d), W(i.s)); }
   void emit(const movzbw& i) { a->Uxtb(W(i.d), W(i.s)); }


mxw:
This change makes things a lot more consistent.  @dave-estes, @ji-saxman, @swalk-cavium, do any of you see any reason not to land this?  I'll do so this afternoon if I don't hear any concerns.

swalk-cavium:
@mxw - As long as the regression tests continue to pass, I have no concerns.
Although, it might make one of the peepholes I added dead code.  See #7593

dave-estes:
I think we still need to sign extend these in order to preserve the sign of narrow loads before doing any compares or arithmetic operations on them. If we don't sign extend a `loadb` of a signed char of 0xff, then even our narrowest 32-bit operations will see it as a positive 255 instead of -1.

cmuellner:
If you do a loadb you will get a Vreg8.
When you want a signed char, then you have it.
When you want a int64_t, then you either need to
sign extend it with `movsx` or directly use `movsbq`.

However, I don't see code in HHVM which does that.

dave-estes:
I'm more worried about the sign flag following arithmetic operations. Consider something like `loadb x, loadb y, cmpb x, y`. For` x = -1` and `y = 0`, without sign extension the generated ARM code would load registers with 0xff and 0x00 and then do a 32-bit compare. The sign flag will not be correctly set.

swalk-cavium:
Wouldn't that be handled in the cmpb lowering?


   1497 void lower(const VLS& e, cmpb& i, Vlabel b, size_t z) {

   1498   lower_impl(e.unit, b, z, [&] (Vout& v) {

   1499     auto s0 = v.makeReg();

   1500     auto s1 = v.makeReg();

   1501     v << movzbl{i.s0, s0}; //<<---

   1502     v << movzbl{i.s1, s1}; //<<---

   1503     v << cmpl{s0, s1, i.sf};

   1504   });

   1505 }

runtime/vm/jit/vasm-arm.cpp


Steve Walk

dave-estes:
Nah, because those `movzbl`s don't sign extend. AND, we'll likely be getting rid of them based on chatter in the mail list.

cmuellner:
You cannot load a signed char with loadb.
You need a loadsb otherwise you make a cast.

If you mean something like this:

```
bool myfun_s(int8_t *ptr1, int8_t *ptr2) {
  return *ptr1 == *ptr2;
}

bool myfun_u(uint8_t *ptr1, uint8_t *ptr2) {
  return (*ptr1 - *ptr2) == 0;//or *ptr1 == *ptr2;
}
```

Then everything is fine (that's what gcc generates):

```
myfun_s(signed char*, signed char*):
        ldrsb   w2, [x0]
        ldrsb   w0, [x1]
        cmp     w2, w0
        cset    w0, eq
        ret
myfun_u(unsigned char*, unsigned char*):
        ldrb    w2, [x0]
        ldrb    w0, [x1]
        cmp     w2, w0
        cset    w0, eq
        ret

```

cmuellner:
BTW: when you want static protection from exchaning signed and unsigned values, we could propose the introduction of VregSN classes (e.g. VregS8). But that would be a feature which is independent of this PR.

swalk-cavium:
Just for grins - x86 generates this code for your previous

code fragment.


  8 myfun_s:

  9 .LFB0:

 10         .cfi_startproc

 11         movb    (%rsi), %al

 12         cmpb    %al, (%rdi)

 13         sete    %al

 14         ret


 28 myfun_u:

 29 .LFB1:

 30         .cfi_startproc

 31         movb    (%rsi), %al

 32         cmpb    %al, (%rdi)

 33         sete    %al

 34         ret



dave-estes:
Do we have a `loadsb` vasm instruction? I think the only way to load a signed char is with a `loadb`.

cmuellner:
If we need one, then we have to introduce one.
Are you having a specific use-case in mind (e.g. a specific  IR)?

cmuellner:
Note, that vasm is not a complete ISA, but a subset, which is just enough for the current IR lowering (to vasm). Besides loadsb also others are missing (e.g. addb).

mxw:
@dave-estes—I'm a little confused about what you were saying w.r.t. `cmpb{}` and sign flags.  Right now, we lower a `cmpb{}` to the following:

```
v << movzbl{i.s0, s0};
v << movzbl{i.s1, s1};
v << cmpl{s0, s1, i.sf};
```

We emit an unsigned byte-extend for `movzbl{}`, so isn't the sign flag already wrong today?

@cmuellner, @dave-estes—Let me try to summarize where I think the conversation has taken us so far.  If I understand correctly, we are going to need to make following changes to the system:

1.  Whenever we define a Vreg8 (or Vreg16) on ARM, guarantee that the unspecified lower 32 bits will be zeros.
2.  For b- or l-form vasm instructions that want the source operands to be zero-extended, we can just emit operations over W-form registers directly, without lowering.
3.  For b- or l-form vasm instructions that want the source operands to be sign-extended, we will still need to explicitly sign-extend them when we lower or emit that instruction.

Is that your take on things as well?  I suppose we could also flip those (sign-extend Vreg8 and Vreg16 by default, then require instructions which use such registers and want zero-extend semantics to do the zero-extending themselves); I don't know which would be better.

Also, thank you, @cmuellner, for driving this effort to clarify our use of these truncation and zero-extension instructions.  Vasm didn't care about width-compatibility until partway into the development of the ARM port, so it's never been very clear how carefully the port adheres to vasm's requirements in this area.  I really appreciate you pushing on this, and I think it will have a good payoff in making our lowered code sequences tighter.

cmuellner:
@mxw - That's essentially what I'm working on here. I currently prefer the zero-extended-by-default approach...

What I would prefer even more are different types for signed/unsigned Vregs (Vregu8, Vregs8). And I'd like to see the same for the Immed class (e.g. Immeds8, Immedu8, etc.). The more expressive vasm's type system is, the easier it gets to create efficient and correct code.

mxw:
Zero-extending by default feels more natural to me, as well.  You said above that it should also be faster, which makes it the obvious choice for the default behavior (since it has to be one or the other).

I'm not sure whether baking signedness into the Vreg type system would be a good idea—mostly because I don't have a great intuition for, in the current landscape of irlower, whether we always interpret a given VregN with the same signedness.  If we aren't consistent w.r.t. how we interpret the signedness of the type and data registers of a TypedValue, in particular, adding this to the type system could be pretty invasive.  Right now, we know any value with type TGen consists of a Vreg64 data component and a Vreg8 type component (unless one of them is known statically).  Things get hairy if we don't have that mapping on hand (and the JIT type system definitely shouldn't care about signedness).

(In the abstract, though, this idea makes sense.)

In the meantime, you could write analysis pass which determines if all of the uses of each Vreg are signed, and then you could rewrite the def of that Vreg to do a signed load or a signed extension or what have you.  This isn't as nice as type enforcement, but it'll get you a similar result, I expect.

edwinsmith:
I believe signed-ness (and how one interprets bits in general, whether signed or unsigned, or as a pointer, or as a float/double, etc) should be associated with operations and not register types. Part of my opinion comes from LLVM, which models integer types only with # of bits (i8, i32, etc) and not by assigning meaning to what the bits really mean.  And part of it comes from the way all CPUs work, where register class only implies data-width and the set of allowed operations.

This does push the sign-vs-zero extension problem upstream, to the point where both instruction and register-widths are chosen, but I think that's the right place for it.  It could mean we will need arch-specific hhir->vasm generation in a handful of places, if vasm-to-vasm lowering is too difficult to manage. that's okay, provided it's relatively isolated.  (you see the same pattern with LLVM and imho its an okay tradeoff).

Lets try to resist getting too attached to the idea that VASM is arch-independent. it doesn't have to be, even though it's desirable to make it mostly generic. Even with arch-specific instructions, we can apply arch-independent optimization passes such as register allocation and copy propagation.

Having Vreg8/16 on an arch with no such register classes inevitably raises the question of whether to generate instructions that operation on those register classes, then lower, or not generate them at all; but only in some places.  coming up with a strong convention (and necessarily arch-specific invariant) around how those classes are to be used, is 100% legit and the right thing to do.

swalk-cavium:
I'm not familiar with any architecture that encodes the signed-ness in the register name.
It's up to the opcode how to interpret the bits.   I agree with @edwinsmith above.

cmuellner:
@edwinsmith - thank's for your input. My desire for signed-ness comes from typed programming languages (e.g. C or C++), where values always are either signed or unsigned. I had a look at LLVM and have seen that they had a type system with explicit types in early versions (e.g. in llvm 1.5 they had e.g. ubyte and sbyte [1]). GCC still provides the signed-ness information to the backend. So I guess my desire is not that odd. Modern approaches like webassembly seem to follow the LLVM model.

If HHVM follows the LLVM model, then I am perfectly fine with that.

Especially since both models are compatible with this PR (ping for @mxw).

[1] http://releases.llvm.org/1.5/docs/LangRef.html

PS: here's an example how GCC and LLVM create different aarch64 assembly based on the knowledge they have about the signed-ness:

```
int8_t mysub1(int8_t v) {
  return v - 123;
}

uint8_t mysub2(uint8_t v) {
  return v - 123;
}
```

GCC 5.4

```
mysub1(signed char):
        sxtb    w0, w0
        sub     w0, w0, #123
        ret

mysub2(unsigned char):
        uxtb    w0, w0
        sub     w0, w0, #123
        ret
```

LLVM 4.0.0

```
mysub1(signed char):                             // @mysub1(signed char)
        and     w8, w0, #0xff
        add     w0, w8, #133            // =133
        ret

mysub2(unsigned char):                             // @mysub2(unsigned char)
        and     w8, w0, #0xff
        add     w0, w8, #133            // =133
        ret
```

dave-estes:
BTW, we're failing a runtime/test related to sign-extending. Maybe you can take a look, @cmuellner. You can cherry pick #7793 and then re-cmake with `-DRUNTIME_TEST_BIN=on`. Then to run the test, you cd down into hphp/runtime/test and execute the GTest.

    $ cd hphp/runtime/test
    $ ./hphp_runtime_test

cmuellner:
@dave-estes - Thanks for the hint. I've already noticed your PR and put it on my to-test-list. Once your PR is merged, I want to integrate the results into the nightly-build output.

mxw:
@cmuellner—I chatted with @edwinsmith offline about this, and I think that having type information in vasm is reasonable, but it should be orthogonal to the VregN register category typesystem.  This corresponds with how machine code instruction operands are constrained—a given instruction requires a register of a particular width category, but the instruction itself is what imposes the signedness.

Again, I think the best way to incorporate signedness on platforms that care is to add a separate annotation pass, where you determine the sign imposed at a def, the sign implicitly required at the use sites, and insert fixup instructions as necessary.  The width-safety of vasm is guaranteed by a width-checking pass.  This is the same idea—you can implement a sign-checking pass, and then make sure that your 8- and 16-bit registers are appropriately zero- or sign-extended when you use them.

cmuellner:
@mxw - we have no use-case for sign-extended values in registers in vasm-arm.cpp. Therefore this patch replaces sign-extensions by zero-extensions. As a side-effect that makes the resulting code faster on some aarch64 CPUs.

As I said before, I have no problem with not having signed-ness information in the VregNs.

cmuellner:
I will merge this PR with #7811.

mxw:
Sounds good—closing out here.

PR #7765, opened 2017-04-04
Allow array_map callback to be nullable

RyanGordon:
According to http://php.net/array_map the callback argument in array_map is nullable. (Example `#4`).

`An interesting use of this function is to construct an array of arrays, which can be easily performed by using NULL as the name of the callback function`

Currently `array_map(null, $arr1, $arr2)` produces a type error: 

```
Invalid argument (Typing[4110])
File "stdlib/builtins_array.hhi", line 117, characters 10-18:
This is a function
File "..."
It is incompatible with a nullable type
```

### DIFF
--- a/hphp/hack/hhi/stdlib/builtins_array.hhi
+++ b/hphp/hack/hhi/stdlib/builtins_array.hhi
@@ -114,7 +114,7 @@ function array_keys<Tk, Tv>(
  * Container<X>          -> R = array<arraykey, Tr>
  * X (unknown type)      -> R = Y (other unknown type)
  */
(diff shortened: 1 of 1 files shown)


mofarrell:
This is invalid syntax.  You can't have a `?` without a type specified.  `mixed` is a type that allows Variant value.  https://github.com/facebook/hhvm/wiki/Extension-API#php-and-c-extensions

simonwelsh:
I suspect this is going to be more difficult than it appears as `array_map` is one of the functions that gets special handling inside the type checker (as the signature of the function changes with the number of arguments)

PR #7768, opened 2017-04-06
Providing lineno information and enabling dump of editable token output

qcmaude:
👋 Hello! 

I've been working on building out a simple tool to perform some static analysis on Hacklang code. 

In order to do this, I've been working off of my own fork of HHVM where I have added an option to dump editable token output (over the minimal token output currently available with the `--dump-full-fidelity-parse`) and tweaked the lexer to add line number output on token generation. 

I reached out to the Facebook HHVM General group about this solution [back in February](https://www.facebook.com/groups/hhvm.general/permalink/780745482078847/) but didn't get around to fully implementing it until just recently (after taking the time to decipher OCaml).

Now this may be useful only to me, but I figured I'd try to get this merged upstream so as to prevent having to patch my own fork indefinitely. Specifically, I _could_ be convinced to merge only the new argument `--dump-full-fidelity-editable-parse` and forgo the changes to the lexer as looking at the code, the omission of line numbers as part of the minimal token records appeared intentional. I hope these changes are superficial enough as to be good additions to the `hh_client`!

Anywho ... looking forward to the feedback! ✨


mofarrell:
Feedback from @ericlippert:
Let me first say what this diff does: it:
* adds an option to dump the editable parse tree as JSON
* adds line number information to the editable parse tree
I have four questions about this diff: is it *unnecessary*? is it *misleading*? is it *incorrect*? and is it *unperformant*?

First: Adding this option is unnecessary. The minimal syntax tree -- which gives the kind, children, and width in characters of each node in the parse tree -- plus the original text is sufficient information to determine the derivation of every character in the file. Obviously there is code written in Ocaml that translates a minimal parse tree into an editable parse tree. So there is no need for code to dump out the editable parse tree; you could write that algorithm in whatever code is consuming the editable parse tree.

That said, I am not in principle opposed to having this option added to the parsing utility. I just want to be clear that this is redundant.

Second: adding line number information to the editable parse tree is *misleading*. The point of the editable parse tree is that *it is easy to reorganize it into a new parse tree* -- hence the name. But the semantics of "line" are "this is the line number of the token in the original parse tree", not in the parse tree that the possibly-rearranged editable tree represents.

That said, this is frequently useful information! When rewriting a parse tree we often wish to know -- for debugging purposes, for instance -- where the original node came from.

I am not in principle opposed to having this information in the editable tree, but I do worry that it will be misunderstood.

Third: the computation of the line number appears at first glance to be incorrect. For example, where is the line number incremented if a /* */ comment spans multiple lines?

I am opposed to having an incorrect computation in the lexer. This should be verified before this diff is checked in.

Fourth: the proposed change increases the size of the lexer, and we allocate a *lot* of lexers in this purely-functional-style algorithm. (I have been meaning for a long time to be more clever about the number of lexers allocated; there is low hanging fruit here.)

I do not have any performance numbers here, but I do worry slightly that increasing the size of the lexer, in a world where so many lexers are created, might cause a perf regression.

qcmaude:
@mofarrell, @ericlippert thanks for the feedback!

I've given this some more thought (and ran my ideas past a few of my coworkers) and wanted to follow-up with your concerns. 

1. I don't think this option is quite as unnecessary as it might seem on the surface. Why? Having access to the editable parse tree as optional output from `hh_client` enables developers to easily create more productivity tools around the Hack environment. Sure, matching up the minimal parse tree output and the source code is a solved problem (and a pretty trivial task), but it would be much more convenient to have this information available from the source. The `hh_client` code _already_ does such a good job of creating this editable parse tree, why not surface it for consumption by the developer?

2. I can see the line number being misleading if the ultimate goal is to have as "lite/flexible" of an output as possible. However, I do believe the line number's role is clear given that in order to obtain the output, the user must specify an entry file or point. I think in regards to utility for consumers of this data, I would prefer to provide more information rather than too little. 

3. Great catch on that oversight! Unfortunately we're not a crew that uses multiline comments in our codebase so that example slipped through the cracks when verifying my changes.

4. This change _does_ increase the size of the lexer but not by much. I don't have any performance stats handy but the version of HHVM we've been running in production for the past few weeks has included these changes. One of my coworkers has been able to quickly put together some static analysis tools with the output with no apparent issues in performance. I can pull up some numbers if you'd like to further ensure that these changes won't be detrimental to the overall performance of the `hh_client`.

PR #7767, opened 2017-04-05
Optimize code generation of CheckRefs.

cmuellner:
* Remove not needed truncation
 * Remove masking, when not needed

This brought a performance boost of 6.7% in
OSS-performance's mediawiki benchmark (comparing the average
transaction rate of 4 subsequent runs each).

No regressions, when running the "all" test set on ARM64.

Example from t-test (vm-perf):

Before:
    (14) CheckRefs<0,255,192> t2:Func, 64 -> B10<Unlikely>
        Main:
              0x24c0c9d0  f9402400              ldr x0, [x0, #72]
              //movtqb w0 <- w0
              0x24c0c9d4  53001c00              uxtb w0, w0
              //andbi w0 <- w0, 255
              0x24c0c9d8  12800001              movn w1, #0x0
              0x24c0c9dc  0a010000              and w0, w0, w1
              //cmpi w0, 192
              0x24c0c9e0  53001c00              uxtb w0, w0
              0x24c0c9e4  128007e1              movn w1, #0x3f
              0x24c0c9e8  6b01001f              cmp w0, w1

              0x24c0c9ec  540000a0              b.eq #+0x14
              0x24c0c9f0  58000052              ldr x18, pc+8 (addr 0x24c0c9f8)
              0x24c0c9f4  d61f0240              br x18
              0x24c0c9f8  28c03948              ldp w8, w14, [x10]
              0x24c0c9fc  00000000              unallocated
(Unallocated)
After:
    (14) CheckRefs<0,255,192> t2:Func, 64 -> B10<Unlikely>
        Main:
              0x1660c9b4  f9402400              ldr x0, [x0, #72]
              //cmpi w0, 192
              0x1660c9b8  53001c00              uxtb w0, w0
              0x1660c9bc  128007e1              movn w1, #0x3f
              0x1660c9c0  6b01001f              cmp w0, w1
              0x1660c9c4  540000a0              b.eq #+0x14
              0x1660c9c8  58000052              ldr x18, pc+8 (addr 0x1660c9d0)
              0x1660c9cc  d61f0240              br x18
              0x1660c9d0  1a603948              unallocated (Unallocated)
              0x1660c9d4  00000000              unallocated (Unallocated)

Signed-off-by: Christoph Muellner <christoph.muellner@theobroma-systems.com>

PR #7772, opened 2017-04-07
jit: irlower-call: Do proper loading in cgCheckRefs to eliminate truncation

cmuellner:
The vasm for CheckRefs distinguishes between 8-bit, 32-bit and 64-bit
masks. However it uses a 64-bit load for all cases in combination with
a movtqX. This patch changes this to use a load which does not require
a truncation for the cost of being little-endian specific (which is already
a requirement for the cgCheckRefs() function).

Testing the all set showed no regressions on ARM64.

### DIFF
--- a/hphp/runtime/vm/jit/irlower-call.cpp
+++ b/hphp/runtime/vm/jit/irlower-call.cpp
@@ -475,20 +475,18 @@ void cgCheckRefs(IRLS& env, const IRInstruction* inst)  {
       if (vals64) cond = CC_E;
     } else {
       auto const bits = v.makeReg();
-      v << load{bitsPtr[bitsOff], bits};
-
-      auto const truncBits = v.makeReg();
       auto const maskedBits = v.makeReg();
 
       if (mask64 <= 0xff && vals64 <= 0xff) {
-        v << movtqb{bits, truncBits};
-        v << andbi{(int8_t)mask64, truncBits, maskedBits, v.makeReg()};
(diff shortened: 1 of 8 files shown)


cmuellner:
This is one of the optimisations derived from #7767.

Note, that cgCheckRefs() already uses the following code,
which is not endian neutral as well:

`hphp/runtime/vm/jit/irlower-call.cpp:469: v << testbim{(int8_t)mask64, bitsPtr[bitsOff], sf};`
`hphp/runtime/vm/jit/irlower-call.cpp:471: v << testlim{(int32_t)mask64, bitsPtr[bitsOff], sf};`


@mxw : please review

swalk-cavium:
Downloaded branch above.  It builds clean.  Starting MOP.
swalk@arm64:/data2/swalk/cmuellner/hhvm$ git branch -v
  master        035b577 Unclown publishOptFunctionInternal
* pr-2017-04-07 1c65d2e jit: irlower-call: Do proper loading in cgCheckRefs to eliminate truncation

swalk-cavium:
MOP finished.  I sent the results to @cmuellner directly via e-mail.  A quick glance
doesn't show any new issues.

cmuellner:
@mxw - I've introduced a loadtql and use it later on. I've implemented the loadtql emit code for x86 and ppc64 as well, however I did not test those. If x86 looks good, we might ping the ppc64 guys for testing as well.

mxw:
lgtm—will land once internal tests finish running.

Thanks for this series of PRs!

Issue #192, opened 2010-08-22
memcache::connect does not give return value


chrisgraham:
<pre>
$MEM_CACHE=new Memcache();
$test=$MEM_CACHE->connect('localhost',11211);
echo gettype($test);
$MEM_CACHE->set('test','x');
echo $MEM_CACHE->get('test',time()-60*60);
</pre>

Code works fine, except $test is NULL. This is a problem as we check the return value of connect to determine whether it was successful (and halt using our memcache support if it was not).

http://php.net/manual/en/memcache.connect.php

I've never actually used memcached normally (wrote our support for it somewhat blind, as just one possible backend for our cache API) so it's possible I am misunderstanding something, but I don't think so.

Issue #72, opened 2010-05-03
Compilation error with eval()s


xenor:
Can it be, that variable evals(); cannot be processed?

[xenor@archftw hiphop]$ ./hphp.sh -k true -o /home/xenor/xBot/bin/ -i /home/xenor/xBot/core.inc.php
[31, {
"UseEvaluation":
[{"c1":["/home/xenor/xBot/core.inc.php",0,0,267,15],"d":"eval($code)","s":0,}
,{"c1":["/home/xenor/xBot/core.inc.php",0,0,395,31],"d":"eval($this->parseeval($cmd))","s":0,}
]
}]
[xenor@archftw hiphop]$

PR #7774, opened 2017-04-07
vasm-simplify: Eliminate masking values with -1. 

cmuellner:
Masking 8/32-bit values with 0xff/0xffffffff is
not necessary. This patch eliminates that.

```
Before:
      load       [%129 + 0x48] => %147              # (14) CheckRefs<0,255,192> t2:Func, 64 -> B10<Unlikely>
      movtqb     %147 => %150
      movzbq     %150 => %151
      movtqb     %151 => %148
      andbi      -1, %148 => %149, %152
      cmpbi      -64, %149 => %145

After:
      load       [%129 + 0x48] => %147              # (14) CheckRefs<0,255,192> t2:Func, 64 -> B10<Unlikely>
      movtqb     %147 => %150
      movzbq     %150 => %151
      movtqb     %151 => %148
      movb       %148 => %149
      cmpbi      -64, %149 => %145
```

Or in ARM64 assembly (already with #7744 applied (uxtb instead of sxtb for movtqb)):

```
Before:
              0x24c0c9d0  f9402400              ldr x0, [x0, #72]
              //movtqb w0 <- w0
              0x24c0c9d4  53001c00              uxtb w0, w0
              //andbi w0 <- w0, 255
              0x24c0c9d8  12800001              movn w1, #0x0
              0x24c0c9dc  0a010000              and w0, w0, w1
              //cmpi w0, 192
              0x24c0c9e0  53001c00              uxtb w0, w0
              0x24c0c9e4  128007e1              movn w1, #0x3f
              0x24c0c9e8  6b01001f              cmp w0, w1

After:
              0x48c0c86c  f9402400              ldr x0, [x0, #72]
              //movtqb w0 <- w0
              0x48c0c870  13001c00              uxtb w0, w0
              //andbi w0 <- w0, 255
              0x48c0c874  2a0003e0              mov w0, w0
              //cmpi w0, 192
              0x48c0c878  53001c00              uxtb w0, w0
              0x48c0c87c  128007e1              movn w1, #0x3f
              0x48c0c880  6b01001f              cmp w0, w1
```

This is another optimisation derived from #7767.

Note, that this PR contains a second commit, which prevents
that `andbi/andli -1` gets lowered in vasm-arm.cpp (before simplify()).

@mxw : please review

### DIFF
--- a/hphp/runtime/vm/jit/vasm-simplify.cpp
+++ b/hphp/runtime/vm/jit/vasm-simplify.cpp
@@ -473,6 +473,29 @@ bool simplify(Env& env, const pop& inst, Vlabel b, size_t i) {
   });
 }
 
+/*
+ * Eliminate masking values with -1 in andXi.
+ * andbi s0=0xff s1=val0 d=val1 -> copy s=val0 d=val1
+ * andli s0=0xffffffff s1=val0 d=val1 -> copy s=val0 d=val1
+ */
+template<typename andi>
+bool simplify_andi(Env& env, const andi& inst, Vlabel b, size_t i) {
+  if (inst.s0.l() != -1 ||
+      env.use_counts[inst.sf] != 0) return false;
+  return simplify_impl(env, b, i, [&] (Vout& v) {
+    v << copy{inst.s1, inst.d};
+    return 1;
+  });
+}
+
+bool simplify(Env& env, const andbi& andbi, Vlabel b, size_t i) {
+  return simplify_andi(env, andbi, b, i);
+}
+
+bool simplify(Env& env, const andli& andli, Vlabel b, size_t i) {
+  return simplify_andi(env, andli, b, i);
+}
+
 ///////////////////////////////////////////////////////////////////////////////
 
 /*
--- a/hphp/runtime/vm/jit/vasm-arm.cpp
+++ b/hphp/runtime/vm/jit/vasm-arm.cpp
@@ -1209,8 +1209,8 @@ void lower(const VLS& e, vasm_opc& i, Vlabel b, size_t z) { \
 
 Y(addli, addl, ldimml, ISAR, i.s0, U2(i.s1, i.d))
 Y(addqi, addq, ldimmq, ISAR, Immed64(value), U2(i.s1, i.d))
-Y(andbi, andb, ldimmb, ISLG(32), i.s0, U2(i.s1, i.d))
-Y(andli, andl, ldimml, ISLG(32), i.s0, U2(i.s1, i.d))
+Y(andbi, andb, ldimmb, (ISLG(32) || (safe_cast<int8_t>(value) == -1)), i.s0, U2(i.s1, i.d))
+Y(andli, andl, ldimml, (ISLG(32) || (safe_cast<int32_t>(value) == -1)), i.s0, U2(i.s1, i.d))
 Y(andqi, andq, ldimmq, ISLG(64), Immed64(value), U2(i.s1, i.d))
 Y(cmpli, cmpl, ldimml, ISAR, i.s0, i.s1)
 Y(cmpqi, cmpq, ldimmq, ISAR, Immed64(value), i.s1)


ptomsich:
Just a handful comments/questions on what exactly is going on here:

> On 08 Apr 2017, at 01:28, Christoph Müllner <notifications@github.com <mailto:notifications@github.com>> wrote:
> 
> After:
>               0x48c0c86c  f9402400              ldr x0, [x0, #72]
>               //movtqb w0 <- w0
>               0x48c0c870  13001c00              uxtb w0, w0
>               //andbi w0 <- w0, 255
>               0x48c0c874  2a0003e0              mov w0, w0
>               //cmpi w0, 192
>               0x48c0c878  53001c00              uxtb w0, w0
This will set w0 as:
	w0 = ((unsigned)value) & 0xff;
I.e.: w0 = 0x000000xy;
>               0x48c0c87c  128007e1              movn w1, #0x3f
Now we have w1 as ~0x3f … i.e. 0xffffffc0.
Note that with the implicit zero-extension the X-form is 0x00000000ffffffc0.
>               0x48c0c880  6b01001f              cmp w0, w1
And finally  a compare between these two W-form values.

Is this really what we want?
From the "CheckRefs<0,255,192> t2:Func, 64 -> B10<Unlikely>” comment it seems this is not what we want.
The HHIR spec says:
> | CheckRefs<firstBit,mask,vals>, ND, S(Func) S(Int), B
> 
>   Perform reffiness guard checks. Operands:
> 
>     S0 - function pointer for the frame
>     S1 - num params expected in the func
>     firstBit - first bit to check, must be a multiple of 64
>     mask - mask to check (RefDeps::Record::m_mask entries)
>     vals - values to check (RefDeps::Record::m_vals entries)

So my understanding of what we want is

	if ( (x0 & mask) == vals )
		goto t2:Func, 64;
	else
		goto B10;

which would translate to AArch64 as

	ldr x0, [x0, #72]
	uxtb x0, x0		// as an alternate to and and "x0, x0, #0xff";
	eors x0, #192		// the “inverted” case of a bitwise compare
	b.ne B10
	[put the likely case in the fall-through path]

which is equivalent to

	if ((x0 & 0xff) != 0x3f)	// can compile to xor
		goto B10;
	
	[likely fall-through case]
		
and could be optimised into (for the little-endian case)

	ldrb x0, [x0, #72]
	eors x0, #192
	b.ne B10


Please let me know if (and where) I am completely wrong, but the original code-sequence still (I had a lenghty discussion about this with Christoph last week) doesn’t look sensible.

Regards,
Philipp.

cmuellner:
Just for reference, here is the complete block (with HHIR):

```
 B9: [profCount=1] (preds B8)
    (11) t2:Func = LdARFuncPtr<IRSPOff 0> t1:StkPtr
        Main: 
              0x4860c868  f85303a0              ldur x0, [x29, #-208]

    (14) CheckRefs<0,255,192> t2:Func, 64 -> B10<Unlikely>
        Main: 
              0x4860c86c  f9402400              ldr x0, [x0, #72]
//movtqb w0 <- w0
              0x4860c870  53001c00              uxtb w0, w0
//andbi w0 <- w0, 255
              0x4860c874  2a0003e0              mov w0, w0
//cmpi w0, 192
              0x4860c878  53001c00              uxtb w0, w0
              0x4860c87c  128007e1              movn w1, #0x3f
              0x4860c880  6b01001f              cmp w0, w1
//fwdJcc CC_NE, B10
              0x4860c884  540000a0              b.eq #+0x14
  -> B11
              0x4860c888  58000052              ldr x18, pc+8 (addr 0x4860c890)
              0x4860c88c  d61f0240              br x18
              0x4860c890  4c6039ac              unimplemented (Unimplemented)
              0x4860c894  00000000              unallocated (Unallocated)

```

I think we should have a look, where this ldur is comming from
and why x0 is not used in the following or why ldur is emitted.

I fully agree with optimizing the lowering of `cmpi w0, 192` to your proposed sequence.
That would be the job of the lowering/emit code in vasm-arm.cpp. I put that on my list.

For the `fwdJcc` there's not much we can do. Maybe eliminate the ldr if the target is close enough, I think that's what the relocation code does/will do.

cmuellner:
@mxw should this be part of removeTrivialNops()?

cmuellner:
Update uses template function and does not create useless mov{rN, rN} (self assignment - see also #7775 ).

hphp/runtime/vm/jit/vasm-simplify.cpp
```diff
@@ -473,6 +473,29 @@ bool simplify(Env& env, const pop& inst, Vlabel b, size_t i) {
...
+ * Eliminate masking values with -1 in andXi.
+ * andbi s0=0xff s1=val0 d=val1 -> movb s=val0 d=val1
+ * andli s0=0xffffffff s1=val0 d=val1 -> movl s=val0 d=val1
+ */
+template<typename mov, typename andi>
+bool simplify_andi(Env& env, const andi& inst, Vlabel b, size_t i) {
+  if (inst.s0.l() != -1) return false;
+  return simplify_impl(env, b, i, [&] (Vout& v) {
+    if (inst.s1 != inst.d)
+      v << mov{inst.s1, inst.d};
```
mxw: We don't allow unbraced multiline conditions in HHVM code (cf. the Apple SSL "goto fail" bug).  You can skim `doc/coding-conventions.md` for other such conventions.

This should be either

```
if (inst.s1 != inst.d) {
  v << mov{inst.s1, inst.d};
}
```

or just

```
if (inst.s1 != inst.d) v << mov{inst.s1, inst.d};
```

hphp/runtime/vm/jit/vasm-simplify.cpp
```diff
@@ -473,6 +473,29 @@ bool simplify(Env& env, const pop& inst, Vlabel b, size_t i) {
...
+    return 1;
+  });
+}
```
mxw: This transformation is unsound.  You also need to ensure that nobody is using the SF output operand of the `and*{}` instruction.  You can look around this file for similar checks.

hphp/runtime/vm/jit/vasm-simplify.cpp
```diff
@@ -473,6 +473,29 @@ bool simplify(Env& env, const pop& inst, Vlabel b, size_t i) {
...
+ * Eliminate masking values with -1 in andXi.
+ * andbi s0=0xff s1=val0 d=val1 -> movb s=val0 d=val1
+ * andli s0=0xffffffff s1=val0 d=val1 -> movl s=val0 d=val1
+ */
+template<typename mov, typename andi>
+bool simplify_andi(Env& env, const andi& inst, Vlabel b, size_t i) {
+  if (inst.s0.l() != -1 ||
+      env.use_counts[andi.sf] != 0) return false;
+  return simplify_impl(env, b, i, [&] (Vout& v) {
+    if (inst.s1 != inst.d) v << mov{inst.s1, inst.d};
```
mxw: Sorry for the back and forth, but thinking on this a bit more—I think this should actually just be a `copy{}`.  You can nix the template `mov` parameter and just use `copy{}` unconditionally.

The `mov*{}` instructions are only used in unique-stubs, and to be honest, I don't remember right now what their specific semantics are intended to be... on x64, they zero the upper bits of the dst register, but that behavior doesn't seem absolutely required, since currently we only use them to copy constant registers, which are always 64 bits, implicitly.  As far as I can tell, they're basically a `copy{}` that tells the vasm emitter that it's allowed to address less than 64 bits of the operands.

Since `copy{}` is transparent to the register allocator, you don't need the condition, either.  (It's probably slightly better to always emit the `copy{}`, since it avoids a vector shift.)

After that we just need to wait for internal tests to pass before landing.

cmuellner: Then I can come with #7775 again. ;) But this time for copy rN rN. => #7788

hphp/runtime/vm/jit/vasm-simplify.cpp
```diff
@@ -473,6 +473,29 @@ bool simplify(Env& env, const pop& inst, Vlabel b, size_t i) {
   });
 }
 
+/*
+ * Eliminate masking values with -1 in andXi.
+ * andbi s0=0xff s1=val0 d=val1 -> movb s=val0 d=val1
+ * andli s0=0xffffffff s1=val0 d=val1 -> movl s=val0 d=val1
+ */
+template<typename mov, typename andi>
```
mxw: You can kill the `mov` parameter now also; it's no longer used.

mxw:
Cool, lgtm.  Internal tests are running.

PR #7778, opened 2017-04-09
jit: generic approach to eliminate load;truncation sequences.

cmuellner:
This is a generic approach to eliminate load;truncation sequences (and replace them by loads of target size).

As this optimization is endianess-dependent I've decided to add a
constexpr-based endianess detection utility to HHVM.

@mxw : please review

### DIFF
--- a/hphp/runtime/vm/jit/vasm-simplify.cpp
+++ b/hphp/runtime/vm/jit/vasm-simplify.cpp
@@ -473,6 +473,32 @@ bool simplify(Env& env, const pop& inst, Vlabel b, size_t i) {
   });
 }
 
+/*
+ * Optimize load followed by truncation
+ * loadq ; movtqb -> loadb (we could use loadtqb as well)
+ * loadq ; movtql -> loadl
+ */
+template<Vinstr::Opcode mov_op, typename loadt>
+bool simplify_load_truncate(Env& env, const load& load, Vlabel b, size_t i) {
+  if (env.use_counts[load.d] != 1) return false;
+  auto const& code = env.unit.blocks[b].code;
+  if (i + 1 >= code.size()) return false;
+
+  return if_inst<mov_op>(env, b, i + 1, [&] (const op_type<mov_op>& mov) {
+    if (load.d != mov.s) return false;
+    return simplify_impl(env, b, i, [&] (Vout& v) {
+      v << loadt{load.s, mov.d};
+      return 2;
+    });
+  });
+}
+
+bool simplify(Env& env, const load& load, Vlabel b, size_t i) {
+  return
+    simplify_load_truncate<Vinstr::movtqb, loadtqb>(env, load, b, i) ||
+    simplify_load_truncate<Vinstr::movtql, loadtql>(env, load, b, i);
+}
+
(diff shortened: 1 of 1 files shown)


cmuellner:
@swalk-cavium exactly the same.

Every loadX->truncateXY sequence, where the load value is only used for the truncation, will be replaced by a loadY in an endianess-aware manner.

hphp/runtime/vm/jit/vasm-simplify.cpp
```diff
@@ -473,6 +474,48 @@ bool simplify(Env& env, const pop& inst, Vlabel b, size_t i) {
...
+    if (endian() == Endian::LITTLE)                                \
+      v << load_opc{load.s, mov.d};                                \
+    else                                                           \
+      v << load_opc{load.s + be_offset, mov.d};                    \
+    return 2;                                                      \
+  });                                                              \
+}
+
+Y(movtqb, loadb, 7)
+Y(movtql, loadl, 4)
```
mxw: This pattern is common in `vasm-arm.cpp`, but we actually prefer to avoid it in general.  Rather than macro-define these functions, you should define a templatized implementation, e.g.

```
template<typename Load, typename mov_op>
bool simplify_load_truncate(Env& env, const load& vload, Vlabel b, size_t i) {
  return if_inst<mov_op>(env, b, i + 1, [&] (const op_type<mov_op>& vmov) {
    // check conditions
    return simplify_impl(env, b, i, [&] (Vout& v) {
      ...
      v << Load{...};
      ...
    });
  });
}

bool simplify(Env& env, const load& vload, Vlabel b, size_t i) {
  return
    simplify_load_truncate<loadtqb, Vinstr::movtqb>(env, vload, b, i) ||
    simplify_load_truncate<loadtql, Vinstr::movtql>(env, vload, b, i);
}
```

(It doesn't look like we actually have a `loadtql{}`, but it should be easy to add one.)

I like the caution you're taking around endianness, but if you look at the implementations of `loadtqb{}` among the different architectures, they're all just identical to `loadb{}` currently.  The endianness check should probably be handled there (though I'm sure there are plenty of other things that would break us on big-endian systems today).

For the scope of this PR, I think it would be best to implement `loadtql{}` for all three architectures, copying the implementation of `loadl{}`, and use `loadtqb{}` and `loadtql{}` here in vasm-simplify.  Feel free to put up a separate PR for endianness checks in `vasm-{arm,ppc64,x64}.cpp` if you want; we'd definitely be happy to take those, too.

cmuellner:
I was thinking about a template function approach, but preferred the macro version, as it does not have to call possible functions (in that return foo<a> || foo<b> ... expression) resulting in less JIT-time. But I'm fine with the template function as well (especially since that's much easier to write for me).

Note that this PR re-uses the exact same commit for introducing the loadtql like in #7772.

mxw:
I'd be surprised if calling the template function cost anything noticeable in terms of JIT time (which is most likely dominated by allocating, iterating, and tearing down a bunch of O(n) data structures over and over)—also, they could easily be inlined.

This diff looks good; I'll land it once tests pass.

jim-saxman:
OSS runs fine for us. LGTM.

PR #7781, opened 2017-04-11
Updates use of perf script to match change to fields argument.

dave-estes:


dave-estes:
@mxw, this is another new PR split from #7661.

dave-estes:
Sure thing. Uploading a new patch now.

PR #7782, opened 2017-04-11
Refactors adjustMetaDataForRelocation to be common across platforms.

dave-estes:
Each platform will still have an implementation of
adjustMetaDataForRelocation(), but it will be reserved for any
additional adjustments for that platform.


dave-estes:
@mxw, this is another PR split from #7661.

mxw:
This diff is taking a long time to land because our test infra has been have issues for days.  There's no action to be taken—I'm just waiting for a clean run.

PR #7788, opened 2017-04-17
arm: Stop emitting moves for `copy rN rN`

cmuellner:
This adds a simple optimization, which is already in place for x86.

### DIFF
--- a/hphp/runtime/vm/jit/vasm-arm.cpp
+++ b/hphp/runtime/vm/jit/vasm-arm.cpp
@@ -386,6 +386,7 @@ void Vgen::patch(Venv& env) {
 ///////////////////////////////////////////////////////////////////////////////
 
 void Vgen::emit(const copy& i) {
+  if (i.s == i.d) return;
   if (i.s.isGP() && i.d.isGP()) {
     a->Mov(X(i.d), X(i.s));
   } else if (i.s.isSIMD() && i.d.isGP()) {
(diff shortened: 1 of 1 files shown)


cmuellner:
@mxw : please review

mxw:
I'm surprised to see that line in `vasm-x64.cpp`...

Did you confirm whether this diff actually has any effect in practice—and if so, can you post an example where you get different code?  I'm pretty sure that the register allocator will resolve any self-copies away.

cmuellner:
@mxw - I think you are right. vasm-xls.cpp:2177 should do that already. However even `void Vgen::emit(const copy2& i)` in vasm-x64.cpp has similar code (`if (s0 != d0)...`). I've just started a test run with an assertion for `i.s != i.d`. Let's if the `all` test set triggers it. If not, should I prepare a patch for x86?

mxw:
It's possible that this code predates that behavior in vasm-xls.

In any case, it's probably better to have this check in the emitter than not.  Having it makes the emitter more independent of the register allocator's particular policy around copy elision.  I was just curious if you actually saw this diff having any effect—that would probably indicate a bug in our XLS implementation.

Let me know what you find, but we can just land this as-is.

cmuellner:
The assertion did not trigger. I.e. the register allocator does its job well.

While the tests were running I had a look at the emit function for copy2 and adapted this to a form, which matches the x86 code (i.e. add the self-assignment protection and remove the use of reg-algorithms). Re-testing with this change as well showed again no regressions on aarch64.

jim-saxman:
Works for me.

PR #7793, opened 2017-04-19
Enable runtime/test build for OSS.

dave-estes:
The GTest based tests under hphp/runtime/test were previously not
built under the OSS build. This patch adds the CMake support which
is enabled with -DRUNTIME_TEST_BIN=on. The results is a binary,
hphp_runtime_test.

Additionally the tests were ported to ARM.

### DIFF
--- a/hphp/runtime/test/CMakeLists.txt
+++ b/hphp/runtime/test/CMakeLists.txt
@@ -1,0 +1,24 @@
+enable_testing()
+find_package(GTest REQUIRED)
+include_directories(${GTEST_INCLUDE_DIRS})
+
+set(CXX_SOURCES)
+auto_sources(files "*.cpp" "RECURSE" "${CMAKE_CURRENT_SOURCE_DIR}")
+list(APPEND CXX_SOURCES ${files} )
+
+set(HEADER_SOURCES)
+auto_sources(files "*.h" "RECURSE" "${CMAKE_CURRENT_SOURCE_DIR}")
+list(APPEND HEADER_SOURCES ${files} "${CMAKE_CURRENT_SOURCE_DIR}/../../hhvm/process-init.cpp")
+
+add_executable(hphp_runtime_test ${CXX_SOURCES} ${HEADER_SOURCES})
+link_object_libraries(hphp_runtime_test ${HHVM_WHOLE_ARCHIVE_LIBRARIES})
+target_link_libraries(hphp_runtime_test ${HHVM_LINK_LIBRARIES} gtest gtest_main gmock gmock_main)
+embed_all_systemlibs(hphp_runtime_test "${CMAKE_CURRENT_BINARY_DIR}/../.."
+                              "${CMAKE_CURRENT_BINARY_DIR}/hphp_runtime_test")
+
+auto_source_group("hphp_runtime_test" "${CMAKE_CURRENT_SOURCE_DIR}"
+  ${CXX_SOURCES} ${HEADER_SOURCES})
+add_dependencies(hphp_runtime_test hphp_system)
+if (ENABLE_COTIRE)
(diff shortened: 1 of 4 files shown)


dave-estes:
@mxw Does this look like a reasonable way to enable these test in the OSS build? I plan to add the tests that we discussed in #7661 next as part of that PR.

hphp/runtime/test/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+enable_testing()
```
Orvid: This should only be required if you use CMake's built-in testing support, but you aren't using it, so this isn't needed.

hphp/runtime/test/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+auto_sources(files "*.cpp" "RECURSE" "${CMAKE_CURRENT_SOURCE_DIR}")
+list(APPEND CXX_SOURCES ${files} )
+
+set(HEADER_SOURCES)
+auto_sources(files "*.h" "RECURSE" "${CMAKE_CURRENT_SOURCE_DIR}")
+list(APPEND HEADER_SOURCES ${files} "${CMAKE_CURRENT_SOURCE_DIR}/../../hhvm/process-init.cpp")
+
+add_executable(hphp_runtime_test ${CXX_SOURCES} ${HEADER_SOURCES})
+link_object_libraries(hphp_runtime_test ${HHVM_WHOLE_ARCHIVE_LIBRARIES})
+target_link_libraries(hphp_runtime_test ${HHVM_LINK_LIBRARIES} gtest gtest_main gmock gmock_main)
```
Orvid: So... GTest is a bit of an absurd beast... All you want here is `gtest_main`, everything else is defining the exact same symbols...

hphp/runtime/test/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+find_package(GTest REQUIRED)
+include_directories(${GTEST_INCLUDE_DIRS})
+
+set(CXX_SOURCES)
+auto_sources(files "*.cpp" "RECURSE" "${CMAKE_CURRENT_SOURCE_DIR}")
+list(APPEND CXX_SOURCES ${files} )
+
+set(HEADER_SOURCES)
+auto_sources(files "*.h" "RECURSE" "${CMAKE_CURRENT_SOURCE_DIR}")
+list(APPEND HEADER_SOURCES ${files} "${CMAKE_CURRENT_SOURCE_DIR}/../../hhvm/process-init.cpp")
```
Orvid: `process-init.cpp` is an interesting name for a header file :)

Stick it in the `CXX_SOURCES` list instead.

dave-estes:
Thanks for the quick reply, @Orvid.

hphp/runtime/test/CMakeLists.txt
dave-estes: Thanks. I tried it with only the `*_main`, but it failed to link. I worked well with just `gtest` and `gmock`.

hphp/runtime/test/vasm-xls-test.cpp
```diff
@@ -67,13 +77,17 @@ template<class T> uint64_t test_const(T val) {
...
+#else
   optimizeX64(vasm.unit(), test_abi, true /* regalloc */);
-  CGMeta fixups;
-
-  emitX64(unit, text, fixups, nullptr);
-  // The above code might use fixups.literals but shouldn't use anything else.
-  fixups.literals.clear();
-  EXPECT_TRUE(fixups.empty());
+  emitX64(unit, text, meta, nullptr);
+#endif
```
mxw: Prefer `if (arch() == ...)` here and everywhere.

PR #7795, opened 2017-04-20
jit: Removal of vasm instructions subb and subbi

cmuellner:
I wanted to verify the correctness of aarch64's emit function for subb and subbi, when I realised that both instructions are not emitted at all. With this PR I propose the elimination of them. Note that subb was not even implemented for x86_64 and ppc64.

The all set showed no regressions on ARM64.

@mxw please review.


mxw:
Sure, seems fine.  Will land once tests pass.

PR #7796, opened 2017-04-20
Set a max amount of time per-request allowed in tc

jscheinblum:
Our application and deployment process doesn't lend itself well to
a warmup script, and especially in our developer/test environments
the jit costs are expensive. This sets a per-request jit budget
that amortizes the jit cost across requests.

Tested with the following config flags:

	hhvm.jit_max_request_translation_time = 50000
	hhvm.jit_timer = 1

Background:
We have a large and broad PHP application that we deploy to dozens of times per-day. Because of a variety of deployment constraints that are difficult for us to change in the near-term, it's difficult for us to run a warm-up script before traffic is served.

Because of this, we observe high tail-latencies on requests after a deployment. Because we have a vary large pool of hosts in the fleet that we serve round-robin and have some infrequently accessed pages, some users end up hitting the jit penalty repeatedly.

Desired change:
We'd like to set a per-request budget for the amount of time to spend in the jit, to amortize that cost across requests.  We'd like to tradeoff a longer time getting code to be fully JITted for a smoother response-time upper bound post-deploy.  We took a stab at adding the limit and our running it on our dev environments, and would like feedback.

Here's what we are seeing after restart for 100 requests, time-to-first-byte vs the size of the tc cache reported in the admin port:

max-request-time-stock

With the translation limit set to 100ms:

max-request-time-with-change

Thanks, and we'd love feedback on the idea and implementation!

### DIFF
--- a/hphp/runtime/base/runtime-option.h
+++ b/hphp/runtime/base/runtime-option.h
@@ -513,6 +513,7 @@ struct RuntimeOption {
   F(uint32_t, JitMaxTranslations,      10)                              \
   F(uint32_t, JitMaxProfileTranslations, 30)                            \
   F(uint64_t, JitGlobalTranslationLimit, -1)                            \
+  F(uint64_t, JitMaxRequestTranslationTime,   -1)                       \
   F(uint32_t, JitMaxRegionInstrs,      1347)                            \
   F(uint32_t, JitProfileInterpRequests, kDefaultProfileInterpRequests)  \
   F(bool, JitProfileWarmupRequests,    false)                           \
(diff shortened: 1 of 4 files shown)


mofarrell:
@jscheinblum This looks awesome.  Since PHP deployment processes rarely include a warmup, a lot of users have this issue when migrating from PHP to HHVM.  As soon as you sign the CLA we can look to upstream this.

hphp/runtime/base/runtime-option.h
```diff
@@ -513,6 +513,7 @@ struct RuntimeOption {
   F(uint32_t, JitMaxTranslations,      10)                              \
   F(uint32_t, JitMaxProfileTranslations, 30)                            \
   F(uint64_t, JitGlobalTranslationLimit, -1)                            \
+  F(uint64_t, JitMaxRequestTranslationTime,   -1)                       \
```
mofarrell: This is an unsigned type.  The comparison in `tc-internal.cpp:130` will always succeed, and the compiler doesn't like that.  The setting above this does the same thing, except in that case -1 becomes effectively unlimited.  The same thing can be done here by removing the check for `maxTransTime >= 0` below.  I prefer making this a signed int as it is more clear.

PR #7797, opened 2017-04-20
arm64 cleanup and performance optimization

cmuellner:


mxw:
@cmuellner—It looks like there are a bunch of local commits that you've rolled into a single PR.  Can you split them into separate PRs instead, with one for each unit of change?  PRs with only a single line changed are totally fine, and easier to digest than one PR basically titled "changes" :]

cmuellner:
I've pushed these patch set as individual PRs and referenced this PR there. There is nothing left to do here. Therefore closing this PR.

PR #7801, opened 2017-04-21
Reduce false sharing in JIT performance counters

swalk-cavium:
The memory for JIT performance counters is currently allocated sequentially.  This
can lead to false sharing for small benchmarks and hosts with large cache lines.
This change slightly improves performance during the profiling interval.

The memory for the performance counter is currently referenced as a doubly indexed
array:  [id / kCountersPerChunk][id % kCountersPerChunk]
Since C/C++ uses row-major ordering the elements from the second index are next to
each other.

This change makes the memory less cache friendly to force the access around different
areas in memory. The second index is multiplied by a small prime and as long as the
modulus variable does not contain a factor there should be no issues.

The standard regression tests were run with 6 option sets.  No additional issues were observed.

### DIFF
--- a/hphp/runtime/vm/jit/prof-data.h
+++ b/hphp/runtime/vm/jit/prof-data.h
@@ -95,6 +95,7 @@ struct ProfCounters {
   ProfCounters& operator=(const ProfCounters&) = delete;
 
   T get(uint32_t id) const {
+    assertx(kCountersPerChunk % 11);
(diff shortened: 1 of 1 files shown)


swalk-cavium:
@ottoni - You send e-mail on 03/17/2017 to the hhvm-arm mailing list regarding this issue.
@mxw - Can you review?

PR #7805, opened 2017-04-25
jit: arm: simplify before invoking lowerForArm().

cmuellner:

### DIFF
--- a/hphp/runtime/vm/jit/vasm-arm.cpp
+++ b/hphp/runtime/vm/jit/vasm-arm.cpp
@@ -1214,8 +1214,8 @@ void lower(const VLS& e, vasm_opc& i, Vlabel b, size_t z) { \
 
 Y(addli, addl, ldimml, ISAR, i.s0, U2(i.s1, i.d))
 Y(addqi, addq, ldimmq, ISAR, Immed64(value), U2(i.s1, i.d))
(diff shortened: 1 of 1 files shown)


cmuellner:
@mxw please review.

mxw:
How did you confirm that this isn't causing us to simplify less than before?  Can you comment with any testing scripts or sample output?

cmuellner:
I don't have a 100% safe answer to your question. I just know of cases where the lowering actually caused less to be simplified. Would it be ok to call simplify() before and after lowering()?

mxw:
@cmuellner—It should be safe, yes.

I think if lowering causes less simplification to happen, though, it probably means that the simplifier could use a few more rules (especially since a lot of lowerToARM() is just clearing upper bits of registers).  It might be worthwhile to actually compare code output, or at least code size (in Vinstr count) with the different orderings, to try to identify what's actually changing.

For the former, you can just use TRACE and then diff the outputs.  For the latter, you can take a look at `util/struct-log.h`, and according to the API defined there, you can implement a very simple StructuredLog routine which just writes to stdout.  (Vasm is already hooked up to StructuredLog to dump statistics on various Vunits; it's configurable through a RuntimeOption—see `logTranslation()` in `vm/jit/tc-record.cpp`.)

PR #7806, opened 2017-04-25
jit: arm64: Drive-by cleanup of vasm fold functions.

cmuellner:
* Replace copy pasta with generic functions
 * Add fold functions for *l vasm instructions
 * Use order: b, bi, w, wi, l, li, q, qi

### DIFF
--- a/hphp/runtime/vm/jit/vasm-fold-imms.cpp
+++ b/hphp/runtime/vm/jit/vasm-fold-imms.cpp
@@ -344,6 +344,7 @@ struct ImmFolder {
     out = imm64;
     return true;
   }
+
   bool logical_imm(Vreg r, int32_t& out) {
     if (!valid.test(r)) return false;
     auto imm64 = vals[r];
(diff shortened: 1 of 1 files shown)


cmuellner:
@mxw please review.

hphp/runtime/vm/jit/vasm-fold-imms.cpp
```diff
@@ -350,65 +351,121 @@ struct ImmFolder {
...
+  void fold(storeqi& in, Vinstr& out) {
+    if (in.s.q() == 0) out = store{PhysReg(vixl::xzr), in.m};
+  }
+  void fold(subb& in, Vinstr& out) {
+    int val;
+    if (arith_imm(in.s0, val)) {
+      if (val == 0 && !used[in.sf]) { // copy sets no flags.
+        out = copy{in.s1, in.d};
+      } else {
+        out = subbi{val, in.s1, in.d, in.sf};
```
jim-saxman: Didn't #7795 remove subbi vasm?

cmuellner:
@jim-saxman I'm happy to rebase this PR (without the subb/subbi folding) once #7795 has landed. As I don't know how to make PRs dependent on each other, I have to come up with correct and complete patches, since I cannot influence the order or merging PR.

cmuellner:
@swalk-cavium reported that he observed a performance issue:

```
   (44) CheckPackedArrayDataBounds t15:Arr=PackedKind, 0 -> B15<Unlikely>
       Main:
             0x2a053770  b9400801              ldr w1, [x0, #8]
             0x2a053774  d2800002              movz x2, #0x0 //should be folded to cmp x1, 0x0
             0x2a053778  eb02003f              cmp x1, x2
             0x2a05377c  540000a8              b.hi #+0x14
             0x2a053780  58000052              ldr x18, pc+8 (addr 0x2a053788)
             0x2a053784  d61f0240              br x18
             0x2a053788  30a247e0              adr x0, #-0xbb703
             0x2a05378c  00000000              unallocated (Unallocated)
```

I've already found the issue (in this PR) and will come up with a corrected version of this PR after testing.

Thanks @swalk-cavium for reporting the issue!

mxw:
@cmuellner—This looks good to me.  Once the PR which removes `subbi{}` lands, you can rebase and update this one, and then I'll upstream it.  (There's a diff that we upstreamed internally recently which changed the `used[]` flags to an array of `uses[]` counts—it shouldn't affect what you're doing here; I just wanted to warn you about the rename.)

One request: in the future, could you add commits to your stack when you update a PR, rather than squashing your changes into the same commit?  When your PR lands, everything will get autosquashed, but keeping the commits separate while the PR is live makes it easier to track changes.  (Otherwise, we lose the ability to diff between different versions on GitHub.)  Thanks!

cmuellner:
@mxw - Ok, will do.

cmuellner:
I've just noticed that both mentioned changes (subb/subbi removal and used[]->uses[] rename) are merged. Therefore I will rebase and (force-)push this PR after confirming that everything is still fine (compiles and no regressions in the all set).

cmuellner:
Just for the record: this patch shows a speedup of 1.64% in OSS-performance's wikimedia benchmark (comparing the average RPSs of 15 runs each).

PR #7807, opened 2017-04-25
immed: Fix unsigned values retrieval.

cmuellner:
The Immed class offers a single constructor,
which assumes the immediate value to be int32_t.
All caller sides in HHVM decided therefore to
cast their values to int32_t. However that
causes safe_cast<uint8_t> to fail in 50% of
all possible values (i.e. 0x80-0xff).

This patch performs the correct cast sequence
for obtaining unsigned values from the Immed class
by performing a safe_cast<int8_t> and casting the
resulting 8-bit value then to uint8_t.

### DIFF
--- a/hphp/util/immed.h
+++ b/hphp/util/immed.h
@@ -80,8 +80,9 @@ struct Immed {
   int64_t q() const { return m_int; }
   int32_t l() const { return safe_cast<int32_t>(m_int); }
   int16_t w() const { return safe_cast<int16_t>(m_int); }
+  uint16_t uw() const { return safe_cast<int16_t>(m_int); }
   int8_t  b() const { return safe_cast<int8_t>(m_int); }
-  uint8_t ub() const { return safe_cast<uint8_t>(m_int); }
+  uint8_t ub() const { return safe_cast<int8_t>(m_int); }
 
   bool fits(int sz) const { return deltaFits(m_int, sz); }
 


cmuellner:
@mxw please review.

mxw:
Can we just give Immed more constructors instead?  If I understand correctly, the issue here is that if I construct an `Immed(uint8_t{0xff})`, the Immed will actually store `0xfffffff`, and then when I try to retrieve my `ub()` value, it will fail.  We should probably just fix the representation.

mxw:
We've been chatting a bit about this internally, and there's some confusion on what the problem is.  Is the issue that `Immed(uint8_t{0xff}) == Immed(0xffffffff)`?  If so, that's surprising, and seems like a bug at the constructor.  If the issue is that `Immed(int8_t{0xff}) == Immed(0xffffffff)`, that's as expected, and it makes sense for `ub()` to fail in that cast, because you can't represent `-1` as an unsigned value of any width.

cmuellner:
I was inspired from the existing code in `cgCheckRefs()`:

`v << andbi{(int8_t)mask64, truncBits, maskedBits, v.makeReg()};`

If mask64 is 0xff, then immed will contain -1 (when obtained with l(), w() or b()).
For me that totally ok.

However when you call now ub(), then in debug mode the safe_cast<> will trigger
a runtime exception, because -1 cannot be cast to an unsigned value.

markw65:
I think thats just because all immediates on x64 are signed; so we're intentionally forcing it to be a signed 8 bit value to match what the assembler expects (but it really doesn't matter, because we only care about the 8 bit result, so signedness is irrelevant).

cmuellner:
Here is the code to reproduce:

#include <stdio.h>
#include <string>
#include <assert.h>

#include <boost/numeric/conversion/cast.hpp>

static constexpr int debug = 1;

void safe_cast_failure(const char* fn,
                       const char* what) {
        printf("%s: %s\n", fn, what);
        assert(false);
}

template<typename To, typename From>
To always_safe_cast(From val) {
  try {
    return boost::numeric_cast<To>(val);
  } catch (std::bad_cast& bce) {
    safe_cast_failure(
      // The function is always safe_cast, but will indicate which
      // types were involved.
      __PRETTY_FUNCTION__,
      bce.what()
    );
  }
}

template<typename To, typename From>
To safe_cast(From val) {
  return debug ? always_safe_cast<To>(val) : static_cast<To>(val);
}


struct Immed {
        /* implicit */ Immed(int i) : m_int(i) {}
        /* implicit */ Immed(unsigned i) = delete;
        /* implicit */ Immed(long i) = delete;
        /* implicit */ Immed(unsigned long i) = delete;
        /* implicit */ Immed(long long i) = delete;
        /* implicit */ Immed(unsigned long long i) = delete;

        uint8_t ub() const { return safe_cast<uint8_t>(m_int); }

        private:
        int32_t m_int;
};

edwinsmith:
those 8-bit masking ops in cgCheckRefs() should probably be (uint8_t) casts.  the masks are all uint64_t to start with and the 8-bit versions are just x64 micro-optimizations

markw65:
Making them uint8_t will break x64 (as will this pull request), because asm-64 uses safe_cast<int8_t> on all 8 bit immediates.

markw65:
Actually, I suppose this pull request doesn't break x64, because for the most part it doesn't use ub() or uw().

cmuellner:
@markw65 how will this PR break x86?

markw65:
I take it back. It *will* break x64, if there's an ldimmb with a value >= 128... assuming its possible to get one of those.

cmuellner:
I'm not fully familiar with the x64 backend, but wouldn't it help to remove the static_cast<> from
`void Vgen::emit(const ldimmb& i)` in vasm-x64.cpp along with this PR?

markw65:
I don't see how - we'll already have aborted on the "auto val = i.s.ub()" (now *why* ldimmb takes a uint8_t, while all the other 8 bit immediates are int8_t, is not clear to me - but lots of code would need changing to fix that).

markw65:
Can you explain exactly what it is you're trying to solve?

cmuellner:
I want an unsigned 8-bit value from the Immed class. I can achieve that with `(uint8_t)imm.b()`, but I don't like the ugly C-cast there.

I pass the uint8_t value later to vixl (arm assembler), which has an API like:
`void Mov(const Register& rd, uint64_t imm);`.

When e.g. the vasm `ldimmb{(int8_t)0xff, dest}` is created, then I would like to pass `imm.ub()` as `imm` to the vixl API without any casts.

cmuellner:
Also note, that the current Immed implementation behaves different on debug and release builds (in the release builds the safe_casts are just C-casts, i.e. what my PR enforces for debug builds as well).

You can simulate this with the example code above, by setting debug to 0.

cmuellner:
Any new thoughts/ideas about this?
Can this land if it does not show regressions on x86_64 builds (release/debug) when testing with the all test set (./hphp/hhvm/hhvm hphp/test/run all)?

mxw:
@cmuellner—I chatted with @markw65 and he convinced me that we should retain the current behavior.  If you create an `Immed(int8_t{0xff})`, like we're (sometimes) doing from `cgCheckRefs()`, then you should not be able to get an unsigned byte back from it, because an unsigned value can't represent `-1`, which is what `int8_t{0xff}` is.

For `cgCheckRefs()` in particular you could theoretically just use `uint8_t` instead of `int8_t`—but the problem with that is that all vasm operations (except for `ldimmb{}` and `ldimmq{}`, for what I suspect is no particular reason) expect signed immediates (which I didn't realize until recently; it's not documented in the vasm layer at all, but it's implicitly required by how we're using vasm with the X64Assembler).

So the right solution to your problem here is, I think, to just do `static_cast<uint8_t>(imm.b())` when you need to take a signed 8-bit Immed (which is what most Immeds are) and zero-extend it for use with a W register.  You could add a short helper in `vasm-arm.cpp` or wherever else to avoid the verbosity.

Let me know if that was confusing or unclear, and I'll try to articulate better.

cmuellner:
I saw, that more or less all instantiation sites of the Immed class are using casts to intN_t. But I was not aware that ldimm{} expects the instantiation site to use uintN_t (which does not hold in all cases - see below). This information makes me understand your point. Proposal 1: what about unifying the behaviour (i.e. assume signed immediates for all vasm instructions including `ldimmN{}`)? Note, that in `vasm-xls.cpp` the Immed class gets instantiated after a cast to `int32_t` for `ldimml` and after a cast to `uint8_t` for `ldimmb`:

```
hphp/runtime/vm/jit/vasm-xls.cpp:            return ldimmb{uint8_t(ivl->var->val.val), dst};
hphp/runtime/vm/jit/vasm-xls.cpp:            return ldimml{int32_t(ivl->var->val.val), dst};
```

My initial motivation for this PR was to *not* have the need for a cast-wrapper function in vasm-arm.cpp, but rather have a flexible infrastructure code, which fits for all. Proposal 2 (in case proposal 1 is not accepted): can we get a method `uint8_t Immed::bAsUb()` into the Immed class, which does a `safe_cast` to `int8_t` and a `static_cast` to `uint8_t`?

I think that both proposals will be nack'ed, therefore I'll implement a function `uint8_t bAsUb(int8_t sb)` in vasm-arm.cpp, which will do the static cast. I'll add this to #7811.

mxw:
Like I said, only `ldimmb{}` and `ldimmq{}` in particular take signed immediates, not the others.  I'm not clear as to why the behavior is different, but it is.  That said, I can't see anything that depends on the unsignedness, so it should probably just be changed, which will give us the blanket assertion that all vasm instructions take signed immediates, but I don't think this affects what you're trying to do here much.

I don't think this would fix your problem though, right?  If I understand correctly, you're trying to extract a uint8_t from an Immed that _is_ supposed to be signed (one that comes from an `andbi{}`), and more generally, you want to zero-extend immediates, not sign-extend them—so you'll still need a helper that does the conversion correctly.

Given that this helper is only used for ARM, I don't see a particular need for it to be a member function on the Immed class.  Not all operations that affect a given struct need to be members of that struct; minimalism there is better.

PR #7808, opened 2017-04-25
vixl: Backport 32-bit immediate handling.

cmuellner:
When using negative signed values as immediate for logical or
move operations (e.g. And, Eor, etc.) then the conversion to
uint64_t (argument type of the immedate defined by the vixl API)
might end up with the upper 32 bits set.

This is problematic when using W-registers (32 bit only).

The current arm64 code solves the problem by using the
MSKTOP() function, which seems to be derived from
upstream vixl. This patch puts the code to HHVM's vixl,
which allows to remove MSKTOP() in vasm-arm.cpp.

### DIFF
--- a/hphp/vixl/a64/constants-a64.h
+++ b/hphp/vixl/a64/constants-a64.h
@@ -38,6 +38,8 @@ constexpr int kFirstCalleeSavedRegisterIndex = 19;
 constexpr int kNumberOfCalleeSavedFPRegisters = 8;
 constexpr int kFirstCalleeSavedFPRegisterIndex = 8;
 
+const int kWRegSizeInBits = 32;
+
 constexpr byte kSimulatorStackJunk = 0x8b;
 
 #define REGISTER_CODE_LIST(R)                                                  \
--- a/hphp/vixl/a64/macro-assembler-a64.cc
+++ b/hphp/vixl/a64/macro-assembler-a64.cc
@@ -91,16 +91,22 @@ void MacroAssembler::LogicalMacro(const Register& rd,
   if (operand.IsImmediate()) {
     int64_t immediate = operand.immediate();
     unsigned reg_size = rd.size();
-    assert(rd.Is64Bits() || is_uint32(immediate));
 
     // If the operation is NOT, invert the operation and immediate.
     if ((op & NOT) == NOT) {
       op = static_cast<LogicalOp>(op & ~NOT);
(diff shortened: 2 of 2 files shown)


cmuellner:
@mxw please review.

cmuellner:
I just noticed, that this give the following warning in non-debug (i.e. release) builds:

```
/mnt/cmuellner/hhvm/hphp/vixl/a64/macro-assembler-a64.cc: In member function ‘void vixl::MacroAssembler::Mov(const vixl::Register&, uint64_t)’:
/mnt/cmuellner/hhvm/hphp/vixl/a64/macro-assembler-a64.cc:235:13: warning: unused variable ‘i64’ [-Wunused-variable]
     int64_t i64 = (int64_t)imm;
```

Will be fixed soon.

mxw:
I have no problem with this diff, but I also don't fully understand it.  Why do we need to leave the top bits alone?  Is it because W registers are actually u32, and don't have any higher bits?  Or is it an optimization?

cmuellner:
In short: this patch allows to eliminate the MSKTOP() function in vasm-arm.cpp.

Long background:
The Immed struct (hphp/util/immed.h) stores values as int32_t's. When obtaining them via Immed::l() we get a int32_t. The vixl macro assembler API wants a uint64_t. when we put the value `0xffffffff` into the immediate, then Immed::l() returns `-1`. A (implicit) cast to uint64_t results in `0xffffffffffffffff`.

On aarch64 we cannot encode all immediates into the instruction. When this is not possible, the immediates need to be loaded into a register first (using movz/movk). When we want to emit e.g. the vasm orli{}, this might lead to an immediate being loaded into a register first. On aarch64 typical instructions are either for 32-bit registers or 64-bit registers. However, vixl does not check if the uint64_t immediate is a 32-bit value for 32-bit register operations and might build up 64-bit values in registers for 32-bit operations.

mxw:
I see—so if the 32-bit immediate implicitly casts to a 64-bit immediate with 0's in the upper bits, vixl will do the right thing.  But if it casts to, e.g. 0xfffffffffffffff, per your example, vixl will assume the full 64 bits are needed even if it's an input to only 32-bit instructions, and will do something inefficient.  Is that correct?

If so, this change makes sense, and looks all set to land.

cmuellner:
Yes exactly.

PR #7809, opened 2017-04-25
Implement StringData::hashHelper in AArch64 assembly

jim-saxman:
StringData::hashHelper() is implmented in x64 assembly and used
when NO_M_DATA is set, which in turn is controlled by USE_LOWPTR.
This patch provides an equivalent version for Aarch64.

There are no new errors in the unit tests, which were tested with
{-m interp, -m pgo, -r, and empty} and OSS Performance suite still
runs fine, both with and without USE_LOWPTR and ENABLE_AARCH64_CRC
defined.

### DIFF
--- a/hphp/runtime/base/string-data-arm.S
+++ b/hphp/runtime/base/string-data-arm.S
@@ -1,0 +1,52 @@
+#include "hphp/runtime/base/string-data-macros.h"
+#include "hphp/util/etch-helpers.h"
+#include "hphp/util/hphp-config.h"
+
+#if defined(ENABLE_AARCH64_CRC) && !defined(NO_HWCRC) && defined(NO_M_DATA) && \
+    !defined(_MSC_VER)
+
+        .file     "hphp/runtime/base/string-data-arm.S"
+
+/*
+ * HPHP::StringData::hashHelper() is defined here only when ENABLE_AARCH64_CRC
+ * is explicitly specified, and NO_M_DATA is used in StringData.
+ */
+
+ETCH_SECTION(_ZNK4HPHP10StringData10hashHelperEv)
+.globl    _ZNK4HPHP10StringData10hashHelperEv
+ETCH_TYPE(_ZNK4HPHP10StringData10hashHelperEv, @function)
+ETCH_ALIGN16
+_ZNK4HPHP10StringData10hashHelperEv:
+        CFI(startproc)
+        ldr     w3, [x0, #SD_LEN]
+        add     x13, x0, #SD_HASH
+        mov     w9, #-1
+        ldr     w10, [x13]
+        cbz     x3, ETCH_LABEL(hend)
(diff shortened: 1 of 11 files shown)


jim-saxman:
@mxw @binliu19 Please review. I'd like to know if we should ever USE_LOWPTR for AArch64. If not, than this patch is irrelevant and I'll quit working on MixedArray::NvGetStr().

hphp/runtime/base/string-data-arm.S
```diff
@@ -0,0 +1,52 @@
...
+        mov     x11, #0xdfdfdfdfdfdfdfdf
+        b       ETCH_LABEL(hheader)
+
+ETCH_LABEL(hloop):
+        crc32cx w9, w9, x1
+ETCH_LABEL(hheader):
+        ldr     x12, [x2], #8
+        adds    x3, x3, #8
+        and     x1, x11, x12
+        bcc     ETCH_LABEL(hloop)
```
apinski-cavium: Can you swap around the adds and and here?  Some processors will fuse the adds and b.cc together.

edwinsmith:
I don't know if @mxw has mentioned this, but it's not required for the ARM and x64 hash functions to be bitwise compatible. Only that they both do the case-insensitive masking.  Handwriting those two functions in x64 asm was a small win (<1% if I recall) but was very dependent on oddities of how gcc compiled c++ to x64; @binliu19 was avoiding spills which may not happen on ARM.

If you can do just as well with arm-specific C++, no need to go the asm route.

hphp/runtime/base/string-data-arm.S
jim-saxman: Sure.

binliu19:
The code looks fine to me, but I don't know ARM assembly well enough to really give meaningful comments.  It looks like a straightforward translation from the x64 version.  So if tests are OK I think it is fine.

Since we already have asm version of the hash_string_i(), implementing hashHelper() using asm is a very small win.
I tried to inline hashHelper() into MixedArray::NvGetStr() and avoid spilling, and it was worth about 0.7%.

jim-saxman:
@edwinsmith @binliu19 Yes, I noticed these functions were near the top of Mediawiki's profile, so I assumed that assembly would be a bigger improvement that what they turned out to be. Now that I've been through the x64 to aarch64 translation of MixedArray::NvGetStr() enough to understand it - and get a working version - I can see that @binliu19's goal was to inline the hashHelper.

I'm going to spend a little bit more time to get an optimized version of NvGetStr().

jim-saxman:
@edwinsmith @binliu19 Thanks for the heads-up on NvGetStr(). The translation from x64 to aarch64 was pretty easy, but after applying a handful of tiny optimizations, I could only get the scores up to the baseline. I won't submit a patch for NvGetStr().

PR #7810, opened 2017-04-26
jit: irlower: Mask type with kDataTypeMask before comparing.

cmuellner:
When using 64-bit registers in cmpbi{} vasm instructions,
a preceding truncation to 8-bits is required in general.
This rule seems to be violated in some parts of the irlower
code. With the help of the arm64 backend, these locations
could be detected.

This patch uses masking with KindOfRefCountThreshold (0x7f)
to address this issue.

No regressions with the 'all' test set on arm64.

Note, that this patch is required in order to allow
a performance improvement patch for arm64 (drop emitting
uxtb instructions before cmpbi{} and cmpb{}).

### DIFF
--- a/hphp/runtime/vm/jit/irlower-internal-inl.h
+++ b/hphp/runtime/vm/jit/irlower-internal-inl.h
@@ -189,15 +189,14 @@ void emitTypeTest(Vout& v, IRLS& env, Type type,
 
   auto const cc = [&] {
 
-    auto const mask_cmp = [&] (int mask, int bits, ConditionCode cc) {
+    auto const mask_cmp = [&] (int mask, int kind, ConditionCode cc) {
       auto const masked = emitMaskTVType(v, mask, typeSrc);
-      emitCmpTVType(v, sf, bits, masked);
+      emitCmpTVType(v, sf, kind, masked);
       return cc;
     };
 
-    auto const cmp = [&] (DataType kind, ConditionCode cc) {
-      emitCmpTVType(v, sf, kind, typeSrc);
-      return cc;
+    auto const cmp = [&] (int kind, ConditionCode cc) {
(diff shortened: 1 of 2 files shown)


cmuellner:
@mxw please review.
Note also, that this is a correct and refined version of
https://github.com/facebook/hhvm/pull/7797/commits/03125aea12f3e5c8fa84e14f10ffb564b1f95630

cmuellner:
Here is a short summary of the underlying issue:

The only correct way to create Vreg8's is to use vasm instructions,
which will emit code via vasm-arm.cpp. All other (incorrect) ways
to create Vreg8's involve casting of a non-Vreg8 to a Vreg8.
Such casts are violations of the type system.

If all creations of Vreg8's are correct (as defined above), then we
automatically get the guarantee to have zeros in the upper bits,
because vasm-arm.cpp does not emit code, where a Vreg8 has other
bits than 7..0 set.

Now we are running into the problem, that HHVM contains code,
which creates Vreg8's from registers, which have other bits
than 7..0 set. This can only be generated by casting a
non-Vreg8 to a Vreg8 without truncation, which is always a
violation of the type system. It is quite hard to find the
actual violation, since more or less all functions use the
type Vreg in the function signature. So the functions might
be correct for one caller (having a correct Vreg8), but
incorrect for another one (not having a correct Vreg8).

For x64, this is just a cosmetic issue, because there exists
a 8-bit register access (implicit truncation on access).
For arm64 such a mechanism does not exist. We either have to
explicitly truncate using uxtb instructions (expensive) or
we get a guarantee that a Vreg8 is a register, which has
only bits 7..0 set (cheap).

mxw:
@cmuellner—I think we want to close this without landing, per our email thread; does that sound right?

cmuellner:
Agreed.

PR #7820, opened 2017-05-01
AARCH64 - movzbl/shrli -> ubfmr2m6 peephole

swalk-cavium:
This creates an ARMv8 specific Vinstr to represent a commonly used form of the
ubfm/ubfx instruction and the peephole to detect it.  This pattern was seen
in several of the regression tests.  It is generated during the lowering of the callm
instruction.

Before
=====
              0x434009c4  53001c21              uxtb w1, w1
              0x434009c8  53027c21              lsr w1, w1, #2
              0x434009cc  d37df021              lsl x1, x1, #3
              0x434009d0  d297c302              movz x2, #0xbe18 
              0x434009d4  f2a1c322              movk x2, #0xe19, lsl #16 
              0x434009d8  f8626821              ldr x1, [x1, x2]
              0x434009dc  d63f0020              blr x1  

After
=====
              0x40400994  53021c21              ubfx w1, w1, #2, #6
              0x40400998  d37df021              lsl x1, x1, #3
              0x4040099c  d2965e02              movz x2, #0xb2f0 
              0x404009a0  f2a081a2              movk x2, #0x40d, lsl #16 
              0x404009a4  f8626821              ldr x1, [x1, x2]
              0x404009a8  d63f0020              blr x1  

This was seen 169 times in hphp/test/quick/all_type_comparison_test.php and
230 times in hphp/test/zend/good/ext/intl/tests/grapheme.php.

The standard regression tests were run with 6 option sets.  No new regressions seen.

### DIFF
--- a/hphp/runtime/vm/jit/vasm-arm.cpp
+++ b/hphp/runtime/vm/jit/vasm-arm.cpp
@@ -340,7 +340,7 @@ struct Vgen {
   void emit(const fcvtzs& i) { a->Fcvtzs(X(i.d), D(i.s)); }
   void emit(const mrs& i) { a->Mrs(X(i.r), vixl::SystemRegister(i.s.l())); }
   void emit(const msr& i) { a->Msr(vixl::SystemRegister(i.s.l()), X(i.r)); }
-  void emit(const ubfm2r7ml& i) { a->ubfm(W(i.d), W(i.s), 2, 7); }
+  void emit(const ubfmli& i) { a->ubfm(W(i.d), W(i.s), i.mr.w(), i.ms.w()); }
 
   void emit_nop() { a->Nop(); }
 
--- a/hphp/runtime/vm/jit/vasm-arm.cpp
+++ b/hphp/runtime/vm/jit/vasm-arm.cpp
@@ -340,6 +340,7 @@ struct Vgen {
   void emit(const fcvtzs& i) { a->Fcvtzs(X(i.d), D(i.s)); }
   void emit(const mrs& i) { a->Mrs(X(i.r), vixl::SystemRegister(i.s.l())); }
   void emit(const msr& i) { a->Msr(vixl::SystemRegister(i.s.l()), X(i.r)); }
+  void emit(const ubfm2r7ml& i) { a->ubfm(W(i.d), W(i.s), 2, 7); }
 
   void emit_nop() { a->Nop(); }
 
--- a/hphp/runtime/vm/jit/vasm-simplify.cpp
+++ b/hphp/runtime/vm/jit/vasm-simplify.cpp
@@ -321,25 +321,6 @@ bool simplify(Env& env, const pop& inst, Vlabel b, size_t i) {
 
 ///////////////////////////////////////////////////////////////////////////////
 
-bool simplify(Env& env, const movzbl& inst, Vlabel b, size_t i) {
-  // movzbl{s, d}; shrli{2, s, d} --> ubfmli{2, 7, s, d}
-  return if_inst<Vinstr::shrli>(env, b, i + 1, [&](const shrli& sh) {
-    if (!(arch() == Arch::ARM &&
-      sh.s0.l() == 2 &&
-      env.use_counts[inst.d] == 1 &&
-      env.use_counts[sh.sf] == 0 &&
-      inst.d == sh.s1)) return false;
-
-    return simplify_impl(env, b, i, [&] (Vout& v) {
-      v << copy{inst.s, inst.d};
-      v << ubfmli{2, 7, inst.d, sh.d};
-      return 2;
-    });
-  });
-}
-
-///////////////////////////////////////////////////////////////////////////////
-
 /*
  * Perform peephole simplification at instruction `i' of block `b'.
  *
--- a/hphp/runtime/vm/jit/vasm-simplify.cpp
+++ b/hphp/runtime/vm/jit/vasm-simplify.cpp
@@ -327,12 +327,13 @@ bool simplify(Env& env, const movzbl& inst, Vlabel b, size_t i) {
     if (!(arch() == Arch::ARM &&
       sh.s0.l() == 2 &&
       env.use_counts[inst.d] == 1 &&
+      env.use_counts[sh.sf] == 0 &&
       inst.d == sh.s1)) return false;
 
-  return simplify_impl(env, b, i, [&] (Vout& v) {
-    v << copy{inst.s, inst.d};
-    v << ubfmli{2, 7, inst.d, sh.d, sh.sf, sh.fl};
-    return 2;
+    return simplify_impl(env, b, i, [&] (Vout& v) {
+      v << copy{inst.s, inst.d};
+      v << ubfmli{2, 7, inst.d, sh.d};
+      return 2;
     });
   });
 }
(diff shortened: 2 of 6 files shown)


swalk-cavium:
@mxw - Max, can you review this one?  Thanks.

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -343,6 +343,7 @@ struct Vgen {
   void emit(const fcvtzs& i) { a->Fcvtzs(X(i.d), D(i.s)); }
   void emit(const mrs& i) { a->Mrs(X(i.r), vixl::SystemRegister(i.s.l())); }
   void emit(const msr& i) { a->Msr(vixl::SystemRegister(i.s.l()), X(i.r)); }
+  void emit(const ubfm2r7ml& i) { a->ubfm(W(i.d), W(i.s), 2, 7); }
```
cmuellner: I would prefer a generic ubfm{immr, imms} instruction over a specialised one.

hphp/runtime/vm/jit/vasm-simplify.cpp
```diff
@@ -524,6 +524,24 @@ bool simplify(Env& env, const pop& inst, Vlabel b, size_t i) {
 
 ///////////////////////////////////////////////////////////////////////////////
 
+bool simplify(Env& env, const movzbl& inst, Vlabel b, size_t i) {
+  // movzbl{s, d}; shrli{2, s, d} --> ubfm2r7ml{s, d}
```
cmuellner: Would be nice if "movzXY{s, tmp} ; shrYi{N, tmp, d} -> ubfm{r, n, s, d}".

hphp/runtime/vm/jit/vasm-simplify.cpp
```diff
@@ -524,6 +524,24 @@ bool simplify(Env& env, const pop& inst, Vlabel b, size_t i) {
...
+  return if_inst<Vinstr::shrli>(env, b, i + 1, [&](const shrli& sh) {
+    if (!(arch() == Arch::ARM &&
+      sh.s0.l() == 2 &&
+      env.use_counts[inst.d] == 1 &&
+      inst.d == sh.s1)) return false;
+
+  return simplify_impl(env, b, i, [&] (Vout& v) {
+    v << copy{inst.s, inst.d};
```
cmuellner: why copying first and not `ubfm{inst.s, sh.d}`?

swalk-cavium: Matching up the Vreg types.  These are cleaned up later.

mxw: What do you mean by "matching up the Vreg types" here?  Both of these are Vreg32...

swalk-cavium:
@mxw - Wouldn't adding immr/imms fields require adding more Immediates to the new intrinsic?
That would be the first Vinstr with multiple immediates.  If I added a different type, I'd have to update the visitor machinery.  Right?

hphp/runtime/vm/jit/vasm-simplify.cpp
swalk-cavium: struct movzbl { Vreg8 s; Vreg32 d; };
struct shrli { Immed s0; Vreg32 s1, d; VregSF sf; Vflags fl; };

The source in the move is Vreg8, the destination in the shrli is a Vreg32,
The copy uses Vreg.  Isn't that the correct way to match up the types?
Or do I not have to worry about that.

cmuellner:
@swalk-cavium you could use a helper function to calculate the values for immr and imms. I think vixl has such a function.

hphp/runtime/vm/jit/vasm-simplify.cpp
cmuellner: A common solution for that is to use `Vreg`, but I don't know if that is appropriate here.

mxw:
@swalk-cavium—I can't think of what would make vasm-visit fail on instructions with multiple immediate operands.  If it does, then yes, you'll have to fix it... but if it already supports it, then you don't.

hphp/runtime/vm/jit/vasm-simplify.cpp
mxw: Oh, I see.  No, `copy{}` is not the right way to do that—the only way is `movzbl{}`.  You'll have to have your new instruction take a Vreg8 src and a Vreg32 dst operand if that's your intention.

swalk-cavium:
@cmuellner - I'm not sure which vixl helper you're referring to.  Can you give me a reference?
We're crossing a level of abstraction and I chose to include information in the name.  Apparently
that's not the right way.

hphp/runtime/vm/jit/vasm-simplify.cpp
cmuellner: In this case we could make ubfmbi{} ubfmwi{} ubfmli{} ubfmqi{}.

cmuellner:
@swalk-cavium I just checked. There is indeed no helper. But I think it is not needed. `immr = sh.s0.l()` and imms depends on the truncation instruction.

edwinsmith:
Many instructions already have 2+ immediates. As long as you have I(whatever) for each one in the instruction's visitor macro, should be fine.

hphp/runtime/vm/jit/vasm-simplify.cpp
mxw: Agreed; that sounds good to me.

swalk-cavium:
@mxw , @cmuellner - Updated per comments above.  Regression tests with 6 option sets run.  No new issues seen.

hphp/runtime/vm/jit/vasm-simplify.cpp
```diff
@@ -524,6 +524,24 @@ bool simplify(Env& env, const pop& inst, Vlabel b, size_t i) {
...
+  // movzbl{s, d}; shrli{2, s, d} --> ubfmli{2, 7, s, d}
+  return if_inst<Vinstr::shrli>(env, b, i + 1, [&](const shrli& sh) {
+    if (!(arch() == Arch::ARM &&
+      sh.s0.l() == 2 &&
+      env.use_counts[inst.d] == 1 &&
+      inst.d == sh.s1)) return false;
+
+  return simplify_impl(env, b, i, [&] (Vout& v) {
+    v << copy{inst.s, inst.d};
+    v << ubfmli{2, 7, inst.d, sh.d, sh.sf, sh.fl};
```
mxw: Does this have the same effect on the status flags as the `shrli{}`?

hphp/runtime/vm/jit/vasm-simplify.cpp
```diff
@@ -524,6 +524,24 @@ bool simplify(Env& env, const pop& inst, Vlabel b, size_t i) {
...
 
+bool simplify(Env& env, const movzbl& inst, Vlabel b, size_t i) {
+  // movzbl{s, d}; shrli{2, s, d} --> ubfmli{2, 7, s, d}
+  return if_inst<Vinstr::shrli>(env, b, i + 1, [&](const shrli& sh) {
+    if (!(arch() == Arch::ARM &&
+      sh.s0.l() == 2 &&
+      env.use_counts[inst.d] == 1 &&
+      inst.d == sh.s1)) return false;
+
+  return simplify_impl(env, b, i, [&] (Vout& v) {
```
mxw: Please fix the indentation here and below.

hphp/runtime/vm/jit/vasm-simplify.cpp
swalk-cavium: @mxw - That's a good question.  I'm not seeing anywhere that the flag is being set/used
by either of the shrli/shrqi instructions.  The underlying ARM instruction doesn't touch any
flags.  I'll check further, but it's probably safe to remove from ubfmli.

Issue #56, opened 2010-02-28
date() function requires explicit cast


serphen:
Hi

<?php echo date('m/Y', mktime(0, 0, 0, (date('m') - 1), date('d'), date('Y'))); ?>
works on Zend Engine, but not on HipHop (because date() is a string), it returns a BadType error.

However it is supposed to be automatically casted like this:

<?php echo date('m/Y', mktime(0, 0, 0, (intval(date('m')) - 1), date('d'), date('Y')));?>

PR #7823, opened 2017-05-02
Aarch64 peephole using csinc

swalk-cavium:
This change introduces an Aarch64 specific pseudo instruction which
is a variation of the conditional move.  One pattern seen in the OOPSLA'14
benchmark tests involved the expression  like this in C

    (cond) ? exp : 1;

The hard-coded zero register can be used with the csinc instruction to
create the 1.

Before
======
     0x146010d0  39c0e361             ldrsb w1, [x27, #56]
     0x146010d4  72001c3f             tst w1, #0xff
     0x146010d8  52800022             movz w2, #0x1   //<<---
     0x146010dc  1a810041             csel w1, w2, w1, eq

After
=====
      0x362011d0  38e16b61              ldrsb w1, [x27, x1] 
      0x362011d4  72001c3f              tst w1, #0xff
      0x362011d8  1a9f1421              csinc w1, w1, wzr, ne  

This saves 1 instruction per instance.

The standard regression suite was run with 6 option sets.  No new regressions
were introduced.

### DIFF
--- a/hphp/runtime/vm/jit/vasm-instr.h
+++ b/hphp/runtime/vm/jit/vasm-instr.h
@@ -295,6 +295,10 @@ struct Vunit;
   O(sarq, I(fl), UH(s,d), DH(d,s) D(sf))\
   O(shlq, I(fl), UH(s,d), DH(d,s) D(sf))\
   /* arm instructions */\
+  O(csoneb, I(cc), U(sf) U(t), D(d))\
+  O(csonew, I(cc), U(sf) U(t), D(d))\
+  O(csonel, I(cc), U(sf) U(t), D(d))\
+  O(csoneq, I(cc), U(sf) U(t), D(d))\
   O(fcvtzs, Inone, U(s), D(d))\
   O(mrs, I(s), Un, D(r))\
   O(msr, I(s), U(r), Dn)\
@@ -1105,6 +1109,10 @@ struct shlq { Vreg64 s, d; VregSF sf; Vflags fl; }; // uses rcx
 /*
  * arm intrinsics.
  */
+struct csoneb { ConditionCode cc; VregSF sf; Vreg8 t, d; };
+struct csonew { ConditionCode cc; VregSF sf; Vreg16 t, d; };
+struct csonel { ConditionCode cc; VregSF sf; Vreg32 t, d; };
+struct csoneq { ConditionCode cc; VregSF sf; Vreg64 t, d; };
 struct fcvtzs { VregDbl s; Vreg64 d;};
 struct mrs { Immed s; Vreg64 r; };
 struct msr { Vreg64 r; Immed s; };
--- a/hphp/runtime/vm/jit/vasm-instr.h
+++ b/hphp/runtime/vm/jit/vasm-instr.h
@@ -295,10 +295,10 @@ struct Vunit;
   O(sarq, I(fl), UH(s,d), DH(d,s) D(sf))\
   O(shlq, I(fl), UH(s,d), DH(d,s) D(sf))\
   /* arm instructions */\
-  O(csoneb, I(cc), U(sf) U(t), D(d))\
-  O(csonew, I(cc), U(sf) U(t), D(d))\
-  O(csonel, I(cc), U(sf) U(t), D(d))\
-  O(csoneq, I(cc), U(sf) U(t), D(d))\
+  O(csincb, I(cc), U(sf) U(f) U(t), D(d))\
+  O(csincw, I(cc), U(sf) U(f) U(t), D(d))\
+  O(csincl, I(cc), U(sf) U(f) U(t), D(d))\
+  O(csincq, I(cc), U(sf) U(f) U(t), D(d))\
   O(fcvtzs, Inone, U(s), D(d))\
   O(mrs, I(s), Un, D(r))\
   O(msr, I(s), U(r), Dn)\
@@ -1109,10 +1109,10 @@ struct shlq { Vreg64 s, d; VregSF sf; Vflags fl; }; // uses rcx
 /*
  * arm intrinsics.
  */
-struct csoneb { ConditionCode cc; VregSF sf; Vreg8 t, d; };
-struct csonew { ConditionCode cc; VregSF sf; Vreg16 t, d; };
-struct csonel { ConditionCode cc; VregSF sf; Vreg32 t, d; };
-struct csoneq { ConditionCode cc; VregSF sf; Vreg64 t, d; };
+struct csincb { ConditionCode cc; VregSF sf; Vreg8 f, t, d; };
+struct csincw { ConditionCode cc; VregSF sf; Vreg16 f, t, d; };
+struct csincl { ConditionCode cc; VregSF sf; Vreg32 f, t, d; };
+struct csincq { ConditionCode cc; VregSF sf; Vreg64 f, t, d; };
 struct fcvtzs { VregDbl s; Vreg64 d;};
 struct mrs { Immed s; Vreg64 r; };
 struct msr { Vreg64 r; Immed s; };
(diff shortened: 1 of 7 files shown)


swalk-cavium:
@mxw - Hi Max, Can you review this one?

hphp/runtime/vm/jit/vasm-instr.h
```diff
@@ -1101,6 +1105,10 @@ struct shlq { Vreg64 s, d; VregSF sf; Vflags fl; }; // uses rcx
 /*
  * arm intrinsics.
  */
+struct csoneb { ConditionCode cc; VregSF sf; Vreg8 t, d; };
```
cmuellner: Why not a real csinc{} and use `PhysReg(vixl::zr)` as the second source register?
`Conditional Select Increment returns, in the destination register, the value of the first source register if the condition is TRUE, and otherwise returns the value of the second source register incremented by 1.`

hphp/vixl/a64/macro-assembler-a64.h
```diff
@@ -422,7 +421,6 @@ class MacroAssembler : public Assembler {
     assert(allow_macro_instructions_);
     assert(!rd.IsZero());
     assert(!rn.IsZero());
```
cmuellner: You might want to remove the assertion `assert(!rn.IsZero());` as well.
Upstream vixl has this code:

```
void MacroAssembler::Csinc(const Register& rd,
                           const Register& rn,
                           const Register& rm,
                           Condition cond) {
  DCHECK(allow_macro_instructions_);
  DCHECK(!rd.IsZero());
  DCHECK((cond != al) && (cond != nv));
  csinc(rd, rn, rm, cond);
}
```

hphp/runtime/vm/jit/vasm-instr.h
swalk-cavium: I thought we couldn't have PhysRegs in the vasm-instrs.  

The peephole flips the condition code if necessary and only creates the
pseudo if semantically correct.

cmuellner: We are doing the PhysReg() trick in the lowering of storeNi{} in vasm-arm.cpp.

cmuellner: Also vasm-fold-imms.cpp is full of that. E.g.:
```
  void fold(storeb& in, Vinstr& out) {
    if (zero_imm(in.s)) out = storeb{PhysReg(vixl::wzr), in.m};
  }
```

swalk-cavium: In vasm-fold-imms.cpp they're in a 'namespace arm'.  The other PhysRegs are either in
target specific or register specific files.   The only target specific code in vasm-simplify are the tests against arch().  There are Vinstrs labeled as intrinsics, so I thought it safe to
create one of those.  You have to pick something which avoids using the reg in the initial
move otherwise it won't be eliminated as dead code.  The pseudo instruction seemed like the right place to emit the xzr.

mxw:
Whoops, sorry for the tag spam; I had the wrong PR.

hphp/runtime/vm/jit/vasm-instr.h
mxw: I just split out the ARM-specific simplifications into a separate file (in 16710e2), so you can move all your new simplifications there and just define `csinc{}`.

hphp/runtime/vm/jit/vasm-simplify.cpp
```diff
@@ -196,6 +196,37 @@ bool simplify(Env& env, const setcc& vsetcc, Vlabel b, size_t i) {
   });
 }
 
+template <typename Reg>
+bool operand_one(Env& env, Reg op) {
```
mxw: Why did you template this function?  Why can't you just use `Vreg`?

hphp/runtime/vm/jit/vasm-simplify.cpp
```diff
@@ -196,6 +196,37 @@ bool simplify(Env& env, const setcc& vsetcc, Vlabel b, size_t i) {
...
+// Reduce use of immediate one possibly removing def as dead code.
+// Specific to ARM using hard-coded zero register.
+template <typename Out, typename Inst>
+bool cmov_fold_one(Env& env, const Inst& inst, Vlabel b, size_t i) {
+  if (arch() != Arch::ARM) return false;
+  if (operand_one(env, inst.f)) {
+    return simplify_impl(env, b, i, [&] (Vout& v) {
+      v << Out{inst.cc, inst.sf, inst.t, inst.d};
+      return 1;
+      });
```
mxw: nit: indentation

hphp/runtime/vm/jit/vasm-simplify.cpp
```diff
@@ -196,6 +196,37 @@ bool simplify(Env& env, const setcc& vsetcc, Vlabel b, size_t i) {
...
+    return simplify_impl(env, b, i, [&] (Vout& v) {
+      v << Out{inst.cc, inst.sf, inst.t, inst.d};
+      return 1;
+      }); 
+  }
+  if (operand_one(env, inst.t)) {
+    return simplify_impl(env, b, i, [&] (Vout& v) {
+      v << Out{ccNegate(inst.cc), inst.sf, inst.f, inst.d};
+      return 1;
+      });
```
mxw: nit: indentation

hphp/runtime/vm/jit/vasm-simplify.cpp
swalk-cavium: @mxw - I'm not sure I understand your comment about csinc{} above.  The csoneb still has
to go through the rest of the passes to get real registers.  I don't think you can make it opaque to the other passes.  I can move the actual peep into the new arm simplify file.

mxw: @swalk-cavium—What I mean is that you can do as @cmuellner suggested and implement `csinc*{}` instead of `csone*{}`—you can just specify the `vixl::wzr` operand explicitly.

swalk-cavium: @mxw - Ok, got it.  I thought you were suggesting something else.  Having a separate file will make the boundaries cleaner.

hphp/runtime/vm/jit/vasm-simplify-arm.cpp
```diff
@@ -33,6 +35,57 @@ bool simplify(Env&, const Inst& inst, Vlabel b, size_t i) { return false; }
...
+      v << Out{ccNegate(inst.cc), inst.sf, PhysReg(vixl::wzr), inst.f, inst.d};
+      return 1;
+    });
+  }
+  return false;
+}
+
+bool simplify(Env& env, const cmovb& inst, Vlabel b, size_t i) {
+  if (cmov_fold_one<csincb>(env, inst, b, i)) return true;
+  return false;
```
mxw: Change this to just `return cmov_fold_one<csincb>(env, inst, b, i);`, and likewise below.

hphp/runtime/vm/jit/vasm-simplify.cpp
```diff
@@ -120,10 +120,8 @@ bool simplify(Env& env, const setcc& vsetcc, Vlabel b, size_t i) {
   });
 }
 
-/*
- * Fold a cmov of a certain width into a copy if both values are the same
- * register or have the same known constant value.
- */
+// Fold a cmov of a certain width into a copy if both values are the same
+// register or have the same known constant value.
```
mxw: Can you undo this change?

hphp/runtime/vm/jit/vasm-simplify-arm.cpp
swalk-cavium: @mxw - Sure.  I was expecting other tests, but we can address them at the time.

swalk-cavium:
@max - Hi Max, Updated again.

PR #7824, opened 2017-05-04
test/slow/ext_shmop/ext_shmop.php is not a hack specific test

swalk-cavium:
This regression test is failing because the reference output doesn't match.
It cries not running the hack typechecker and suggests the option to add to
the command line.  There is nothing hack specfic in the test.


swalk-cavium:
@mxw - Hi Max, Can you review?  Thanks.

PR #7832, opened 2017-05-08
[Aarch64] Make stxr instruction predictable

jim-saxman:
The stxr instruction should not use register s == t. Using s == t
is constrained unpredictable behavior according the the ARMv8
specification.


jim-saxman:
@apinski-cavium @mxw Please review.

apinski-cavium:
Yes this looks ok.

Here is the psedu-code from the ARM ARMv8 document that says the instruction is constrainted to be unpredictable if rs == rt:
 if memop == MemOp_STORE && excl then
 if s == t || (pair && s == t2) then
 Constraint c = ConstrainUnpredictable();

PR #7833, opened 2017-05-08
jit: vasm-lower: Add ARM specific truncation for type values.

cmuellner:
When using 64-bit registers in cmpbi{} vasm instructions,
a preceding truncation to 8-bits is required in general.
This rule seems to be violated in some parts of the irlower
code. With the help of the arm64 backend, these locations
could be detected.

Note, that this patch is required in order to allow
a performance improvement patch for arm64 (drop emitting
uxtb instructions before cmpbi{} and cmpb{}).

### DIFF
--- a/hphp/runtime/vm/jit/vasm-lower.cpp
+++ b/hphp/runtime/vm/jit/vasm-lower.cpp
@@ -189,15 +189,19 @@ void lower_vcall(Vunit& unit, Inst& inst, Vlabel b, size_t i) {
       static_assert(offsetof(TypedValue, m_type) == 8, "");
 
       if (dests.size() == 2) {
-        if (!(arch() == Arch::ARM)) {
-          v << copy2{rret(0), rret(1), dests[0], dests[1]};
-        } else {
-          //For ARM64 we are truncating the aux data
-          //from the type. That allows to use the resulting
-          //register values to be used in type comparisons
-          //without the need for truncation there.
-          v << copy{rret(0), dests[0]};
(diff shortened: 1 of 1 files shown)


cmuellner:
This PR fixes the root cause of the problem mentioned in #7810.

No regressions with the 'all' test set on arm64.

@mxw please review.

hphp/runtime/vm/jit/vasm-lower.cpp
```diff
@@ -189,7 +189,16 @@ void lower_vcall(Vunit& unit, Inst& inst, Vlabel b, size_t i) {
       static_assert(offsetof(TypedValue, m_type) == 8, "");
 
       if (dests.size() == 2) {
-        v << copy2{rret(0), rret(1), dests[0], dests[1]};
+        if (!(arch() == Arch::ARM)) {
```
mxw: Could you make these switch on architecture instead?  I doubt we have plans to add any more soon, but if we ever do, it would be good to force auditing of this site.

hphp/runtime/vm/jit/vasm-lower.cpp
```diff
@@ -189,7 +189,16 @@ void lower_vcall(Vunit& unit, Inst& inst, Vlabel b, size_t i) {
...
+          v << copy2{rret(0), rret(1), dests[0], dests[1]};
+        } else {
+          //For ARM64 we are truncating the aux data
```
mxw: nit: `// Space after the slashes`, here and below.

PR #7849, opened 2017-05-21
aarch64: add emitter for cmovl

swalk-cavium:
This adds an emitter for the cmovl Vinstr.  Recent improvements to the
earlier part of the compiler exposed this missing case.  The commit which
exposed this was 

17cade0 Lower some CufIter operations to more specific instructions

This change returns the regression tests to previous baseline.


swalk-cavium:
@mxw - Hi Max, Can you review this one?

swalk-cavium:
PR#7852 is more complete.  This covered the immediate problem.

PR #7851, opened 2017-05-22
    jit: arm64: Fix regression in async testcases (repo auth)

cmuellner:
A recent PR (#7833/61b3b0d) addresses the problem, that the arm64
lowering/emitter code need to truncate the operands of the vasm
instructions cmpb{} and cmpbi{}. The actual problem is, that the
given values are not Vreg8 values, but treated as Vreg8 values
by follow-up code. The fix in the named PR is to insert a movtqb{}
whenever a type value is lowered.

As reported by @swalk-cavium this breaks some async tests when
running in repo authoritative mode.

This patch addresses this issue by clearing bits 9-31 of the
type value, but keep bits 32-63 to keep the AUX value in the
register (clearing that bits leads to the regression in repo
authoritative mode).
    
The patch relies on the fact that the type value will later
be interpreted as 32-bit (W) register, i.e. ignoring the AUX
value anyways. Therefore a cmpb{} and cmpbi{} will not see
the AUX value.

Running the 'all' test set showed no regressions with and without
authoritative mode (all previously failing async tests pass now).

Thanks @mxw for the idea for this PR.

### DIFF
--- a/hphp/runtime/vm/jit/vasm-lower.cpp
+++ b/hphp/runtime/vm/jit/vasm-lower.cpp
@@ -195,12 +195,12 @@ void lower_vcall(Vunit& unit, Inst& inst, Vlabel b, size_t i) {
             v << copy2{rret(0), rret(1), dests[0], dests[1]};
             break;
           case Arch::ARM:
-            // For ARM64 we are truncating the aux data
-            // from the type. That allows to use the resulting
-            // register values to be used in type comparisons
-            // without the need for truncation there.
+            // For ARM64 we need to clear the bits 8..31 from the type value.
+            // That allows to use the resulting register values to be used
+            // in type comparisons without the need for truncation there.
             v << copy{rret(0), dests[0]};
-            v << movtqb{rret(1), dests[1]};
+            v << andq{v.cns(0xffffffff000000ff),
+                      rret(1), dests[1], v.makeReg()};
             break;
         }
       } else {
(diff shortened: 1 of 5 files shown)


mxw:
Question: on ARM, does this new `andqi64{s0, s1, d, ...}` instruction compile down to the same instructions that `andq{v.cns(s0), s1, d, ...}` would?  If so, can we just do that instead?

If not, and it's necessary to create a separate instruction, can you declare it as ARM-specific?  This is an artifact of behavior on x64, but the canonical way to represent a 64-bit immediate in a Vinstr is to materialize it in a constant register and use the non-immediate form of the instruction.  On x64, e.g., `andqi64{s0, s1, d, ...}` and `andq{v.cns(s0), s1, d, ...}` _would_ be identical.

cmuellner:
Ok, I was not aware of cns()/makeConst(Vconst{}).
I've just started the tests and will push an update when I can confirm that there are no regressions.

The resulting instruction for aarch64 should be `and     x0, x0, #0xffffffff000000ff`.
Note, that the immediate can be embedded into the instruction, which keeps the
penalty low.

hphp/runtime/vm/jit/vasm-lower.cpp
```diff
@@ -195,12 +195,12 @@ void lower_vcall(Vunit& unit, Inst& inst, Vlabel b, size_t i) {
             v << copy2{rret(0), rret(1), dests[0], dests[1]};
             break;
           case Arch::ARM:
-            // For ARM64 we are truncating the aux data
-            // from the type. That allows to use the resulting
-            // register values to be used in type comparisons
-            // without the need for truncation there.
+            // For ARM64 we need to clear the bits 8..31 from the type value.
+            // That allows to use the resulting register values to be used
```
mxw: `s/That allows to use the resulting register values to be used/That allows us to use the resulting register values/`, and reflow—here and below.

Can you also mention that we leave bits 32..63 alone in order to preserve the m_aux member?

PR #7853, opened 2017-05-23
Removed projects that do no support HHVM anymore

fabpot:
I think it's better to remove projects that do not support HHVM anymore to avoid confusion. Anyway, tested versions are now very old and not maintained anymore, so I don't see the point in keeping them here.

In this PR, I've only removed my projects for which I'm the lead dev.

### DIFF
--- a/hphp/test/frameworks/frameworks.yaml
+++ b/hphp/test/frameworks/frameworks.yaml
@@ -418,33 +418,6 @@
     commit: 4d76935edae27bd9ed9e87f839efd5af4a8015d1
     install_root: stash
     test_root: stash
-  symfony:
-    url: https://github.com/symfony/symfony.git
-    branch: v2.4.8
-    commit: 3da923989cce41c149c6f20b7723c052ac251b16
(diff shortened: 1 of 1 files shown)


fredemmott:
HHVM testing a project does not imply the upsstreqm project supports it; if something says otherwise, that should be changed. This suite is used to check for regressions in HHVMs PHP5 support; state of upstream support or staleness of the version is not relevant to regression testing.

Removing them seems to have little benefit, and will be harmful - increasing the chance that HHVMs PHP support gets worse.

hopeseekr:
This PR is needed, because **WE ARE NOT IN A PHP5 WORLD ANY LONGER*. 

The HHVM team, and Facebook Corp., need to be *galvanized* to committing to stay compatible with PHP7 or they need to break all pretenses and start marking Hack Language as its own thing.

Otherwise, once all PHP5 code is collecting dust, relatively soon, all of your and my HHVM efforts will have been in vain!

*Hopefully this PR is a wakeup call!*

PR #7857, opened 2017-05-25
Aarch64 - Use ARMv8.1 LSE Atomics if available

swalk-cavium:
This change allows the JIT to generate an ldadd instruction in the emitter for  Vinstr::decqmlock
if  HHVM is running on a linux machine with the HWCAP_ATOMICS attribute.  This instruction
is from the ARMv8.1 Large System Extentions.  This change also  updates the vixl disassembler 
for all forms of this instruction.  A command line option is added to disable if  needed.  
To disable, use:
   -vEval.JitArmLse=0

The standard regression tests were run with several option sets.  No new issues were introduced.
Here is a partial example of what the disassembler can now produce:

23287               0x22400040  38300070              ldaddb w16, w16, [x3] 
23288               0x22400044  38b00070              ldaddab w16, w16, [x3] 
23289               0x22400048  38700070              ldaddlb w16, w16, [x3] 
23290               0x2240004c  38f00070              ldaddalb w16, w16, [x3] 
23291               0x22400050  78300070              ldaddh w16, w16, [x3] 
23292               0x22400054  78b00070              ldaddah w16, w16, [x3] 
23293               0x22400058  78700070              ldaddlh w16, w16, [x3] 
23294               0x2240005c  78f00070              ldaddalh w16, w16, [x3] 
23295               0x22400060  b8300070              ldadd x16, x16, [x3] 
23296               0x22400064  b8b00070              ldadda x16, x16, [x3] 
23297               0x22400068  b8700070              ldaddl x16, x16, [x3] 
23298               0x2240006c  b8f00070              ldaddal x16, x16, [x3] 
23299               0x22400070  f8300070              ldadd x16, x16, [x3] 
23300               0x22400074  f8b00070              ldadda x16, x16, [x3] 
23301               0x22400078  f8700070              ldaddl x16, x16, [x3] 
23302               0x2240007c  f8f00070              ldaddal x16, x16, [x3]


mxw:
cc/ @cmuellner, @dave-estes, @jim-saxman, @Orvid

hphp/runtime/base/runtime-option.cpp
```diff
@@ -487,6 +492,14 @@ static inline bool alignMacroFusionPairs() {
   return false;
 }
 
+static inline bool armLseDefault() {
+#ifdef __linux__
+  return (getauxval(AT_HWCAP) & HWCAP_ATOMICS) != 0;
```
jim-saxman: Possibly check the kernel version for when HWCAP_ATOMICS were added.

apinski-cavium: Actually I think the #ifdef should be:
#if defined (__linux__) && defined (__aarch64__)
Because HWCAP_ATOMICS is not defined on all targets.

mxw: Agreed with @pinski-cavium.  Strange that `armLseDefault()` would not check whether the `arm` part holds.

swalk-cavium:
@mxw - Hi Max,  Just a heads up.  The MOP finished last night and there were a couple of unexpected failures in the option settings using PGO.  It's similar to the issue reported on the
Facebook hhvm-general page  (Keith Adams, May 23rd).

hphp/vixl/a64/assembler-a64.cc
```diff
@@ -1135,6 +1135,13 @@ void Assembler::ldr(const FPRegister& ft, double imm) {
 }
 
 
+void Assembler::ldadd(const Register& rt, const MemOperand& src) {
```
apinski-cavium: Can you add stadd here  which emits the same as ldadd except Rt is ZR (aka 31)?

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -671,24 +671,24 @@ void Vgen::emit(const imul& i) {
...
-Y(decqmlock, Sub)
-
-#undef Y
+void Vgen::emit(const decqmlock& i) {
+  auto adr = M(i.m);
+  /* Use VIXL's macroassembler scratch regs. */
+  a->SetScratchRegisters(vixl::NoReg, vixl::NoReg);
+  if (RuntimeOption::EvalJitArmLse) {
+    a->Mov(rVixlScratch0, -1);
+    a->ldadd(rVixlScratch0, adr);
```
apinski-cavium: And use stadd here.

PR #7862, opened 2017-05-30
[AArch64] Add retranslate all support.

dave-estes:


dave-estes:
@mxw, this PR is ready for review. I've worked out the last few known issues. I've been testing retranslate all by running the server and then hitting it with siege over a list of all of tests in test/quick. I'll run OSS next and report back if I run into any errors.

@swalk-cavium, it might be helpful to run it through the MoP too. This code won't necessarily be executed by unit tests, but it's still a good idea. I typically run through the unit tests but only with a simple JIT configuration.

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -179,9 +183,13 @@ bool relocateSmashable(Env& env, Instruction* src, Instruction* dest,
...
    */
-  auto target = smashableJmpTarget(srcAddr);
+  auto target = smashableJmpTarget(srcAddrActual);
   if (target) {
-    auto sl = getSmashableFromTargetAddr(srcAddr + kSmashJmpTargetOff);
+    // Look up the smashable (jmp, jcc, etc) from the target. Note this is
+    // an actual address, and so we'll use the offset with the (potentially)
+    // virtual srcAddr when we search through smashableLocation.
+    auto slActual = getSmashableFromTargetAddr(srcAddrActual + kSmashJmpTargetOff);
+    auto sl = srcAddr + (slActual - srcAddrActual);
```
mxw: I'd prefer either having getSmashableFromTargetAddr take an actual and a virtual address or adding a function to CodeBlock to do the reverse mapping, this is a little confusing.

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -209,10 +217,8 @@ bool relocateSmashable(Env& env, Instruction* src, Instruction* dest,
   }
   env.updateInternalRefs = true;
   FTRACE(3,
-         "Relocated smashable at src 0x{:08x} ",
-         "with target 0x{:08x} to 0x{:08x} (0x{:08x})\n",
-         (uint64_t)srcAddr, (uint64_t)target,
-         *((uint32_t*)destAddr), (uint64_t)destAddr);
+         "Relocated and optimized smashable at src {} with target {} to {}.\n",
+         srcAddrActual, target, env.destBlock.toDestAddress(destAddr));
```
mxw: You might want to print the virtual source too, the metadata won't reference the "real" source.

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -231,9 +237,16 @@ bool relocateSmashable(Env& env, Instruction* src, Instruction* dest,
...
                         size_t& srcCount, size_t& destCount) {
-  auto const destAddr = reinterpret_cast<TCA>(dest);
+
+  auto const srcFrom = Instruction::Cast(srcAddr);
+  auto const destFrom = Instruction::Cast(destAddr);
+
+  auto const srcAddrActual = env.srcBlock.toDestAddress(srcAddr);
+  auto const destAddrActual = env.destBlock.toDestAddress(destAddr);
+  auto const src = Instruction::Cast(srcAddrActual);
+  auto const dest = Instruction::Cast(destAddrActual);
```
mxw: Is updating Instruction to track a virtual and real address out of the question? We did that for DecodedInstruction on the x64 side and I think it made stuff like this a lot less verbose.

hphp/runtime/vm/jit/retranslate-all.md
```diff
@@ -0,0 +1,108 @@
+Overview
+--------
```
mxw: This is fantastic, but it should be moved to hphp/doc/hackers-guide, and could use some more detail about the ARM side of things (and also looks like it's somewhat out of date).

hphp/runtime/vm/jit/retranslate-all.md
```diff
@@ -0,0 +1,108 @@
...
+
+The retranslate all feature performs many retranslations simultaneously. These
+threads need space to translate into before statically relocate to the final
```
mxw: `threads need space to translate into before they statically relocate to the final`

hphp/runtime/vm/jit/retranslate-all.md
```diff
@@ -0,0 +1,108 @@
...
+    *   Use toDestAddress() when casting to Instruction.
+    *   Fix uses of SetImmPCOffset() and ImmPCOffset(). (details in next section)
+*   Update all uses of __builtin__clear_cache().
+    *   align-arm.cpp
+    *   func-guard-arm.cpp
+    *   relocation-arm.cpp
+    *   smashable-instr-arm.cpp
+    x   tc-relocate.cpp
+    *   unique-stubs-arm.cpp
+    *   vasm-arm.cpp
```
mxw: It looks like a bunch of these were covered in this diff. Are these still TODOs?

hphp/runtime/vm/jit/retranslate-all.md
```diff
@@ -0,0 +1,108 @@
...
+instructions which includes their targets. The trick here is that these targets
+point into the weeds. The idea of using the virtual address space is that once
+we relocate them into their resting point, the instructions *should* still be PC
+relative. Therefore when emitting these PC relative instruction, we use a branch
+distance from the virtual address to the target in the actual code cache.
+However, that distance applied to to physical address in the thread local code
+cache will point into the sticks. To counteract this, two key vixl::Instruction
+helper functions are modified to take a `from` argument. If that argument is
+non-null, then all offsets are calculated as if they were from that `from`
+address and not the actual address.
```
mxw: This seems like another reason we'd want to have the Instruction just track its virtual address the way the x64 assembler and DecodedInstruction do.

hphp/runtime/vm/jit/retranslate-all.md
```diff
@@ -0,0 +1,108 @@
...
+
+### ARM PC Relative Instruction Details
```
mxw: I'm a little unclear on the point of this section

hphp/runtime/vm/jit/retranslate-all.md
```diff
@@ -0,0 +1,108 @@
...
+    -   Instruction::ImmPCOffsetTarget()
+    -   These are used in:
+        -   smashable-instr-arm.cpp: Code is published, so no updates needed.
+	-   Label::bind() - Needs updating. (DONE)
+	-   relocation-test.cpp: No updates needed.
+	-   relocation-arm.cpp: Needs updating.
+
+
+Scratch Work
+------------
```
mxw: ?

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -219,20 +219,22 @@ struct Vgen {
     , catches(env.catches)
   {}
   ~Vgen() {
-    auto begin = reinterpret_cast<char*>(base);
-    auto end = reinterpret_cast<char*>(a->frontier());
-    __builtin___clear_cache(begin, end);
+    auto begin = env.cb->toDestAddress(base);
+    auto end = env.cb->toDestAddress(a->frontier());
+    __builtin___clear_cache(reinterpret_cast<char*>(begin),
+                            reinterpret_cast<char*>(end));
```
mxw: Maybe a helper that takes a CodeBlock, and begin and end addresses and clears the cache is in order...

hphp/vixl/a64/assembler-a64.h
```diff
@@ -556,10 +556,23 @@ class Label {
...
   // link_ points to the latest branch instruction generated branching to this
   // branch.
+
+  // link_ and target_ will hold virtual addresses if the CodeBlock is
+  // configured as such. When binding a Label, the physical instructions will
+  // be updated with targets computed with the virtual address of the linked
+  // instructions and the virtual target_. This will cause calls to
+  // Instruction::ImmPCOffsetTarget() for the physical instruction to return
+  // target in the weeds, and so the code in relocation-arm.cpp takes this
+  // into account. Whe the madness? Because translations in high memory of
```
mxw: `// into account. Why the madness? Because translations in high memory of`

hphp/vixl/a64/assembler-a64.h
```diff
@@ -1789,7 +1802,6 @@ class Assembler {
 
   // Emit the instruction at pc_.
   void Emit(Instr instruction) {
-    assert(cb_.canEmit(sizeof(instruction)));
```
mxw: ?

mxw:
@dave-estes—Can you also rebase this PR onto latest master?  There are some conflicts.

dave-estes:
Sure will. I wanted to give you a chance to review my follow up changes before I do a forced git push. How about I rebase now but hold off on the push until I get an OK from you?

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -417,35 +412,48 @@ struct Vgen {
...
+  }
+  return nullptr;
+}
+
+static CodeAddress toReal(Venv& env, CodeAddress a) {
+  CodeBlock* b = getBlock(env, a);
+  if (b) {
+    return b->toDestAddress(a);
   }
+
```
mxw: `return b != nullptr ? b->toDestAddress(a) : a;`

dave-estes:
Reverted the changes to `DataBlock::sync`. Also merged up. I'll work those changes to `DataBlock::sync` into a new PR shortly, but there are a lot of conflicts with this PR. Once this lands, I'll push that PR.

mxw:
Oh, I always forget that GitHub doesn't let you make dependent PRs...

PR #7888, opened 2017-06-18
fix xdebug coverage (part1): `xdebug_get_code_coverage` return value format for executed line

### LINKED ISSUE #1589: Reported code coverage differs from what PHP/Xdebug reports
While working on code coverage support for PHPUnit on HHVM, I discovered that the reported code coverage differs from what Xdebug reports. This will probably not be the only difference between HHVM's and Xdebug's code coverage but it's the first I found.

To reproduce, simply run PHPUnit's `master` on the test suite of https://github.com/sebastianbergmann/money (`phpunit -c build`).

PHP and Xdebug report 100% code coverage:

php

HHVM reports less than 100% code coverage:

hhvm-1

hhvm-2

hhvm-3

xKerman:
This pull request is related to #1589 .
(I will submit another pull request for supporting `XDEBUG_CC_UNUSED` and `XDEBUG_CC_DEAD_CODE` later)

`xdebug_get_code_coverage()` return value format is described in https://xdebug.org/docs/code_coverage :

> The returned values for each line are:
> * 1: this line was executed
> * -1: this line was not executed
> * -2: this line did not have executable code on it

According to this description, `xdebug_get_code_coverage()` should return value like:

```
[
    "/path/to/foo.php" => [
          3 => 1,  // executed line 3
          5 => 1,  // executed line 5
    ]
]
```

However, hhvm's `xdebug_get_code_coverage()` currently returns value like:

```
[
    "/path/to/foo.php" => [
          3 => 2,  // executed line 3 (2 times)
          5 => 3,  // executed line 5 (3 times)
    ]
]
```

In this pull request, hhvm's `xdebug_get_code_coverage()` return value format for executed line is changed to follow description in https://xdebug.org/docs/code_coverage .


xKerman:
Thanks!

PR #7891, opened 2017-06-20
[AArch64] Guarantees smashableMovq adjustment during relocation.

dave-estes:
smashableMovqs often have an internal reference which needs to be
adjusted. While that adjustment code is present, it's only
triggered when the update internal reference flag is set. This
patch insures that the flag is set whenever a smashableMovq is
relocated.

### DIFF
--- a/hphp/runtime/vm/jit/relocation-arm.cpp
+++ b/hphp/runtime/vm/jit/relocation-arm.cpp
@@ -590,6 +590,12 @@ size_t relocateImpl(Env& env) {
         // Do nothing, as the instruction word was initially copied above
       }
 
+      // If we just copied the first instruction of a smashableMovq, then it may
+      // with an internal reference that'll need to be adjusted below.
+      if (!literals.count(src) && isSmashableMovq(srcAddr)) {
+	env.updateInternalRefs = true;
+      }
+
       if (srcAddr == env.start) {
         /*
          * For the start of the range, we only want to overwrite the "after"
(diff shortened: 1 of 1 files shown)


dave-estes:
Ping @mxw

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -590,6 +590,12 @@ size_t relocateImpl(Env& env) {
         // Do nothing, as the instruction word was initially copied above
       }
 
+      // If we just copied the first instruction of a smashableMovq, then it may
+      // with an internal reference that'll need to be adjusted below.
```
mxw: `s/with/have`?

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -590,6 +590,12 @@ size_t relocateImpl(Env& env) {
...
+      if (!literals.count(src) && isSmashableMovq(srcAddr)) {
+	env.updateInternalRefs = true;
```
mxw: You're off by one space.

hphp/runtime/vm/jit/relocation-arm.cpp
dave-estes: Fixed

hphp/runtime/vm/jit/relocation-arm.cpp
dave-estes: Sorry

PR #7894, opened 2017-06-26
[AArch64] - capture n:immr:imms encoding for AND (immediate)

swalk-cavium:
This sequence was seen multiple times in several of the regression tests
particularly hphp/test/quick/all_type_comparison_test.php

    b2609ff3              mov x19, #0xffffffff000000ff
    8a130021              and x1, x1, x19

Some ARMv8 instructions utilize BitMask encoding to represent both
32 and 64-bit values.  The AND (immediate) form of the instruction can represent
the  immediate value above and  the constant can be propagated.  The  mov instruction 
is removed as dead.

    92609c21              and x1, x1, #0xffffffff000000ff

This immediate value is hard-coded in the lowering of vcall, vinvoke, syncvmret, and
defvmretdata on ARM hosts.

The current implementation of Vinstr::andqi can only support 32-bit values.  This
change will catch the BitMask encoding ( n:immr:imms) into a new ARM specific intrinsic.

The standard regression tests were run with six option sets.  No additional failures were
observed.

### DIFF
--- a/hphp/runtime/vm/jit/vasm-arm.cpp
+++ b/hphp/runtime/vm/jit/vasm-arm.cpp
@@ -294,6 +294,7 @@ struct Vgen {
   void emit(const andli& i) { a->And(W(i.d), W(i.s1), i.s0.l(), UF(i.fl)); }
   void emit(const andq& i) { a->And(X(i.d), X(i.s1), X(i.s0), UF(i.fl)); }
   void emit(const andqi& i) { a->And(X(i.d), X(i.s1), i.s0.q(), UF(i.fl)); }
+  void emit(const andzi& i) { a->And(X(i.d), X(i.s1), i.s0.Q(), UF(i.fl)); }
   void emit(const cmovb& i) { a->Csel(W(i.d), W(i.t), W(i.f), C(i.cc)); }
   void emit(const cmovw& i) { a->Csel(W(i.d), W(i.t), W(i.f), C(i.cc)); }
   void emit(const cmovl& i) { a->Csel(W(i.d), W(i.t), W(i.f), C(i.cc)); }
--- a/hphp/runtime/vm/jit/vasm-arm.cpp
+++ b/hphp/runtime/vm/jit/vasm-arm.cpp
@@ -294,7 +294,7 @@ struct Vgen {
   void emit(const andli& i) { a->And(W(i.d), W(i.s1), i.s0.l(), UF(i.fl)); }
   void emit(const andq& i) { a->And(X(i.d), X(i.s1), X(i.s0), UF(i.fl)); }
   void emit(const andqi& i) { a->And(X(i.d), X(i.s1), i.s0.q(), UF(i.fl)); }
-  void emit(const andzi& i) { a->And(X(i.d), X(i.s1), i.s0.Q(), UF(i.fl)); }
+  void emit(const andqi64& i) { a->And(X(i.d), X(i.s1), i.s0.q(), UF(i.fl)); }
   void emit(const cmovb& i) { a->Csel(W(i.d), W(i.t), W(i.f), C(i.cc)); }
   void emit(const cmovw& i) { a->Csel(W(i.d), W(i.t), W(i.f), C(i.cc)); }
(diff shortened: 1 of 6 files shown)


hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -294,6 +294,7 @@ struct Vgen {
   void emit(const andli& i) { a->And(W(i.d), W(i.s1), i.s0.l(), UF(i.fl)); }
   void emit(const andq& i) { a->And(X(i.d), X(i.s1), X(i.s0), UF(i.fl)); }
   void emit(const andqi& i) { a->And(X(i.d), X(i.s1), i.s0.q(), UF(i.fl)); }
+  void emit(const andzi& i) { a->And(X(i.d), X(i.s1), i.s0.Q(), UF(i.fl)); }
```
mxw: Why can't this just use `q()`?

hphp/runtime/vm/jit/vasm-fold-imms.cpp
```diff
@@ -354,6 +354,14 @@ struct ImmFolder {
     return true;
   }
 
+  bool logical_BitMask(Vreg r, uint64_t& out) {
```
mxw: Avoid mixing case conventions in the same symbol.

hphp/runtime/vm/jit/vasm-fold-imms.cpp
```diff
@@ -413,6 +420,17 @@ struct ImmFolder {
...
   void fold(cmpq& in, Vinstr& out) { return fold_cmp<cmpqi>(in, out); }
 
+  void fold(andq& in, Vinstr& out) {
+    int val;
+    if (logical_imm(in.s0, val)) { out = andqi{val, in.s1, in.d, in.sf}; }
+    else if (logical_imm(in.s1, val)) { out = andqi{val, in.s0, in.d, in.sf}; }
+    else {
+      uint64_t bm;
+      if (logical_BitMask(in.s0, bm)) { out = andzi{bm, in.s1, in.d, in.sf}; }
+      else if (logical_BitMask(in.s1, bm)) { out = andzi{bm, in.s0, in.d, in.sf}; }
```
mxw: This line is > 80 chars long.

I think it's probably better to just change everything in this function to this sort of thing:

```
if (...) {
  // stuff
} else if (...) {
  // stuff
} else {
  // stuff
}
```

etc.

hphp/runtime/vm/jit/vasm-instr.h
```diff
@@ -1109,6 +1110,7 @@ struct shlq { Vreg64 s, d; VregSF sf; Vflags fl; }; // uses rcx
 /*
  * arm intrinsics.
  */
+struct andzi { Immed64 s0; Vreg64 s1, d; VregSF sf; Vflags fl; };
```
mxw: What does the `z` stand for?

swalk-cavium: @mxw - Hi Max,  The intent was to pick something  not 'q'.  The current
naming scheme seems to be for encoding the size of the result.  This is the
second Vinstr to use Immed64.  Do you have any suggestion?

mxw: In @cmuellner's original implementation of this, he used `andqi64{}`, which I thought was fairly descriptive, whereas `z` is not.  I would go with that.

(cf. https://github.com/facebook/hhvm/pull/7851)

hphp/runtime/vm/jit/vasm-arm.cpp
mxw: In https://github.com/facebook/hhvm/pull/7851, @cmuellner just used `q()`, so I think that should be fine here, and you can revert the implementation of `Q()`.

hphp/runtime/vm/jit/vasm-instr.h
swalk-cavium: @mxw - Ok, yes.  That sounds better.

mxw:
Sorry for the delay.  This looks good; I'm going to land once all internal tests pass.

PR #7895, opened 2017-06-27
[PPC64] Fix runtime: new vasms and r30 usage

gut:
1) Fix PPC64's register allocation on lowering

When debugging, this assert was failing:
Assertion failure:
[...]/hhvm/hphp/runtime/vm/jit/vasm-lower-inl.h:76:
HPHP::jit::vasm_lower(HPHP::jit::Vunit&, Vlower)::<lambda(HPHP::jit::Vlabel)> [with Vlower = HPHP::jit::{anonymous}::lowerForPPC64(HPHP::jit::Vunit&)::<lambda(const HPHP::jit::VLS&, HPHP::jit::Vinstr&, HPHP::jit::Vlabel, size_t)>]:
assertion `allow || unit.next_vr == next' failed.

This is related to this change:
22c005b

Fixing it for PPC64 as done in ARM.

2) Lower cmovl, cmovb and cmovw to cmovq on PPC64

These vasms are being emitted now and quick test suite fails due to
unimplemented vasm.

3) Avoid non-volatile r30 register as it's not saved

Starting on gcc v5.4.1, the r30 register can't be clobbered and the ABI
still flags it as a non-volatile register.

Recently some tests started to fail as the value of r30 could not be
lost when executing the VM. For some reason the compiler was not
generating a code which r30 was being used so this fault was not
happening.

### DIFF
--- a/hphp/runtime/vm/jit/abi-ppc64.cpp
+++ b/hphp/runtime/vm/jit/abi-ppc64.cpp
@@ -38,19 +38,20 @@ const RegSet kGPCallerSaved =
 
 const RegSet kGPCalleeSaved =
   reg::r14 | reg::r15 | reg::r16 | reg::r17 | reg::r18 | reg::r19 | reg::r20 |
-  reg::r21 | reg::r22 | reg::r23 | reg::r24 | reg::r25 | reg::r26;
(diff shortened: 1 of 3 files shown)


gut:
do you need any input from my side? It's been 5 days on this "import" state and nothing happened...

PR #7897, opened 2017-06-29
Enable hot static code mapping onto huge pages

octmoraru:
This PR enables mapping of hot static code onto huge pages for open
source builds. It defines `__hot_start` and `__hot_end` at link time using
the first and last functions from the linker script. This feature can be
turned off by setting the CMake option `MAP_TEXT_HUGE_PAGES` to `Off`.

`hugifyText` was aligned to 2MB so that we maximize the amount of hot code
we can map if `__hot_end` is placed just before it.

It also fixes a few issues with the hfsort collection process:
* Pass `--ffunctions-sections` to C code in third-party modules. Otherwise
the linker will not find hot C functions if they are referenced in the
linker script.
* Only collect user space stacks in `hf-prod-collect`. Since hfsort only
looks at the topmost two functions on the stack, having kernel functions
there will cause less accurate information to be used.
* Add possibility to pass `HHVM_PID` to `hf-prod-collect` through an env
var. This is to account for the case where we use oss-performance to
train. When we do that, `pgrep -o hhvm` matches the HHVM process that
runs `perf.php` instead of the HHVM under test.
* Regenerate linker script to include all these changes. oss-performance
WordPress was uses as the training workload.


octmoraru:
@ottoni, @markw65 - could one of you (or someone else) take a look at this?

octmoraru:
@ottoni - Yes, I'm seeing a 1.5-2% improvement in RPS on oss-performance (WordPress, Drupal7, MediaWiki). This is on an Intel Xeon Platinum 8180 system.

octmoraru:
@ottoni - Is there anything else needed for this to be imported into Phabricator?

octmoraru:
@mxw, @ottoni - any updates?

octmoraru:
@mxw @ottoni It's been 23 days since this diff was imported. Is there anything preventing it from landing?

PR #7902, opened 2017-07-05
Add asm optimization on hash for PPC64

lbianc:
Added PPC64 specific code on hash initialization.


mxw:
lgtm.  I'll land once sanity tests pass.

PR #7903, opened 2017-07-05
Revert avoid optimization to PPC64

lbianc:
Reverted change that removed optimization parameters for PPC64.


gut:
@lbianc : Why are you reverting that?

lbianc:
It was causing a stackoverflow.

PR #7906, opened 2017-07-07
Allow getopt to return optind by reference.

TysonAndre:
PHP 7.1 changed the signature of getopt by adding a third parameter
returning the index where parameter parsing stopped.

`function getopt(string $options, array $longopts = null, int &$optind = null)`

Haven't added getopt_006-009 from php-src yet,
waiting to see if the new parameter is a feature that could be accepted
into HHVM.

This is useful for parsing CLI arguments in standalone scripts
without relying on third party libraries.

See http://php.net/getopt

### DIFF
--- a/hphp/runtime/ext/std/ext_std_options.cpp
+++ b/hphp/runtime/ext/std/ext_std_options.cpp
@@ -595,7 +595,8 @@ static req::vector<opt_struct> parse_opts(const char *opts, int opts_len) {
 }
 
 static Array HHVM_FUNCTION(getopt, const String& options,
-                                   const Variant& longopts /*=null */) {
+                                   const Variant& longopts /*=null */,
+                                   OutputArg optind /*= null */) {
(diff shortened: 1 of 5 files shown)


TysonAndre:
It seems like I took the wrong approach using references. This probably should be `<<__OutOnly>>`, like I see for preg_match_all_with_matches.

This PR is over 3 years old and has no reviews. After it was created, there was also an announcement that HHVM's APIs and syntax would diverge from PHP's APIs, so my motivation for including this isn't as strong, so I'm going to assume that getopt is a low priority; there are definitely better libraries in composer.

If anyone wants to resurrect this, feel free to include any of my code.

```php
function preg_match_all_with_matches(string $pattern,
                                     string $subject,
                                     <<__OutOnly>>
                                     inout mixed $matches,
                                     int $flags = 0,
                                     int $offset = 0): mixed;
```

Issue #3925, opened 2014-10-08
test/slow/ext_icu/uspoof.php consistently fails


rrh:
On my ubuntu 14.04 system this test
   test/run test/slow/ext_icu/uspoof.php
consistently fails.

I have run
  sudo locale-gen zh_CN.utf8 
which is part of how FB's travis jig https://github.com/facebook/hhvm/blob/master/.travis.yml#L22 is setup.

I'm told that all tests pass on FB's travis system, so presumably this test passes there, so I presume there's something missing from my environment.  Perhaps I need another locale that's not documented in travis.yml?

Where did the test code
  https://github.com/facebook/hhvm/blob/master/hphp/test/slow/ext_icu/uspoof.php
come from?

However I'm wondering if the test itself is bogus.
  https://github.com/facebook/hhvm/blob/master/hphp/test/slow/ext_icu/uspoof.php#L120
is a test that 2 Han Chinese characters are "suspiciously" present when SINGLE_SCRIPT semantics holds sway.  Both of these characters  http://www.fileformat.info/info/unicode/char/8005/index.htm and http://www.fileformat.info/info/unicode/char/5fcd/index.htm are labeled as "Han Chinese", so I'd think (I am not an i18n person) they could co-exist without suspicion.

This piece of php code also fails on ubuntu 14.04's default php5-intl installation, eg php5.5.9 with intl support ICU version 52.1.

vielmetti:
This test fails on aarch64 with Ubuntu 16.04, with ICU 55:

```
hhvm@ed-2a-sjc-dcos:/home/src/github.com/facebook/hhvm/hphp$ ldd hhvm/hhvm | grep -i icu      
        libicui18n.so.55 => /usr/lib/aarch64-linux-gnu/libicui18n.so.55 (0x0000ffffb5240000) 
        libicuuc.so.55 => /usr/lib/aarch64-linux-gnu/libicuuc.so.55 (0x0000ffffb50a2000)      
        libicudata.so.55 => /usr/lib/aarch64-linux-gnu/libicudata.so.55 (0x0000ffffb2291000)
```

PR #7912, opened 2017-07-11
AArch64 Fix getFrameRegs assertion during indirect fixup

jim-saxman:
On AA4ch64 plaforms when calling getFrameRegs() and an indirect
fixup is found, the RIP needs to be present at the correct
offset. It was not, because the destructor was compiled with a
tail-call. It needs to have a standard stack frame.

By disabling the tail-call via the compiler flag on all destructors
in g_destructors[], we can ensure that all destructors for AArch64
are walked correctly.

This bug occurred sometimes in OSS Mediawiki after a few minutes.
We tried to create a simplified unit test but were unsuccessful.

This patch introduces no new unit test failures for either DebugOpt
or Release build types on AArch64 platforms and doesn't affect x64
or PPC64 destructors.

### DIFF
--- a/hphp/runtime/base/array-data-defs.h
+++ b/hphp/runtime/base/array-data-defs.h
@@ -212,6 +212,7 @@ inline Variant ArrayData::getKey(ssize_t pos) const {
 
 ///////////////////////////////////////////////////////////////////////////////
 
+AARCH64_WALKABLE_FRAME
 inline void ArrayData::release() noexcept {
   assert(hasExactlyOneRef());
   return g_array_funcs.release[kind()](this);
--- a/hphp/runtime/base/array-data-defs.h
+++ b/hphp/runtime/base/array-data-defs.h
@@ -216,7 +216,6 @@ inline void ArrayData::release() noexcept {
   assert(hasExactlyOneRef());
   g_array_funcs.release[kind()](this);
   AARCH64_WALKABLE_FRAME();
-  return;
 }
 
 inline ArrayData* ArrayData::append(Cell v, bool copy) {
--- a/hphp/runtime/base/array-data-defs.h
+++ b/hphp/runtime/base/array-data-defs.h
@@ -212,10 +212,11 @@ inline Variant ArrayData::getKey(ssize_t pos) const {
 
 ///////////////////////////////////////////////////////////////////////////////
 
-AARCH64_WALKABLE_FRAME
 inline void ArrayData::release() noexcept {
   assert(hasExactlyOneRef());
-  return g_array_funcs.release[kind()](this);
+  g_array_funcs.release[kind()](this);
+  AARCH64_WALKABLE_FRAME();
+  return;
 }
 
 inline ArrayData* ArrayData::append(Cell v, bool copy) {
--- a/hphp/runtime/base/mixed-array.cpp
+++ b/hphp/runtime/base/mixed-array.cpp
@@ -468,7 +468,7 @@ ArrayData* MixedArray::MakeDictFromAPC(const APCArray* apc) {
 
 //=============================================================================
 // Destruction
-
+AARCH64_WALKABLE_FRAME
 NEVER_INLINE
 void MixedArray::Release(ArrayData* in) {
   assert(in->isRefCounted());
--- a/hphp/runtime/base/mixed-array.cpp
+++ b/hphp/runtime/base/mixed-array.cpp
@@ -468,7 +468,6 @@ ArrayData* MixedArray::MakeDictFromAPC(const APCArray* apc) {
 
 //=============================================================================
 // Destruction
-AARCH64_WALKABLE_FRAME
 NEVER_INLINE
 void MixedArray::Release(ArrayData* in) {
   assert(in->isRefCounted());
@@ -495,6 +494,7 @@ void MixedArray::Release(ArrayData* in) {
     }
   }
   MM().objFree(ad, ad->heapSize());
+  AARCH64_WALKABLE_FRAME();
 }
 
 NEVER_INLINE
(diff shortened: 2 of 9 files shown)


jim-saxman:
@mxw Hi Max, can you lease review this PR.
@swalk-cavium Are you still seeing the errors in test/quick/*dtor.php unti tests, and does this patch help?

apinski-cavium:
You just use __aarch64__ instead __AARCH64EL__ (not that matters but it is just a style thing).
Also it seems better to do something like:

inline void ArrayData::release() noexcept {
  assert(hasExactlyOneRef());
  g_array_funcs.release[kind()](this);
  AARCH64_WALKABLE_FRAME ();
 }

And have AARCH64_WALKABLE_FRAME (); expand to something like:
__asm__("":::"memory");

I am not a fan of the optimize attribute in this case :).

mxw:
I think `__aarch64__` is more consistent with what we're using elsewhere.

I chatted with @dave-estes offline who seems to like the memory barrier approach.  You can continue ifdeffing that macro to do nothing on x86-64.

hphp/runtime/base/array-data-defs.h
```diff
@@ -214,7 +214,9 @@ inline Variant ArrayData::getKey(ssize_t pos) const {
 
 inline void ArrayData::release() noexcept {
   assert(hasExactlyOneRef());
-  return g_array_funcs.release[kind()](this);
+  g_array_funcs.release[kind()](this);
+  AARCH64_WALKABLE_FRAME();
+  return;
```
mxw: Just delete this return; this function returns `void`.

hphp/runtime/base/mixed-array.cpp
```diff
@@ -468,7 +468,6 @@ ArrayData* MixedArray::MakeDictFromAPC(const APCArray* apc) {
 
 //=============================================================================
 // Destruction
-
```
mxw: Put this back?

swalk-cavium:
@jim-saxman - I ran a full MOP on this change.  

These two tests now pass with these option sets
hphp/test/run -v -m jit -a '-vEval.JitPGO=1'
hphp/test/run -v -m jit -r -a '-vEval.JitPGO=1'
    hphp/test/quick/setg-dtor.php
    hphp/test/quick/sets-dtor.php

hphp/test/slow/async/simple_meth.php passes all option sets.

A bunch of failures have reappeared with this option set
    hphp/test/run -v -m jit -r -a '-vEval.JitPGO=1'
    hphp/test/zend/good/ext/spl/tests/iterator_002.php
    hphp/test/zend/good/ext/spl/tests/iterator_033.php
    hphp/test/zend/good/ext/spl/tests/iterator_040.php
    hphp/test/zend/good/ext/spl/tests/iterator_048.php
    hphp/test/zend/good/ext/spl/tests/observer_001.php
    hphp/test/zend/good/ext/spl/tests/observer_002.php
    hphp/test/zend/good/ext/spl/tests/RecursiveCallbackFilterIteratorTest.php

I think I saw them last between fixes for relocation.

jim-saxman:
Hi @swalk-cavium, Thanks for running tests on this patch. I re-ran all unit tests in all configurations and observed the following:

For PGO with and without repoAuth:
hphp/test/quick/setg-dtor.php
hphp/test/quick/sets-dtor.php
pass both before and after this patch.

hphp/test/slow/async/simple_meth.php
still fails with -r -m pgo

php/test/zend/good/ext/spl/tests/iterator_002.php
hphp/test/zend/good/ext/spl/tests/iterator_033.php
hphp/test/zend/good/ext/spl/tests/iterator_040.php
hphp/test/zend/good/ext/spl/tests/iterator_048.php
hphp/test/zend/good/ext/spl/tests/observer_001.php
hphp/test/zend/good/ext/spl/tests/observer_002.php
hphp/test/zend/good/ext/spl/tests/RecursiveCallbackFilterIteratorTest.php
never failed both before and after this patch, for -r -m pgo.

I performed these tests on both CentOS with GCC 6.2.1 and Ubuntu with GCC 5.4.

swalk-cavium:
Hi @jim-saxman, I'm wondering if one of my machines needs a firmware upgrade.

vielmetti:
Hi @swalk-cavium @jim-saxman - just did a build and test cycle and am seeing these errors:

HipHop VM 3.20.0-dev (rel)
Compiler: heads/master-0-g26a2f8c2795726e89dd636c16b2e1e6316dee7df
Repo schema: c9ccdb2b68ae4d61f3c1c0232b4b2426ff39c26e
gcc version 5.4.0 20160609 (Ubuntu/Linaro 5.4.0-6ubuntu1~16.04.4) 

```
test/slow/ext_icu/uspoof.php
test/slow/ext_image/1794.php
test/slow/ext_posix/ext_posix.php
test/slow/ext_spl_file/1795.php
test/slow/ir_inlining/1.php
test/quick/setg-dtor.php
test/quick/sets-dtor.php
```
System under test is a 96-core Packet Type 2A server, with Cavium ThunderX, using Ubuntu 16.04. Rather than noting a new issue, since the set*-dtor failures seem to be related, I'll just report here.

swalk-cavium:
Hi Ed,


We have yet to have an all green run on the regression tests.

Our Jenkins runs 6 different option sets.


Which options did you use below?


The uspoof ones depend on which version of the internationalization

library is used.  This one always fails our CI testing.  It has something

to do with libicui18n.so


The ext_posix.php test assumes that there is a 'wheel'  group.  Try

sudo groupadd wheel


I was not able to get 1794.php or 1795.php to fail.


I think these are related to the PR in the subject.

ir_inlining/1.php, setg-dtor.php and sets-dtor.php


Theobroma Systems makes their CI available here.  One option

set, multiple OSs.

http://hhvm-arm64.theobroma-systems.com/


Steve Walk




vielmetti:
After applying this patch all of test/quick passes ("SHIP IT").

All of test/slow passes except test/slow/ext_icu/uspoof.php which is described by #3925.

Looking good!

PR #7914, opened 2017-07-12
Aarch64 - update EqDbl to use SIMD/FP instructions

swalk-cavium:
This change updates the EqDbl and NeqDbl sequences to use different instructions.
It reduces the number of instructions needed and removes the dependency on the
status register.

Before
=====
    (29) t8:Bool = EqDbl t7:Dbl, t5:Dbl
        Main:   
              0x51003aec  d53b4210              mrs x16, nzcv
              0x51003af0  1e602020              fcmp d1, d0
              0x51003af4  da9f13f2              csetm x18, eq 
              0x51003af8  9e670240              fmov d0, x18 
              0x51003afc  d51b4210              msr nzcv, x16
              0x51003b00  9e660000              fmov x0, d0
              0x51003b04  53001c00              uxtb w0, w0
              0x51003b08  12000000              and w0, w0, #0x1

After
====
    (29) t8:Bool = EqDbl t7:Dbl, t5:Dbl
        Main:   
              0x1460407c  5e60e420              fcmeq d0, d1, d0
              0x14604080  9e660000              fmov x0, d0
              0x14604084  12000000              and w0, w0, #0x1

This also updates the vixl disassembler to recognize two new instruction formats:
Advanced SIMD Scalar Three Same, and Advanced SIMD Scalar two-register Misc.
Whenever the enumeration clashed with an existing definition a prefix was added.
See STS, and STM respectively.

The regression suite was run with 6 option sets.  No additional failures were observed.


swalk-cavium:
@mxw - Hi Max,  Can you take a look at this one?  Thanks.

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -1545,7 +1533,7 @@ void lower(const VLS& e, movtdb& i, Vlabel b, size_t z) {
   lower_impl(e.unit, b, z, [&] (Vout& v) {
     auto d = v.makeReg();
     v << copy{i.s, d};
-    v << movtqb{d, i.d};
+    v << copy{d, i.d};
```
mxw: What's the purpose of copying through a tmp reg?

swalk-cavium: @mxw - The code generator wouldn't accept it without the extra copy.  Something to do with changing register classes I suspect.

mxw: But the register widths are propagated through copies, and they already don't match since this is a truncation operation...

Can you paste the exact error?

swalk-cavium: @mxw - Looks like I didn't capture that in my daily log.  I'll have to try and recreate.

swalk-cavium: @mxw - I was unable to recreate the issue.  I'll reduce the copy and retest.

jim-saxman:

swalk-cavium:
@mxw - MOP results the same with only 1 copy, so updated pull request.

mxw:
Before I accept this, I'd like someone else to give a second opinion on the change (not just on test results) to confirm it doesn't affect the behavior of the instruction in any way.

swalk-cavium:
@mxw - Hi Max,  Here's how I tested the sequences before incorporating in hhvm
https://pastebin.com/PC6Ym7ZY

swalk-cavium:
@mxw - Here's the table I used for the unordered comparison case.
https://pastebin.com/psgHNebB

SNaN - signalling NaN,
QNaN - quiet NaN

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -1543,9 +1531,7 @@ void lower(const VLS& e, movzlq& i, Vlabel b, size_t z) {
 
 void lower(const VLS& e, movtdb& i, Vlabel b, size_t z) {
   lower_impl(e.unit, b, z, [&] (Vout& v) {
-    auto d = v.makeReg();
-    v << copy{i.s, d};
-    v << movtqb{d, i.d};
+    v << copy{i.s, i.d};
```
dave-estes: Is it safe to do this if we end up reverting f9ecdf1? Might want to split this change out.

hphp/vixl/a64/assembler-a64.cc
```diff
@@ -1352,6 +1352,22 @@ void Assembler::frintz(const FPRegister& fd,
 }
 
 
+void Assembler::fcmeq(const FPRegister& fd,
```
dave-estes: I think it would be more readable if this was named `Assembler::fcmeqz()`.

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -909,29 +909,17 @@ void Vgen::emit(const unpcklpd& i) {
 ///////////////////////////////////////////////////////////////////////////////
 
 void Vgen::emit(const cmpsd& i) {
```
dave-estes: I think renaming the 2 argument `fcmeq` (see next comment) will help, but I also think a line or two of comments will make this more clear too.

swalk-cavium:
@mxw - I think I might have to update this to make it check the hardware capabilities.  I just noticed at least one of the forms says, ARMv8.2.  Something similar to what we did for the LSE Atomics.

swalk-cavium:
@mxw - Um, disregard the previous comment.  SIMD instructions are required in the
ARM SBSA (Server Base System Architecture).

hphp/runtime/vm/jit/vasm-arm.cpp
swalk-cavium: @dave-estes - Hi Dave,  I'm not sure.  If your new version goes in first, I'll have to retest.

mxw:
@swalk-cavium—Cool, thanks for this change.  I'll land it as soon as internal tests pass.

swalk-cavium:
@mxw - Hi Max,  Any word on your testing infrastructure issue?  Can this one land?

PR #7916, opened 2017-07-17
Reduce race condition when smashing on PPC64

lbianc:
Currently on PPC64 architecture all immediates are loaded using up to 5
instructions, which makes the read not atomic.
The smashable instructions may have its immediate changed on the fly, i.e.,
while running or re-emitting the JITted code, which causes race condition,
if reading the immediate is composed with more than one instruction.
The race condition happen at two different moments:

1. The JITted code is being executed and the pipeline stop reading the
instructions group that reads the immediate in the middle (thread
preemption), then another thread smashes the instruction that uses the same
immediate.

2. The Decoder is reading the immediate used in a smashable instruction and
this thread stop being executed (thread preemption), then another thread
smashes the instruction that uses the same immediate.

The item 2 represents the most frequently cause of the race conditions, as
the Decoder code takes much more instructions (written in C++) to read the
entire immediate, when comparing the reading while executing (maximum
of 5 instructions).
For the reason above, this PR is being sent to fix the item 2, which locks
the C++ methods that read/write smachable instructions.
The lock was not extended while executing the JITted code, that would fix
the item 1, because it would create a performance bottleneck. Fixing it will
need a bigger change, as many parts of the code depends on how the immediate
is being written/loaded currently, and will be sent later, after be better
analyzed.

### DIFF
--- a/hphp/runtime/vm/jit/smashable-instr-ppc64.cpp
+++ b/hphp/runtime/vm/jit/smashable-instr-ppc64.cpp
@@ -37,6 +37,14 @@ using ppc64_asm::DecodedInstruction;
 
 ///////////////////////////////////////////////////////////////////////////////
 
+// Locks to avoid read/write when smashing instructions
+static folly::MicroSpinLock s_lockMovq;
+static folly::MicroSpinLock s_lockCmpq;
+static folly::MicroSpinLock s_lockCall;
+static folly::MicroSpinLock s_lockJmp;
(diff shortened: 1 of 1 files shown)


lbianc:
@mxw, as discussed, this is the change to reduce the race conditions on PPC64, until we have the final solution.

lbianc:
Closing this PR as a new solution was implemented in https://github.com/facebook/hhvm/pull/7959.

PR #7918, opened 2017-07-18
correction to ocaml Aarch64 code gen

swalk-cavium:
This corrects an issue in the ocaml compiler where it tries to make
a invalid immediate operand for a subtraction instruction.

This change was tested by rebuilding from scratch and running the
standard regression tests.

I'm not sure why I had to add the --ignore-whitespace option to the patch
command line.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,7 +30,7 @@ if (PATCHES)
   foreach(PATCH ${PATCHES})
     message(STATUS "Applying ${PATCH}")
     execute_process(
-      COMMAND patch -p1 --forward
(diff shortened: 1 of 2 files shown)


Orvid:
Has this change been submitted to upstream OCaml as well?

swalk-cavium:
@Orvid - working on it.

swalk-cavium:
@mxw - Hi Max,  The fix has been submitted upstream to ocaml also.  Aarch64 builds
are blocked without this patch.  Can you inspect?  Thanks.

mxw:
Will land this as soon as tests pass.

PR #7919, opened 2017-07-18
Aarch64 - fold known ANDs

swalk-cavium:
This catches an annoying sequence  seen while working on PR#7894.  The values for
an and immediate are known and can be trivially folded at compile-time.  This was
seen 94 times in hphp/test/quick/all_type_comparison_test.php.

 Before
 ======
` B15: [profCount=1] (preds B14 B13)`
`  (64) RetCtrl<3> t1:StkPtr, t0:FramePtr, InitNull`
`     Main:   `
`    0x5340202c  910083bc              add x28, x29, #0x20 (32)`
`    0x53402030  b2609fe1              mov x1, #0xffffffff000000ff`
`    0x53402034  92400021              and x1, x1, #0x1 `   //<<---
`    0x53402038  a9407bbd              ldp x29, x30, [x29]`
`    0x5340203c  d65f03c0              ret  `

 After
 =====
` B15: [profCount=1] (preds B14 B13)`
`    (64) RetCtrl<3> t1:StkPtr, t0:FramePtr, InitNull`
`        Main:   `
`              0x3a000d24  910083bc              add x28, x29, #0x20 (32)`
`              0x3a000d28  d2800021              movz x1, #0x1`
`              0x3a000d2c  a9407bbd              ldp x29, x30, [x29]`
`              0x3a000d30  d65f03c0              ret `

The standard regression tests were run with six option sets.  No additional errors were
seen.

### DIFF
--- a/hphp/runtime/vm/jit/vasm-fold-imms.cpp
+++ b/hphp/runtime/vm/jit/vasm-fold-imms.cpp
@@ -421,12 +421,26 @@ struct ImmFolder {
   void fold(cmpq& in, Vinstr& out) { return fold_cmp<cmpqi>(in, out); }
 
   void fold(andq& in, Vinstr& out) {
+    if (!uses[in.sf] && valid.test(in.s0) && valid.test(in.s1)) {
+      auto imm0 = vals[in.s0];
+      auto result = imm0 & vals[in.s1];
+      out = ldimmq{result, in.d};
+      return;
+    }
     int val;
     uint64_t bm;
     if (logical_imm(in.s0, val)) { out = andqi{val, in.s1, in.d, in.sf}; }
     else if (logical_imm(in.s1, val)) { out = andqi{val, in.s0, in.d, in.sf}; }
     else if (logical_bmsk(in.s0, bm)) { out = andqi64{bm, in.s1, in.d, in.sf}; }
     else if (logical_bmsk(in.s1, bm)) { out = andqi64{bm, in.s0, in.d, in.sf}; }
+  }
+
+  void fold(andqi& in, Vinstr& out) {
+    if (uses[in.sf]) return;
+    if (!valid.test(in.s1)) return;
+    auto imm64 = vals[in.s1];
+    auto result = imm64 & in.s0.q();
+    out = ldimmq{result, in.d};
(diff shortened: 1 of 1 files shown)


swalk-cavium:
@mxw - Hi Max, Can you inspect?  Thanks.

jim-saxman:

hphp/runtime/vm/jit/vasm-fold-imms.cpp
```diff
@@ -421,6 +421,12 @@ struct ImmFolder {
   void fold(cmpq& in, Vinstr& out) { return fold_cmp<cmpqi>(in, out); }
 
   void fold(andq& in, Vinstr& out) {
+    if (!uses[in.sf] && valid.test(in.s0) && valid.test(in.s1)) {
+      auto imm0 = vals[in.s0];
+      auto result = imm0 & vals[in.s1];
+      out = ldimmq{result, in.d};
```
mxw: `auto const`

Also, I think this could be compacted a bit.  I'd prefer

```
auto const result = vals[in.s0] & vals[in.s1];
out = ldimmq{result, in.d};
```

or even just

```
out = ldimmq{vals[in.s0] & vals[in.s1], in.d};
```

For such a concise expression, I don't think having the extra `imm0`, e.g., is worthwhile.

hphp/runtime/vm/jit/vasm-fold-imms.cpp
```diff
@@ -429,6 +435,14 @@ struct ImmFolder {
     else if (logical_bmsk(in.s1, bm)) { out = andqi64{bm, in.s0, in.d, in.sf}; }
   }
 
+  void fold(andqi& in, Vinstr& out) {
+    if (uses[in.sf]) return;
+    if (!valid.test(in.s1)) return;
+    auto imm64 = vals[in.s1];
+    auto result = imm64 & in.s0.q();
+    out = ldimmq{result, in.d};
```
mxw: Same as above; this could be compacted a bit.

swalk-cavium:
@mxw - Updated per inspection, ran MOP w/six option sets.  No additional failures.

PR #7921, opened 2017-07-21
Aarch64 - improve emitter for shl

swalk-cavium:
This improves the emitter for the shift left instructions in the case
where only the condition code is needed.  An immediate can be  calculated
at compile time which checks only the equivalent bits.  This saves 1 instruction 
per instance.

This happened to show up 82 times in the all_type_comparison_test.php
regression test.

Before
======
` 196: Lt `
`    (30) t9:Bool = ConvDblToBool t5:Dbl`
`      Main:   `
`          0x51001ec8  d37ff800              lsl x0, x0, #1`
 `         0x51001ecc  ea3f001f              bics xzr, x0, xzr `

After
=====
` 196: Lt `
`    (30) t9:Bool = ConvDblToBool t5:Dbl`
`      Main:   `
`          0x534023e8  f240f81f              tst x0, #0x7fffffffffffffff`

The standard regression tests were run with six option sets.  No new regressions
were seen.

### DIFF
--- a/hphp/runtime/vm/jit/vasm-arm.cpp
+++ b/hphp/runtime/vm/jit/vasm-arm.cpp
@@ -1080,9 +1080,8 @@ void Vgen::emit(const vasm_opc& i) {                         \
   } else {                                                   \
     checkSF(i, StatusFlags::NotV);                           \
     if (!flagRequired(i.fl, StatusFlags::C)) {               \
-      /* Perform the shift and set N and Z. */               \
-      a->arm_opc(gpr_w(i.d), gpr_w(i.s1), i.s0.l());         \
-      a->Bic(vixl::zr, gpr_w(i.d), vixl::zr, SetFlags);      \
+      /* Equivalent to shift and test, set N and Z. */       \
+      a->Tst(gpr_w(i.s1), Operand(~(-1LL << (sz - i.s0.l())))); \
     } else {                                                 \
       /* Use VIXL's macroassembler scratch regs. */          \
       a->SetScratchRegisters(vixl::NoReg, vixl::NoReg);      \


swalk-cavium:
@mxw - Hi Max,  Can you inspect?  Thanks.

jim-saxman:

mxw:
As with #7914, before I accept this, I'd like someone else to give a second opinion on the change (not just on test results) to confirm it doesn't affect the behavior of the instruction in any way.

mxw:
@dave-estes, @jim-saxman, @cmuellner, can any of you give a review to confirm this maintains behavior?

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -1080,9 +1080,8 @@ void Vgen::emit(const vasm_opc& i) {                         \
   } else {                                                   \
     checkSF(i, StatusFlags::NotV);                           \
     if (!flagRequired(i.fl, StatusFlags::C)) {               \
-      /* Perform the shift and set N and Z. */               \
-      a->arm_opc(gpr_w(i.d), gpr_w(i.s1), i.s0.l());         \
-      a->Bic(vixl::zr, gpr_w(i.d), vixl::zr, SetFlags);      \
+      /* Equivalent to shift and test, set N and Z. */       \
+      a->Tst(gpr_w(i.s1), Operand(~(-1LL << (sz - i.s0.l())))); \
```
jim-saxman: Hi @swalk-cavium , don't the shlli and shlqi require the destination i.d to be left shifted by i.s0 ?

swalk-cavium: @jim-saxman - Hi Jim,  this was the case where the actual result of the shift was the status flag.  If you look at the previous sequence the destination of the Bic was vixl::zr.  

I think you might be right in that we cannot know for certain that gpr_w(i.d) is dead.  It may be
that the ones I noticed were only using the status flags.

swalk-cavium:
Looks like the emitter is too late.  The sequence was introduced in cgConvDblToBool().
Maybe we can catch it earlier.

 62 void cgConvDblToBool(IRLS& env, const IRInstruction* inst) {
 63   auto const dst = dstLoc(env, inst, 0).reg();
 64   auto const src = srcLoc(env, inst, 0).reg();
 65   auto& v = vmain(env);
 66 
 67   auto const movtdq_res = v.makeReg();
 68   auto const sf = v.makeReg();
 69   v << movtdq{src, movtdq_res};
 70   // 0.0 stays zero and -0.0 is now 0.0
 71   v << shlqi{1, movtdq_res, v.makeReg(), sf};
 72   // lower byte becomes 1 if dst != 0
 73   v << setcc{CC_NE, sf, dst};
 74 }
runtime/vm/jit/irlower-conv.cpp

swalk-cavium:
I will resubmit a peephole change to catch this one.

PR #7922, opened 2017-07-21
Avoid race condition of PPC64 decoder

lbianc:
Avoided change data from m_decoder_table, where multiple threads are able
to make changes at the same time.


lbianc:
@mxw

mxw:
lgtm, will land after internal tests pass.

Issue #7564, opened 2016-12-22
HHVM 3.17.0 fails to build with  -DENABLE_ASYNC_MYSQL=OFF


paravoid:
It's currently not possible to build HHVM 3.17.0 with  -DENABLE_ASYNC_MYSQL=OFF (which we need in Debian as webscalesqlclient is too big to bring in as an embedded copy).

This seems to be caused (at least) by commit a14a87da9c4, which introduced a dependency to squangle and `facebook::common::mysql_client::SSLOptionsProviderBase`. squangle isn't enabled and thus not in the include path if async MySQL isn't (cf. third-party/CMakeLists.txt) and therefore the build fails due to the missing include:
```
if(ENABLE_ASYNC_MYSQL)
  find_package(RE2)
  if(NOT RE2_LIBRARY)
    message(STATUS "Building RE2 from third-party")
    list(APPEND THIRD_PARTY_MODULES re2)
  endif()
  list(APPEND THIRD_PARTY_MODULES squangle)
  list(APPEND THIRD_PARTY_MODULES webscalesqlclient)
endif()
```

paravoid:
Same with 3.18. Any ideas about that?

lexidor:
I am going over old issues on this repository, to see which ones apply to the current versions of hhvm.

AsyncMysqlClient is now unconditionally available to Hack. It is not possible to remove this extension anymore. The following code is always valid to hh_client, so hhvm should not allow this class to be removed from the runtime.

```Hack
$pool = AsyncMysqlConnectionPool(dict[]);
```

PR #7925, opened 2017-07-24
fix mysql extension compile error with -DENABLE_ASYNC_MYSQL=OFF

wwbmmm:
Fix issue #7564.


fredemmott:
Not sure why this hasn't landed, started a new test run.

Issue #3148, opened 2014-07-09
file_exists does_not_exist/../ does not work the same way as PHP5


huzhiguang:
replay way:

```
<?php
 $flag = file_exists("a/b/../a.txt");
 var_dump($flag);
 ?>
```

b folder is don't exists.

a.txt real location is a/a.txt 

php  exec result:

bool(false)

hhvm exec result:

bool(true)

jesseschalken:
Simple test: https://3v4l.org/Q59Ms

jesseschalken:
This is also broken in the case of symlinks: if `a/b` is a symlink, `a/b/..` refers to the parent of the symlink destination, not `a/..`.

Example:

```
$ mkdir test && cd test
$ mkdir a
$ mkdir a/b
$ touch a/c
$ ln -s a/b d
$ php -r 'var_dump(file_exists("./d/../c"));'
bool(true)
$ hhvm --php -r 'var_dump(file_exists("./d/../c"));'
bool(false)
$
```
PHP returns `true` (correct) while HHVM returns `false` because it checked `./c` instead of `./a/c`.

As far as I know, the only safe canonicalization that can be done to a POSIX path without consulting the file system is removing trailing `/` chars and collapsing multiple `/` into a single `/`. Replacing `/./` with `/` may also be safe, I'm not sure.

huzhiguang:
@jesseschalken ,we have fixed this bug,I provide PR can solve this problem,I have been verified.

binliu19:
@mofarrell Could you or someone look at @huzhiguang's PR here?  If I understand correctly, this brought a big win for Baidu.

lexidor:
Hi, I am reading HHVM issues to see which issues are unlikely to be fixed.

This issue was created because HHVM's behavior was/is different from PHP's. HHVM stopped striving from php compatibility with the release of version 4.0.0.

The current behavior, although different, is not buggy or broken. This very similar to this realpath #8298 issue. Hhvm will resolve `..` early. I have explained my rationale for closing the issue there.

PR #7932, opened 2017-07-26
fix file_exists relative path bug

huzhiguang:
a.txt real location is a/a.txt

file_exists("a/b/../a.txt");

this code return value is false ,but hhvm process relative path,so this result is true,diff php,and then old way process relative path performance have problem,this modify way performance is improve 40%

replay case:

#3148

### DIFF
--- a/hphp/runtime/ext/std/ext_std_file.cpp
+++ b/hphp/runtime/ext/std/ext_std_file.cpp
@@ -193,7 +193,15 @@ static int accessSyscall(
         return -1;
       }
     }
-    return ::access(File::TranslatePathWithFileCache(path).data(), mode);
+    String filepath = path;
+    if (path.charAt(0) != '/') {
+      String cwd = g_context->getCwd();
+      filepath = cwd + "/" + path;
+      if (!cwd.empty() && cwd[cwd.length() - 1] == '/') {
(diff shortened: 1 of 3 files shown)


hphp/runtime/ext/std/ext_std_file.cpp
```diff
@@ -193,7 +193,15 @@ static int accessSyscall(
         return -1;
       }
     }
-    return ::access(File::TranslatePathWithFileCache(path).data(), mode);
```
wwbmmm: I think this change will disable the StaticFileCache feature.

The cause of this bug is that FileUtil::canonicalize is used in File::TranslatePathWithFileCache and File::TranslatePathKeepRelative. So what's the purpose of FileUtil::canonicalize? It a path is canonicalized, the access() result may be different.

huzhiguang: I think FileCache has no meaning, please see the code (File: TranslatePathWithFileCache),the logic is :
`!translated.empty() && access(translated.data(), F_OK) < 0 && StaticContentCache::TheFileCache`

so the physical file does not exist, and then the file cache won't exist.
I don't think this code is meaningful.

PR #7933, opened 2017-07-26
Correction for tc-print build

swalk-cavium:
Correction for #7703 for case where libxed is NOT installed on x86_64 machine.

This was tested on Aarch64, and x86_64 with/without libxed.


gdbentley:
@swalk-cavium Line 243 of offline-x86-code.cpp should also be changed to #error rather than error.

Issue #1447, opened 2013-12-27
If php source is invoked when it's being created, HHVM will lost the complete code.


wwbmmm:
code to reproduce：

$ cat test.php

```
<?php
$file = 'code.php';
$fd = fopen($file, 'w+');
fwrite($fd, "<?php\n");
echo "Include when file is incomplete\n";
include($file);
fwrite($fd, "var_dump('Hello World');\n");
fclose($fd);

sleep(1); // sleep enough time for file reload

echo "Include when file is complete\n";
include($file);
```

$ hhvm test.php

```
Include when file is incomplete
Include when file is complete
```

$ rm code.php
$ php test.php

```
Include when file is incomplete
Include when file is complete
string(11) "Hello World"
```

When request this test.php in server mode, and then request code.php, you will get an empty response even after a long time, until code.php is modified or hhvm is restarted.

This may happen in actual scene online. For example, there is file a.php at the beginning and do the whole work. After some updates, file b.php is added and included in file a.php . During the online deployment, file a.php is modified first. When file b.php is being created, a request come up and invoke file a.php, that cause HHVM to parse and invoke the incomplete file b.php. After b.php is written completely, HHVM will not reload this file, so the incomplete b.php is always being used.

wwbmmm:
This bug occurs when file system is ext2. It doesn't occur when file system is ext4. Because ext4 file system has higher file stat time resolution.

lexidor:
I am going over old issues on this repository, to see which ones apply to the current versions of hhvm.

In hhvm 4.67 and above, it is not possible to include a file multiple times.
https://hhvm.com/blog/2020/07/20/hhvm-4.67.html
This version of hhvm removed support for top-level code / pseudo-mains. (Code which lives outside a function and executes upon inclusion.)

The only constructs that may appear at the top-level are named entities (classes, constants, functions, types, enums, interfaces). If included more than once, this will throw an `\Error` or a `fatal error:` depending on your configuration.

So in order to include a file more than once, it must have been empty (with the exception of comments / whitespace). If it weren't empty, this would error out on the second include.

PR #7942, opened 2017-07-31
Implement DelayedStatCache

wwbmmm:
Change-Id: I38fe9fe03c6fc81360ea27cd777c3f4348234f20

For non-RepoAuthoritative mode, the stat/access system call are hot functions. For the opensource mediawiki app, the stat() will cost about 2.5% cpu.

DelayedStatCache cache the result of stat/lstat/access/readlink/realpath system call in the memory for fast access. The result may not be the latest status of the file system. If the file system changes, the cache will update within `hhvm.server.delayed_stat_cache_ttl` seconds.

To enable this feature, add `hhvm.server.delayed_stat_cache=On` in you ini config. Set `hhvm.server.delayed_stat_cache_ttl` to a proper value.

After using DelayedStatCache, the cost of include() will be reduced by 30%. For the opensource mediawiki app, the total performance will be increased by about 2%. In some of the production at Baidu, the performance was increased by more than 5%.


wwbmmm:
@binliu19 please look at this

hphp/runtime/base/runtime-option.cpp
```diff
@@ -1400,6 +1403,9 @@ void RuntimeOption::Load(
     Config::Bind(ServerThreadDropStack, ini, config, "Server.ThreadDropStack");
     Config::Bind(ServerHttpSafeMode, ini, config, "Server.HttpSafeMode");
     Config::Bind(ServerStatCache, ini, config, "Server.StatCache", false);
+    Config::Bind(ServerFastStatCache, ini, config, "Server.FastStatCache", false);
+    Config::Bind(ServerFastStatCacheBucketNum, ini, config, "Server.FastStatCacheBucketNum", 0);
+    Config::Bind(ServerFastStatCacheTTL, ini, config, "Server.FastStatCacheTTL", 5);
```
binliu19: Clarify the unit, is the TTL in seconds?

hphp/test/slow/include/fast-stat-cache.php.expect
```diff
@@ -0,0 +1 @@
+OldOldNew
```
binliu19: This is a good example.  faststatcache changed the semantics of the language, right?  Without the faststatcache, I expect output 'OldNewNew' instead.  In a sense, this makes the implementation 'incorrect'.  Do you agree?

If that is the case, I think we need to document clearly that the option makes unsafe assumptions and should only be enabled when people know what they are doing.  Maybe rename the option to 'UnsafeStatCache' instead.

hphp/test/slow/include/fast-stat-cache.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+$file = tempnam('/tmp', 'test-fast-stat-cache');
+file_put_contents($file, "<?php\necho 'Old';");
+include($file);
+unlink($file);
+sleep(1);
```
binliu19: Also, if `sleep()` changes the output, it feels like a race condition.

hphp/runtime/base/runtime-option.cpp
wwbmmm: yes, it is in seconds

hphp/test/slow/include/fast-stat-cache.php.expect
wwbmmm: Yes, it changed the semantics. This should be clearly documented.

hphp/test/slow/include/fast-stat-cache.php
wwbmmm: sleep() is to wait for the stat cache to expire.

binliu19: I understand the intention of sleep() here.  However, it is racy because the operating system may decide to wake up the process later than the specified time.  In the unlikely case that the OS scheduling delay is more than a second, the result could be different from expected here (and this may happen without the first sleep()). 

If we clear the cache any time we make changes to the file system (create/delete a file, as opposed to checking existence), would things work better in this aspect?

hphp/test/slow/include/fast-stat-cache.php.expect
SiebelsTim: FWIW #1447 shows caching effects of included files.

hphp/runtime/base/runtime-option.cpp
wwbmmm: I had changed ServerFastStatCacheTTL to ServerFastStatCacheExpireSeconds in the new commit.

hphp/test/slow/include/fast-stat-cache.php
wwbmmm: I add a new commit that reduce a sleep() call.

_If we clear the cache any time we make changes to the file system (create/delete a file, as opposed to checking existence), would things work better in this aspect?_

I had considered this idea, but there are so many ways to change a file. File system functions like fwrite, other extension function like Imagick::writeImage, and even another process may change the file. Inotify can handle all this case(that's what StatCache do), but it's slow, after turning on Server.StatCache the performance decrease by 5% on wordpress.

hphp/test/slow/include/fast-stat-cache.php.expect
wwbmmm: #1447 is because the file system mtime resolution is seconds in the old ext2 file system, so that a file may be changed but it's mtime if not changed. I had added a PR to fix this: #7926.

wwbmmm: 'UnsafeStatCache' is a little strange, maybe 'DelayedStatCache'?

SiebelsTim: I just wanted to point out that there is already an unpredictable behaviour considering included files.

binliu19: @wwbmmm 'DelayedStatCache' is fine.  Can you add some documentation about the option in hphp/doc/options.compiled?  Especially about the semantic change and when the option is helpful.

I think if you update the source code without restarting HHVM, having this cache may help consistency.  If the file updates all happen within the cache window, HHVM will either see the set of old files, or the set of new files, but never a mixture of both.

wwbmmm: All right, I will add some documentation.

Yes, it helps consistency. And we add a clear cache command in Admin Server, after we deploy new files, we can call this command at once, so the cache will update as soon as possible, without the disadvantage of delay.

wwbmmm: I have add a new commit that renamed it to DelayedStatCache and add some documentation.

fredemmott:
Closing as stale; please comment if you want it re-opened.

PR #7944, opened 2017-08-02
Fix PPC64 compilation

lbianc:
After the PR https://github.com/facebook/hhvm/pull/7703, the method
getArchName was missing in the offline-ppc64-code.cpp file.
The missing method was added and the return value for collectJmpTargets as
well, in order to avoid warning when compiling.


lbianc:
@mxw

PR #7947, opened 2017-08-03
[AArch64] Shrinks embedded TCA targets to 32 bits.

dave-estes:
Smashable jmps/jccs and patchable jmps/jccs embed the target in
the instruction stream. When using low memory for the translation
cache, those targets only need to use 32 bits to store the TCA as
a literal in the instruction stream. This reduces D$ pressure.

### DIFF
--- a/hphp/runtime/vm/jit/relocation-arm.cpp
+++ b/hphp/runtime/vm/jit/relocation-arm.cpp
@@ -137,8 +137,9 @@ InstrSet findLiterals(Instruction* start, Instruction* end) {
           if (lit->IsLoadLiteral()) {
             auto oops = Instruction::Cast(lit->LiteralAddress());
             literals.erase(oops);
-            if (lit->Mask(LoadLiteralMask) == LDR_x_lit)
+            if (lit->Mask(LoadLiteralMask) == LDR_x_lit) {
               literals.erase(oops->NextInstruction());
+	    }
           }
         }
       };
@@ -720,12 +721,13 @@ size_t relocateImpl(Env& env) {
            */
           if (src->IsLoadLiteral()) {
             if (src->Mask(LoadLiteralMask) == LDR_w_lit) {
-              auto addr = reinterpret_cast<TCA32*>(dest->LiteralAddress());
+              auto addr = reinterpret_cast<uint32_t*>(dest->LiteralAddress());
               auto target = *addr;
               auto adjusted =
 		env.rel.adjustedAddressAfter(reinterpret_cast<TCA>(target));
-              if (adjusted) patchInstr(reinterpret_cast<TCA>(addr),
-				       makeTCA32(adjusted));
+              if (adjusted) {
+		patchTarget32(reinterpret_cast<TCA>(addr), adjusted);
+	      }
             } else {
               auto addr = reinterpret_cast<TCA*>(dest->LiteralAddress());
               auto target = *addr;
@@ -734,10 +736,14 @@ size_t relocateImpl(Env& env) {
                 // Consider the case of a non-initialized mcprep smashableMovq
                 target = reinterpret_cast<TCA>((uint64_t(target) >> 1));
                 adjusted = env.rel.adjustedAddressAfter(target);
-                if (adjusted) adjusted =
-				reinterpret_cast<TCA>((uint64_t(adjusted) << 1)
-						      | 1);
-                if (adjusted) patchInstr(reinterpret_cast<TCA>(addr), adjusted);
+                if (adjusted) {
+		  adjusted = reinterpret_cast<TCA>(
+		    (uint64_t(adjusted) << 1) | 1
+		  );
+		}
+		if (adjusted) {
+		  patchTarget64(reinterpret_cast<TCA>(addr), adjusted);
+		}
               }
             }
           } else if (src->IsMovz()) {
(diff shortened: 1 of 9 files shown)


dave-estes:
FYI: @mxw, @swalk-cavium, @jim-saxman.

hphp/runtime/vm/jit/unique-stubs-arm.cpp
```diff
@@ -214,10 +215,10 @@ TCA emitCallToExit(CodeBlock& cb, DataBlock& data, const UniqueStubs& us) {
   auto const start = cb.frontier();
 
   // Jump to enterTCExit
-  a.Ldr(rAsm, &target_data);
+  a.Ldr(rAsm_w, &target_data);
   a.Br(rAsm);
   a.bind(&target_data);
-  a.dc64(us.enterTCExit);
+  a.dc64(makeTCA32(us.enterTCExit));
```
dave-estes: Should be a dc32. Will fix.

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -692,23 +694,26 @@ size_t relocateImpl(Env& env) {
...
+              auto addr = reinterpret_cast<TCA*>(dest->LiteralAddress());
+              auto target = *addr;
+              auto adjusted = env.rel.adjustedAddressAfter(target);
+              if (!adjusted) {
+                // Consider the case of a non-initialized mcprep smashableMovq
+                target = reinterpret_cast<TCA>((uint64_t(target) >> 1));
+                adjusted = env.rel.adjustedAddressAfter(target);
+                if (adjusted) adjusted =
+				reinterpret_cast<TCA>((uint64_t(adjusted) << 1)
+						      | 1);
```
mxw: I think this is more readable as

```
if (adjusted) {
  adjusted = reinterpret_cast<TCA>(
    ...
  );
}
```

hphp/runtime/vm/jit/smashable-instr-arm.h
```diff
@@ -77,6 +77,22 @@ constexpr size_t kSmashJccTargetOff = 12;
...
 
+template<typename T>
+  static void patchInstr(TCA inst, T target) {
+  *reinterpret_cast<T*>(inst) = target;
+  auto const end = reinterpret_cast<char*>(inst + sizeof(T));
+  auto const begin = end - sizeof(T);
+  __builtin___clear_cache(begin, end);
+}
+
+using TCA32 = uint32_t;
```
mxw: This is pretty weird, since `TCA` is a pointer type.  Can you just use `uint32_t` directly?  Many of the places you're using this type call something generic like `patchInstr()` with it, which really just expects any 32-bit immediate.

hphp/runtime/vm/jit/smashable-instr-arm.h
```diff
@@ -77,6 +77,22 @@ constexpr size_t kSmashJccTargetOff = 12;
...
+
+template <class T>
+static uint32_t makeTCA32(T t) {
```
mxw: You've declared these functions in a header file, so what's the point of the `static`?

hphp/runtime/vm/jit/smashable-instr-arm.cpp
```diff
@@ -272,18 +268,20 @@ TCA smashableJccTarget(TCA inst) {
...
-      ldr->Mask(LoadLiteralMask) == LDR_x_lit &&
+      ldr->Mask(LoadLiteralMask) == LDR_w_lit &&
       ldr->ImmPCOffsetTarget() == target &&
       br->Mask(UnconditionalBranchToRegisterMask) == BR &&
       br->Rn() == rd) {
     assertx((reinterpret_cast<uintptr_t>(target) & 3) == 0);
-    return *reinterpret_cast<TCA*>(target);
+    const uint32_t target32 = *reinterpret_cast<uint32_t*>(target);
+    const uint64_t target64 = target32;
+    return reinterpret_cast<TCA>(target64);
```
mxw: Can you not reinterpret_cast a `uint32_t` directly as a `TCA`?

hphp/runtime/vm/jit/smashable-instr-arm.h
```diff
@@ -77,6 +77,22 @@ constexpr size_t kSmashJccTargetOff = 12;
 
 ///////////////////////////////////////////////////////////////////////////////
 
+template<typename T>
+  static void patchInstr(TCA inst, T target) {
+  *reinterpret_cast<T*>(inst) = target;
+  auto const end = reinterpret_cast<char*>(inst + sizeof(T));
+  auto const begin = end - sizeof(T);
+  __builtin___clear_cache(begin, end);
+}
```
mxw: I think it's be nicer to have differently-named versions of this for `uint64_t` and `uint32_t`, since we really want to be clear which one we're using at every point.

hphp/runtime/vm/jit/smashable-instr-arm.h
mxw: Also, I think you could probably get away with just doing a `safe_cast<uint32_t>` instead of having this function at all.

mxw:
I came here thinking I'd forgotten to review this, but fortunately past-me had my back.  I think this needs rebasing over your recent cache-syncing refactor, but otherwise is good to go modulo comments.

dave-estes:
haha...yep yep, I'm working in rebasing now and then addressing the comments.

hphp/runtime/vm/jit/smashable-instr-arm.h
mxw: Why would including the header in multiple files require `static` for an inline template function?  Maybe I'm being dense.

dave-estes: Actually, yeah. I didn't realize that since this was a template'd function that it didn't doesn't need to be explicitly marked as static or inline to avoid a linking problem. There are two other functions there, that aren't template'd and I did need to make them static. I'm about to push a patch which makes all three explicitly inline. Let me know what you think?

PR #7954, opened 2017-08-07
Fixes for building with icc.

gdbentley:


CMake/HPHPCompiler.cmake
```diff
@@ -36,7 +36,7 @@ CHECK_CXX_SOURCE_COMPILES("
 int main() { return 0; }" IS_PPC64)
 
 # using Clang or GCC
-if (${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang" OR ${CMAKE_CXX_COMPILER_ID} STREQUAL "GNU")
+if (${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang" OR ${CMAKE_CXX_COMPILER_ID} STREQUAL "GNU" OR ${CMAKE_CXX_COMPILER_ID} STREQUAL "Intel")
```
Orvid: You'll want to drop the other Intel section if it's no longer going to be used.

Wilfred:
@Orvid this looks OK to me now, any objections to merging?

hphp/runtime/vm/native.h
```diff
@@ -231,6 +231,7 @@ bool coerceFCallArgs(TypedValue* args,
  * calling.
  */
 template<bool usesDoubles>
+__attribute__((noinline))
```
Wilfred: Why is this changing? I would have thought changing how we inline `callFunc` could impact performance.

alshamiri5 (approved):

PR #7955, opened 2017-08-08
Fix DateTime constructor arguments

bcopeland:
According to php documentation, first DateTime argument is a
nullable string, not mixed.  Update the prototype accordingly
so that non-strings passed into DateTime constructor will be
flagged by typechecker.

### DIFF
--- a/hphp/hack/hhi/stdlib/builtins_datetime.hhi
+++ b/hphp/hack/hhi/stdlib/builtins_datetime.hhi
@@ -100,7 +100,7 @@ class DateTime implements DateTimeInterface {
   const RFC3339 = '';
   const RSS = '';
   const W3C = '';
-  public function __construct(mixed $time = 'now', ?DateTimeZone $timezone = null);
+  public function __construct(string $time = 'now', ?DateTimeZone $timezone = null);
   public function add(DateInterval $interval);
   public function modify(string $modify);
   public function getOffset();


bcopeland:
Signed the CLA just then.

SiebelsTim:
Seems to be that this cannot be null. It's rather a string with a default argument set to `"now"`
https://3v4l.org/53s19

bcopeland:
Hmm, it works in php5 mode though, according to that link.  Php docs (http://php.net/manual/en/datetime.construct.php) says:

"Enter NULL here to obtain the current time when using the $timezone parameter."

I don't really care about nullability, though, so I can resend without the '?' if you want.  I just ran into an issue in our project where someone passed a DateTime which only got caught at runtime and would like to catch that.

SiebelsTim:
You are right. I just skimmed through the documentation, looking at the function signature.
Funny enough, PHP7 gets this wrong https://3v4l.org/DKdFT

SiebelsTim:
The PHP guys updated the docs. NULL is not supposed to be  passed

https://bugs.php.net/bug.php?id=75052

bcopeland:
Very well, dropped the '?'

bcopeland:
Oops I better fix commitlog too

SiebelsTim:
Thanks :)

bcopeland:
Just dropping this as I'm not working in hack these days and don't want to spend add'l time on it.

PR #7956, opened 2017-08-08
Aarch64,ppc64 - avoid reintroducing previously lowered instructions

swalk-cavium:
Commit 66f6498 introduced a simplification load/cmp -> cmpm.  Unfortunately,
those instructions do not exist on all hosts.  They had already been lowered for
Aarch64 and PPC64.

This change introduces the obvious correction to get hhvm running on all targets again.
Suggestions for refactoring are welcome.

The standard regression tests were run with six option sets.  No new failures were detected.


swalk-cavium:
@mxw - Hi Max, Can you comment?

edwinsmith:
Can the arch-test be in foldable_load() instead?

swalk-cavium:
@edwinsmith - I think some of these are valid xforms for PPC64.  Sinking the test 
would loose that opportunity.

PR #7959, opened 2017-08-10
Fix smashable instructions in PPC64

lbianc:
Currently on PPC64 architecture all immediate are loaded using up to 5
instructions, which makes the read not atomic.
The smashable instructions may have its immediate changed on the fly, i.e.,
while running or re-emitting the JITted code, which causes race condition,
if reading the immediate is composed with more than one instruction.
The race condition happen at two different moments:

1. The JITted code is being executed and the pipeline stop reading the
instructions group that reads the immediate in the middle (thread
preemption), then another thread smashes the instruction that uses the same
immediate.

2. The Decoder is reading the immediate used in a smashable instruction and
this thread stop being executed (thread preemption), then another thread
smashes the instruction that uses the same immediate.

For the reason above, this PR is being sent to enable the concept of TOC
(Table Of Contents) in the JITted code, where all immediate are save in
the "TOC" area, keeping the entire value in a slot of 64 bits, making the
read/write atomic, as only one operation is needed.
The TOC memory is an vector, where the pointer to it will be placed in the
R2 register, using up to two instructions to calculate its offset, which
keeps fixed, as only the position is TOC memory is changes when smashing.

### DIFF
--- a/hphp/ppc64-asm/asm-ppc64.cpp
+++ b/hphp/ppc64-asm/asm-ppc64.cpp
@@ -39,7 +39,7 @@ VMTOC::~VMTOC() {
 
 int64_t VMTOC::pushElem(int64_t elem, bool elemMayChange) {
   int64_t offset;
-  if(elemMayChange) {
+  if (elemMayChange) {
     offset = allocTOC(elem);
   }
   else {
--- a/hphp/ppc64-asm/asm-ppc64.cpp
+++ b/hphp/ppc64-asm/asm-ppc64.cpp
@@ -19,35 +19,37 @@
 #include "hphp/ppc64-asm/decoder-ppc64.h"
 #include "hphp/runtime/base/runtime-option.h"
 #include "hphp/util/trace.h"
+#include <folly/MicroSpinLock.h>
 
 #include <type_traits>
 
 TRACE_SET_MOD(asmppc64);
 
 namespace ppc64_asm {
+
+// Lock to protect TOC when writing.
+static folly::MicroSpinLock s_TOC;
+
+//////////////////////////////////////////////////////////////////////
+
 VMTOC::~VMTOC() {
-  FTRACE(1, "Number of values stored in TOC: {}\n",
+  FTRACE(1, "Number of values if 64bits stored in TOC: {}\n",
     std::to_string(m_last_elem_pos));
 }
 
-int64_t VMTOC::pushElem(int64_t elem) {
-  auto& map_elem = m_map[elem];
-  if (map_elem) return map_elem;
+int64_t VMTOC::pushElem(int64_t elem, bool elemMayChange) {
+  int64_t offset;
+  if(elemMayChange) {
+    offset = allocTOC(elem);
+  }
+  else {
+    auto& map_elem = m_map[elem];
+    if (map_elem) return map_elem;
+    offset = allocTOC(elem);
+    map_elem = offset;
+  }
 
-  auto offset = allocTOC(static_cast<int32_t>(elem & 0xffffffff), true);
-  map_elem = offset;
-  allocTOC(static_cast<int32_t>((elem & 0xffffffff00000000) >> 32));
-  m_last_elem_pos += 2;
-  return offset;
-}
-
-int64_t VMTOC::pushElem(int32_t elem) {
-  auto& map_elem = m_map[elem];
-  if (map_elem) return map_elem;
-
-  auto offset = allocTOC(elem);
-  map_elem = offset;
-  m_last_elem_pos++;
+  m_last_elem_pos += 1;
   return offset;
 }
 
@@ -72,15 +74,15 @@ int64_t VMTOC::getValue(int64_t index, bool qword) {
   return ret_val;
 }
 
-int64_t VMTOC::allocTOC(int32_t target, bool align) {
-  HPHP::Address addr = m_tocvector->frontier();
-  if (align) {
-    forceAlignment(addr);
-    always_assert(reinterpret_cast<uintptr_t>(addr) % 8 == 0);
-  }
+uint64_t* VMTOC::getAddr(int64_t index) {
+  return reinterpret_cast<uint64_t*>(
+      static_cast<intptr_t>(index) + getPtrVector());
+}
 
-  m_tocvector->assertCanEmit(sizeof(int32_t));
-  m_tocvector->dword(target);
+int64_t VMTOC::allocTOC(int64_t target) {
+  folly::MSLGuard g{s_TOC};
+  HPHP::Address addr = m_tocvector->frontier();
+  m_tocvector->qword(target);
   return addr - (m_tocvector->base() + INT16_MAX + 1);
 }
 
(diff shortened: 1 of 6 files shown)


hphp/ppc64-asm/asm-ppc64.cpp
```diff
@@ -19,33 +19,35 @@
...
+  FTRACE(1, "Number of values if 64bits stored in TOC: {}\n",
+    std::to_string(m_last_elem_pos));
 }
 
-int64_t VMTOC::pushElem(int32_t elem) {
-  auto& map_elem = m_map[elem];
-  if (map_elem) return map_elem;
+int64_t VMTOC::pushElem(int64_t elem, bool elemMayChange) {
+  int64_t offset;
+  if(elemMayChange) {
```
mxw: `if (elemMayChange) {`

hphp/runtime/vm/jit/smashable-instr-ppc64.cpp
```diff
@@ -73,74 +73,59 @@ TCA emitSmashableCall(CodeBlock& cb, CGMeta& fixups, TCA target,
...
-  if (target > inst && target - inst <= smashableJmpLen()) {
-    a.emitNop(target - inst);
-  } else {
-    a.branchFar(target);
-  }
+  // Smash TOC value
+  const DecodedInstruction di(inst);
+  assertx(di.isLoadingTOC() && di.isSmashable(reinterpret_cast<uint64_t>(target)));
+  uint64_t* imm_address = di.decodeTOCAddress();
+  if(imm_address) *imm_address = reinterpret_cast<uint64_t>(target);
```
mxw: `if (imm_address)`

hphp/runtime/vm/jit/smashable-instr-ppc64.cpp
```diff
@@ -73,74 +73,59 @@ TCA emitSmashableCall(CodeBlock& cb, CGMeta& fixups, TCA target,
 }
 
 TCA emitSmashableJmp(CodeBlock& cb, CGMeta& fixups, TCA target) {
-  return EMIT_BODY(cb, fixups, branchFar, target);
+  return EMIT_BODY(cb, fixups, branchFar, target, ppc64_asm::BranchConditions::Always,
+                   ppc64_asm::LinkReg::DoNotTouch, ppc64_asm::ImmType::TocOnly, true);
```
mxw: Our linter requires all lines to be <= 80 chars.  Please reflow all over-long lines in this file (and others you've touched).

hphp/ppc64-asm/asm-ppc64.h
```diff
@@ -261,10 +255,21 @@ struct VMTOC {
...
+  uint64_t* getAddr(int64_t index);
+
+  /*
+   * Return TOC index of the element.
+   */
+  int64_t getIndex(uint64_t elem);
 
 private:
-  int64_t allocTOC (int32_t target, bool align = false);
+  //int64_t allocTOC (int32_t target, bool align = false);
```
mxw: Delete this line.

hphp/ppc64-asm/asm-ppc64.h
```diff
@@ -784,11 +789,12 @@ struct Assembler {
...
     if (CallArg::Internal == ca) {
       // tries best performance possible
-      branchAuto(target, BranchConditions::Always, LinkReg::Save);
+      branchAuto(target, BranchConditions::Always, LinkReg::Save, true);
     } else {
       // branchFar is not only smashable but also it will use r12 so an
       // external branch can correctly set its TOC address appropriately.
-      branchFar(target, BranchConditions::Always, LinkReg::Save);
+      branchFar(target, BranchConditions::Always, LinkReg::Save,
+                ImmType::TocOnly, true);
```
mxw: Why are calls always `immMayChange`?

hphp/ppc64-asm/decoded-instr-ppc64.cpp
```diff
@@ -425,6 +410,51 @@ uint8_t DecodedInstruction::decodeImm() {
...
+    }
+  }
+  return 0;
+
+}
+
+bool DecodedInstruction::isSmashable(uint64_t elem) const {
+  auto elemOffset = decodeTOCOffset();
+  auto offset = VMTOC::getInstance().getIndex(elem);
+  return offset!=elemOffset;
```
mxw: `offset != elemOffset`

hphp/ppc64-asm/decoded-instr-ppc64.cpp
```diff
@@ -438,6 +468,13 @@ bool DecodedInstruction::isLimmediatePossible() const {
   return isLi64Possible();
 }
 
+bool DecodedInstruction::isLoadingTOC() const {
+  if (m_dinfo.isLd(true) || m_dinfo.isLwz(true) || m_dinfo.isAddis(true)) {
+    return true;
+  }
+  return false;
```
mxw: `return m_dinfo.isLd(true) || ...;`

You don't need to check a boolean for true; just return it.

hphp/ppc64-asm/asm-ppc64.cpp
mxw: Also—can you explain to me what is the consequence of not caching the offset in `m_map`?

hphp/ppc64-asm/asm-ppc64.h
lbianc: The ```immMayChange``` is used to optimize the TOC allocation, avoiding duplicated values on it, but it cannot be used to immediate that may be changed when patching or smashing.
In this case I was defensive, forcing all calls to allocate one TOC slot, avoiding be referenced by another part of the code that may change it.

hphp/ppc64-asm/asm-ppc64.cpp
lbianc: Not caching means that the just added value may not be shared with another part of the code. That is necessary to avoid a part of the code to reference the same location of a smashable or patchable location.

hphp/ppc64-asm/asm-ppc64.h
lbianc: Done.

hphp/ppc64-asm/decoded-instr-ppc64.cpp
lbianc: Done.

hphp/ppc64-asm/decoded-instr-ppc64.cpp
lbianc: Done.

hphp/runtime/vm/jit/smashable-instr-ppc64.cpp
lbianc: Done.

hphp/runtime/vm/jit/smashable-instr-ppc64.cpp
lbianc: Done.

lbianc:
@mxw This PR and two others are taking too long to be merged, is there anything that I could help?

PR #7966, opened 2017-08-18
Add junit xml output to test script

gut:
Use hphp/test/run --xml to get an junit-like xml output. Useful on e.g
Jenkins as it can plot those results to a time series, so a regression
is easily spotted.

### DIFF
--- a/hphp/test/run
+++ b/hphp/test/run
@@ -496,6 +496,7 @@ function get_options($argv) {
     'verbose' => 'v',
     'fbmake' => '',
     'testpilot' => '',
+    'xml' => '',
     'threads:' => '',
     'args:' => 'a:',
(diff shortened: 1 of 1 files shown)


gut:
@mxw or whoever can review this

fredemmott:
Can you use a more generic format instead (eg Test Anything Protocol/TAP)? I don't like the idea of adding an option that's meant to emulate a tool that none of us use.

If there isn't a more tool-agnostic format that can be used instead, please rename the option to --junit-xml or similar, not just --xml, in case we have other needs in the future.

gut:
Ok, I'll take a closer look. Thanks for the hints

alshamiri5 (approved):

PR #7973, opened 2017-08-26
[AArch64] Sign-extend before cmp instructions.

dave-estes:
Additionally fix immediate casting. Immediates are stored in a
signed 32-bit int, which cannot be safe_casted to an unsigned int
using Boost.

### DIFF
--- a/hphp/runtime/vm/jit/vasm-arm.cpp
+++ b/hphp/runtime/vm/jit/vasm-arm.cpp
@@ -24,11 +24,15 @@
  *  A Vreg64 is a X-register with a u64 value.
  *
  * This allows to omit truncation instructions for sub-32-bit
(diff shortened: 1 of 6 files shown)


dave-estes:
@mxw, This fixes the invalid cast that we're seeing in OSS Mediawiki as well as the changes to sign extend cmp[bw][i]{}.

swalk-cavium:
@dave-estes - Testing results?

dave-estes:
Just rebased this PR. It has changes to PPC64 after renaming some vasm. Builds fine, but I haven't tested on PPC64.

dave-estes:
@mxw, @swalk-cavium: FYI, once #7947 lands you may start seeing an assert in Debug builds on OSS mediawiki. This PR addresses that.

swalk-cavium:
@dave-estes - Thanks for the warning.  I usually only do release builds with oss-performance
so I would have missed it.

PR #7978, opened 2017-08-30
Fix retranslate all feature for PPC64

lbianc:
The concept of resizable buffers was not added to PPC64 specific files,
which is used for the retranslate all feature.
The decoder was changed to receive the destination and buffer addresses,
and the calls for it, in the relocation method, was changed to add the
buffer address.
Also methods to obtain the buffer address, based on the destination code
address, were added.


lbianc:
@mxw

lbianc:
@lbianc Ok to test.

lbianc:
@paulbiss could you review this?

PR #7983, opened 2017-09-02
Refactor use of __builtin__clear_cache and build into DataBlock.

dave-estes:


dave-estes:
@mxw, this is the portion of #7862 that I split out.

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -877,8 +874,8 @@ void adjustInstruction(RelocationInfo& rel, Instruction* instr,
     auto adjusted = rel.adjustedAddressAfter(target);
     if (adjusted) {
       *addr = adjusted;
-      __builtin___clear_cache(reinterpret_cast<char*>(addr),
-                              reinterpret_cast<char*>(addr) + 8);
+      auto const begin = reinterpret_cast<TCA>(addr);
+      DataBlock::syncDirect(begin, begin + 8);
```
mxw: Should we use `2 * vixl::kInstructionSize` here and above where we're using `8` directly?

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -422,35 +417,48 @@ struct Vgen {
...
+  }
+  return nullptr;
+}
+
+static CodeAddress toReal(Venv& env, CodeAddress a) {
+  CodeBlock* b = getBlock(env, a);
+  if (b) {
+    return b->toDestAddress(a);
   }
+
```
mxw: `return b == nullptr ? a : b->toDestAddress(a);`

PR #7984, opened 2017-09-02
Adds option for faster, estimated PGO counters.

dave-estes:
Benchmarks with small amounts of code generate a lot of contention for that code's PGO counters. Using a `decqm` instead of a `decqmlock` removes the lock, easing the contentions but also leading to accuracy issues. This new switch (`Eval.JitPGORacyProfiling`) changes the counters to use `decqm` if accuracy isn't required.

### DIFF
--- a/hphp/runtime/base/runtime-option.cpp
+++ b/hphp/runtime/base/runtime-option.cpp
@@ -538,7 +538,8 @@ static inline uint32_t profileBCSizeDefault() {
 }
 
 static inline uint32_t resetProfCountersDefault() {
-  return RuntimeOption::EvalJitPGORacyProfiling ? UINT32_MAX
+  return RuntimeOption::EvalJitPGORacyProfiling
+    ? std::numeric_limits<uint32_t>::max()
     : RuntimeOption::EvalJitConcurrently ? 250 : 1000;
 }
 
--- a/hphp/runtime/base/runtime-option.cpp
+++ b/hphp/runtime/base/runtime-option.cpp
@@ -538,7 +538,8 @@ static inline uint32_t profileBCSizeDefault() {
 }
 
 static inline uint32_t resetProfCountersDefault() {
-  return RuntimeOption::EvalJitConcurrently ? 250 : 1000;
+  return RuntimeOption::EvalJitPGORacyProfiling ? UINT32_MAX
+    : RuntimeOption::EvalJitConcurrently ? 250 : 1000;
 }
 
 static inline int retranslateAllRequestDefault() {
--- a/hphp/runtime/base/runtime-option.h
+++ b/hphp/runtime/base/runtime-option.h
@@ -656,6 +656,7 @@ struct RuntimeOption {
   F(uint32_t, JitPGORelaxUncountedToGenPercent, 20)                     \
   F(uint32_t, JitPGORelaxCountedToGenPercent, 75)                       \
   F(bool,     JitPGODumpCallGraph,     false)                           \
+  F(bool,     JitPGOFastProfiling,     false)                           \
   F(uint64_t, FuncCountHint,           10000)                           \
   F(uint64_t, PGOFuncCountHint,        1000)                            \
   F(uint32_t, HotFuncCount,            4100)                            \
--- a/hphp/runtime/base/runtime-option.h
+++ b/hphp/runtime/base/runtime-option.h
@@ -656,7 +656,7 @@ struct RuntimeOption {
   F(uint32_t, JitPGORelaxUncountedToGenPercent, 20)                     \
   F(uint32_t, JitPGORelaxCountedToGenPercent, 75)                       \
   F(bool,     JitPGODumpCallGraph,     false)                           \
-  F(bool,     JitPGOFastProfiling,     false)                           \
+  F(bool,     JitPGORacyProfiling,     false)                           \
   F(uint64_t, FuncCountHint,           10000)                           \
   F(uint64_t, PGOFuncCountHint,        1000)                            \
   F(uint32_t, HotFuncCount,            4100)                            \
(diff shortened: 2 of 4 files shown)


dave-estes:
@mxw, @swalk-cavium, this patch provides a runtime option to replace the use of decqmlock{} with decqm{} for PGO counters. These counters have caused problems for ARM in the past, hence the change to use ARM atomics extensions for decqmlock{} (#7857) and the change to minimize cache collisions (#7801).

The retranslate all feature runs PGO translations for 3000 requests collecting profiles until it retranslates all of the translations. While this does have an impact on all of OSS, the impact for most of the targets is just longer warmup periods. However, it can be incredibly slow for fibonacci because all 200 concurrent users are executing code in tight loops which are vying for only a few counters. On some older machines, we've seen siege timeout while the requests are still processing and then kick off 200 more make the problem even worse with 400, 600, 800...concurrent requests.

This option is defaulted to off. We experimented with it on, and less accurate PGO counters don't affect the scores on OSS. We recommend other folks try it on different architectures to see if it helps their warmup times without affecting their scores.

mxw:
This PR includes tabs, but shouldn't.

dave-estes:
Doh, sorry. Fixed.

hphp/runtime/base/runtime-option.h
```diff
@@ -637,6 +637,7 @@ struct RuntimeOption {
   F(uint32_t, JitPGORelaxUncountedToGenPercent, 20)                     \
   F(uint32_t, JitPGORelaxCountedToGenPercent, 75)                       \
   F(bool,     JitPGODumpCallGraph,     false)                           \
+  F(bool,     JitPGOFastProfiling,     false)                           \
```
mxw: @ottoni: nit: I'd prefer "Racy" instead of "Fast"

hphp/runtime/vm/jit/prof-data.h
```diff
@@ -358,8 +358,10 @@ struct ProfData {
   int64_t transCounter(TransID id) const {
     folly::SharedMutex::ReadHolder lock{m_transLock};
     assertx(id < m_transRecs.size());
-    auto const counter = m_counters.get(id);
+    auto counter = m_counters.get(id);
     auto const initVal = m_counters.getDefault();
+    if (RuntimeOption::EvalJitPGOFastProfiling &&
+        (counter > initVal)) counter = 0;
```
mxw: You have a tab here.  Please try to convert tabs to spaces on all files before putting up a PR (or maybe configure your editor to always use spaces).

mxw: @ottoni: I don't see how this can happen. Even with racy decrements, counter should never be larger than initVal.

dave-estes:
@mxw, @ottoni, @paulbiss: Hey folks. I finally got back to addressing the review feedback. Initially I thought the consensus was that I should make the counters thread local. Seemed like a fair amount of work, so I postponed it. Now that I'm reading the comments more closely, they read like my initial command-line option is reasonable after addressing the code-specific comments from @mxw and @ottoni. Is this correct? Thanks a bunch...

ottoni:
@dave-estes, yep, I think this is mostly OK. However, other than my previous comments, one more issue occurred to me since I originally commented on this change. The JIT resets these profile counters after a given number of requests (controlled by Eval.JitResetProfCountersRequest) to avoid large discrepancies due to when a given block gets JITed (which can be significant due to locking). Doing non-atomic operations may cause the reseting of those counters to be lost. So, when using these racy counters, it may be better to disable the reset of the counters by setting a very large value here:
https://github.com/facebook/hhvm/blob/master/hphp/runtime/base/runtime-option.cpp#L546

dave-estes:
Thanks a bunch, @ottoni. Will push a new patch with the changes soon-ish.

hphp/runtime/vm/jit/prof-data.h
```diff
@@ -358,8 +358,12 @@ struct ProfData {
   int64_t transCounter(TransID id) const {
     folly::SharedMutex::ReadHolder lock{m_transLock};
     assertx(id < m_transRecs.size());
-    auto const counter = m_counters.get(id);
+    auto counter = m_counters.get(id);
     auto const initVal = m_counters.getDefault();
+    // RacyProfiling can lead to counters underflowing, so reset to zero.
+    if (RuntimeOption::EvalJitPGORacyProfiling && (counter > initVal)) {
```
mxw: These parens are spurious.

hphp/runtime/base/runtime-option.cpp
```diff
@@ -538,7 +538,8 @@ static inline uint32_t profileBCSizeDefault() {
 }
 
 static inline uint32_t resetProfCountersDefault() {
-  return RuntimeOption::EvalJitConcurrently ? 250 : 1000;
+  return RuntimeOption::EvalJitPGORacyProfiling ? UINT32_MAX
```
ottoni: C++ nit: use `numeric_limits<uint32_t>::max()` instead of `UINT32_MAX`

hphp/runtime/vm/jit/prof-data.h
ottoni: Thinking more about this, I believe you were hitting it because of the counter resetting logic not always working due to the racy updates. Now that you disabled counter resetting when racy counters are on, I don't see how you can hit this. Can you please try to remove this logic and retest? And if you still hit the assert below, we should investigate it further. Simply setting the counter to 0 here will silently throw the counters way off, which can hurt performance.

dave-estes: I added an assert(false) within this underflow correction code, and both OSS and unit tests pass without hitting it. I then removed the code that disables counter reseting, and it still passes. I can't reproduce the underflow situation any longer. This puzzles me, but since the behavior fits what you expect, @ottoni, then I'm OK with just removing this underflow correction code completely. If the existing assert starts popping in Debug[Opt] builds, then I'll revisit.

Issue #20, opened 2010-02-21
Missing sources in generated CMakeLists.txt


carsonmcdonald:
When generating from php that contains classes it appears that the resulting CMakeLists.txt file is missing a couple directories needed to compile. Compiling the source from gist demonstrates the issue:

http://gist.github.com/310277

I created a fix here, although I'm not sure if I need both directories or if there are others that may show up: aec0f563bcb2216ca3c7b8dc03b10a2c063bd2e0

Issue #19, opened 2010-02-21
Compile error


Majkl578:
<code>...
[ 22%] Building CXX object src/CMakeFiles/hphp_runtime_static.dir/cpp/base/hphp_ffi.cpp.o
In file included from /storage/hiphop/hiphop-php/src/cpp/base/zend/zend_functions.h:22,
                 from /storage/hiphop/hiphop-php/src/cpp/base/type_string.h:23,
                 from /storage/hiphop/hiphop-php/src/cpp/base/array/array_element.h:20,
                 from /storage/hiphop/hiphop-php/src/cpp/base/type_array.h:22,
                 from /storage/hiphop/hiphop-php/src/cpp/base/hphp_ffi.cpp:18:
/storage/hiphop/hiphop-php/src/util/hash.h:53: warning: integer constant is too large for ‘long’ type
/storage/hiphop/hiphop-php/src/util/hash.h:89: warning: integer constant is too large for ‘long’ type
/storage/hiphop/hiphop-php/src/util/hash.h:93: warning: integer constant is too large for ‘long’ type
/storage/hiphop/hiphop-php/src/util/hash.h:103: warning: integer constant is too large for ‘long’ type
/storage/hiphop/hiphop-php/src/util/hash.h:130: warning: integer constant is too large for ‘long’ type
/storage/hiphop/hiphop-php/src/util/hash.h:307: warning: integer constant is too large for ‘long’ type
In file included from /storage/hiphop/hiphop-php/src/cpp/base/zend/zend_string.h:22,
                 from /storage/hiphop/hiphop-php/src/cpp/base/type_string.h:24,
                 from /storage/hiphop/hiphop-php/src/cpp/base/array/array_element.h:20,
                 from /storage/hiphop/hiphop-php/src/cpp/base/type_array.h:22,
                 from /storage/hiphop/hiphop-php/src/cpp/base/hphp_ffi.cpp:18:
/storage/hiphop/hiphop-php/src/cpp/base/type_variant.h:95: error: ‘HPHP::Variant::Variant(ssize_t)’ cannot be overloaded
/storage/hiphop/hiphop-php/src/cpp/base/type_variant.h:92: error: with ‘HPHP::Variant::Variant(int)’
In file included from /storage/hiphop/hiphop-php/src/cpp/base/hphp_ffi.cpp:18:
[...]

Issue #18, opened 2010-02-21
Extended ASCII not supported in Variables names


serphen:
Very low priority

Variables names with accents are not supported:
$modalité = 'Hello World';

Using variables names with accents is stupid, I agree, but this is legal syntax:
http://www.php.net/manual/en/language.variables.basics.php

compiling and linking CPP files...
In file included from /home/m_build/build/php/footer.h:6,\n                 from /home/m_build/build/php/wall.cpp:2:\n/home/m_build/build/sys/global_variables.h:1878: error: stray \xE2\x80\x98\351\xE2\x80\x99 in program\nmake[2]: **\* [CMakeFiles/program.dir/php/wall.cpp.o] Error 1\nmake[1]: **\* [CMakeFiles/program.dir/all] Error 2\nmake: **\* [all] Error 2\n
compiling and linking CPP files took 1'16" (76073 ms) wall time

PR #7986, opened 2017-09-05
Emit note during cmake if aarch64 host supports crc instructions, but option not selected.

swalk-cavium:
The ability to use the ARMv8 crc instructions was added by PR#7556.  They
are enabled only if an option is added to the cmake command line. 

This makes a minor change to the HPHPCompiler.cmake file to emit a note
if the build host supports the crc instruction and it has not been selected.

### DIFF
--- a/hphp/tools/gdb/gdbutils.py
+++ b/hphp/tools/gdb/gdbutils.py
@@ -126,7 +126,11 @@ def string_data_val(val, keep_case=True):
     else:
         data = val['m_data']
 
-    s = data.string('utf-8', 'ignore', val['m_len'])
+    try:
+        s = data.string('utf-8', 'ignore', val['m_len'])
+    except:
+        s = "Unknown"
+
     return s if keep_case else s.lower()
 
 
@@ -300,3 +304,11 @@ def deref(val):
         return val.cast(rawtype(val.type))
     else:
         return deref(p.referenced_value())
+
+def get_arch(self):
+    try:
+        arch = gdb.selected_frame().architecture().name()
+    except:
+        arch = re.search('\(currently (.*)\)', gdb.execute('show architecture', to_string=True)).group(1)
+    return arch
+
--- a/CMake/HPHPCompiler.cmake
+++ b/CMake/HPHPCompiler.cmake
@@ -35,6 +35,22 @@ CHECK_CXX_SOURCE_COMPILES("
 #endif
 int main() { return 0; }" IS_PPC64)
 
+function(AARCH64_HAS_CRC HAS_CRC)
+  if(NOT AARCH64_TARGET_CPU)
+    set(CMAKE_REQUIRED_FLAGS "-mcpu=native")
+  endif()
+  CHECK_CXX_SOURCE_COMPILES("
+    int main() {
+      unsigned x;
+      unsigned ret;
+      __asm__ (\"crc32cx %w0, %w1, %x1\"
+      : \"=r\"(ret)
+      : \"r\"(x)
+     );  return ret;
+  }" FOUND_CRC)
+  set(HAS_CRC ${FOUND_CRC} PARENT_SCOPE)
+endfunction()
+
 # using Clang or GCC
 if (${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang" OR ${CMAKE_CXX_COMPILER_ID} STREQUAL "GNU")
   # Warnings to disable by name, -Wno-${name}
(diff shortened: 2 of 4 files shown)


hphp/tools/gdb/gdbutils.py
```diff
@@ -300,3 +304,11 @@ def deref(val):
         return val.cast(rawtype(val.type))
     else:
         return deref(p.referenced_value())
+
```
mxw: +1 line whitespace here to match the rest of the file.

hphp/tools/gdb/gdbutils.py
```diff
@@ -300,3 +304,11 @@ def deref(val):
...
+def get_arch(self):
```
mxw: This function needs to be documented with a triple-quote docstring, as all the functions above.

hphp/tools/gdb/gdbutils.py
```diff
@@ -300,3 +304,11 @@ def deref(val):
...
+    try:
+        arch = gdb.selected_frame().architecture().name()
```
mxw: Just return it here...

hphp/tools/gdb/gdbutils.py
```diff
@@ -300,3 +304,11 @@ def deref(val):
...
+    except:
+        arch = re.search('\(currently (.*)\)', gdb.execute('show architecture', to_string=True)).group(1)
```
mxw: ...and here...

hphp/tools/gdb/gdbutils.py
```diff
@@ -300,3 +304,11 @@ def deref(val):
...
+    return arch
```
mxw: And you can skip having an `arch` variable.

hphp/tools/gdb/gdbutils.py
mxw: Also, wrap to 80 chars.  Maybe

```python
arch = re.search(
    '\(currently (.*)\)',
    gdb.execute('show architecture', to_string=True)
).group(1)
```

hphp/tools/gdb/gdbutils.py
```diff
@@ -126,7 +126,11 @@ def string_data_val(val, keep_case=True):
     else:
         data = val['m_data']
 
-    s = data.string('utf-8', 'ignore', val['m_len'])
+    try:
+        s = data.string('utf-8', 'ignore', val['m_len'])
+    except:
+        s = "Unknown"
+
```
mxw: Undo this change.

swalk-cavium: @mxw - Actually, without this change the lookup fails hard and stops crawling the stack.
Example below:

#18 0xffffffffd4b0 @ 0x05c2c91c: HPHP::ObjectData::release() at /mnt/cdn/hhvm/play.083017a/hhvm/hphp/runtime/base/object-data.cpp:203
#19 0xffffacefff90 @ 0x14000228: [PHP] foo() at /mnt/cdn/hhvm/play.083017a/hhvm/hphp/test/quick/setg-dtor.php
#20 0xffffacefffc0 @ 0x0f000918: [PHP] <pseudomain>() at /mnt/cdn/hhvm/play.083017a/hhvm/hphp/test/quick/setg-dtor.php

Traceback (most recent call last):
  File "/mnt/cdn/hhvm/play.083017a/hhvm/hphp/tools/gdb/gdbutils.py", line 55, in wrapped
    return func(*args, **kwds)
  File "/mnt/cdn/hhvm/play.083017a/hhvm/hphp/tools/gdb/stack.py", line 115, in invoke
    idx=i, ar=fp.cast(ar_type), rip=rip)))
  File "/mnt/cdn/hhvm/play.083017a/hhvm/hphp/tools/gdb/frame.py", line 163, in create_php
    func_name = nameof(func)
  File "/mnt/cdn/hhvm/play.083017a/hhvm/hphp/tools/gdb/nameof.py", line 34, in nameof
    return string_data_val(deref(sd))
  File "/mnt/cdn/hhvm/play.083017a/hhvm/hphp/tools/gdb/gdbutils.py", line 130, in string_data_val
    s = data.string('utf-8', 'ignore', val['m_len'])
OverflowError: signed integer is greater than maximum

Python Exception <class 'OverflowError'> signed integer is greater than maximum: 
Error occurred in Python command: signed integer is greater than maximum

swalk-cavium:
Didn't create branch properly.  Intertwingled two separate commits.  These comments will
be applied to PR#7985.  The cmake file changes will be submitted separately.

PR #7987, opened 2017-09-06
aarch64 - Emit note if crc instructions available and not used

swalk-cavium:
The ability to use the ARMv8 crc instructions was added by PR#7556. They
are enabled only if an option is added to the cmake command line.

This makes a minor change to the HPHPCompiler.cmake file to emit a note
if the build host supports the crc instruction and it has not been selected.

### DIFF
--- a/CMake/HPHPCompiler.cmake
+++ b/CMake/HPHPCompiler.cmake
@@ -35,6 +35,22 @@ CHECK_CXX_SOURCE_COMPILES("
 #endif
 int main() { return 0; }" IS_PPC64)
 
+function(AARCH64_HAS_CRC HAS_CRC)
+  if(NOT AARCH64_TARGET_CPU)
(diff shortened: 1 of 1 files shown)


swalk-cavium:
@mxw - Hi Max, Can you inspect?

swalk-cavium:
@jim-saxman, @dave-estes - Any comments?  It didn't seem right to change options automatically on the user, so I figured a note would be sufficient.

jim-saxman:
@swalk-cavium Works for me with our standard CMake command line. However, we're using the -DAARCH64_TARGET_CPU=qdf24xx in conjunction with the -DENABLE_AARCH64_CRC=1. Can I ask what your cmake looks like?

swalk-cavium:
@jim-saxman - I had been letting it default.  That's when I noticed the one regression test slowing down.

Issue #7950, opened 2017-08-04
Memcached extension hhi out of sync


bakks:
This is an issue on master and going back at least as far as 3.18.

The [.hhi file](https://github.com/facebook/hhvm/blob/master/hphp/hack/hhi/stdlib/builtins_memcached.hhi) for the built in memcached extension is out of sync with the [extension code itself](https://github.com/facebook/hhvm/blob/master/hphp/runtime/ext/memcached/ext_memcached.php). For example, the extension implements a `resetServersList()` method that doesn't appear in the hhi definition. Another example: the extension implements a 4 argument `increment()` method, while the hhi file defines this method as having 2 arguments.

This causes problems if you're working with the memcached extension while in strict mode. At the very least the hhi should agree with the implementation. It would also be nice if the Memcached hhi declared types on its methods.

lexidor:
I am going over old issues on this repository, to see which ones apply to the current versions of hhvm.

This issue was addressed by f284e0a as the author mentioned.

PR #7989, opened 2017-09-07
Syncing memcached.hhi with ext_memcached.php

RyanGordon:
This should address #7950 by syncing the methods from ext_memcached.php to memcached.hhi

I also noticed there were also some strict-type inconsistencies in ext_memcached.php like: https://github.com/facebook/hhvm/blob/master/hphp/runtime/ext/memcached/ext_memcached.php#L341

But since this file appears to be auto-generated I didn't try to fix it there.


kevankids1013:
Graph.facebook. user=  true

kevankids1013:
Graph.facebook. user=  true

fredemmott:
Sorry, assumed the spam/accidentla commennts were feedback requests.

fredemmott:
Reverted as it caused internal test issues, will investigate and re-land, but probably not today

alshamiri5 (approved):

PR #7996, opened 2017-09-11
AArch64 Fix gold linker to work on AArch64

jim-saxman:
Add variables to control usage of the gold linker options --icf=
and ensure the final hhvm binary is not marked in the ELF header
with an executable stack section.

Assembly files, even empty, are now marked with the directive that
prevents them from having an executable stack. Besides locking down
security somewhat, it also brings total runtime of the unit test
suites back in line with the old ld.bfd linker (which does not
mark the ELF as having an executable stack section).

The --icf=all option currently works on AArch64 Ubuntu but not
AArch64 CentOS. When cmaking the hhvm project on CentOS, one can
-DICF="-Wl,--icf=none" to prevent a broken binary. CentOS currently
crashes with --icf=all in boost.*program_options.* when parsing the
the command line parameters. Who knows what else is broken due to
ICF after that. There are 77 symbols ICF'd from/into boost's
program_options, so using the --keep-unique SYMBOL gold linker
option is fairly unreasonable.


jim-saxman:
@mxw This is a starting point. 
@swalk-cavium Can you test the gold linker? You should see a nice performance improvement on OSS.

jim-saxman:
HI @swalk-cavium . Have you had a chance to test this PR? Please let me know if you have questions or issues with it. I updated it so that it will use the tool /usr/bin/lsb_release to query which ARM distro it's running on, and only use the --icf=all flag if not CentOS, since that's where I and @cmuellnerTheobromaSystems had issues trying to enable ld.gold.

jim-saxman:
Hi @mxw, ping.

swalk-cavium:
@jim-saxman - Hi Jim, Does this depend on a certain version of gcc?  I'm getting a link-time error
on ubuntu 16.04.

Generating Repo Schema ID and Compiler ID
[100%] Linking CXX executable tc-print
Generating Repo Schema ID and Compiler ID
Embedding php in tc-print
[100%] Built target tc-print
Erratum 843419 found and fixed at "../compiler/libhphp_analysis.a(analysis_result.cpp.o)", section 6801, offset 0x000001fc.
Erratum 843419 found and fixed at "../runtime/libhphp_runtime.a(bytecode.cpp.o)", section 11708, offset 0x00000180.
Erratum 843419 found and fixed at "../runtime/libhphp_runtime.a(unwind-itanium.cpp.o)", section 7131, offset 0x00000084.
Erratum 843419 found and fixed at "../compiler/libhphp_analysis.a(analysis_result.cpp.o)", section 6801, offset 0x000001fc.
Erratum 843419 found and fixed at "../runtime/libhphp_runtime.a(bytecode.cpp.o)", section 11708, offset 0x00000180.
Erratum 843419 found and fixed at "../runtime/libhphp_runtime.a(unwind-itanium.cpp.o)", section 7131, offset 0x00000084.
/usr/bin/ld.gold: internal error in try_fix_erratum_843419_optimized, at ../../gold/aarch64.cc:2007
collect2: error: ld returned 1 exit status
hphp/hhvm/CMakeFiles/hhvm.dir/build.make:434: recipe for target 'hphp/hhvm/hhvm' failed
make[2]: *** [hphp/hhvm/hhvm] Error 1

 gcc --version
gcc (Ubuntu/Linaro 5.4.0-6ubuntu1~16.04.4) 5.4.0 20160609
Copyright (C) 2015 Free Software Foundation, Inc.

jim-saxman:
Hi @swalk-cavium. We are using the same GCC 5.4 on Ubuntu 16.04, however we also pass the -DAARCH_TARGET_CPU=qdf24xx. The error you are seeing can be controlled/disabled by the -mno-fix-cortex-a53-843419 flag. See CMake/HPHPCompiler.cmake linke 226 for the logic. I doubt anyone is using a buggy A53 CPU for running HHVM. Perhaps we should be throwing the -mno-fix-cortex-a53-843419 all of the time?

PR #7997, opened 2017-09-12
aarch64 - correction for LSE Atomics

swalk-cavium:
Running a MOP on a larger lab machine exposed unexpected failures when -vEval.JitPGO=1 was used.  The issue was traced to the use of the LSE Atomic instructions.  The implementation of
PR#7857 was not complete.  The decqmlock Vinstr was expected to set the PSW.  Without this
change arbitrary PSW values were returned when the hardware supported the LSE instructions.

The hphp/test/quick/setg-dtor.php test reliably exposed the problem.

### DIFF
--- a/hphp/runtime/vm/jit/vasm-arm.cpp
+++ b/hphp/runtime/vm/jit/vasm-arm.cpp
@@ -735,8 +735,8 @@ void Vgen::emit(const decqmlock& i) {
   a->SetScratchRegisters(vixl::NoReg, vixl::NoReg);
   if (RuntimeOption::EvalJitArmLse) {
     a->Mov(rVixlScratch0, -1);
-    a->stadd(rVixlScratch0, adr);
-    a->Msr(NZCV, vixl::xzr);
+    a->ldaddal(rVixlScratch0, rVixlScratch0, adr);
+    a->Sub(rAsm, rVixlScratch0, 1, SetFlags);
   } else {
     vixl::Label again;
     a->bind(&again);
--- a/hphp/runtime/vm/jit/vasm-arm.cpp
+++ b/hphp/runtime/vm/jit/vasm-arm.cpp
@@ -736,6 +736,7 @@ void Vgen::emit(const decqmlock& i) {
   if (RuntimeOption::EvalJitArmLse) {
     a->Mov(rVixlScratch0, -1);
     a->stadd(rVixlScratch0, adr);
+    a->Msr(NZCV, vixl::xzr);
   } else {
     vixl::Label again;
     a->bind(&again);
--- a/hphp/vixl/a64/assembler-a64.cc
+++ b/hphp/vixl/a64/assembler-a64.cc
@@ -1136,11 +1136,11 @@ void Assembler::ldr(const FPRegister& ft, double imm) {
 }
 
 
-void Assembler::stadd(const Register& rt, const MemOperand& src) {
+void Assembler::ldaddal(const Register& rs, const Register& rt, const MemOperand& src) {
   assert(src.IsImmediateOffset() && (src.offset() == 0));
-  // ldadd alias
-  uint32_t op = rt.Is64Bits() ? LSELD_ADD_x : LSELD_ADD_w;
-  Emit(op | Rs(rt) | Rt(xzr) | RnSP(src.base()));
+  // aquire/release semantics
+  uint32_t op = rt.Is64Bits() ? LSELD_ADD_alx : LSELD_ADD_alw;
+  Emit(op | Rs(rs) | Rt(rt) | RnSP(src.base()));
 }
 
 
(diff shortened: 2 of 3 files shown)


swalk-cavium:
Regression tests run with 6 option sets.  Jit'd and interpretted results identical.

swalk-cavium:
@mxw - Hi Max, can you inspect?

mxw:
This could use review from other ARM experts—@dave-estes, @jim-saxman?

hphp/runtime/vm/jit/vasm-arm.cpp
```diff
@@ -736,6 +736,7 @@ void Vgen::emit(const decqmlock& i) {
   if (RuntimeOption::EvalJitArmLse) {
     a->Mov(rVixlScratch0, -1);
     a->stadd(rVixlScratch0, adr);
+    a->Msr(NZCV, vixl::xzr);
```
dave-estes: I'm not sure if just forcing the status flags to zero is the correct thing to do. The flags have to be set based on the arithmetic operation here, which is a subtraction of 1. Maybe instead you can `cmp` the result of the operation to zero? ...actually that's going to be hard seeing as how that result was set atomically in memory.

You might try making the use of `stadd` here an optimization for use only when the status flags don't need to be def'd. Otherwise if there is a use of the status flags that require this def, then revert to the original implementation. You can use the existing undef SF pass for guidance.

I have a suspicion that the status flags are always checked after a `decqmlock{}` as the primary use for this vasm instruction is during profiling for PGO. In that case, it might not be a huge deal if we just always use the old approach. Sure atomics will be faster, but this profiling only happens during warmup. AND #7984 actually introduces an option to just use a simple `decqm{}` to estimate the counters thus bypassing the use of `decqmlock{}` altogether.

swalk-cavium:
@dave-estes - Hi Dave, Thanks for your comment.  According to the latest ARM ARM I could find,
ARM DDI 0487B.a (ID033117), the results of the stxr instruction is the ExclusiveMonitorsStatus()
zero extended.  It has nothing to do with the arithmetic operation.  It's looping until the data is written.  This is similar to the PowerPC stdcx loop.  I haven't been able to find good information about the lock prefix on dec for x86 yet.  Some older docs suggest it's a hard-stop to wait for memory.  That must have changed by now.

The description for the ARMv8 stadd instruction says 'All observers in the shareability domain
observe the following load and store atomically' (p.C6-837).  There's no status returned.
I think indicating success is the appropriate change.

So I guess the real question is what is the code in cgCheckCold() in runtime/vm/jit/irlower-intrinsic.cpp expecting.  Is it that the counter goes to zero?  Or just that the value was written.
I couldn't find any documentation about that a JitFilterLease might be.

swalk-cavium:
Hi,  I went back through all the docs I have access to, and talked to some of the hardware guys.
The one line change is still correct, but there's another change to the instruction encoding that will be needed.  I expect to update this pull request some time tomorrow.

dave-estes:
I'm still confused how zeroing out the status flags addresses this `decqmlock{}` instruction's need to set the status flags based on the arithmetic operation. Let us know what you find, and maybe what the hardware guys said. Is an `mrs` of `Xzr` some sort of a weird way of triggering the status flags to be set after the atomic?

jim-saxman:
@swalk-cavium I agree with @dave-estes ; I think we need to see/set the flags for vasm decqm{lock} , and set those. Not the result of the stadd.

hphp/runtime/vm/jit/vasm-arm.cpp
jim-saxman: The NZCV flags are set by the ARM sub(s) instruction based on the SetFlags parameter on line 744.

apinski-cavium:
Ok, I think understand the issue now.
Before we did a subs for the subtraction, for the LSE case we don't set the flags register.
What we need is the following:
mov rVixlScratch0, -1
ldadd rVixlScratch0, rVixlScratch0, [adr]
subs vixl::xzr, rVixlScratch0, 1 (or cmp rVixlScratch0m, 1 since that is the same instruction)

Thanks,
Andrew

dave-estes:
That'll work.

swalk-cavium:
Ok, I understand where I made a wrong turn.  Thanks for all the comments.

swalk-cavium:
@mxw - Hi Max,  This updates the LSE Atomics change.  The original issue was that I was using the wrong form of the instruction.  This uses the acquire/release form of the ldadd instruction.  It's my understanding that this only holds the cache line exclusive once where the ldxr/stxr sequence could
allow other cache activity.  This lets you make full use of hardware that supports these instructions.
No need for a special option.

The regression suite was run with 6 different option sets.  The previous memory overwrites
are now corrected.

mxw:
Thanks for the review, @dave-estes and @apinski-cavium.  If you're happy, I'm happy—will land shortly.

PR #8014, opened 2017-10-02
AArch64 speedup warmup times by reading cpufreq files

jim-saxman:


jim-saxman:
Hi @mxw, @swalk-cavium. Would you please review this PR.

swalk-cavium:
@jim-saxman - I'm not seeing MHz entries in /proc/cpuinfo on our machines.  @apinski-cavium recalled that changing the contents was Nak'd on the ARM kernel mailing list.
http://lists.infradead.org/pipermail/linux-arm-kernel/2017-September/533228.html

Which distro are you testing with?

jim-saxman:
Hi @swalk-cavium This patch checks  /sys/device/system/cpu/cpu*/cpufreq/cpuinfo_max_freq if available, on ARM, just like /proc/cpuinfo is read on x64. If you don't have cpufreq, then this patch will work as before (no change in performance).

jim-saxman:
This pull request no longer provides any performance benefit, so I'm closing it.

PR #8017, opened 2017-10-04
Fix emitting optimized version for min2/max2 with splat operator

### LINKED ISSUE #7998: max with splat emitting wrong opt
### HHVM Version
3.18-3.22
```php
<?php

$values = [1, 2];
var_dump(max(3, ...$values));
```
### Expected result
```
int(3)
```
### Actual result
```
Warning: __SystemLib\max2() expects exactly 2 parameters, 3 given in /in/tuOuo on line 4
NULL
```

SiebelsTim:
fixes #7998


fredemmott:
We're working on a replacement frontend as part of hack - can you either:
- preferable: add support for it (currently at https://github.com/facebook/hhvm/blob/dfdbe60e016090afe6967b0bd2c16f6fb5a2fe8f/hphp/hack/src/hhbc/emit_expression.ml ); you can test this with `test/run --compare-hh-codegen`
- blacklist the test in https://github.com/facebook/hhvm/blob/dfdbe60e016090afe6967b0bd2c16f6fb5a2fe8f/hphp/test/hhcodegen_failing_tests_slow so it's on our todo list

SiebelsTim:
Can't do the implementation right now. Maybe later (if I wrap my head around OCaml). Let's blacklist it for now.

fredemmott:
Sorry - for the conflicts, and general maintainance, can you put the blacklist entry in alphabetically? Also, I think it cares about the EOL/no EOL at the end of the file.

SiebelsTim:
Oh well, I should've guessed that it's alphabetically when looking at that file.

PR #8054, opened 2017-11-27
Update XHP+namespace resolution in the typechecker to match the runtime.

TJ09:
Right now (and, as far as I can tell, since the days of XHP as an extension of PHP), the runtime always looks for XHP classes in the "global" namespace. This means that defining an XHP class within in a namespace is a futile effort; even within that same namespace references to that class will fail.

The typechecker, on the other hand, has the exact opposite behavior: it always looks for XHP classes in the current namespace (like any other class). This mismatch means that using XHP in namespaced code is impossible in strict mode (though partial mode works perfectly fine; hh doesn't know what class the code is referencing but the runtime is happy to do what it's always done and resolve it globally). Since trying to `use` an XHP class is unsupported by the parser, partial mode is more-or-less the best that can be accomplished today.

This PR updates the typechecker to always assume an empty namespace context when resolving an XHP name. By doing so, we get the same behavior the runtime has (referencing XHP from a namespace will match the global class, but not a class in the same namespace). The included tests should hopefully demonstrate the behavioral change. If nothing else, having the typechecker match the runtime here allows for safer typing, since the typechecker will now be able to validate usages of XHP, and namespaced code that uses XHP can be converted to strict mode.

I have little faith that this will be accepted as-is--if not because this is my first foray into ocaml and I have no clue what I'm doing, then because I'm pretty sure this isn't the direction the HHVM team wants to go with XHP typing. Nonetheless, it's worth a try and IMO this is the right thing to do.

### DIFF
--- a/hphp/hack/src/naming/naming.ml
+++ b/hphp/hack/src/naming/naming.ml
@@ -502,10 +502,22 @@ end = struct
       (GEnv.gconst_pos genv.tcopt)
       x
 
+  let get_namespace_env_with_xhp_global_fallback genv x =
+    (* XHP should always exist in the global namespace (this has been the
+     * runtime behavior since it was introduced). Even an XHP class defined in
+     * a namespace should not be automatically accessible from within that
+     * namespace.
+     * So, we pretend there's no namespace if the class looks like XHP. *)
+    let (pos, name) = x in
+    match name with
+      | name when String_utils.string_starts_with name ":" -> Namespace_env.empty genv.tcopt
+      | _ -> genv.namespace
+
   let type_name (genv, _) x ~allow_typedef =
     (* Generic names are not allowed to shadow class names *)
     check_no_runtime_generic genv x;
-    let (pos, name) as x = NS.elaborate_id genv.namespace NS.ElaborateClass x in
+    let namespace_env = get_namespace_env_with_xhp_global_fallback genv x in
+    let (pos, name) as x = NS.elaborate_id namespace_env NS.ElaborateClass x in
     match GEnv.type_info genv.tcopt name with
     | Some (_def_pos, `Class) ->
       (* Don't let people use strictly internal classes
--- a/hphp/hack/test/typecheck/namespace_fallback1.php
+++ b/hphp/hack/test/typecheck/namespace_fallback1.php
@@ -1,0 +1,11 @@
+<?hh // strict
+
+namespace {
+  class :my-xhp {}
+}
+
+namespace Foo {
+  function test(): :my-xhp {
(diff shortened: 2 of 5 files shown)


fredemmott:
cc @ngavalas, @KendallHopkins , @kmeht 

>  the runtime always looks for XHP classes in the "global" namespace

This is an unintentional bug - this behavior is *only* meant to exist for functions, and only because of PHP - we're hoping to kill it for functions too in Hack.

I'm 99% sure we don't want to accept this behavior - it will be adding another inconsistency to the language.

fredemmott:
See also past discussion and suggestions on hhvm/xhp-lib#64 - namespace support is something we want to fix, and it's largely waiting for someone with enough ocaml expertise to create and advocate for a solution that is consistent with the rest of the language.

TJ09:
>This is an unintentional bug - this behavior is only meant to exist for functions, and only because of PHP - we're hoping to kill it for functions too in Hack.

Unintentional bug or not, it's the only thing that makes XHP even a little useful outside of FB. Because in OSS-land, namespaces are widely used, and within that scope the global fallback is the only way to use XHP (by which I mean "write code that runs and passes the typechecker").

Namespaced XHP is definitely a better end goal, as you point out. But hhvm/xhp-lib#64 has been around for over three and a half years, and there's no sign that it's going to be addressed any time soon, so in the medium-term, this makes XHP more accessible to use. Removing the global fallback once namespace support exists is fine, but leaving XHP in this awkward (and IMO broken) state because there's no resources to do the "right" thing feels like a waste.

fredemmott:
The supported way is to not have the parts of your code that directly use XHP be in namespaces.

We generally don't use lack of time to do things right as a reason to make the language more inconsistent - even when it's to aid usability. As well as introducing another inconsistency of a class we already want to clean up, this also makes some of the proposed long-term fixes impractical, where in parts they're based on the assumption that as this raises errors, and is advised against in the documentation, certain classes of BC-breaking changes do not matter at all.

fredemmott:
> trying to use an XHP class is unsupported by the parser

I'd be more interested in changing this and making the runtime agree with the typechecker. This should be relatively straightforward to do, and I can find the time if there's agreement that it's a good change to do without the other changes mentioned in the xhp-lib; that said, I don't think it is, because it will still make the long-term fixes less likely to happen by turning the "this technical BC break is fine because that code isn't supported anyway" into significant BC breakages.

fredemmott:
Closing: XHP with namespaces is something we want to sort out in the next few months, but we're not going to be taking this as it makes the language more inconsistent, and would require an additional breaking change in the future.

PR #8055, opened 2017-11-28
[AArch64] Cleanup whitespace and use of kInstructionSize.

dave-estes:


dave-estes:
@mxw, this is the first of a few patches that I've got coming. It's just foundational cleanup. Sorry I let those whitespace diffs in before.

PR #8056, opened 2017-11-28
[AArch64] Correctly pad out service request stubs when relocating.

dave-estes:


dave-estes:
@mxw, this addresses an issue where service request stubs were changing size during relocation. Theoretically the other platforms can have the same problem arise. It's probably not worth implementing this, though an assert might be prudent if there isn't one already.

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -571,6 +572,61 @@ size_t relocateImmediate(Env& env, TCA srcAddr, TCA destAddr,
...
+  src = src->NextInstruction();
+  env.rewrites.insert(src);
+  while (src->Mask(ExceptionMask) == BRK) {
+    src = src->NextInstruction();
+    env.rewrites.insert(src);
+    srcCount++;
+  }
+
+  // Pad out the remainder of the dest service request stub
+  auto const destAddrBegin = (srcAddrBegin == env.start) ?
```
mxw: Spurious parens.

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -571,6 +572,61 @@ size_t relocateImmediate(Env& env, TCA srcAddr, TCA destAddr,
...
+    destStart : env.rel.adjustedAddressAfter(srcAddrBegin);
```
mxw: question: Is `destStart != env.rel.adjustedAddressAfter(srcAddrBegin)` if `srcAddrBegin == env.start`?

PR #8057, opened 2017-11-28
[AArch64] Consider multiple smashables per target when relocating.

dave-estes:
When finding the smashable from its target, we need to consider all possible smashables instead of just stopping at the first match. The rationale is that the first possible smashable that we find might not be an actual smashable. Consider a case where a smashable `jmp` is coincidentally preceded by a conditional branch. This can make it look like a smashable `jcc` which is the first possible smashable found. Without considered all possible smashables, we'd miss this `jmp`.

### DIFF
--- a/hphp/runtime/vm/jit/smashable-instr-arm.cpp
+++ b/hphp/runtime/vm/jit/smashable-instr-arm.cpp
@@ -337,22 +337,31 @@ ConditionCode smashableJccCond(TCA inst) {
  *       sequence is a full jcc even though a jcc actually uses the same
  *       sequence as a jmp in its implementation.
  */
-TCA getSmashableFromTargetAddr(TCA addr) {
+hphp_hash_set<vixl::Instruction*> getSmashablesFromTargetAddr(TCA addr) {
+  using InstrSet = hphp_hash_set<vixl::Instruction*>;
   using namespace vixl;
+
+  InstrSet smashables;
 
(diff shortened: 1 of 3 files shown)


dave-estes:
@mxw, this patch provides a little more robustness for a situation where a smashable is emitted into an longer instruction sequence that might lead to ambiguity as to whether the smashable is a shorter jmp or a longer jcc.

hphp/runtime/vm/jit/smashable-instr-arm.cpp
```diff
@@ -337,22 +337,31 @@ ConditionCode smashableJccCond(TCA inst) {
  *       sequence is a full jcc even though a jcc actually uses the same
  *       sequence as a jmp in its implementation.
  */
-TCA getSmashableFromTargetAddr(TCA addr) {
+hphp_hash_set<vixl::Instruction*> getSmashablesFromTargetAddr(TCA addr) {
```
mxw: I think these days we just prefer `std::unordered_set` here since newer compilers don't force-malloc when they're constructed.

That said, maybe a `std::array<TCA,3>` would be better here so that we return on the stack instead of allocating memory at all?

mxw:
By the way, can you add the descriptions into the commit description so that it can end up in the final commit message?

hphp/runtime/vm/jit/smashable-instr-arm.cpp
dave-estes: Will do.

PR #8058, opened 2017-11-28
[AArch64] Correctly handle adjustment of MOV/MOVK sequences.

dave-estes:


dave-estes:
@mxw, this patch fixes issues with relocating MOV/MOVK sequences as well as the resulting adjustments. There were bugs regarding the handling of the extra NOP padding. I rewrote `adjustInstruction()` to use a temporary CodeBlock so that I could just fall back on VIXL's `Mov()` instead of trying to delicately update the immediate fields in each instruction...which included the accidental updating of an immediate field in a NOP. *facepalm

PR #8059, opened 2017-11-28
[AArch64] Adds additional support for shortening LDR literal jmps.

dave-estes:
During relocation, jmps which use LDR (literal) and BR sequences
(smashed smashable jmps/jccs and patchable jmps/jccs) are shortened
to use simple PC relative B instructions. This patch extends this
to also shorten these sequences to absolute branches using
MOVZ/MOVK and BR instructions. The effect is to reduce D$ pressure
and to decrease instruction count.

### DIFF
--- a/hphp/runtime/vm/jit/relocation-arm.cpp
+++ b/hphp/runtime/vm/jit/relocation-arm.cpp
@@ -213,13 +213,27 @@ bool relocateSmashable(Env& env, TCA srcAddr, TCA destAddr,
   if (!adjusted) adjusted = target;
   int imm = (adjusted - destAddr) >> kInstructionSizeLog2;
 
-  if (!is_int26(imm)) return false;
-
   vixl::MacroAssembler a { env.destBlock };
   env.destBlock.setFrontier(destAddr);
-  a.b(imm);
+  destCount--;
+
+  if (is_int26(imm)) {
+    a.b(imm);
+    destCount += 1;
+  } else {
+    auto const tmp = rVixlScratch0;
+    a.SetScratchRegisters(vixl::NoReg, vixl::NoReg);
+    a.Mov(tmp, adjusted);
+    // Pad out to two instructions for adjustment later
+    if ((env.destBlock.frontier() - destAddr) == kInstructionSize) {
+      a.Nop();
+    }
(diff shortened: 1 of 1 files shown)


dave-estes:
@mxw, this optimization uncovered the issues that are fixed in #8056, #8057, and #8058. We shouldn't land this PR until after those are resolved.

edwinsmith:
nice

dave-estes:
@edwinsmith, well I think this is still short of it's potential since most smashables are not smashed at the time of static relocation. However, this change did catch more smashable-like jumps/jccs that could be optimized. Previously this code would only optimize if the distance was close enough to be encoded into a PC relative branch. Now it catches longer jumps and encodes as a MOV/MOVK/BR...which uncovered a few bugs. But hey, relocation is much more robust now. :-)

I'm still considered the path of dynamic relocation as an opportunity to catch more of these smashed smashables. I think all that's left is to build a list of unsmashed smashables from the debug info (reloc map). I back-burnered the effort since dynamic relocation isn't likely to be used any longer with retranslate all in place. Actually I still need to study more about the interplay with smashables during retranslate all. I've always assumed that retranslate all would still generate lots of smashable jumps/jccs to service request stubs. But if PGO info is used, then they might be direct jumps now already... Is that how it works?

edwinsmith:
> But if PGO info is used, then they might be direct jumps now already... Is that how it works?

Good question. @paulbiss ?

mxw:
@dave-estes—Here's what @paulbiss said:

```
[18:12:17] @paulbiss:	mwang: even if we know the call target we still emit a bindcall right now and patch it up later 
[18:12:48] @paulbiss:	mwang: when we relocate functions into the TC the relocation is per-function so if you wanted to burn in the addresses for target prologues you'd need to track those relocations 
```

That said, we shouldn't be emitting many bindjmp{}s or retranslate{}s because the optimized translations will be large contiguous regions.  So those jumps will be direct already (or just omitted as fallthrus), but the branches used for function calls will still get smashed over, as things are today.

dave-estes:
Thanks for the follow-up, @mxw.

mxw:
@dave-estes Can you rebase this PR?

PR #8060, opened 2017-11-28
[AArch64] Change callToExit to MOVZ/K, BR.

dave-estes:


dave-estes:
@mxw, this is a small optimization similar to #8059.

mxw:
Just to clarify my understanding—this _allows_ your relocation machinery to kick in to rewrite this branch, right?  If this itself is the final form we're aiming for, I think I may be missing something.

dave-estes:
@mxw, this is just a spot where we previously used an LDR/BR sequence instead of the slightly better MOV/MOVK/BR. I don't think unique stubs are ever relocated, and so it was never optimized. There's definitely no reason to use the LDR/BR here as the embedded target is never patched or anything like that.

PR #8063, opened 2017-12-01
[AArch64] Fixes int overflow when encoding immediates.

dave-estes:


dave-estes:
@mxw, this fixes some transient unit test failures. Specifically, those tests that load an immediate float of 1.2345. The problem arises when the code cache's base is 0x12000000.

mxw:
@dave-estes Can you rebase this PR also?

PR #8070, opened 2017-12-11
[AArch64] Remove extra syncing of CodeBlock from base to frontier.

dave-estes:
The VGen dtor was mistakenly syncing from the CodeBlock's base
repeatedly instead of just the new section emitted.


dave-estes:
@mxw, this is a minor fix restoring the more targeted CodeBlock syncing that we had before ddc02cf81a9d9133722ae67d2adc8bd4d08a38e2.

Issue #8089, opened 2018-01-05
segfault on include/require with registered user 'file' stream wrapper


pprasse:
### HHVM Version
tested in 3.22.0, still in master
at least since 3.2.0

### Operating System and Version
Ubuntu 16.04 (probably all operating systems)

### Analysis

In https://github.com/facebook/hhvm/blob/HHVM-3.22/hphp/runtime/base/unit-cache.cpp#L493 we get the stream wrapper and execute stat on it.
If stat fails, we continue to use the stream wrapper we got before stat. This gets problematic if the stream wrapper is messed with (unregistered, restored, registered) in the stat call.

### Standalone code, or other way to reproduce the problem
[segfault-user-stream-wrapper.zip](https://github.com/facebook/hhvm/files/1607876/segfault-user-stream-wrapper.zip)
```
unzip segfault-user-stream-wrapper.zip
hhvm main.php
```
### Actual result

```
START FileIsolationStreamWrapper::url_stat array (
  0 => '/home/patrickprasse/sandboxes/polymer/actindo3/_ERP-1130/Autoload1.php',
  1 => 2,
)
strlen($_text)=8388608
END FileIsolationStreamWrapper::url_stat array (
  0 => 87,
  1 => 1073273987,
  2 => 33188,
  3 => 1,
  4 => 10010,
  5 => 10010,
  6 => 0,
  7 => 45,
  8 => 1515162286,
  9 => 1515162287,
  10 => 1515162287,
  11 => 1048576,
  12 => 8,
  'dev' => 87,
  'ino' => 1073273987,
  'mode' => 33188,
  'nlink' => 1,
  'uid' => 10010,
  'gid' => 10010,
  'rdev' => 0,
  'size' => 45,
  'atime' => 1515162286,
  'mtime' => 1515162287,
  'ctime' => 1515162287,
  'blksize' => 1048576,
  'blocks' => 8,
)

START FileIsolationStreamWrapper::url_stat array (
  0 => '/home/patrickprasse/sandboxes/polymer/actindo3/_ERP-1130/Autoload2.php',
  1 => 2,
)
strlen($_text)=16777216
END FileIsolationStreamWrapper::url_stat array (
  0 => 87,
  1 => 1073272912,
  2 => 33188,
  3 => 1,
  4 => 10010,
  5 => 10010,
  6 => 0,
  7 => 45,
  8 => 1515162285,
  9 => 1515162285,
  10 => 1515162285,
  11 => 1048576,
[...]

PR #8092, opened 2018-01-08
fix segfault on include/require with registered user 'file' stream wrapper

pprasse:
I fixed issue #8089 by returning false in case "stat" from the user stream wrapper is unsuccessful.


fredemmott:
Please change this to target master; once it's there, it can be cherry-picked.

pprasse:
@fredemmott just changing target produces conflicts. I will open a new Pull-RQ.

pprasse:
@fredemmott Here's the follow-up pull request https://github.com/facebook/hhvm/pull/8095

PR #8095, opened 2018-01-09
fix segfault on include/require with registered user 'file' stream wrapper

pprasse:
I fixed issue #8089 by returning false in case "stat" from the user stream wrapper is unsuccessful.


fredemmott:
Hey - I won't be able to take a proper look at this until after 3.24; that said:
 - can you add a small isolated test to test/slow that reproduces the problem
 - from a quick look, it seems unlikely to be such a simple fix, given that there is further explict handling of non-file streams below

alshamiri5 (approved):

PR #8128, opened 2018-02-16
Enabling hash functions that require SSE4.2 support

sushmakt:
Hash functions for HHVM are currently guarded by __SSE4_2__ flag which are not
enabled by default. The fallback versions for "hash int" functions are "C"
implementations leaving some room for performance. This change enables the SSE4.2
support by default allowing all the hash functions to take the corresponding
assembly routines. The change also leaves the ENABLE_SSE4_2 flag configurable so
it can be turned off in case run on really old processors.

This gives ~0.4% performance improvement on oss-performance/WordPress and Drupal7

### DIFF
--- a/CMake/Options.cmake
+++ b/CMake/Options.cmake
@@ -6,6 +6,7 @@ option(STATIC_CXX_LIB "Statically link libstd++ and libgcc." OFF)
 option(ENABLE_AVX2 "Enable the use of AVX2 instructions" OFF)
 option(ENABLE_AARCH64_CRC "Enable the use of CRC instructions" OFF)
 option(ENABLE_FASTCGI "Enable the FastCGI interface." ON)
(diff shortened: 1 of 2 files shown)


CMake/Options.cmake
```diff
@@ -6,6 +6,7 @@ option(STATIC_CXX_LIB "Statically link libstd++ and libgcc." OFF)
 option(ENABLE_AVX2 "Enable the use of AVX2 instructions" OFF)
 option(ENABLE_AARCH64_CRC "Enable the use of CRC instructions" OFF)
 option(ENABLE_FASTCGI "Enable the FastCGI interface." ON)
+option(ENABLE_SSE4_2 "Enable SSE4.2 supported code." ON)
```
fredemmott: Sorry for the delay; pease switch the default to OFF.

sushmakt: Thanks for the feedback. Done switching it to OFF

PR #8143, opened 2018-03-01
[easy] Make Mac build 1/3rd the size and 300x faster

fredemmott:
Test Plan:

sandcastle for Linux, and local builds on both FB and non-FB macs.


hphp/runtime/base/program-functions.cpp
```diff
@@ -838,6 +838,16 @@ static void pagein_self(void) {
   FILE *fp;
 
   auto mapped_huge = false;
+#ifdef __APPLE__
+  // The normal implemention is theroetically safe on non-Linux platforms,
```
swtaarrs: theoretically

PR #8164, opened 2018-03-28
X64 code generation using the Intel XED library

catalin-manciu:
This adds a complete reimplementation of the **X64Assembler** class based on the Intel XED library. The newly added **XedAssembler** class offers the same functionality as the one it aims to replace, **X64Assembler**, but makes it easier to expand HHVM’s supported instruction set by delegating all the encoding-specific details to the XED library. XED is actively maintained so any instruction that gets added to newer CPU generations will immediately be supported by the library.
It also provides a runtime option - **Eval.UseXedAssembler** - which allows specifying which of the two X64 assembler classes will be used for code generation (either the XED-based one or the previous version) for more efficient functional and performance evaluation:
-	**Eval.UseXedAssembler=true** – the XedAssembler class will be used for X64 code generation
-	**Eval.UseXedAssembler=false** – the X64Assembler class will be used for X64 code generation, **this is the default value**

## Requirements:
This requires a recent version of the open source Intel XED library, available here: https://github.com/intelxed/xed.

## Building the XED library:
- You will need the need the mbuild Python module, available here: https://github.com/intelxed/mbuild
  - To install the mbuild module go to its source directory and run `sudo python setup.py install`
- Go to the XED source dir, run `python mfile.py install --shared` - this will build the library and generate the header files in `xed_sources_dir/kits/xed-install-base-[build_date]-[OS]-[arch]/`. In this folder you will find the `/include` directory and the `/lib` directory containing libxed.so, use these paths when running `./configure` as described below.

## Building HHVM with XED code generation support:
In order to use this new feature, the following two options must be specified for `./configure`:
- `-DLibXed_INCLUDE_DIR=/path_to/include/xed`
- `-DLibXed_LIBRARY=/path_to/libxed.so`

If the two options aren’t correctly set, HHVM will not be able to use XED for code generation and the **Eval.UseXedAssembler** runtime option will not be available.

## Running HHVM with XED support:
Use **Eval.UseXedAssembler** to enable/disable the XED-based assembler, it is **disabled** by default.
The code generated by the **XedAssembler** class is mostly identical to the one generated by the **X64Assembler** class – there is a known difference: `cmp/xor/test (imm, reg8)` - and possibly similar others - are emitted 1 byte shorter by the XED implementation.
All regression tests that pass for the unpatched version also pass for the patched one with XED enabled (when running `./run all`).

There is no significant difference in throughput when running oss-performance’s Drupal7, WordPress or Mediawiki targets when using the XED-based assembler class.

Development and tests were conducted on an Intel Xeon E5-2699 v4 machine, running Ubuntu 16.04.3 with GCC version 5.4.0.


fredemmott:
FYI, to include this in external builds, we will require:
- static linking, not dynamic
- it adding to https://github.com/hhvm/hhvm-third-party , including support for our existing CMake build system

catalin-manciu:
Static linking is currently possible, with a couple of small changes to the steps provided in the previous comment:

1. When building XED:
    - `python mfile.py install` ~~--shared~~ - you no longer need to specify the `--shared` option for mfile.py - this will just build the static library in the same `xed_sources_dir/kits/xed-install-base-[build_date]-[OS]-[arch]/lib` folder
2. When running `./configure`
    - `-DLibXed_LIBRARY=/path_to/libxed.a` - the path should be that of the static library.

I’ll add XED to the HHVM CMake build pipeline as a separate pull request to https://github.com/hhvm/hhvm-third-party.

XED is compatible with MacOS, but I don’t have access to a MacOS machine to validate the pull request. If anyone can run this on a Mac and provide feedback, it would be very useful.

fredemmott:
I can try it out once the PR for hhvm-third-party is up.

hphp/util/asm-x64-intelxed.h
```diff
@@ -0,0 +1,1031 @@
...
+
+  inline int reduceImmSize(int64_t value, int allowedSizes, immFitFunc func) {
+    for (int crtSize = sz::byte; crtSize < sz::qword; crtSize <<= 1) {
+      if ((allowedSizes & crtSize) && (*func)(value, crtSize))
+        return crtSize;
+      }
+    assert((allowedSizes & sz::qword) &&
+           "Could not find an optimal size for Immed");
+    return sz::qword;
+  }
```
swtaarrs: Something is off with the indentation and/or the block-ending `}` between 163 and here.

hphp/util/asm-x64-intelxed.h
```diff
@@ -0,0 +1,1031 @@
...
+#define XED_MEMREF(m, memSize)            XedOperand(m, memSize).op
+#define XED_MEMREF_RIP(m, memSize)        XedOperand(m, memSize).op
+#define XED_BRREL(p, size)                XedOperand((CodeAddress)p, size).op
+
+///////////////////////////////////////////////////////////////////////////////
+
+struct XedAssembler : public X64AssemblerBase {
+private:
+  constexpr static xed_state_t s_xedState = {
+        XED_MACHINE_MODE_LONG_64,
```
swtaarrs: 2-space indentation

hphp/util/asm-x64-intelxed.h
```diff
@@ -0,0 +1,1031 @@
...
+   *
+   * Addressing with index registers:
+   *
+   *   a.  movl   (index, ecx);
+   *   a.  loadq  (*rax, rbx);
+   *   a.  storeq (rbx, rbx[rcx*8]);
+   *   a.  call   (rax);            // indirect call
+   *
+   */
+#define XED_INVERSE(x, y)   y, x
```
swtaarrs: Why not just pass the args in the right order?

hphp/util/asm-x64-intelxed.h
```diff
@@ -0,0 +1,1031 @@
...
+    assert(c != CC_None);
+    static xed_iclass_enum_t cmovs[] = CC_TO_XED_ARRAY(CMOV);
+    return cmovs[(int)c];
+  }
+
+  //Xed Emit funcs
+
+#define DECLARE_UNUSED(type, name)  type name; static_cast<void>(name)
+
+#define XED_EMIT_BEGIN()                                                    \
```
swtaarrs: This should be doable without macros by having one `xedEmit` function that has the prologue and epilogue, and takes a lambda that does the call to the right `xed_inst*()`.

hphp/util/asm-x64-intelxed.h
```diff
@@ -0,0 +1,1031 @@
...
+    }
+    // Emit remaining NOPs (if any)
+    if (n) {
+      xedInstr(nops[n], 0);
+    }
+  }
+
+  void pad() {
+    auto remaining = available();
+    static uint32_t int3 = 0;
```
swtaarrs: You're emitting to a `uint32_t` rather than `uint8_t[XED_MAX_INSTRUCTION_BYTES]` because these instructions both fit within 4 bytes, right? That's hardcoding some information about the size of them, so at that point you might as well just do this instead:
```c++
static uint8_t int3[1];
static uint8_t ud2[2]; // or whatever their sizes actually are
```
That should also let you get rid of the casts on 722 and 726.

catalin-manciu:
@fredemmott Added XED to third-party in https://github.com/hhvm/hhvm-third-party/pull/128. https://github.com/facebook/hhvm/pull/8164/commits/4077df8aa7376b7faa7a7784061c5a2f5eb4a83b updated the HHVM CMake scripts to reflect that change.

hphp/util/asm-x64-intelxed.h
```diff
@@ -0,0 +1,1031 @@
...
+
+struct XedInit {
+  XedInit() {
+      xed_tables_init();
+  }
+};
+
+//////////////////////////////////////////////////////////////////////
+
+#define SZ_TO_BITS(sz)                    (sz << 3)
```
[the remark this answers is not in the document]
catalin-manciu: I've changed `SZ_TO_BITS` and `BITS_TO_SZ` to `inline constexpr` functions in https://github.com/facebook/hhvm/pull/8164/commits/b0f08a27f56e94240848f527b4f5c87c043aa306. For `XED_REG(...)` and similar macros (lines 169 - 174), there are several approaches I could go for:
- leave them unchanged - I think it makes the code more readable
- change them to inline functions - adds more code that doesn't help with neither readability nor functionality
- replace the macros with their content - since each of them is a one-liner, it would not make the code too bloated, but I think it will still be less readable
- drop the `XedOperand` class, since all it does is conversion from HHVM objects (such as `Reg8/16/32/64`, `Immed`, `Immed64`) to a `xed_encoder_operand_t` instance which it contains as a class member, and replace it with a set of conversion functions that take one of the mentioned HHVM objects and return a `xed_encoder_operand_t`

Which one would you prefer?

catalin-manciu:
@swtaarrs - I've addressed your comments in the last two commits (https://github.com/facebook/hhvm/pull/8164/commits/ccd788956f88b93b697a637cb7ebb64090d48e9b & https://github.com/facebook/hhvm/pull/8164/commits/b0f08a27f56e94240848f527b4f5c87c043aa306)  - all except the one I replied to.

hphp/util/asm-x64-intelxed.h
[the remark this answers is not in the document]
swtaarrs: How would making them inline functions add much more code? The macros do have shorter definitions, but inline functions wouldn't require any more code at each usage site, and they're generally much better when it comes to compiler errors and debugging. These macros aren't doing anything weird like concatenating tokens together or referencing a variable not passed as an argument. `XED_REG` could be something like:
```c++
template<typename Reg>
xed_encoder_operand_t xed_reg(const Reg& r) {
  return XedOperand(r).op;
}

// Then, unless I'm missing something, you can call xed_reg just like you'd
// use the XED_REG macro:
XED_REG(blah)
// becomes
xed_reg(blah)
```
But then at that point you're not too far away from your last suggestion of replacing `XedOperand` with a set of conversion functions that don't use an intermediate object. I think going all the way there would be best.

swtaarrs:
Thanks for the update! I'm about to go on vacation until next Friday and unfortunately I won't have time to finish this before leaving. Hopefully someone else will be able to pick up where I left off.

catalin-manciu:
@swtaarrs I've now addressed all your comments and, additionally, I replaced most macros with an appropriate C++ construct (in https://github.com/facebook/hhvm/pull/8164/commits/1bb7ff072b868835eb0f26bd71f01a203924b8ce, https://github.com/facebook/hhvm/pull/8164/commits/05ed773153c80b1680d8ff18853ef8f58d9b3a49 and https://github.com/facebook/hhvm/pull/8164/commits/64f55b0aae9a18509d7039b6b7bc924b17a4845f).
It looks like there's a minor conflict with one of the master commits, let me know if you prefer rebasing or a merge commit.

hphp/util/asm-x64-intelxed.h
```diff
@@ -0,0 +1,1066 @@
...
+    return cmovs[(int)c];
+  }
+
+  // XED emit functions
+
+#define DECLARE_UNUSED(type, name)  type name; static_cast<void>(name)
+
+  ALWAYS_INLINE
+  uint32_t xedEmitImpl(xed_iclass_enum_t instr, CodeAddress destination,
+                       std::function<void(xed_encoder_instruction_t*)>
```
swtaarrs: Unless there's a very strong reason to use `std::function<>`, we use templates to take lambdas:
```c++
template<typename L>
uint32_t xedEmitImpl(xed_iclass_enum_t instr, CodeAddress destination,
                     L xedFunc) {
```
It's much more concise, and we don't have to worry about the compiler not being able to properly eliminate the `std::function` overhead.

hphp/util/asm-x64-intelxed.h
```diff
@@ -0,0 +1,1066 @@
...
+  }
+
+  ALWAYS_INLINE
+  uint32_t xedEmit(xed_iclass_enum_t instr,
+                   const xed_encoder_operand_t& op_1,
+                   const xed_encoder_operand_t& op_2,
+                   const xed_encoder_operand_t& op_3,
+                   xed_uint_t effOperandSizeBits,
+                   CodeAddress destination) {
+    return xedEmitImpl(instr, destination,
```
swtaarrs: We should be able to collapse the `xedEmit` flavors that do and don't take a dest into one function for each number of operands, with something like this:
```c++
uint32_t xedEmit(..., CodeAddress destination = nullptr) {
  auto const size = xedEmitImpl(instr, destination ? destination : dest(), ...);
  if (!destination) codeBlock.moveFrontier(size);
```

hphp/util/asm-x64-intelxed.h
```diff
@@ -0,0 +1,1066 @@
...
+               const xed_encoder_operand_t& op_1,
+               const xed_encoder_operand_t& op_2,
+               const xed_encoder_operand_t& op_3,
+               xed_uint_t effOperandSizeBits = 0) {
+    codeBlock.moveFrontier(xedEmit(instr, op_1, op_2, op_3, effOperandSizeBits,
+                                   dest()));
+  }
+
+public:
+#define INT3_SIZE   (sz::byte)
```
swtaarrs:
```c++
static auto constexpr kInt3Size = sz::byte;
```
same for ud2 as well

hphp/util/asm-x64-intelxed.h
```diff
@@ -0,0 +1,1066 @@
...
+#define UD2_SIZE    (sz::word)
+
+  void emitInt3s(int n) {
+    static uint8_t instr = 0;
+    if (n == 0) return;
+    if (instr == 0) {
+      xedEmit(XED_ICLASS_INT3, sz::byte, &instr);
```
swtaarrs: This isn't thread-safe. Something like this should do it:
```c++
static auto const instr = []{
  uint8_t tmp;
  xedEmit(XED_ICLASS_INT3, sz::byte, &tmp);
  return tmp;
}();
```

hphp/util/asm-x64-intelxed.h
```diff
@@ -0,0 +1,1066 @@
...
+      XED_ICLASS_NOP4,
+      XED_ICLASS_NOP5,
+      XED_ICLASS_NOP6,
+      XED_ICLASS_NOP7,
+      XED_ICLASS_NOP8,
+      XED_ICLASS_NOP9,
+    };
+    // While n >= 9, emit 9 byte NOPs
+    while (n >= 9) {
+      xedInstr(XED_ICLASS_NOP9, 0);
```
swtaarrs: Use `sz::nosize` rather than `0` here and on line 585.

hphp/util/asm-x64-intelxed.h
```diff
@@ -0,0 +1,1066 @@
...
+    if (n) {
+      xedInstr(nops[n], 0);
+    }
+  }
+
+  void pad() {
+    auto remaining = available();
+    static uint8_t padInstructs[UD2_SIZE + INT3_SIZE] = {0};
+    if (padInstructs[0] == 0) {
+      xedEmit(XED_ICLASS_UD2, 0, padInstructs);
```
swtaarrs: Same comment about thread safety as line 557.

hphp/util/asm-x64-intelxed.h
```diff
@@ -0,0 +1,1066 @@
...
+    auto instrLen = xedFunc();
+    lenCache->insert({key, instrLen});
+    return instrLen;
+  }
+
+  static constexpr uint32_t xedLenCacheKey(xed_iclass_enum_t instr,
+                                           uint32_t size) {
+    // 16 bits should fit a xed_iclass_enum_t value (there are currently ~1560
+    // distinct values).
+    return uint32_t(instr) | (size << 16);
```
swtaarrs: `size` will completely represent any potential differences in the size of immediates?

hphp/util/asm-x64.h
```diff
@@ -893,141 +627,125 @@ struct X64Assembler {
 #undef BYTE_LOAD_OP
 #undef BYTE_STORE_OP
 #undef BYTE_REG_OP
+#undef IMM64_STORE_OP
+#undef IMM64R_OP
```
ottoni: Drop this line, it's already undef in 622 above.

hphp/util/asm-x64-legacy.h
```diff
@@ -0,0 +1,1389 @@
...
+#undef IMM64R_OP
+#undef FULL_OP
+#undef REG_OP
+#undef STORE_OP
+#undef LOAD_OP
+#undef BYTE_LOAD_OP
+#undef BYTE_STORE_OP
+#undef BYTE_REG_OP
+#undef IMM64_STORE_OP
+#undef IMM64R_OP
```
ottoni: Drop this -- same as line 295 above.

hphp/util/asm-x64-intelxed.h
```diff
@@ -0,0 +1,1066 @@
...
+                                     return *this; }
+
+private:
+  /*
+   * The following section contains conversion methods that take a Reg8/32/64,
+   * RegXMM, MemoryRef, RipRelative struct and convert it to a
+   * xed_encoder_operand_t.
+   */
+
+  static constexpr int sizeToBits(int sz) {
```
ottoni: nit: maybe call this "bytesToBits" to be clearer

hphp/util/asm-x64-intelxed.h
```diff
@@ -0,0 +1,1066 @@
...
+    return (sz << 3);
+  }
+
+  static constexpr int bitsToSize(int bits) {
```
ottoni: nit: and this "bitsToBytes"

hphp/runtime/vm/jit/vasm-x64.cpp
```diff
@@ -1097,7 +1150,15 @@ void optimizeX64(Vunit& unit, const Abi& abi, bool regalloc) {
 
 void emitX64(Vunit& unit, Vtext& text, CGMeta& fixups,
              AsmInfo* asmInfo) {
-  vasm_emit<Vgen>(unit, text, fixups, asmInfo);
+#ifdef HAVE_LIBXED
+  if (RuntimeOption::EvalUseXedAssembler) {
+    vasm_emit<Vgen<XedAssembler>>(unit, text, fixups, asmInfo);
```
ottoni: You can return here and #endif to avoid duplicate lines 1157 and 1160.

hphp/util/asm-x64.cpp
```diff
@@ -38,6 +38,11 @@ const char* show(RoundDirection rd) {
   not_reached();
 }
 
+#ifdef HAVE_LIBXED
+  static XedInit xi;
+  constexpr xed_state_t XedAssembler::s_xedState;
```
ottoni: I think you're relying on the initialization order here (ie, xed_state_t should be created after xed_tables_init, which is called by XedInit's constructor). I believe the initialization order is guaranteed in this case, but it's worth adding a comment here calling out this implicit dependence.

hphp/util/asm-x64.h
```diff
@@ -787,102 +557,66 @@ struct X64Assembler {
...
-  FULL_OP(add, instr_add)
-  FULL_OP(xor, instr_xor)
-  FULL_OP(sub, instr_sub)
-  FULL_OP(and, instr_and)
-  FULL_OP(or,  instr_or)
-  FULL_OP(test,instr_test)
-  FULL_OP(cmp, instr_cmp)
-  FULL_OP(sbb, instr_sbb)
+#define BYTE_LOAD_OP(name)                                    \
+  virtual void name##b(MemoryRef m, Reg8 r) = 0;
```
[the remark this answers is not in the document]
ottoni: Yep, I'm not a fan of the virtual functions either. But hopefully this is just temporary: if we get rid of the legacy assembler, then we can combine the xed assembler with the base class.

hphp/runtime/base/runtime-option.h
```diff
@@ -860,6 +860,7 @@ struct RuntimeOption {
   F(uint32_t, ProfBranchSampleFreq, 0)                                  \
   /* Sampling frequency for profiling packed array accesses. */         \
   F(uint32_t, ProfPackedArraySampleFreq, 0)                             \
+  F(bool, UseXedAssembler, false)
```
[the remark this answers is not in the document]
catalin-manciu: Addressed in https://github.com/facebook/hhvm/pull/8164/commits/4d4aa256e469a02b132b0343066cc6886b40b45d

hphp/runtime/vm/jit/vasm-x64.cpp
```diff
@@ -340,19 +344,20 @@ template<class Inst> void Vgen::commute(Inst& i) {
 /*
  * Helper for emitting instructions whose Vptr operand specifies a segment.
  */
-X64Assembler& prefix(X64Assembler& a, const Vptr& ptr) {
+template<class X64Asm>
+X64Asm& prefix(X64Asm& a, const Vptr& ptr, MemoryRef* mref) {
```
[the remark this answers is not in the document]
catalin-manciu: Addressed in https://github.com/facebook/hhvm/pull/8164/commits/4d4aa256e469a02b132b0343066cc6886b40b45d

hphp/runtime/vm/jit/vasm-x64.cpp
```diff
@@ -340,19 +344,20 @@ template<class Inst> void Vgen::commute(Inst& i) {
...
+    a.gs(mref);
   }
   return a;
 }
 
 ///////////////////////////////////////////////////////////////////////////////
 
 /*
- * Returns true iff the status flags necessary to take a j<a> imply that a j<b>
+ * Returns true if the status flags necessary to take a j<a> imply that a j<b>
```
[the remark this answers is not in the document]
catalin-manciu: Addressed in https://github.com/facebook/hhvm/pull/8164/commits/4d4aa256e469a02b132b0343066cc6886b40b45d

hphp/runtime/vm/jit/vasm-x64.cpp
```diff
@@ -1097,7 +1150,15 @@ void optimizeX64(Vunit& unit, const Abi& abi, bool regalloc) {
...
-  vasm_emit<Vgen>(unit, text, fixups, asmInfo);
+#ifdef HAVE_LIBXED
+  if (RuntimeOption::EvalUseXedAssembler) {
+    vasm_emit<Vgen<XedAssembler>>(unit, text, fixups, asmInfo);
+  } else {
+    vasm_emit<Vgen<X64Assembler>>(unit, text, fixups, asmInfo);
+  }
+#else
+  vasm_emit<Vgen<X64Assembler>>(unit, text, fixups, asmInfo);
+#endif
```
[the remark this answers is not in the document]
catalin-manciu: Addressed in https://github.com/facebook/hhvm/pull/8164/commits/ac4a93085f536b08c1f366528bb51a4599d8b290

hphp/util/asm-x64-intelxed.h
```diff
@@ -0,0 +1,1066 @@
...
+
+///////////////////////////////////////////////////////////////////////////////
+
+typedef bool(*immFitFunc)(int64_t, int);
+
+///////////////////////////////////////////////////////////////////////////////
+
+struct XedAssembler : public X64AssemblerBase {
+private:
+  constexpr static xed_state_t s_xedState = {
```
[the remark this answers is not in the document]
catalin-manciu: Addressed in https://github.com/facebook/hhvm/pull/8164/commits/ac4a93085f536b08c1f366528bb51a4599d8b290

hphp/util/asm-x64-intelxed.h
```diff
@@ -0,0 +1,1066 @@
...
+  ALWAYS_INLINE
+  xed_iclass_enum_t ccToXedCMov(ConditionCode c) {
+    assertx(c != CC_None);
+    static const xed_iclass_enum_t cmovs[] = CC_TO_XED_ARRAY(CMOV);
+    return cmovs[(int)c];
+  }
+
+  // XED emit functions
+
+#define DECLARE_UNUSED(type, name)  type name; static_cast<void>(name)
```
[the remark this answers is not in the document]
catalin-manciu: Addressed in https://github.com/facebook/hhvm/pull/8164/commits/ac4a93085f536b08c1f366528bb51a4599d8b290

hphp/util/asm-x64-intelxed.h
catalin-manciu: Addressed in https://github.com/facebook/hhvm/pull/8164/commits/6923adab7c8900c24fbe55f12632909549f13627

hphp/util/asm-x64-intelxed.h
catalin-manciu: Addressed in https://github.com/facebook/hhvm/pull/8164/commits/41f6fc4dee0b2a2b3e3ea52d1c583202b583425f

hphp/util/asm-x64-intelxed.h
catalin-manciu: Addressed in https://github.com/facebook/hhvm/pull/8164/commits/ac4a93085f536b08c1f366528bb51a4599d8b290

hphp/util/asm-x64.h
catalin-manciu: Addressed in https://github.com/facebook/hhvm/pull/8164/commits/ac4a93085f536b08c1f366528bb51a4599d8b290

hphp/util/asm-x64-legacy.h
catalin-manciu: https://github.com/facebook/hhvm/pull/8164/commits/ac4a93085f536b08c1f366528bb51a4599d8b290

hphp/util/asm-x64-intelxed.h
catalin-manciu: Addressed in https://github.com/facebook/hhvm/pull/8164/commits/ffa4555156393d19cfa7b8b0ca6f51db5253dae2

hphp/util/asm-x64-intelxed.h
catalin-manciu: Addressed in https://github.com/facebook/hhvm/pull/8164/commits/ffa4555156393d19cfa7b8b0ca6f51db5253dae2

hphp/runtime/vm/jit/vasm-x64.cpp
catalin-manciu: Addressed in https://github.com/facebook/hhvm/pull/8164/commits/ac4a93085f536b08c1f366528bb51a4599d8b290

hphp/util/asm-x64-intelxed.h
```diff
@@ -0,0 +1,1066 @@
...
+  };
+
+  CodeAddress dest() const {
+    codeBlock.assertCanEmit(XED_MAX_INSTRUCTION_BYTES);
+    return codeBlock.toDestAddress(codeBlock.frontier());
+  }
+
+  static constexpr auto nullrip = RIPRelativeRef(DispRIP(0));
+  static constexpr auto immfitSigned = deltaFits;
+  static constexpr auto immfitUnsigned = (immFitFunc)magFits;
```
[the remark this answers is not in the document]
catalin-manciu: It's not currently used. The only reason I currently have for keeping this here is to support any future instruction that might require it.

hphp/util/asm-x64-intelxed.h
catalin-manciu: The cache stores instructions sizes and each entry is uniquely identified by: the instruction itself, and the size of the immediate (which is only used for branch-relative jumps). There's a cache for each xedInstr* function that requires it, which means that the each cached size is defined by: the instruction, the immediate size (in some cases), and the instruction's argument types (since there's a cache for each xedInstr* function which means there's a cache for each combination of instruction argument types) - which should be sufficient.

hphp/util/asm-x64.cpp
catalin-manciu: There's no dependency between kXedState and the Xed initialization.  kXedState is just used later as a mandatory argument to the xed_inst* method.

ottoni:
@catalin-manciu, it seems that there are some conflicts. Can you take a look at those and rebase this branch on current master?

catalin-manciu:
@ottoni Rebased the commits

hphp/util/asm-x64.h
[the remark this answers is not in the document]
catalin-manciu: Classes marked as final in https://github.com/facebook/hhvm/pull/8164/commits/1d5d659f057cabcd04ef79992d0cf775c37a9717

hphp/util/asm-x64-intelxed.h
catalin-manciu: Addressed in https://github.com/facebook/hhvm/pull/8164/commits/ec24a9fdfb463fda92b86f6ac4e1061488fa4d06

ottoni:
@catalin-manciu, this is currently blocked on the missing third-party support.

fredemmott:
PR here, AIUI we're waiting for Intel legal

catalin-manciu:
Rebased PR, fixed conflicts and a recently found issue (an incorrect size being passed to XED when encoding MOVSX(src = reg16/8, dest = reg64/32) and MOVSX(src = mem8, dest = reg64).

catalin-manciu:
Commit https://github.com/facebook/hhvm/pull/8164/commits/78986fea52d5439a975b19d72894eb8728c4b381 addresses an older comment by @swtaarrs:
> [...]can you delete this and the immFitFunc arguments passed to a bunch of other functions?[...]

and fixes Clang build (validated with Clang 6.0).

CMake/FindLibXed.cmake
```diff
@@ -1,4 +1,42 @@
...
+      // we're only interested in the major version number
+      return !(atoi(version + 1) >= 10);
+    }"
+    LibXed_GOOD_VERSION)
+  set(CMAKE_REQUIRED_INCLUDES)
+  set(CMAKE_REQUIRED_LIBRARIES)
+
+  if (NOT LibXed_GOOD_VERSION)
+    message(STATUS "LibXed version check failed")
+  endif()
```
fredemmott: You need to do one out of:
- completely remove support for system installations of libxed
- only use system libxed if explicitly specified
- make it so that if not LibXED_GOOD_VERSION, we use third-party/, even if LibXed is found

In particular, if libxed.so is present, it currently seems more likely to be a library for the X-Apps Editor (https://github.com/linuxmint/xed) than the intel library

CMake/FindLibXed.cmake
```diff
@@ -1,4 +1,42 @@
...
+endif()
+
+include(FindPackageHandleStandardArgs)
+FIND_PACKAGE_HANDLE_STANDARD_ARGS(LibXed
+                                  DEFAULT_MSG
+                                  LibXed_LIBRARY
+                                  LibXed_INCLUDE_DIR
+                                  LibXed_GOOD_VERSION)
+if (NOT LibXed_FOUND)
+  message(STATUS "Using third-party bundled LibXed")
```
fredemmott: This is one of the checks that needs updating

CMake/HPHPFindLibs.cmake
```diff
@@ -81,10 +81,14 @@ endif()
...
 # libXed
-find_package(LibXed)
-if (LibXed_INCLUDE_DIR AND LibXed_LIBRARY)
-  include_directories(${LibXed_INCLUDE_DIR})
+if (NOT DISABLE_XED)
+  find_package(LibXed)
+  if (LibXed_FOUND)
+    include_directories(${LibXed_INCLUDE_DIR})
+  endif()
   add_definitions("-DHAVE_LIBXED")
```
fredemmott: - you probably want this when using third-party
- it's currently being defined even when xed is not found, and third-party is not available, leading to build failrues

CMake/FindLibXed.cmake
catalin-manciu: In `CMake/FindLibXed.cmake` there is a test-run of the system libxed. If it fails, `LibXed_GOOD_VERSION` will be set accordingly.
`FIND_PACKAGE_HANDLE_STANDARD_ARGS` (line 31) will set `LibXed_FOUND` only if: 
* valid paths were found (`LibXed_LIBRARY` and `LibXed_INCLUDE_DIR` are valid) **and**
* `LibXed_GOOD_VERSION` is valid

If there is another library with the same name on the system, the test-run will fail, `LibXed_GOOD_VERSION` will be set to FALSE and `FIND_PACKAGE_HANDLE_STANDARD_ARGS` will **not** set `LibXed_FOUND`.
So, currently, `LibXed_FOUND` is only set if a system libxed exists and it's valid. If `LibXed_FOUND` is not set, third-party is used. This is the same behavior as libLZ4 and follows your third bullet.

CMake/HPHPFindLibs.cmake
catalin-manciu: In summary, if the system libxed was not validated and there's no third-party directory, the build should not fail? I could add checks using `if(TARGET xed)` to verify that there's a third-party xed available, but this is not a pattern widely used by other third party libraries. Do you think that's the way to go?

fredemmott:
So, ultimately, the problem is we're not really able to do pre-commit testing of the third-party rev. We've decided to accept there, and if there's issues, either fix forward quickly or revert, depending on if we need to do any other changes to third-party before this is resolved.

1. I have merged the PR to the third-party repository
1. This PR needs updating to also update the submodule
1. We expect this to have a reasonably high chance of breaking some OSS builds, so, once it's ready to merge (hopefully just the submodule update needed), I will merge it on a monday through wednesday, on a day when the previous days nightly builds succeeded.

catalin-manciu:
Rebased & updated third-party to latest. Thank you for merging the third-party PR!

catalin-manciu:
Rebased the PR, made it compatible with the new ENABLE_XED option: [https://github.com/facebook/hhvm/pull/8164/commits/01b6de196c9706df073cc2b9bd50e8b9fd89e18b](https://github.com/facebook/hhvm/pull/8164/commits/01b6de196c9706df073cc2b9bd50e8b9fd89e18b) and fixed a MacOS runtime error (which didn't occur on Ubuntu): [https://github.com/facebook/hhvm/pull/8164/commits/cd9bdc00bcaf5cac9b68c4d3ed3511f1269bb35a](https://github.com/facebook/hhvm/pull/8164/commits/cd9bdc00bcaf5cac9b68c4d3ed3511f1269bb35a).
Building on MacOS will result in a binary that passes the regression tests (`./run -a "-vEval.UseXedAssembler=true/false" all`). I've followed [this guide](https://docs.hhvm.com/hhvm/installation/building-from-source#building-hhvm__macos) and didn't encounter any issues when building for this platform.

Happy New Year, everyone! :sparkles:

catalin-manciu:
Rebased the PR, fixed conflict in asm-x64.h

PR #8166, opened 2018-03-30
Add C flags to CMake that are currently required on MacOS

fredemmott:
Trying to get a flag-free build working on MacOS.

LZ4 flag is needed because of clang, ICU4C because of ICU4C 61.1 having
a breaking change.

Test Plan:

hhvm-mac, manual build on my macs.


Stork526:
mute the thread

PR #8167, opened 2018-04-02
Disable non-mac extensions by default on macos

fredemmott:
Test plan:

did a mac build with no CMake arguments, it worked.

### DIFF
--- a/CMake/Options.cmake
+++ b/CMake/Options.cmake
@@ -26,7 +26,11 @@ option(CPACK_GENERATOR "Enable build of distribution packages using CPack" OFF)
 
 option(ENABLE_COTIRE "Speed up the build by precompiling headers" OFF)
 option(ENABLE_ASYNC_MYSQL "Build the async_mysql extension" ON)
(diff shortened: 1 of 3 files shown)


CMake/Options.cmake
```diff
@@ -26,7 +26,11 @@ option(CPACK_GENERATOR "Enable build of distribution packages using CPack" OFF)
 
 option(ENABLE_COTIRE "Speed up the build by precompiling headers" OFF)
 option(ENABLE_ASYNC_MYSQL "Build the async_mysql extension" ON)
-option(ENABLE_MCROUTER "Build the mcrouter library and extension" ON)
+if (APPLE)
+  option(ENABLE_MCROUTER "Build the mcrouter library and extension" OFF)
```
Orvid: Should probably just remove this variable entirely. I left it when I did the migration 2.5 years ago in order to not break things.

fredemmott: I'd rather do that separately so this is a clean cherry-pick for the release branches; replacing ENABLE_MCROUTER requires changing third-party

PR #8170, opened 2018-04-06
Fix compilation error (numa_node_set undefined) for non-NUMA builds

kristapsk:
Fixes compilation error for non-NUMA builds which was introduced with afa3f0811556226a202b724aab8d41c94ae55675.


fredemmott:
Huh. This makes sense, though we have done non-numa builds since then :/

PR #8203, opened 2018-05-16
Avoiding the GC to crash when listing the modules supported.

dcarlier-afilias:


Wilfred:
@dcarlier-afilias can you advise how to reproduce the crash?

dcarlier-afilias:
hhvm --modules

PR #8213, opened 2018-05-25
[AArch64] Explicitly turn PIE off in the event that GCC has it on

dave-estes-QCOM:
Some GCC toolchains are being built with PIE defaulted to on.
PIE breaks the ability to use sbrk() to reserve low memory. This
patch ensures that PIE is off regardless of how it was configured.


dave-estes:
@swalk-cavium, this is the change that I had to make to get around the issue when building on a Debian.
Ping @mxw.

alshamiri5 (approved):

PR #8237, opened 2018-06-21
aarch64 - patch for folly F14

swalk-cavium:
This patch updates the folly library so that hhvm continues
  to build on aarch64 hosts.  A recent change (6a19df5)  made 
  use of capabilities that are not yet present in the third-party 
  folly library.  One commit is cherry-picked  (19c316a).

  This patch should be re-evaluated if the third-party folly
  is updated.

  The regression tests were run with 6 option sets.  No new issues were
   observed.


swalk-cavium:
@mxw - Can you review?

Issue #8263, opened 2018-07-17
tc-print will not link


catalin-manciu:
### HHVM Version
All when building with `-DLibXed_INCLUDE_DIR=/valid/dir` and `-DLibXed_LIBRARY=/valid/file`.
### Operating System and Version
Ubuntu 16.04
### Standalone code, or other way to reproduce the problem
Commit 98b7c7cf01a7d9900c400ddd8b2e61cfce09be02 added `hphp/tools/tc-print/std-logger.cpp`, but that file was not referenced in `hphp/tools/tc-print/CMakeLists.txt`.
Adding `std-logger.cpp` to `CMakeLists.txt`, as a dependency for the **tc-print** target will fix this (lines 6 and 19).
### Actual result
tc-print linking fails with: `undefined reference to 'vtable for HPHP::StdLogger'`.
### Expected result
tc-print is built successfully.

catalin-manciu:
Submitted a [pull request](https://github.com/facebook/hhvm/pull/8267) which fixes this.

catalin-manciu:
This was fixed

PR #8267, opened 2018-07-19
Fix tc-print linking error

catalin-manciu:
Update tc-print/CMakeLists.txt and add std-logger.cpp to the tc-print target.


catalin-manciu:
This fixes https://github.com/facebook/hhvm/issues/8263.

catalin-manciu:
This has been fixed

PR #8274, opened 2018-07-31
Redirect people to new Emacs hack-mode plugin

jra3:
Summary:
There is a new hack-mode plugin with its own OSS. Here I've
removed the old plugin and slightly changed the language in the README

Reviewed By: tomasreimers

Differential Revision: D9075757


fredemmott:
@jra3: this was merged a few days ago, but not marked as such because of the internal edits made to the diff summary, removing the tracking metadata

PR #8277, opened 2018-08-08
Add hackfmt support for single line if statements

alisha:
Currently, a single line if statement will be split into 2 lines after hackfmt is run. For example,
```
if (true) print("Hello, world!");
```
becomes
```
if (true)
  print("Hello, world!");
```

This PR changes hackfmt so it keeps single line if statements, as long as the line isn't too long. If the line exceeds the specified limit, then it'll be split into 2 lines.

This was tested on a file with both single line if statements (of various lengths) and multi-line if statements.

### DIFF
--- a/hphp/hack/src/hackfmt/hack_format.ml
+++ b/hphp/hack/src/hackfmt/hack_format.ml
@@ -814,7 +814,7 @@ let rec t (env: Env.t) (node: Syntax.t) : Doc.t =
       t env kw;
       Space;
       transform_condition env left_p condition right_p;
-      handle_possible_compound_statement env if_body;
+      transform_consequence t env if_body right_p;
       handle_possible_list env elseif_clauses;
       t env else_clause;
       Newline;
@@ -829,7 +829,7 @@ let rec t (env: Env.t) (node: Syntax.t) : Doc.t =
       t env kw;
       Space;
       transform_condition env left_p condition right_p;
-      handle_possible_compound_statement env body;
+      transform_consequence t env body right_p;
     ]
   | Syntax.ElseClause x ->
     Concat [
@@ -840,7 +840,7 @@ let rec t (env: Env.t) (node: Syntax.t) : Doc.t =
           t env x.else_statement;
           Space;
         ]
-      | _ -> handle_possible_compound_statement env x.else_statement
+      | _ -> transform_consequence t env x.else_statement x.else_keyword;
     ]
   | Syntax.AlternateIfStatement {
       alternate_if_keyword = kw;
@@ -3360,6 +3360,20 @@ and transform_xhp_leading_trivia triv =
 and node_has_trailing_newline node =
   let trivia = Syntax.trailing_trivia node in
   List.exists trivia ~f:(fun x -> Trivia.kind x = TriviaKind.EndOfLine)
+
+and transform_consequence t (env: Env.t) (node_body: Syntax.t) (node_newline: Syntax.t) = 
+  match Syntax.syntax node_body with
+  | Syntax.CompoundStatement _ -> handle_possible_compound_statement env node_body
+  | _ -> Concat [
+      Space;
+      if has_newline (Syntax.trailing_trivia node_newline)
+      then Concat [Newline; Nest [t env node_body]]
+      else WithRule (Rule.Parental, Nest [ Span [
+        Space;
+        Split;
+        t env node_body;
+      ]]);
+    ]
 
 let transform (env: Env.t) (node: Syntax.t) : Doc.t =
   t env node
--- a/hphp/hack/test/hackfmt/tests/allow_single_line_ifs.php
+++ b/hphp/hack/test/hackfmt/tests/allow_single_line_ifs.php
@@ -1,0 +1,89 @@
+<?hh // strict
+
+// Should be on a single line
+if (true) print("hi");
+
+// Should not be on a single line
+if (true)
+print("hi");
+
+if (true)
+  print("hi");
+
+if (true) {
+print("this is a");
+print("multi line if");
+}
+
+if (true)
+print("this is not a");
+print("multi line if");
+
+// Should all have the if body on a new line
+if (true) this_is_a_very_long_function_call_that_really_exceeds_the_line_length_lorem_ipsum_dolor_sit_amet_consectetur_adipiscing_elit();
+
+if (true)
+this_is_a_very_long_function_call_that_really_exceeds_the_line_length_lorem_ipsum_dolor_sit_amet_consectetur_adipiscing_elit();
+
+if (true)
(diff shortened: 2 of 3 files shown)


fredemmott:
I'll need others to evaluate the implementation, but this will need tests adding (hphp/hack/test/hackfmt/tests)

hphp/hack/src/hackfmt/hack_format.ml
```diff
@@ -822,7 +822,17 @@ let rec t (env: Env.t) (node: Syntax.t) : Doc.t =
...
       Space;
       transform_condition env left_p condition right_p;
-      handle_possible_compound_statement env if_body;
+      (match Syntax.syntax if_body with
+      | Syntax.CompoundStatement {
+        compound_left_brace;
+        compound_statements;
+        compound_right_brace; } ->
+          handle_possible_compound_statement env if_body;
+      | _ -> Concat [
```
fredemmott: please test how this works with format-as-you-type (e.g. nuclide, vscode with `"editor.formatOnType": true` set)

hphp/hack/test/hackfmt/tests/allow_single_line_ifs.php.exp
```diff
@@ -0,0 +1,24 @@
...
+if (true) {
+	print("this is a");
+	print("multi line if");
+}
+
+if (true) print("this is not a");
+print("multi line if");
+
+if (true)
+  this_is_a_very_long_function_call_that_really_exceeds_the_line_length_lorem_ipsum_dolor_sit_amet_consectetur_adipiscing_elit();
```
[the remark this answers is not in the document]
alisha: Good catch! That was a mistake from my editor using tabs. I'll fix that!

hphp/hack/src/hackfmt/hack_format.ml
```diff
@@ -822,7 +822,18 @@ let rec t (env: Env.t) (node: Syntax.t) : Doc.t =
...
+          Space;
+          Split;
```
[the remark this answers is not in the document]
alisha: Thanks for the feedback! I fixed the unused variable bindings and the `Split`, and also added tests for the example you posted.

fredemmott:
hphp/hack/test/hackfmt/difftests/multifile.diff needs updating

fredemmott:
To run these:

```
difftests fredemmott$ hackfmt --test --diff-dry-run --root root < multifile.diff
```

hphp/hack/src/hackfmt/hack_format.ml
```diff
@@ -822,7 +822,15 @@ let rec t (env: Env.t) (node: Syntax.t) : Doc.t =
...
       Space;
       transform_condition env left_p condition right_p;
-      handle_possible_compound_statement env if_body;
+      (match Syntax.syntax if_body with
+      | Syntax.CompoundStatement _ ->
+          handle_possible_compound_statement env if_body;
+      | _ -> Concat [
+          Space;
+          SplitWith Cost.Base;
+          Nest [t env if_body];
```
rjbailey: @alisha, I'm seeing some pretty funky-looking results from some of our internal test cases.

Code that was previously formatted like this:

```
if ($this->head->isEmpty())
  return new MyQueue($this->head->push($item), $this->tail);
else
  return new MyQueue($this->head, $this->tail->push($item));
```

Is now formatted like this:

```
if ($this->head->isEmpty()) return
  new MyQueue($this->head->push($item), $this->tail);
else return new MyQueue($this->head, $this->tail->push($item));
```

We could probably deal with this case:

```
if ($cond) return
  $x;
```

by using a `Parental` rule to control the split before the body instead of an independent split (`SplitWith`):

```
WithRule (Rule.Parental, Nest [ Span [
  Space;
  Split;
  t env if_body;
]]);
```

Using a `Parental` rule means that if any line break is added within the body, we'll add a line break at the `Split` preceding the body, too. Sorry for suggesting `SplitWith` earlier--I hadn't considered this possible outcome!

Even if we deal with that case, to be frank, I think this formatting:

```
if ($this->head->isEmpty())
  return new MyQueue($this->head->push($item), $this->tail);
else
  return new MyQueue($this->head, $this->tail->push($item));
```

Is much more readable than this:

```
if ($this->head->isEmpty())
  return new MyQueue($this->head->push($item), $this->tail);
else return new MyQueue($this->head, $this->tail->push($item));
```

Or this:

```
if ($empty) return new MyQueue($this->head->push($item), $this->tail);
else return new MyQueue($this->head, $this->tail->push($item));
```

So I've become a little more uneasy about this change! I feel like some human input is useful in deciding when it's appropriate to put a single-statement `if` on one line.

One way we could deal with this is by preserving single-line `if`s when they already exist (and fit within the line length limit), and not forcing other single-statement `if`s onto one line if they aren't already. We just need to find out if there was already a newline after the condition's right parenthesis:

```
if has_newline (Syntax.trailing_trivia right_p)
then Newline
else Split;
```

Since this is starting to be a fair bit of code which is duplicated across `IfStatement`, `ElseifClause`, and `ElseClause`, perhaps we should move it to a new helper function called something like `transform_consequence` :)

alisha:
Thanks for the feedback @rjbailey! I've changed the PR so that single line if statements are preserved when we're not over the line limit, but newlines are no longer removed. I followed your suggestions to use the Parental rule, and moved all this logic into the `transform_consequence` function. I also updated the test files and the behavior looks correct now, but please let me know if I missed anything.

I also did a rebase to make sure there are no merge conflicts with the upstream master branch, and squashed my commits into one to make this PR a bit cleaner :)

hphp/hack/test/hackfmt/tests/allow_single_line_ifs.php
```diff
@@ -0,0 +1,82 @@
...
+
+if (true)
+this_is_a_very_long_function_call_that_really_exceeds_the_line_length_lorem_ipsum_dolor_sit_amet_consectetur_adipiscing_elit();
+
+if (true)
+  this_is_a_very_long_function_call_that_really_exceeds_the_line_length_lorem_ipsum_dolor_sit_amet_consectetur_adipiscing_elit();
+
+// Should preserve the single line if
+f(g($a, $b), () ==> {
+  if (true) print('hi');
```
pittsw: Can we also add a test here where we have a multiline `if` statement that's inside a block?  Something like this:

```
  if (true) print('hi');
  if (true)
  print('hi');
```

This way we can ensure that the parental rule is activating properly :)

pittsw:
Looks good to me!  Just left one comment.

hphp/hack/test/hackfmt/tests/allow_single_line_ifs.php
alisha: Thanks for the feedback! I added the multi-line if statement, as well as a single-line if statement inside a multi-line if statement below

pittsw:
Looks good to me!  Once internal tests finish I can merge it in :)

PR #8289, opened 2018-08-20
Fix some misspellings

carusogabriel:


fredemmott:
@carusogabriel : thanks - FYI, "Haswell" is an Intel architecture, and should not be updated to "Has well". I'm fixing that for this PR, but you might want to exclude that in your script.

carusogabriel:
@fredemmott Anything left here to do for its merge?

carusogabriel:
Closing by the number of conflicts and the time this is opened.

PR #8297, opened 2018-08-25
Escape Hatch Extension: Simple, Fast, Async IPC

y3llowcake:
Escape Hatch: A simple asynchronous IPC mechanism intended for use with a
sidecar process. Supports request+response or one-way requests over pooled
[unix domain] sockets. Every message in either direction is framed with
an 8 byte header:
- Magic bytes 'E' then 'H'
- A single byte; 0x00 for oneway, or 0x01 for response requested.
- Another byte, ignored.
- An unsigned 32 bit big endian integer indicating the length of the
  payload to come.

This is a dirty PR that isn't meant to be merged as is. Per Fred's suggestion, I'm sending this out to see if Facebook would be willing to take this upstream once cleaned. We've been using this extension in production under medium load for the last few months. I'd like to reduce the burden of tracking HHVM/Folly changes and maintaining the build tooling for this extension.

### DIFF
--- a/hphp/runtime/ext/escape_hatch/ext_escape_hatch.cpp
+++ b/hphp/runtime/ext/escape_hatch/ext_escape_hatch.cpp
@@ -36,10 +36,6 @@ namespace HPHP {
  * - Another byte, ignored.
  * - An unsigned 32 bit big endian integer indicating the length of the
  *   payload to come.
- *
- * There are no timeouts, and possibly other gotchas. Careful!
- *
- * Grab some data and let's get the hell outta Dodge.
  */
 
 static const auto optPool = String::FromCStr("pool");
--- a/hphp/runtime/ext/escape_hatch/ext_escape_hatch.cpp
+++ b/hphp/runtime/ext/escape_hatch/ext_escape_hatch.cpp
@@ -1,0 +1,504 @@
+#include <folly/Chrono.h>
+#include <folly/io/async/AsyncSocket.h>
+#include <folly/io/IOBuf.h>
+#include <folly/io/IOBufQueue.h>
+
+#include "hphp/runtime/ext/asio/asio-external-thread-event.h"
+#include "hphp/runtime/ext/asio/socket-event.h"
+#include "hphp/runtime/ext/extension.h"
+
+#include "hphp/util/compatibility.h"
+#include "hphp/util/logger.h"
+
+using std::shared_ptr;
+using folly::AsyncSocket;
+using folly::AsyncTimeout;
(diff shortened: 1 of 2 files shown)


fredemmott:
Currently on hold: recent bugfixes to ASIO (which affected `stream_await()` to a huge extent) under libevent2 might mean that Hack code is a better approach.

If the native impl is still wanted in 3.28+, then we'll look into merging this - just don't want to merge more native code if this no longer adds things.

I'm not looking for "this can't be done in Hack" - I'm looking for "we don't think Hack is the best approach for this because ..." . This was easy to answer below 3.24.8/3.27.1, but hopefully is a harder question now :)

fredemmott:
I'm closing this, as this is the kind of problem we expect to be solveable in pure Hack soon.

Based on conversations with @y3llowcake , I believe the requirements are:
- get the HSL IO landed (~ 2 weeks, waiting for 170f3b9ca28304c3047ab0853ae2d8d0ebe9d839 to be deployed @ FB)
- extend the HSL IO - including new native functionality - as covered in https://github.com/hhvm/hsl-experimental/issues/21

This will hopefully be possible for 3.30 (December), though my priority through then is to remove the need for PHP dependencies from our tools and libraries. If it's not 3.30, it should be the release after.

It does not seem worthwhile to merge this when we're hoping to make it obsolete before the first release that could potentially include it.

That said:
- we're open to merging something similar to this if we can't make pure Hack a viable alternative
- we would require thorough unit tests before merging this: we wouldn't be using it, and having an untested thing in-tree is likely to make things harder for everyone

hphp/runtime/ext/escape_hatch/ext_escape_hatch.cpp
```diff
@@ -0,0 +1,500 @@
...
+ * [unix domain] sockets. Every message in either direction is framed with
+ * an 8 byte header:
+ * - Magic bytes 'E' then 'H'
+ * - A single byte; 0x00 for oneway, or 0x01 for response requested.
+ * - Another byte, ignored.
+ * - An unsigned 32 bit big endian integer indicating the length of the
+ *   payload to come.
+ */
+
+static const auto optPool = String::FromCStr("pool");
```
fredemmott: the usual pattern is `const StaticString s_optPool("pool")`

`static` /could/ be added there, but we generally wrap the file-local stuff in `namespace {` instead

y3llowcake: done

PR #8299, opened 2018-08-27
Aarch64 - Prevent unnecessary cache syncs

jim-saxman:
Summary:
On Aarch64, ___builtin__clear_cache() is used to ensure I and D
cache coherency when writing out JIT'd instructions. Sometimes (6%)
Vgen is destroyed with no modified code, however the expensive
___builtin__clear_cache() function is still called. This patch
prevents unncessary cache syncs when no bytes have been written to
the data-block.

There are no new unit test case regressions with this patch, and
the time to run the unit test suite is reduced.

### DIFF
--- a/hphp/util/data-block.h
+++ b/hphp/util/data-block.h
@@ -284,7 +284,7 @@ struct DataBlock {
   }
 
   static void syncDirect(Address begin,  Address end) {
-    if (arch() == Arch::ARM) {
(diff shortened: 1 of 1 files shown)


jim-saxman:
Hi @mxw, @swalk-cavium can you please review?

swalk-cavium:
@jim-saxman - LGTM

PR #8301, opened 2018-08-27
Aarch64 Sync I and D caches after modifying instructions

jim-saxman:
Summary:
When writing instructions on Aarch64, the I and D caches need to
be made cohorent, so call sync() and syncDirect().


ottoni:
Do you have any example that breaks without this? 
I'm not sure these changes are needed. Both handleLiterals() and the optimizeSmashed*() functions run on code that hasn't been published (ie, made reachable) yet, so none of that code should be in the I-caches.
Also, you need to rebase as I recently got rid of some of those optimizeSmashed* functions in https://github.com/facebook/hhvm/commit/a675885a848505d3f8871a507148e85a9196e150.

jim-saxman:
Hi @ottoni, thanks for the quick review. I don't have an example that breaks. After reviewing your a675885 I see what's going on and your SmashInst() properly covers the cache coherency part I was was trying to address. I just closed this PR.

ottoni:
Cool, thanks.

PR #8303, opened 2018-08-27
Aarch64 Use immediate forms of branch-and-link if possible

jim-saxman:
Summary:
When emiting a call, try and use a pc-relative bl instruction if
possible.

### DIFF
--- a/hphp/runtime/vm/jit/relocation-arm.cpp
+++ b/hphp/runtime/vm/jit/relocation-arm.cpp
@@ -46,6 +46,7 @@ TRACE_SET_MOD(mcg);
  *   PC Relative
  *     ADR/ADRP - Builds PC relatative addresses.
  *     B[.<cc>] (immediate) - Branch to a PC relative address.
+ *     BL (immediate) - Branch with link to a PC relative address.
  *     LDR (literal) - Loads a literal from a PC relative address.
  *
  *   Immediates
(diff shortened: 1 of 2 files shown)


jim-saxman:
Hi @mxw, @swalk-cavium, can you please review?

ottoni:
I think this should be done in the relocator instead, because code is typically not emitted in its final place.

jim-saxman:
Hi @ottoni, thanks for the quick review. I was targeting the call's emitted in the unique-stubs*, which are not relocated, correct?

ottoni:
@jim-saxman, I see. Yeah, I think the unique stubs are not relocated.

One concern about this is that, for code that is relocated, I'm not sure the relocator currently knows how to turn the call back into a BLR in case the offset doesn't fit into the immediate after the code is relocated. Can you please check this and, if needed, add such support to the relocator?

jim-saxman:
Hi @ottoni, The relevant code in relocation-arm.cpp is around the use of IsUncondBranchImm(). Good thing you asked me to check, since this function returns true for both B (imm) and BL (imm). The only difference in these two instructions' encoding is the 31st bit.

In relocatePCRelative() if IsUncondBranchImm() is true and is_int26(imm) is false the relocator will always use a BR instruction. I have a fix which will output a BLR instruction if that's what is called for. I will post it after the unit testing completes.

hphp/runtime/vm/jit/relocation-arm.cpp
```diff
@@ -610,6 +612,7 @@ bool relocatePCRelative(Env& env, TCA srcAddr, TCA destAddr,
         isRelative = false;
       }
     } else if (src->IsUncondBranchImm()) {
+      // Handle both B and BL forms of UncondBranchImm.
       imm >>= kInstructionSizeLog2;
       if (!is_int26(imm) || env.far.count(src)) {
         env.destBlock.setFrontier(destAddr);
```
jim-saxman: I couldn't exercise this code as is during unit testing or OSS/Mediawiki. However, I forced it by using `if (!is_int6(imm) || env.far.count(src)) {` on the previous line.

PR #8304, opened 2018-08-27
Aarch64 - Disable executable stacks

jim-saxman:
Summary:
This patch marks assembly files as having a non-executable stack.
It also updates the Aarch64 gold linker's command line to force
the stack to be non-execuatable and throw a warning if the stack
becomes executable in the future.

There are no new unit test case regressions with this patch, and
the time to run the unit test suite is greatly reduced.


jim-saxman:
Hi @mxw, @swalk-cavium, can you please review?

swalk-cavium:
@jim-saxman - LGTM

PR #8308, opened 2018-08-30
Update MySQL stuff in third-party

fredemmott:
- update to latest versions of squangle and fb-mysql
- rename the webscalesqlclient targets/directories to fb-mysql to reflect which project we're using - it's been fb-mysql for a while now.


danirod1386:
nice as shit

PR #8315, opened 2018-09-05
aarch64 - loadb/movsbl->loadsbl peephole

swalk-cavium:
While examining the tracefile for mediawiki, I saw an opportunity to remove
some sign-extension instructions.  There is a form of the load byte instruction
which sign-extends the value.  

Before
======
```
    (01) t1:StkPtr = DefSP<FPInvOff 2> t0:FramePtr
    (04) CheckLoc<Uninit,1> t0:FramePtr -> B2<Unlikely>
        Main:        
               0xf6002b4  385e83a0              ldurb w0, [x29, #-24]
               0xf6002b8  13001c00              sxtb w0, w0  //<<---
               0xf6002bc  7100001f              cmp w0, #0x0 (0) 
               0xf6002c0  540003a1              b.ne #+0x74
```

After
=====
```
    (01) t1:StkPtr = DefSP<FPInvOff 2> t0:FramePtr
    (04) CheckLoc<Uninit,1> t0:FramePtr -> B2<Unlikely>
        Main:  
               0xf601a5c  38de83a0              ldursb w0, [x29, #-24]  //<<---
               0xf601a60  7100001f              cmp w0, #0x0 (0) 
               0xf601a64  540003a1              b.ne #+0x74 
```

The standard regression tests were run with 6 option sets.  No new regressions were
observed.  The oss-performance mediawiki test was run before/after.  A (very) small
improvement was seen.

### DIFF
--- a/hphp/runtime/vm/jit/vasm-simplify-arm.cpp
+++ b/hphp/runtime/vm/jit/vasm-simplify-arm.cpp
@@ -85,7 +85,7 @@ bool simplify(Env& env, const cmovq& inst, Vlabel b, size_t i) {
 ///////////////////////////////////////////////////////////////////////////////
 
 template<Vinstr::Opcode op>
-bool is_next(const Env& env, Vlabel b, size_t i)
+bool is_inst(const Env& env, Vlabel b, size_t i)
 {
   auto const& code = env.unit.blocks[b].code;
   if (i > code.size()) return false;
@@ -93,7 +93,7 @@ bool is_next(const Env& env, Vlabel b, size_t i)
 }
 
 bool simplify(Env& env, const loadb& inst, Vlabel b, size_t i) {
-  if (is_next<Vinstr::movzbl>(env, b, i + 1)) {
+  if (is_inst<Vinstr::movzbl>(env, b, i + 1)) {
     return if_inst<Vinstr::movzbl>(env, b, i + 1, [&] (const movzbl& mov) {
       // loadb{s, tmp}; movzbl{tmp, d}; -> loadzbl{s, d};
       if (!(env.use_counts[inst.d] == 1 &&
@@ -104,7 +104,7 @@ bool simplify(Env& env, const loadb& inst, Vlabel b, size_t i) {
         return 2;
       }); });
   }
-  if (is_next<Vinstr::movsbl>(env, b, i + 1)) {
+  if (is_inst<Vinstr::movsbl>(env, b, i + 1)) {
     return if_inst<Vinstr::movsbl>(env, b, i + 1, [&] (const movsbl& mov) {
       // loadb{s, tmp}; movsbl{tmp, d}; -> loadsbl{s, d};
       if (!(env.use_counts[inst.d] == 1 &&
--- a/hphp/runtime/vm/jit/vasm-simplify-arm.cpp
+++ b/hphp/runtime/vm/jit/vasm-simplify-arm.cpp
@@ -84,17 +84,38 @@ bool simplify(Env& env, const cmovq& inst, Vlabel b, size_t i) {
 
 ///////////////////////////////////////////////////////////////////////////////
 
-bool simplify(Env& env, const loadb& inst, Vlabel b, size_t i) {
-  return if_inst<Vinstr::movzbl>(env, b, i + 1, [&] (const movzbl& mov) {
-    // loadb{s, tmp}; movzbl{tmp, d}; -> loadzbl{s, d};
-    if (!(env.use_counts[inst.d] == 1 &&
-          inst.d == mov.s)) return false;
+template<Vinstr::Opcode op>
+bool is_next(const Env& env, Vlabel b, size_t i)
+{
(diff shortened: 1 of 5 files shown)


jim-saxman:
LGTM. I measured approximately 75,000 fewer sxtb instructions during OSS/Mediawiki with this change request.

hphp/runtime/vm/jit/vasm-simplify-arm.cpp
```diff
@@ -84,17 +84,38 @@ bool simplify(Env& env, const cmovq& inst, Vlabel b, size_t i) {
 
 ///////////////////////////////////////////////////////////////////////////////
 
-bool simplify(Env& env, const loadb& inst, Vlabel b, size_t i) {
-  return if_inst<Vinstr::movzbl>(env, b, i + 1, [&] (const movzbl& mov) {
-    // loadb{s, tmp}; movzbl{tmp, d}; -> loadzbl{s, d};
-    if (!(env.use_counts[inst.d] == 1 &&
-          inst.d == mov.s)) return false;
+template<Vinstr::Opcode op>
```
ottoni: I don't see a reason for op being a template parameter instead of a function parameter. But I guess this is following the same approach used by if_inst (although I don't see why use a template there either), so I'm ok.

hphp/runtime/vm/jit/vasm-simplify-arm.cpp
```diff
@@ -84,17 +84,38 @@ bool simplify(Env& env, const cmovq& inst, Vlabel b, size_t i) {
...
+bool is_next(const Env& env, Vlabel b, size_t i)
```
ottoni: Can you rename this to is_inst?
This function is testing the same instruction i that it takes, not the next one, so is_next is misleading.

ottoni:
What happened that so many files changed here now? Was it a bad rebase?

swalk-cavium:
@ottoni - rebased against Tot so I could build and lightly test.  No sure why so much is now reported.

swalk-cavium:
@ottoni - looks like I did a pull instead of a rebase.  Fixing now.

hphp/runtime/vm/jit/vasm-simplify-arm.cpp
```diff
@@ -84,17 +84,38 @@ bool simplify(Env& env, const cmovq& inst, Vlabel b, size_t i) {
...
+  return code[i].op == op;
+}
 
-    return simplify_impl(env, b, i, [&] (Vout& v) {
-      v << loadzbl{inst.s, mov.d};
-      return 2;
-    });
-  });
+bool simplify(Env& env, const loadb& inst, Vlabel b, size_t i) {
+  if (is_next<Vinstr::movzbl>(env, b, i + 1)) {
```
markw65: Why add the new helper? Why not just write

  if (if_inst<...>(
    ...
  ) {
    return true;
  }
  return if_inst<...>(...);

hphp/runtime/vm/jit/vasm-simplify-arm.cpp
```diff
@@ -84,17 +84,38 @@ bool simplify(Env& env, const cmovq& inst, Vlabel b, size_t i) {
...
-bool simplify(Env& env, const loadb& inst, Vlabel b, size_t i) {
-  return if_inst<Vinstr::movzbl>(env, b, i + 1, [&] (const movzbl& mov) {
-    // loadb{s, tmp}; movzbl{tmp, d}; -> loadzbl{s, d};
-    if (!(env.use_counts[inst.d] == 1 &&
-          inst.d == mov.s)) return false;
+template<Vinstr::Opcode op>
+bool is_next(const Env& env, Vlabel b, size_t i)
+{
+  auto const& code = env.unit.blocks[b].code;
+  if (i > code.size()) return false;
```
markw65: Should be >=

But I think you should just drop it, and use if_inst instead.

markw65:
I'll also note that we have optimizations for cmpq (in vasm-simplify.cpp) that would get rid of the sign extension in your example (and most compares) altogether (eg there's never any need to sign extend if you're comparing for equality against zero). Not sure why those weren't extended to cover all compares - I think I must have forogotten; it looks pretty straightforward.

Its orthogonal to this optimization, but it struck me that your example isn't the best one for this particular optimization...

swalk-cavium:
Hello @markw65,
The pattern I saw above was seen in the mediawiki trace from oss-performance.  I agree that
there's cases comparing against eq/ne to zero don't need the sign extension.  This seemed the
smallest change that was likely to be accepted.

ottoni:
@swalk-cavium, we provide all the feedback here, but sometimes people just see the change once it's imported (as in this case). That's not a problem: you should just keep updating the code as requested, and we'll re-import it. 
Thanks for checking.

markw65:
As far as not wanting to update the diff and invalidate testing goes... you're doing out of bounds vector accesses. All the testing *is* invalid.

swalk-cavium:
Is there a clang tidy specification somewhere?

PR #8321, opened 2018-09-15
Modify `is_null` HHI signature to accept Disposable

acrylic-origami:
A function with nullable Disposable argument[s] can't use them because
`is_null` trips Typing[4188].


acrylic-origami:
Wait, not so simple. Actually following through with the type refinement results in `Disposable objects must only be assigned in a 'using' statement (Typing[4179])`.

MVE:

```hack
<?hh // strict
class A implements \IDisposable {
	public function __dispose(): void {}
	public function foo(): void { echo __FUNCTION__; }
}
function foo(<<__AcceptDisposable>>?A $a): void {
	if(!\is_null($a)) {
		$a->foo();
	}
}
```

Issue #8314, opened 2018-09-04
No typechecker or runtime error when trying to append to a const collection


PranayAgarwal:
### HHVM Version

```
HipHop VM 3.27.2 (rel)
hh-e6d25982534cd57e1e7fa1e990dcb876c981b957-3.27.2 Wed Aug 15 18:09:53 2018
```

### Operating System and Version

macOS High Sierra

### Standalone code, or other way to reproduce the problem

```
const arr = ['a'];
arr[] = 'b';
var_dump(arr);
```

### Actual result

Output:
```
array(1) {
  [0]=>
  string(1) "a"
}
```

with no errors. Same behavior when called from inside a function.

### Expected result

Should raise a typechecker error and a runtime fatal.

fredemmott:
Also repros on 3.28

PranayAgarwal:
`dict` as well

PR #8328, opened 2018-09-25
RFC: Check lvalue of Array_get in ast_check

fangyi-zhou:
Summary: See #8314 for context
In the Array_get lvalue, we only checked whether the array being modified
is a class const member and raised a parse error, so we were able to
reject code like this

```
class Foo {
    const varray<int> DATA = varray[42];
    public function foo(): void {
        Foo::DATA[] = -1;
    }
}
```

In order for an Array_get lvalue to be valid, the array must be a valid
lvalue. This commit addresses this issue by adding this requirement.

For reference, the following code raises runtime fatal in PHP
(https://3v4l.org/bWGMO)

```
<?php

const FOO = array(42);
FOO[] = -1;
```

but was accepted by HackC (tested locally)

Side Note: We allow return values from functions to be a valid lvalue
for the moment. There probably needs to be checks by the HHVM runtime
elsewhere in addition to this parser change. This is necessary because
Hack Collections don't have value semantics and we can construct code
like to exploit this.

Luckily PHP doesn't accept lvalue that is a conditional expression -
this is possible in C:

Valid C Code:
```
int foo(bool cond) {
    int a = 0 , b = 0;
    (cond ? a : b) = 42;
    return a;
}
```

Whereas its PHP equivalent is parse error: https://3v4l.org/Yk0Ac
```
function foo($cond) {
    ($cond ? $a : $b) = 42;
    return $a;
}
```


fangyi-zhou:
Could I get someone from typechecker team to look at it?

fangyi-zhou:
#m pinguping

hphp/system/php/spl/datastructures/SplObjectStorage.php
```diff
@@ -364,7 +364,8 @@ public function unserialize($serialized) {
    * @return     mixed   No value is returned.
    */
   public function setInfo($data) {
-    current(&$this->__storage)['inf'] = $data;;
+    $storage = current(&$this->__storage);
```
fredemmott: Was this already broken, or is there a way to make this syntax work in PHP Code?

fangyi-zhou: I reenabled the syntax in the second commit but forgot to revert this change. Will revert this change.

fangyi-zhou:
Rebased and reverted the change in systemlib

fangyi-zhou:
#m pingping

fredemmott:
This is a smell, but:

```Hack
class :foo {
  attribute Map<string, string> bar;

  public function doStuff(): void {
    $this->:bar['baz'] = 'herpderp'; // incorrectly raises lvalue error
  }
}
```

PR #8329, opened 2018-09-26
aarch64 - sxtb/lsl peephole

swalk-cavium:
This adds a peephole which captures a common pattern seen in Mediawiki traces.

Before
=====
```
0xcb00a04  93401c21     sxtb x1, w1
0xcb00a08  d37ef421     lsl x1, x1, #2
```
After
====
```
0xcb00a04  937e1c21     sbfiz  x1, x1, #2,  #8
```

The standard regression tests were run with 6 option sets.  No new issues were seen.
The mediawiki oss-performance test was run without incident.


swalk-cavium:
@mxw - Can you inspect?

alshamiri5 (approved):

PR #8335, opened 2018-10-02
zlib.inflate stream filter should respect window param

### LINKED ISSUE #8334: HHVM does not respect window param on zlib.inflate stream filter
This is PHP5/7 incompatibility bug. 
`stream_filter_append` [accepts 3rd argument](http://php.net/manual/en/filters.compression.php) (array)params which is in case of inflate filter, [allows to pass windowBits to](https://github.com/php/php-src/blob/5177db127dee6ef7291083c31bd815e6329c23c9/ext/zlib/zlib_filter.c#L318-L331) zlib's `inflateInit2` and in turn make zlib capable of detecting gzip(RFC1952) or zlib(RFC1950) header in the stream.
However, HHVM has this [hardcoded](https://github.com/facebook/hhvm/blob/master/hphp/runtime/ext/zlib/ext_zlib.cpp#L445) to `-MAX_WBITS` which forces zlib to treat stream as raw deflate format (no header).

### HHVM Version
HipHop VM 3.29.0-dev (rel)
### Operating System and Version
Ubuntu 16.04 using Docker.
Travis-ci as well.

```php
$content = gzencode('rfc1952');
$fh = fopen('php://memory', 'r+b');
fwrite($fh, $content);
rewind($fh);

stream_filter_append($fh, 'zlib.inflate', STREAM_FILTER_READ, ['window' => 15 + 32]);

var_dump(fread($fh, 1024));
fclose($fh);
```
### Actual result
```
Warning: Failed to extract chunk
string(0) ""
```
### Expected result
```
string(7) "rfc1952"
```

kkopachev:
Fixes #8334

### DIFF
--- a/hphp/runtime/ext/zlib/ext_zlib.cpp
+++ b/hphp/runtime/ext/zlib/ext_zlib.cpp
@@ -442,15 +442,22 @@ struct ChunkedInflator {
   ChunkedInflator(): m_eof(false) {
     m_zstream.zalloc = (alloc_func) Z_NULL;
     m_zstream.zfree = (free_func) Z_NULL;
-    int status = inflateInit2(&m_zstream, -MAX_WBITS);
-    if (status != Z_OK) {
-      raise_error("Failed to init zlib: %d", status);
-    }
   }
 
   ~ChunkedInflator() {
     if (!eof()) {
       inflateEnd(&m_zstream);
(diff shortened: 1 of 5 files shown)


cerber717:
👍

cerber717:
😿

PR #8338, opened 2018-10-05
Rainbow of Locks to Improve NonRepoAuth Unit Cache Performance Under High Contention

enjmusic:
## Problem Statement
While deploying code in NonRepoAuth mode, we noticed that requests were often timing out and causing 503 errors. These spikes subsided quickly after deploys and seemed to often be worst after a full restart of hhvm.

## Observations
We noticed that the data structure used for caching code units in NonRepoAuth mode, a subclass of `tbb::concurrent_hash_map`, caused a large quantity of CPU time to be used spinning on the `spin_rw_mutex` in a bucket containing some bytecode during a deploy. During this time, the following conditions must have been true:
- Contention was high on these locks, which is the worst case scenario for their performance
- These locks were being held by writers
- Due to the aggressive spinning in the implementation of the `spin_rw_mutex`, most CPU time was taken up by reader threads performing no work, further exacerbating the issue by allowing less resources to the writer thread (the increased amount of context switching could be observed in `perf top`, which showed as much as 45% CPU overhead for `finish_task_switch`.

## Changes
This PR introduces an additional layer of less performance-degrading locks on top of all `find/insert/erase` operations in the CHM using a new subclass, `RankedBucketCHM`. This modified CHM has the same accessors as `RankedCHM` but also incorporates a rainbow of `SmallLock`s (the # of which is specified as a template parameter) which essentially eliminate the problem of `spin_rw_mutex` contention. These locks are hashed to using the same hashing function as the CHM, and are only acquired for the duration of the `find/insert/erase` operations.

## Risks/Safety
The risks of this approach are low, since acquiring and releasing a lock at the beginning and the end of `find/insert/erase` method calls does not change the overall functionality of the `NonRepoUnitCache`, only moving most of the thread contention to an earlier, less CPU-intensive lock. 

Since only one thread can hold any particular lock in the rainbow at a time, there will be a maximum of one additional request thread attempting to acquire the `spin_rw_mutex` inside the CHM bucket, which will only spin in the case that a reader acquired the outer lock previously. This ensures that accessing cached units is still thread-safe, and also ensures that when all accessors are readers, the outer `SmallLock` is released quickly, preventing too extreme of a bottleneck.

## Performance
Using ApacheBench with a concurrency of 120 threads, the performance of the current build of `hhvm` and a modified build using this change were measured both during a simulated deploy, and a couple minutes after the simulated deploy, once `hhvm` was warmed up and in a more steady state (all trials performed in NonRepoAuth mode).

In the trials during a simulated deploy, the build of `hhvm` with this change did not time out, while the current build of `hhvm` did in the first couple thousand cumulative requests. This change demonstrated a vastly reduced amount of context switching, with `finish_task_switch` at less than 1% CPU overhead vs. the aforementioned 45% overhead of the current build.

In the steady-state trials, the ApacheBench results were as follows:
(for `ab -n 10000 -c 120 ...`, connecting to a commonly used test API endpoint on localhost)

### Current Build
```
Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    1   2.3      0      33
Processing:    34  429 521.9    362   10210
Waiting:       34  405 509.0    349   10207
Total:         34  430 521.9    363   10210

Percentage of the requests served within a certain time (ms)
  50%    363
  66%    457
  75%    516
  80%    557
  90%    708
  95%    897
  98%   1226
  99%   1444
 100%  10210 (longest request)
```

### Modified Build With Proposed Changes
```
Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    1   2.9      0      28
Processing:    52  435 296.0    363    3252
Waiting:       49  404 264.1    343    3252
Total:         53  436 296.3    364    3252

Percentage of the requests served within a certain time (ms)
  50%    364
  66%    464
  75%    543
  80%    600
  90%    812
  95%   1026
  98%   1262
  99%   1461
 100%   3252 (longest request)
```

With less than 1ms difference for the median response time and a consistently lower max response time, this change appears to be reasonable from a performance standpoint. That said, some future work on this matter could be to tune the size of the rainbow of locks.

### DIFF
--- a/hphp/test/zend/good/ext/xsl/tests/xslt010.php.expect
+++ b/hphp/test/zend/good/ext/xsl/tests/xslt010.php.expect
@@ -74,18 +74,18 @@ Test 10: EXSLT Support
     second-in-minute     : NaN
   
   Test Date : -0004-02-29
-    year                 : -4
-    leap-year            : true
-    month-in-year        : 2
-    month-name           : February
-    month-abbreviation   : Feb
-    week-in-year         : 9
-    day-in-year          : 60
-    day-in-month         : 29
-    day-of-week-in-month : 5
-    day-in-week          : 1
-    day-name             : Sunday
-    day-abbreviation     : Sun
+    year                 : NaN
+    leap-year            : NaN
+    month-in-year        : NaN
+    month-name           : 
+    month-abbreviation   : 
+    week-in-year         : NaN
+    day-in-year          : NaN
+    day-in-month         : NaN
+    day-of-week-in-month : NaN
+    day-in-week          : NaN
+    day-name             : 
+    day-abbreviation     : 
     time                 : 
     hour-in-day          : NaN
     minute-in-hour       : NaN


hphp/util/mutex.h
```diff
@@ -20,12 +20,14 @@
 #include <pthread.h>
 #include <time.h>
 #include <thread>
+#include <mutex> // For std::unique_lock
```
kmaslack: Local style seems to avoid comments about why we're including system headers.

hphp/util/mutex.h
```diff
@@ -326,6 +328,104 @@ struct RankedCHM : tbb::concurrent_hash_map<K, V, H> {
...
+template<typename K, typename V, typename H=K, Rank R=RankUnranked, uint32_t N=DEFAULT_CHM_LOCK_BUCKET_COUNT>
+struct RankedBucketCHM : tbb::concurrent_hash_map<K, V, H> {
+private:
+  typedef tbb::concurrent_hash_map<K, V, H> RawCHM;
+
+  /*
+   * InternalSmallLock is a union to avoid issues with default initialization of an
+   * array or container of SmallLocks - base_int is never actually used, only lock
+   */
+  typedef union { uint32_t base_int; mutable SmallLock lock; } InternalSmallLock;
```
kmaslack: Is it worth a `static_assert(sizeof(unit32_t) == sizeof(SmallLock))`?

hphp/util/mutex.h
```diff
@@ -326,6 +328,104 @@ struct RankedCHM : tbb::concurrent_hash_map<K, V, H> {
...
+    bool freed;
+  };
+
+  /*
+   * All find/insert/erase operations hold an outer lock, which prevents
+   * many threads from spinning aggressively on the spin_rw_mutex inside
+   * of tbb::concurrent_hash_map
+   */
+  bool find(const_accessor& a, const K& k) const {
+    std::unique_lock<SmallLock> outerLock(locks[outerLockIndex(k)].lock);
```
kmaslack: I'm no longer sure I understand why this works :(.

I was imagining the the outer critical section would coincide with the lifetime of the accessor. With this scheme, it seems like it should be possible (?) for a caller to `find(a, k)`, get an `a` that points to nothing and is a lock, and then have any number of readers also `find(a, k)` and burn CPU in the body of `tbb::chm::find()` waiting for the writer to destroy its accessor.

What am I missing?

hphp/util/mutex.h
enjmusic: Sounds like a good way to ensure at compile time that this union always works!

markw65:
We just fixed the equivalent issue for repo mode. I think I have a better solution that will work for both.

hphp/util/mutex.h
kmaslack: OK. While I see why we're seeing the benefit now, I feel like it would be easier to reason about, more robust, and possibly free up that one extra core at T2 in Jamie's writeup, to scope the SmallLock critical section to the accessor lifetime (or, accessor acquire/release() span anyway). Does that seem reasonable?

enjmusic:
@markw65 What does your solution involve? From what you said it sounds like it will also be replacing the fix for repo mode; is that correct?

enjmusic:
Also, newest commit to make the outer lock lifetime the same as the lifetime of the accessor seems to have improved average performance but increased the p95-and-up times.

### Unmodified `hhvm` at steady state:
```Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.8      0       8
Processing:    92  332  92.3    321    1551
Waiting:       91  329  90.5    318    1550
Total:         92  333  92.3    321    1551

Percentage of the requests served within a certain time (ms)
  50%    321
  66%    347
  75%    367
  80%    381
  90%    426
  95%    479
  98%    555
  99%    608
 100%   1551 (longest request)
```

### new version of this patch, steady state:
```Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    1   3.0      0      40
Processing:    28  329 685.4    206   16038
Waiting:       28  300 667.0    198   16037
Total:         29  330 685.5    207   16038

Percentage of the requests served within a certain time (ms)
  50%    207
  66%    279
  75%    348
  80%    406
  90%    619
  95%    829
  98%   1176
  99%   1501
 100%  16038 (longest request)
```

hphp/util/mutex.h
enjmusic: A vector will work but sadly not a vector of raw `SmallLocks` because their default constructor has been deleted. The constructor argument is a good point, and I'll be switching to that pattern.

hphp/util/mutex.h
```diff
@@ -326,6 +328,81 @@ struct RankedCHM : tbb::concurrent_hash_map<K, V, H> {
...
+   * InternalSmallLock is a union to avoid issues with default initialization of an
+   * array or container of SmallLocks - base_int is never actually used, only lock
+   */
+  typedef union { uint32_t base_int; mutable SmallLock lock; } InternalSmallLock;
+
+  const uint32_t nLocks = N;
+  mutable InternalSmallLock locks[N];
+public:
+  SmallLock* getLockForPath(const K& path) {
+    size_t lockIndex = (RawCHM::my_hash_compare.hash(path) % nLocks);
```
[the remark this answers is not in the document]
enjmusic: `auto` instead of `size_t`?

hphp/util/mutex.h
```diff
@@ -326,6 +328,81 @@ struct RankedCHM : tbb::concurrent_hash_map<K, V, H> {
...
+    return &(locks[lockIndex].lock);
+  }
+
+  /*
+   * Same accessors as RankedCHM
+   */
+  struct accessor : RawCHM::accessor {
+    accessor(SmallLock* lock) : freed(false) {
+      pushRank(R);
+      outerLock = std::unique_lock<SmallLock>(*lock);
```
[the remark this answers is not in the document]
enjmusic: That would probably be wise. That said, `pushRank` doesn't seem to do anything synchro-related on its own, and `SmallLock` doesn't use `Rank`.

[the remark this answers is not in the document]
ajmacd: If there's no reason _not_ to lock it first, I would just do that, following the principle of least surprise.

hphp/util/mutex.h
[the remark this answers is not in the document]
ajmacd: Right, I prefer to use `auto` wherever possible. And really this should be `const auto` but now we're splitting some pretty fine hairs.

hphp/util/mutex.h
ajmacd: I had a look at `smalllocks.h`. I don't think there's any reason to rely on that. You don't have a particular size concern with only 137 objects (in the default case). Just use an array of `std::mutex`s.

PR #8362, opened 2018-10-19
Update hphp.7.tab.cpp

dainmiller:
Use `std::equal` due to other uses of `std::` in file & functional instantiation


fredemmott:
FYI
- we need the CLI signing before we can look at this
- based on the title: this is practically dead code, and we're hoping to delete it shortly; all lexing, parsing, and everything else up through bytecode generation is now handled by ocaml code

fredemmott:
Sorry for not getting to this sooner; the code is dead and this PR is no longer applicable.

Issue #8386, opened 2018-11-19
build fails due to missing lz4


florinpapa:
Building HHVM fails due to missing lz4 on the system, although lz4 is present in third-party and should be taken from there if not present on the system. Changing the build system to dune in commit b99195e4498dd316728773ebf937e4879a76f641 seems to be the cause.

### HHVM Version

Master branch, since this commit on 11/13: b99195e4498dd316728773ebf937e4879a76f641

### Operating System and Version

Ubuntu 16.04, bare metal.

### Standalone code, or other way to reproduce the problem

```
./configure
make -j `nproc`
(cd _build/default/heap && /home/slo/git/hhvm-perf/workspace-2018-11-16_08.45.23/hhvm/third-party/ocaml/build/bin/ocamlc.opt -g -ccopt -g -o hh_shared.o hh_shared.c)
hh_shared.c:110:10: fatal error: lz4.h: No such file or directory
 #include <lz4.h>
          ^~~~~~~
compilation terminated.
Makefile:18: recipe for target 'build-hack' failed
make[3]: *** [build-hack] Error 1
hphp/hack/CMakeFiles/hack.dir/build.make:57: recipe for target 'hphp/hack/CMakeFiles/hack' failed
make[2]: *** [hphp/hack/CMakeFiles/hack] Error 2
CMakeFiles/Makefile2:1483: recipe for target 'hphp/hack/CMakeFiles/hack.dir/all' failed
make[1]: *** [hphp/hack/CMakeFiles/hack.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

swalk-cavium:
@florinpapa - try adding liblz4-dev

sudo apt-get install -y liblz4-dev

It looks like the packaging info hasn't been updated yet.

florinpapa:
@swalk-cavium - thank you for your suggestion, the build works when liblz4-dev is installed.

If the liblz4-dev package is now a system requirement, shouldn't lz4 be removed from third-party, as it is no longer used?

swalk-cavium:
@florinpapa - you'll have to ask FB.

fredemmott:
KP, we're working on it - sorry for the delay

cc @Wilfred, @CatherineGasnier

PR #8390, opened 2018-11-28
link to built lz4 and sqlite3

fredemmott:
Summary:
This should fix the nightly build.

Make dune build with the libraries built elsewhere with CMake if there is no system version or if the system version is not new enough.

Fixes facebook/hhvm#8386

Differential Revision: D13234118


fredemmott:
Broken.

PR #8394, opened 2018-12-04
Add new native functions for accessing GET and POST variables

fredemmott:
Summary:
The current interfaces all lose data in some situation. In particular:
- the handling of `.` and `[]` is a phpism
- HTTP allows multiple K-V pairs with identical K
- if I specify `foo=bar` and `foo[]=baz` what should the value be?

Resolve these by keeping values verbatim, and returning
`dict<string,vec<string>>`. This is already the format used by HPHP::Transport.

Reviewed By: kmeht

Differential Revision: D13312805


fredemmott:
Not doing this.

fredemmott:
To recap all the past problems (as I'm sharing a link here to a few other people):
- http allows multiple k-v pairs with the same k for $_POST and $_GET
- php has magic for `foo.bar` (`dict['foo' => dict['bar' => ...]]`) and `foo[]` (`dict['foo'] => vec[ ... ])
- the right thing to do for a replacmenet for $_POST is not obvious when multipart/form-data is considered: it's probably surprising that a replacement for $_POST['foo'] potentially reading in a multi-gigabyte file upload

Proposed solution:
- dict<string, vec<string>> is probably the best representation for these k-v paris
- HHVM does as little as possible and leave it up to frameworks.
- in pure-php terms, this means frameworks use php://input, $_SERVER['QUERY_STRING'], and $_SERVER['CONTENT_TYPE']
- HHVM only exposes strings and file handles, no dicts. Frameworks decide what to do about huge POST data
- we need replacements for those php implementation details at some point

Issue #8396, opened 2018-12-05
fix call stub clobbers introduces approximately 100 regression test failures on aarch64


swalk-cavium:
### HHVM Version

swalk@lab-3:/data2/swalk/play.102418c/hhvm$ hphp/hhvm/hhvm --version
HipHop VM 3.29.0-dev (rel)
Compiler: tags/gcc-4.6-22496-g98c4f446b683756cf7f98aef8fa9ac97d791b4c1
Repo schema: a8cefb59c6b78f29f979d79ac2188365c07529d7

### Operating System and Version

swalk@lab-3:/data2/swalk/play.102418c/hhvm$ uname -a
Linux lab-3 4.2.0 #1 SMP Wed Nov 25 14:00:54 CST 2015 aarch64 aarch64 aarch64 GNU/Linux

### Standalone code, or other way to reproduce the problem

Standard regression tests run with 6 different option sets, then post processed.
Columns are:
  hphp/test/run -v -m interp all
  hphp/test/run -v -m interp -r all
  hphp/test/run -v -m jit -a '-vEval.JitPGO=0' all
  hphp/test/run -v -m jit -r -a '-vEval.JitPGO=0' all
  hphp/test/run -v -m jit -a '-vEval.JitPGO=1' all
  hphp/test/run -v -m jit -r -a '-vEval.JitPGO=1' all
  
### Actual result

Here's a sample of the output.

hphp/test/quick/closure_class.php | . | . | F | F | F | F | 
hphp/test/quick/closure_clone_2.php | . | . | F | F | F | F | 
hphp/test/quick/closure_gen.php | . | . | F | F | F | F | 
hphp/test/quick/closure_gen_scope.php | . | . | F | F | F | F | 
hphp/test/quick/closure_get_class.php | . | . | F | F | F | F | 
hphp/test/quick/closure_in_generator.php | . | . | F | F | F | F | 
hphp/test/quick/closure_in_generator2.php | . | . | F | F | F | F | 
hphp/test/quick/closure_lifetime.php | . | . | F | F | F | F | 
hphp/test/quick/closure_private.php | . | . | F | F | F | F | 
hphp/test/quick/closure_recursive.php | . | . | F | F | F | F | 
hphp/test/quick/closure_static.php | . | . | F | F | F | F | 
hphp/test/quick/closure_static_vars.php | . | . | F | F | F | F | 
hphp/test/quick/hh_async_closure_return_type_0.php | . | . | F | F | F | F | 
[...]

PR #8401, opened 2018-12-12
revise callstubs defs for aarch64 target

swalk-cavium:
This commit https://github.com/facebook/hhvm/commit/98c4f446b683756cf7f98aef8fa9ac97d791b4c1
changed the registers presented to the allocator for the callstub Vinstr.  The comment says
it did not change the register sets for aarch64.  This is not correct.  This change caused an
internal assert to fire when it was trying to create spill space on the stack.  This sounds like
the problem seen on other hosts.

The standard regression tests were run with 6 option sets.  This brings the results back to
the normal baseline for the lab machine I use.

This fixes issue https://github.com/facebook/hhvm/issues/8396

### DIFF
--- a/hphp/runtime/vm/jit/abi-arm.cpp
+++ b/hphp/runtime/vm/jit/abi-arm.cpp
@@ -92,10 +92,16 @@ const RegSet kSF = RegSet(RegSF{0});
 ///////////////////////////////////////////////////////////////////////////////
 
 /*
+ * Registers that can safely be used within a prologue.
+ */
+const RegSet kPrologueRegs =
+  kSIMDCallerSaved | (kGPUnreserved - arm::vm_regs_with_sp());
+
+/*
  * Registers that can safely be used for scratch purposes in-between traces.
  */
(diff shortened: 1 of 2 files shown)


swalk-cavium:
@mxw , @ricklavoie - Can you review?

ricklavoie:
This isn't the right fix. This change was made originally because the defs set was incorrect and a new register allocator we're working on exposed this bug. This pull request just undoes the fix for ARM, which means it will trigger that bug. The right fix is to do the same thing we did for x86 and remove registers (one is probably all thats required) from the kScratchCrossTraceRegs to reduce the register pressure.

swalk-cavium:
Hi @ricklavoie,
I modeled the change after the x64 version and reduced the kScratchCrossTraceRegs
until the regression tests went back to baseline.

The regression tests were run with 6 option sets. The internal assert is gone.

swalk-cavium:
Is there anything else I need to do for this one?

PR #8402, opened 2018-12-13
avoid specific length tests

swalk-cavium:
Test case introduced within the last day is failing all option sets tested.
Per conversation on the slack channel - HackLang - build-and-dev, the
length fields are fudged a little.


swalk-cavium:
This test is still failing as of a build on 01/28/2019.

PR #8406, opened 2018-12-20
[HHI] add missing closure methods interface

azjezz:
HHVM supports `Closure::bind` and `Closure::bindTo`.

this fixes an issue with the type checker.


fredemmott:
We intentionally do not allow this in Hack code; changing the type - and scope kind - of a closure basically breaks type safety.

Wilfred:
The problem here is that `$this` can be anything if you allow users to bind a closure to different objects. This is a deliberate omission I'm afraid.

PR #8410, opened 2018-12-28
Update block_statement.h

mattnappo:


dowlandaiello:
Think you might need to add a few lines...

PR #8427, opened 2019-02-01
Fix crash in tc-print when using '-e list' option.

dave-estes-ampere:
Crash was due the the use of the g_logger before it was initialized.
The '-B list' option was broken in the same way.


dave-estes-ampere:
The new (forced) commit just edits the author email address and a typo in the commit.

PR #8436, opened 2019-02-13
Consider a DebugOpt and Debug build equivalent in terms of the -DNDEBUG.

dave-estes-ampere:


dave-estes-ampere:
Ping @fredemmott, mind taking a look at this? We use DebugOpt builds a fair amount, and this is a convenient change for additionally getting TRACE output.

PR #8439, opened 2019-02-13
Fix issue where static locals in closures are dropped to BoxedInitCell.

dave-estes-ampere:
FrameStateMgr drops references of TBoxedCell to TBoxedInitCell for locals.
Unfortunately, static locals in closures are managed the same way (see
"Static locals" commend in rds-util.h). The result is that
dropLocalsRefsInnerTypes() will incorrectly drop the type, which
triggers an assert for type mismatch in check_invariants() because
the type has been initialized with the initial execution of the
closure.

This issue was found when running hhvm as a server in a method similar
to the OSS performance suite and then using siege with a single URL
running the unit test: quick/static_closure_gen.php.

### DIFF
--- a/hphp/runtime/vm/jit/frame-state.cpp
+++ b/hphp/runtime/vm/jit/frame-state.cpp
@@ -774,7 +774,8 @@ void FrameStateMgr::update(const IRInstruction* inst) {
     }
 
     assertx(!extra.smashesAllLocals || extra.nChangedLocals == 0);
-    if (extra.smashesAllLocals || inst->marker().func()->isPseudoMain()) {
+    if (extra.smashesAllLocals || inst->marker().func()->isPseudoMain() ||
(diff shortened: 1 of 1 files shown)


dave-estes-ampere:
Hey @ottoni , mind taking a look at this PR? It's an issue that we uncovered via stress testing using a modified OSS perf suite with a target consisting of a large chunk of HHVM unit tests. It's been uncovering a few items, but this is the only fix we have ready to push. Thx.

fredemmott:
@dave-estes-ampere I'm pretty sure that this is no longer needed given that HHVM no longer supports static locals, can you confirm?

dave-estes-ampere:
Thanks for following up, @fredemmott. You're right. I tried to reproduce this on the tip and the issue is no longer there. Will close. Thanks again.

PR #8447, opened 2019-02-20
Update opam to opam 2.0.2

vsiles:
Depends on https://github.com/vsiles/hhvm-third-party/commit/0920692428dc5ab245f07d11d7fda30252303d33

In order to support opam 2, a few modifications are necessary:
- version bump for dune since the dune file format has changed
- update of the opam commands
- not mandatory: bump of pcre version, because I could :D

Test plan: must not affect the build at all. Existing commands must
complete as usual


vsiles:
Depends on https://github.com/facebook/hhvm/pull/8447

Wilfred:
@vsiles I can't actually see a PCRE upgrade, neither in this PR nor in https://github.com/hhvm/hhvm-third-party/pull/138 . What changed for the new PCRE version?

vsiles:
https://github.com/facebook/hhvm/pull/8447/commits/19632d8c813837f213cfb94777f8f655529ab6a6#diff-2c4e4781e542e91acf32f097217dd05dR29 I bumped the version number of the ocaml mappings in opam_setup.sh

fredemmott:
I'm a little nervous about that - the .84 bump has led to segfaults in HHVM itself

vsiles:
i'll try to revert PCRE and see if it's compatible with 2.0

vsiles:
With the old pcre version I have the following compilation error:
```
File "_none_", line 1:
Error: Files src/typing/typing_regex.cmx
       and /home/vsiles/hhvm/hphp/hack/_build/.opam/default/lib/pcre/pcre.cmxa
       make inconsistent assumptions over interface Pcre
Command exited with code 2.
Hint: Recursive traversal of subdirectories was not enabled for this build,
  as the working directory does not look like an ocamlbuild project (no
  '_tags' or 'myocamlbuild.ml' file). If you have modules in subdirectories,
  you should add the option "-r" or create an empty '_tags' file.
  
  To enable recursive traversal for some subdirectories only, you can use the
  following '_tags' file:
  
      true: -traverse
      <dir1> or <dir2>: traverse
      
Compilation unsuccessful after building 1710 targets (1709 cached) in 00:00:36.
Makefile.ocamlbuild:200: recipe for target '_build/hack.stamp' failed
```

Everything is ok with 7.4.0. Should I keep the updated version and only update the third-party revision ?

vsiles:
I confirm that I don't manage to build with pcre 7.3.4, but it's ok with 7.4.0

vsiles:
`make clean` was not enough to test the version change. After cleaning in depth everything, I managed to build without the pcre version modification. I backtraced it.
@fredemmott I'm not used to github PR, but everything should be as discussed now: subrepo points to the right one and pcre version in restored

vsiles:
Oups that was unintended... I wanted to update my fork to test locally some building scenarios...

PR #8449, opened 2019-02-22
Fix autocomplete, outline, go-to-definition for .hack files that contain <?hh

fredemmott:
For example:
```
function do_some_codegen(): void {
  file_put_contents('foo.txt', '<?hh');
  //                            ^ this is not the start of code, this is
  //                            a string literal. If it weren't a .hack file,
  //                            it would be the start of code
}
```

Right now, how this is parsed entirely depends on the file extension.
Various parts of the IDE support create instances of the parser with
dummy filenames, so "is this a .hack file?" isn't a working approach.

This is not a good fix - but it is a small fix that I /think/ is
compatible with existing usage, and is suitable for committing to a `.z`
release.

This PR:
- makes several IDE entrypoints to the parser effectively set the `force_hh` flag
- changes the lexer to look at a global byref flag for 'ide mode' - if so, we assume it's a hack file,  and `<?` must either be the first line, or the second line, with the first line being a shebang line. If neither's two, we parse it as a .hack file.

This is based on two assumptions:

non-.Hack Hack files are either of this form:

```
<?hh
foo();
```

... or this form...

```
#!/usr/bin/env hhvm
<?hh
```

Definitely not of the form:

```
Foo. please echo this content.
<?hh
```

... which is valid PHP.

The second assumption is that IDE utiltities
are invoked for real PHP files.


fredemmott:
*DO NOT DIRECTLY MERGE THIS PULL REQUEST* - it needs to go the staging repository first.

fredemmott:
:/ this fixes outline everywhere AFAICT, but autocomplete *only* works in files containing '<?hh'

hphp/hack/src/server/serverIdeUtils.ml
```diff
@@ -84,7 +84,7 @@ let declare_and_check_ast ?(path=path) ?content ~make_ast ~f tcopt =
   Errors.do_ @@ make_then_revert_local_changes begin fun () ->
     Fixmes.HH_FIXMES.(remove_batch @@ KeySet.singleton path);
     Fixmes.DECL_HH_FIXMES.(remove_batch @@ KeySet.singleton path);
-    let ast = make_ast () in
+    let ast = ParserOptions.with_ide_mode ~f:(fun () -> make_ast ()) in
```
Wilfred: No big deal for a patch release, but this could be:

```suggestion
    let ast = ParserOptions.with_ide_mode ~f:make_ast in
```

hphp/hack/src/parser/full_fidelity_ast.ml
```diff
@@ -3671,6 +3671,7 @@ let defensive_program
       ~elaborate_namespaces
       ~keep_errors:(keep_errors || (not fail_open))
       ~parser_options
+      ~enable_hh_syntax:true
```
Wilfred: Why has this changed? Can you use `!is_ide_mode` here too?

fredemmott: ah woops. I initially did this by keying on force_hh_syntax instead of the new global flag - but that turned out to be a bad idea given that hh_server is used for non-IDE queries too and we need to preserve the meaning

fredemmott:
Merged

PR #8457, opened 2019-03-05
Allow resetting stat_cache from Hack

jscheinblum:
We'd like to enable the stat_cache, but are worried about any drift of the in-memory view of the stat_cache with what's on disk.

Background on our deploy:
• We rsync files onto the we serving hosts into A or B directory, and flip the serving root back and forth.
• We drain each host on deploy, and then run a quick warm up script
• We do not restart hhvm between deploys
• We do no currently run with the stat_cache

What we'd like:
• We'd like to enable the stat_cache in prod and clear the cache on each deploy without restarting hhvm (operationally, this is complicated, although we do aim to do that this year before switching to repo auth)
• On each deploy, we'll run a script before putting traffic on the server, which would call clearstatcache()

We're looking for feedback on this approach, or any hints of alternate ways to solve this. To be clear, we've not seen any problems with the StatCache, we're just nervous at the risk that it gets out of sync on deploy.

Additionally, we're not quite sure when it's safe to reset the stat_cache, so we defer that until the requestInit function, partially because of the comment that the work in StatCache::requestInit() that must be done before ExecutionContext::requestInit. https://github.com/facebook/hhvm/blob/HHVM-3.30/hphp/runtime/base/program-functions.cpp#L2617

Thanks!

### DIFF
--- a/hphp/runtime/base/stat-cache.cpp
+++ b/hphp/runtime/base/stat-cache.cpp
@@ -420,7 +420,7 @@ StatCache::NodePtr StatCache::Node::getChild(const std::string& childName,
 // StatCache.
 
 StatCache::StatCache()
-  : m_lock(false /*reentrant*/, RankStatCache), m_ifd(-1), m_should_clear(false),
+  : m_lock(false /*reentrant*/, RankStatCache), m_ifd(-1), m_shouldClear(false),
     m_lastRefresh(time(nullptr)) {
 }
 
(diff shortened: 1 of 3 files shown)


fredemmott:
This will need to be landed against master first, then potentially backported. Trying to import this with the merge-base ignored. If that doesn't work, will need this PR to be rebased first.

fredemmott:
Nope, that doesn't work. Needs a rebase for feedback on the actual diff.

For feedback on the approach, the Facebook groups are probably best.

jscheinblum:
Opened a new PR here: https://github.com/facebook/hhvm/pull/8459

PR #8459, opened 2019-03-06
Allow resetting stat_cache from Hack

jscheinblum:
We'd like to enable the stat_cache, but are worried about any drift of the in-memory view of the stat_cache with what's on disk.

Background on our deploy:
• We rsync files onto the we serving hosts into A or B directory, and flip the serving root back and forth.
• We drain each host on deploy, and then run a quick warm up script
• We do not restart hhvm between deploys
• We do no currently run with the stat_cache

What we'd like:
• We'd like to enable the stat_cache in prod and clear the cache on each deploy without restarting hhvm (operationally, this is complicated, although we do aim to do that this year before switching to repo auth)
• On each deploy, we'll run a script before putting traffic on the server, which would call clearstatcache()

We're looking for feedback on this approach, or any hints of alternate ways to solve this. To be clear, we've not seen any problems with the StatCache, we're just nervous at the risk that it gets out of sync on deploy.

Additionally, we're not quite sure when it's safe to reset the stat_cache, so we defer that until the requestInit function, partially because of the comment that the work in StatCache::requestInit() that must be done before ExecutionContext::requestInit. https://github.com/facebook/hhvm/blob/HHVM-3.30/hphp/runtime/base/program-functions.cpp#L2617

Thanks!

### DIFF
--- a/hphp/runtime/base/stat-cache.cpp
+++ b/hphp/runtime/base/stat-cache.cpp
@@ -420,7 +420,7 @@ StatCache::NodePtr StatCache::Node::getChild(const std::string& childName,
 // StatCache.
 
 StatCache::StatCache()
-  : m_lock(false /*reentrant*/, RankStatCache), m_ifd(-1), m_shouldClear(false),
+  : m_lock(false /*reentrant*/, RankStatCache), m_ifd(-1),
     m_lastRefresh(time(nullptr)) {
 }
 
@@ -648,13 +648,6 @@ void StatCache::removeLPath(const std::string& path, Node* node) {
 void StatCache::refresh() {
 #ifdef __linux__
   SimpleLock lock(m_lock);
-
-  // Check if we should reset the cache
-  // as part of this refresh
-  if (m_shouldClear) {
-    m_shouldClear = false;
-    reset();
-  }
 
   if (m_ifd == -1) {
     return;
@@ -949,7 +942,7 @@ void StatCache::clearCache() {
   if (!RuntimeOption::ServerStatCache) return;
 
   SimpleLock lock(s_sc.m_lock);
-  s_sc.m_shouldClear = true;
+  s_sc.reset();
 }
 
 ///////////////////////////////////////////////////////////////////////////////
--- a/hphp/runtime/base/stat-cache.cpp
+++ b/hphp/runtime/base/stat-cache.cpp
@@ -420,7 +420,7 @@ StatCache::NodePtr StatCache::Node::getChild(const std::string& childName,
 // StatCache.
 
 StatCache::StatCache()
-  : m_lock(false /*reentrant*/, RankStatCache), m_ifd(-1),
+  : m_lock(false /*reentrant*/, RankStatCache), m_ifd(-1), m_shouldClear(false),
     m_lastRefresh(time(nullptr)) {
 }
 
@@ -648,6 +648,13 @@ void StatCache::removeLPath(const std::string& path, Node* node) {
 void StatCache::refresh() {
 #ifdef __linux__
   SimpleLock lock(m_lock);
+
+  // Check if we should reset the cache
+  // as part of this refresh
+  if (m_shouldClear) {
+    m_shouldClear = false;
+    reset();
+  }
 
   if (m_ifd == -1) {
     return;
@@ -914,6 +921,7 @@ StatCache StatCache::s_sc;
 
 void StatCache::requestInit() {
   if (!RuntimeOption::ServerStatCache) return;
+  TRACE(5, "StatCache: requestInit refresh");
   s_sc.refresh();
 }
 
@@ -935,6 +943,13 @@ std::string StatCache::readlink(const std::string& path) {
 std::string StatCache::realpath(const char* path) {
   if (!RuntimeOption::ServerStatCache) return realpathLibc(path);
   return s_sc.realpathImpl(path);
+}
+
+void StatCache::clearCache() {
+  if (!RuntimeOption::ServerStatCache) return;
+
+  SimpleLock lock(s_sc.m_lock);
+  s_sc.m_shouldClear = true;
 }
 
 ///////////////////////////////////////////////////////////////////////////////
--- a/hphp/runtime/base/stat-cache.h
+++ b/hphp/runtime/base/stat-cache.h
@@ -129,7 +129,6 @@ struct StatCache {
 
   SimpleMutex m_lock;       // Protects the following fields.
   int m_ifd;
-  bool m_shouldClear;      // True if we should clear the cache on the next request
 #ifdef __linux__
   static const size_t kReadBufSize = 10 * (sizeof(struct inotify_event)
                                            + NAME_MAX + 1);
--- a/hphp/runtime/base/stat-cache.h
+++ b/hphp/runtime/base/stat-cache.h
@@ -78,6 +78,7 @@ struct StatCache {
     int m_wd;                // Watch descriptor; -1 if a file.
 
     bool m_valid;            // True if m_stat/m_lstat are currently valid.
+
     struct stat m_stat;      // Cached stat() result.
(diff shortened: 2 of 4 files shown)


alexeyt:
This is weird.
PHP's clearstatcache has immediate effect, and this version happens the next time _a_ request starts, which is timed non-deterministically relative to this request (but the clearing _may_ affect results seen by this request).
The operation uses reset, so it not only clears the cache but then immediately re-inits it, meaning the kernel/hhvm immediately starts doing work again to keep the cache up to date (whereas other code here does init lazily in order to avoid this).
IIUC the refresh at the start of each request should bring the in-memory view up-to-date, so unless we have bugs this change accomplishes nothing?
"we've not seen any problems with the StatCache, we're just nervous" is not a good way to do software development IMO. If you think there are bugs, you should create a repro that shows the bugs happening, then start changing code.

jscheinblum:
Hey thanks for the feedback! Much appreciated. We are seeing multiple millisecond improvements at p50 with the stat_cache enabled on the canary hosts we enabled the stat_cache on, which is huge! We're quite eager to get it turned on in production, we're just trying to build up confidence that the stat_cache won't skew or that we can detect it if it does. Especially since we deploy often to a large number of hosts, and rarely restart the hhvm process.

> PHP's clearstatcache has immediate effect, and this version happens the next time a request starts, which is timed non-deterministically relative to this request (but the clearing may affect results seen by this request).

Yea, our preference would be to have it reset as soon as `clearstatcache` is called, I was just confused by the comment in the code that `the work in StatCache::requestInit() that must be done before ExecutionContext::requestInit.` If we can reset this inline safely, I will update the PR.

> IIUC the refresh at the start of each request should bring the in-memory view up-to-date, so unless we have bugs this change accomplishes nothing?

That's my understanding from the code read of stat_cache as well. And agreed this is belt-and-suspenders, and the code should work correctly. The risk of the stat_cache being out of sync from the filesystem is pretty high, and best I can tell, it never re-checks to see if it has the updated stat other than based on an inotify event, and no way to check if there's any skew. If that's an incorrect understanding of the code, that would be good to know and we'll adjust.

Would this be better implemented as an admin command to flush the cache?

kmaslack:
Hey, Alexey. I'm also on the Slack team, and reviewed Jamie's diff internally. I agree it's surprising that the flushing occurs on the _next_ request, and Jamie and I had a back and forth about it internally. I encouraged him to see what upstream would say.

But it seems like most of your objection would apply even if the flush were performed eagerly.  RepoAuth is not currently able to run our application, so we are running in sandbox mode while we get RepoAuth-ready, and while upstream figures out why it sometimes branches to negative offsets in the hhbc file. When running like this, we are getting eaten alive by kernel CPU utilization in stat(2) calls. Hence our interest in stat cache.

Reading between the lines of stat cache not being on by default, it feels like upstream doesn't have total confidence in stat cache. I know that I did not have total confidence in it the last time I tried to use it, and my recollection was that the long tail of weird, frustrating bugs that led stat cache to drop an update, and therefore pin a stale file in cache, led Jason Evans to disable it by default. But that was back when Obama was in office.

If all this ancient history is irrelevant, and there is a ton of operational experience with stat cache's perfection, we can drop this change and see what we see trying it in production. Have you all heavily used stat cache? Either in prod or in dev?

fredemmott:
I think there's a few concerns here:

## When it happens

I think the admin server would be better here

## When is this useful?

I think there's two cases here:
- not trusting the inotify mechanism
- not trusting statcache as a whole

I don't think this really helps in the second case - if there's non-inotify problems, you're going to need to restart HHVM to be safe;  I guess you're not doing that as you want to keep it as warm as possible? The other main alternative is to roll it out to a small subset of hosts, watch them closely, and bug report + revert if it fails - if there is such as thing as a small enough subset of hosts for failures to be acceptable while still being statistically significant.

If it's the inotify/invalidation stuff that's most concerning, perhaps it could be better to tie this to an option to completely disable the inotify support? i.e. statcache will definitely be stale while files are being modified, until you flush it. This might also help you more generally - say an update ends up looking like:

1. upload new foo.php which depends on new bar.php
2. upload new bar.php
3. [optional] flush cache

Both without statcache and with just this, a request between 1 & 2 will fail; if it were possible to enable statcache without inotify, a request anywhere between 1 and 3 would get the old version consistently and succeed - no new code will be used until 3 happens

## Is StatCache tested in production?

I can't find references, but I think that Wikipedia ended up using it for a while, and there were a lot of bugfixes a /long/ while back - but it seems that was ~ 5 years ago at least.

jscheinblum:
We've updated the PR to switch to an admin-command `/statcache-clear`, and undid the `clearstatcache` change.

We'll work on disabling the inotify separately.

hphp/runtime/base/stat-cache.cpp
```diff
@@ -649,6 +649,13 @@ void StatCache::refresh() {
 #ifdef __linux__
   SimpleLock lock(m_lock);
 
+  // Check if we should reset the cache
+  // as part of this refresh
+  if (m_shouldClear) {
+    m_shouldClear = false;
+    reset();
```
alexeyt: Are you using reset() here rather than clear() because you don't want to hit the early return on line 660 below? If we are going to do this clearing in a deferred way based on m_shouldClear (which I'm not sure is the best thing to do), then why not have this whole if be below the one on line 659-661? That way if we enter refresh() with m_shouldClear set but m_ifd == -1 we won't needlessly init the stat cache.

jscheinblum: Thanks, that’s the question we have as well- we were unclear if we had to do the full reset before ExecutionContext. If not, we’ll clear it inline. 

> Additionally, we're not quite sure when it's safe to reset the stat_cache, so we defer that until the requestInit function, partially because of the comment that the work in StatCache::requestInit() that must be done before ExecutionContext::requestInit. https://github.com/facebook/hhvm/blob/HHVM-3.30/hphp/runtime/base/program-functions.cpp#L2617

alexeyt: There is only one stat cache that is shared by all requests, so even if you're doing it before the start of ExecutionContext::requestInit in one request, it can easily be in the middle of a request for other requests that are running concurrently.
IIUC the comment in program-functions.cpp is trying to say that we want to process any pending cache invalidations via StatCache::requestInit before we potentially start using the StatCache while initializing the rest of the request (ExecutionContext::requestInit, etc.). IIUC this ordering is needed so that, if you update a file on disk and then issue a request to the webserver, the request will definitely see that file change when it is initializing. If you're going to invalidate the entire cache by calling clear, then AFAICT it doesn't matter when it happens relative to the lifetime of the request that triggers it or other requests that are running in parallel.

jscheinblum: Outstanding. I’ll clean up the pr tonight to simplify and call inline. 

Thanks!

hphp/runtime/base/stat-cache.cpp
```diff
@@ -937,5 +938,12 @@ std::string StatCache::realpath(const char* path) {
   return s_sc.realpathImpl(path);
 }
 
+void StatCache::clearCache() {
+  if (!RuntimeOption::ServerStatCache) return;
+
+  SimpleLock lock(s_sc.m_lock);
+  s_sc.reset();
```
alexeyt: This is still calling .reset(), not .clear()
Would you mind changing it on the PR? If that's inconvenient I can do that locally before landing it.

jscheinblum: Ahh shoot - I can make that change and re-test here. If you think it's safe, feel free to make that change locally and merge it, and I'll do the same in our branch. Otherwise, happy to test with that change here first. From a static code read, I *think* it's fine.

alexeyt: Automated testing takes a long time on my end, so I'm not in a hurry. Please update when you're ready and I'll re-import it again.

PR #8474, opened 2019-03-28
Fix opam2/arm64 build

vsiles:
The current ocaml used to build hhvm is from opam2, which is lacking
the patches/ocaml.patch patch. This PR updates the opam switch creation
so that ocaml is patched before it is built.

The old ocaml is still present to have a simple PR. Next PR will remove
the old ocaml from third-party and the toplevel ocaml-patch


vsiles:
Tested and validated on arm64 (see issue linked just above).

swalk-cavium:
@fredemmott - Any chance of this one landing?

PR #8503, opened 2019-05-16
Add JitPgoPseudomain command line option to allow retranslation.

dave-estes-ampere:
This new flag changes the behavior of shouldPGOFunc() additionally
allowing pseudomains to be PGO'd. This allows pseudomains to be
retranslated as a TransOptimize translation instead of the a simple
TransLive translation. This leads to more optimized pseudomains,
but more importantly, any call sites within the pseudomain that
originally pointed to a TransProfile translation will now point to
a TransOptimize translation.

Experiments using tcprint on the OSS Mediawiki target showed
a significant number of perf samples in aProf without the new
JitPgoPseudomain flag. Setting the existing JitPseudomain to false
led to profiles with no perf samples in aProf because the
pseudomains were always always interpretted and thus the call sites
were never "stale". The performance was mildly better, but
intpretting pseudomains was slow. Using the new JitPgoPseudomain
gave the strongest performance increase also with no perf samples
in aProf.

### DIFF
--- a/hphp/runtime/base/runtime-option.h
+++ b/hphp/runtime/base/runtime-option.h
@@ -803,6 +803,7 @@ struct RuntimeOption {
                                                                         \
   F(bool, JitDisabledByHphpd,          false)                           \
   F(bool, JitPseudomain,               true)                            \
+  F(bool, JitPGOPseudomain,            false)                           \
   F(uint32_t, JitWarmupStatusBytes,    ((25 << 10) + 1))                \
   F(uint32_t, JitWarmupMaxCodeGenRate, 20000)                           \
   F(uint32_t, JitWarmupRateSeconds,    64)                              \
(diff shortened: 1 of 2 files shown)


fredemmott:
This is probably simple enough for us to review and merge, but in general pseudomain optimizations are extremely low priority for us, given that they are banned in strict Hack (replaced by `<<__EntryPoint>>` functions), and PHP is not supported by recent versions of HHVM

dave-estes-ampere:
Thanks @fredemmott. Are FB internal workloads typically strict Hack? Just wondering, because it might be worth trying out the flag on some FB internal workloads that use pseudomains. I was startled at the gain on Mediawiki.

fredemmott:
Yes, strict is typical, and we have 0 or near-zero pseudomains

PR #8508, opened 2019-05-25
Split security policy out from README.md to SECURITY.md

azjezz:
SECURITY.md is an increasingly common convention, and now has special recognition on github


fredemmott:
Retitled, and added description

PR #8513, opened 2019-06-06
use SIGPROF for Xenon

fredemmott:
Summary: Currently SIGVTALRM is used by both Xenon and request timer.  This tris to use SIGPROF for Xenon, and handle it in the dedicated signal handling thread.

Differential Revision: D15696983

### DIFF
--- a/hphp/runtime/ext/xenon/ext_xenon.cpp
+++ b/hphp/runtime/ext/xenon/ext_xenon.cpp
@@ -27,6 +27,7 @@
 #include "hphp/runtime/vm/vm-regs.h"
 
 #include "hphp/util/struct-log.h"
+#include "hphp/util/sync-signal.h"
 #include "hphp/util/thread-local.h"
 
 #include <signal.h>
@@ -145,7 +146,7 @@ Array parsePhpStack(const Array& bt) {
 // If in always on mode, the Xenon Surprise flags have to be on for each thread
 // and are never cleared.
 // For timer mode, when start is invoked, it adds a new timer to the existing
-// handler for SIGVTALRM.
+// handler for SIGPROF.
 
 Xenon& Xenon::getInstance() noexcept {
   static Xenon instance;
@@ -165,6 +166,13 @@ void Xenon::incrementMissedSampleCount(ssize_t val) {
 int64_t Xenon::getAndClearMissedSampleCount() {
     return std::atomic_exchange<int64_t>(&m_missedSampleCount, 0);
 }
+
+static void onXenonTimer(int signo) {
+  if (signo == SIGPROF) {
+    Xenon::getInstance().onTimer();
+  }
+}
+
 // XenonForceAlwaysOn is active - it doesn't need a timer, it is always on.
 // Xenon needs to be started once per process.
 // The number of milliseconds has to be greater than zero.
@@ -193,9 +201,10 @@ void Xenon::start(uint64_t msec) {
 
     sigevent sev={};
     sev.sigev_notify = SIGEV_SIGNAL;
-    sev.sigev_signo = SIGVTALRM;
-    sev.sigev_value.sival_ptr = nullptr; // null for Xenon signals
+    sev.sigev_signo = SIGPROF;
     timer_create(CLOCK_REALTIME, &sev, &m_timerid);
+
+    sync_signal(SIGPROF, onXenonTimer);
 
     itimerspec ts={};
     ts.it_value.tv_sec = fSec;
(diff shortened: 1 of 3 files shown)


enjmusic:
@binliu19 This doesn't seem to actually work in blocking Xenon's profiling timer signal for quite a while (since there's a solid chunk of time between the Xenon call to `sync_signal` and the call to `block_sync_signals_and_start_handler_thread`. I was able to get the following behavior with a wide range of Xenon sampling periods from 10ms all the way up to 1 second.
```raw
$ hhvm --config test.ini script.php 
Profiling timer expired
```

hphp/runtime/ext/xenon/ext_xenon.cpp
```diff
@@ -193,10 +201,12 @@ void Xenon::start(uint64_t msec) {
 
     sigevent sev={};
     sev.sigev_notify = SIGEV_SIGNAL;
-    sev.sigev_signo = SIGVTALRM;
-    sev.sigev_value.sival_ptr = nullptr; // null for Xenon signals
+    sev.sigev_signo = SIGPROF;
     timer_create(CLOCK_REALTIME, &sev, &m_timerid);
 
+    signal(SIGPROF, SIG_IGN);
```
enjmusic: One more small comment is that `signal` should probably go above `timer_create` to avoid these kinds of circumstances:
- `hhvm` receives a `SIGSTP`/`SIGSTOP` after `timer_create` and is resumed later with `SIGCONT`, potentially causing the realtime timer to trigger immediately before `SIGPROF` is ignored
- Some long time window where other threads execute between `timer_create` and `signal`

binliu19: My understanding is that the timer isn't really running before `timer_settime()`.  If so, there is no need to move it above.

enjmusic: You're completely right, I got mixed up. Sorry for the confusion!

enjmusic:
@binliu19 Nice work! This updated patch provides the same fix as #8506 as demonstrated by the following test that I ran on both PRs' builds (for source, see the linked PR):
```raw
Spawning 10 children...
Sleeping in main thread while loop - slept 2 seconds, should have slept 2 seconds.
Sleeping in main thread while loop - slept 2 seconds, should have slept 2 seconds.
Sleeping in main thread while loop - slept 2 seconds, should have slept 2 seconds.
Desired sleep: 5 seconds. Actual sleep: 5 seconds. Gathered 1 samples.
Desired sleep: 6 seconds. Actual sleep: 6.01 seconds. Gathered 1 samples.
Sleeping in main thread while loop - slept 2 seconds, should have slept 2 seconds.
Desired sleep: 7 seconds. Actual sleep: 7 seconds. Gathered 1 samples.
Desired sleep: 8 seconds. Actual sleep: 8 seconds. Gathered 1 samples.
Sleeping in main thread while loop - slept 2 seconds, should have slept 2 seconds.
Desired sleep: 9 seconds. Actual sleep: 9 seconds. Gathered 1 samples.
Desired sleep: 10 seconds. Actual sleep: 10 seconds. Gathered 1 samples.
Sleeping in main thread while loop - slept 2 seconds, should have slept 2 seconds.
Desired sleep: 11 seconds. Actual sleep: 11 seconds. Gathered 1 samples.
Desired sleep: 12 seconds. Actual sleep: 12 seconds. Gathered 1 samples.
Sleeping in main thread while loop - slept 2 seconds, should have slept 2 seconds.
Desired sleep: 13 seconds. Actual sleep: 13 seconds. Gathered 1 samples.
Desired sleep: 14 seconds. Actual sleep: 14 seconds. Gathered 1 samples.
Sleeping in main thread while loop - slept 2 seconds, should have slept 2 seconds.
All children exited!
Main thread gathered 9 samples.
```
Thank you for doing this 😃

binliu19:
@enjmusic Glad it works for you.  I've also seen sporadic failures with 'syscall interrupted by signal' in some occasions.  Hopefully this fixes that too. 

I am trying to confirm that Xenon still works with the change.

We should probably disable Xenon by default in cli mode.

enjmusic:
I think Xenon sampling is disabled by default everywhere and will only turn on if a nonzero `hhvm.xenon.period` is supplied in a `--config` file. The `Gathered 1 samples` from my testing script at least shows that Xenon is gathering samples, although it's definitely good to verify that you're getting samples that look sane if you're worried.

enjmusic:
Also, before this is merged we should update the comments in [hphp/runtime/ext/xenon/ext_xenon.h](https://github.com/facebook/hhvm/blob/master/hphp/runtime/ext/xenon/ext_xenon.h) to correct the comment stating that `Xenon appends a timer to the already existing SIGVTALRM handler`.

PR #8530, opened 2019-07-09
Fix return from MimePart::parse loop if ProcessLine fails

### LINKED ISSUE #8532: Handle large mime files parsing errors
`ProcessLine` method returns error when it identifies a mime that contains more than 300 children. `MimePart::parse` is not handling this error which causes the `ProcessLine` to log an error for every line of the mime.

This bug was fixed in PECL `mailparse-v3.0.3` which was not ported hhvm library.

> Bug - https://bugs.php.net/bug.php?id=63542
> Fix - https://github.com/php/pecl-mail-mailparse/commit/8aa8c3d5195a38cb03f30d9ec89bc62a1b5c11e2

I have tried to simply port these fixes to HHVM.

### HHVM Version

> $ hhvm --version     
> HipHop VM 4.0.4 (rel)
> Compiler: 1558567803_N
> Repo schema: 3db73d7601eff5f54a8e907a42a71f8c7bfae2cb

### Operating System and Version

MacOS Mojave 10.14.5

Pass a large email (with more than 300 attachments) to `MimePart::parse` method. 

> MimePart::parse(`<large file>`, `<size>`)

### Actual result

- Generates lots of `MIME message too complex` errors.
- Full parsed MIME resource is not returned.

### Expected result

- Should generate a single `MIME message too complex` error 
- `MimePart::parse` should return `false`

madhugb:
# Fix #8532 

ProcessLine method returns false if the count of children have exceeded `MAXPARTS`. Since it would do this for all of the following lines, this is safe to break the loop and not care about the rest of the code.

This fix was missing from the php pecl mailparse library

### DIFF
--- a/hphp/runtime/ext/mailparse/mime.cpp
+++ b/hphp/runtime/ext/mailparse/mime.cpp
@@ -621,7 +621,17 @@ bool MimePart::parse(const char *buf, int bufsize) {
     if (len < bufsize && buf[len] == '\n') {
       ++len;
       m_parsedata.workbuf += String(buf, len, CopyString);
-      ProcessLine(req::ptr<MimePart>(this), m_parsedata.workbuf);
+      if (!ProcessLine(req::ptr<MimePart>(this), m_parsedata.workbuf)) {
+        /*
+         * Code ported from php pecl
+         * ProcessLine() only returns FAILURE in case the count of children
+         * have exceeded MAXPARTS and doing so at the very begining, without doing any work.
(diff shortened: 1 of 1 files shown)


zuercher:
@fredemmott Any word on this? Without this fix message with large numbers of attachments produces log storms and end up not being parsed correctly with an incorrect return value from MimePart::parse.

zuercher:
Sure. We'll try to get one written in the next few days.

Issue #8430, opened 2019-02-05
CLI server: client does not exit on script completion


ssandler:
### HHVM Version

```
$ hhvm --version
HipHop VM 3.30.2 (rel)
Compiler: remotes/origin/3.30.2-slack-0-g327d860eb6f5b8c13c31d2674d607bc9badafafe
Repo schema: 1a3521cc3e29cfd81df07e6b64acade1241f6538
```

### Operating System and Version

Ubuntu 16.04

Notable this does not repro on OSX

### Standalone code, or other way to reproduce the problem

```

$ hhvm -m server -d hhvm.unix_server_path=/tmp/mysock &

$ cat test.php
<?hh
echo 'test';

$ hhvm -d hhvm.use_remote_unix_server=true -d hhvm.unix_server_path=/tmp/mysock test.php
test
```

### Actual result

The client process hangs and does not return control after the script completes.

### Expected result

The client process exits when the script completes.

fredemmott:
FYI, unlikely to look at this before 4.0, but I suspect a fix would be a candiate for backporting to both 4.0 and 3.30

zuercher:
For reference, this bug describes the problem that occurs in the cli client: https://bugzilla.kernel.org/show_bug.cgi?id=40852

PR #8520, opened 2019-06-26
fix CLI client exit hang

zuercher:
Summary:
Per #8430, CLI client does not exit upon script termination because the
client's closure of the server's stderr does not trigger a read event
in all cases, leaving the client waiting to join a thread that will never
terminate. Modifies the stderr reader to use folly::AsyncPipeReader to
avoid the hang.

fixes facebook/hhvm#8430

PR #8531, opened 2019-07-10
fix CLI client exit

zuercher:
Summary:
Per #8430, CLI client does not exit upon script termination because the
client's closure of the server's stderr does not trigger a read event
in all cases, leaving the client waiting to join a thread that will never
terminate. Introduce LogThread struct to encapsulate the thread and a
pipe used to signal the thread to exit (aka the self-pipe trick).

fixes facebook/hhvm#8430

### DIFF
--- a/hphp/runtime/vm/extern-compiler.cpp
+++ b/hphp/runtime/vm/extern-compiler.cpp
@@ -181,6 +181,94 @@ struct CompilerOptions {
 
 constexpr int kInvalidPid = -1;
 
+struct Pipe final {
+  explicit Pipe(bool nonblocking = false) {
+    int flags = O_CLOEXEC;
+    if (nonblocking) flags |= O_NONBLOCK;
+
+    if (pipe2(fds, flags) == -1) throwErrno("unable to open pipe");
+  }
+  ~Pipe() {
+    if (fds[0] != -1) close(fds[0]);
+    if (fds[1] != -1) close(fds[1]);
+  }
+  FILE* detach(const char* mode) {
+    auto ret = fdopen(fds[*mode == 'r' ? 0 : 1], mode);
+    if (!ret) throwErrno("unable to fdopen pipe");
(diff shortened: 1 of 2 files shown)


zuercher:
Replaces #8520. I think this change should both be more reliable and hopefully easier understand.

hphp/runtime/vm/extern-compiler.cpp
```diff
@@ -181,6 +181,94 @@ struct CompilerOptions {
...
+  int remoteIn() const { return fds[0]; }
+  int remoteOut() const { return fds[1]; }
+  int fds[2];
+};
+
+std::string readline(FILE* f);
+
+struct LogThread {
+  LogThread(int pid, FILE* file)
+    : m_pid(pid), m_file(file), m_signalPipe(std::make_unique<Pipe>(true))
```
fredemmott: from @paulbiss : Does this need to be a pointer?

zuercher: No, I can convert it to an fd and just keep track of that.

zuercher: Oops. Spoke too soon. It gets passed `readline`, which is expecting `FILE*`.

zuercher: And I don't think it's safe to pass non-pointer FILE around, if that's what you were asking for.

fredemmott: Sorry, to clarify, does `Pipe` need to be a pointer? The FILE* is fine

zuercher: I chose to store it as a pointer so that it can be destroyed, closing the pipe's file descriptors, when the thread is stopped (and destroyed). Strictly speaking, no it doesn't need to be a pointer, though I think I'd want to add a close method to Pipe to close the fds in LogThread::stop.

PR #8540, opened 2019-07-24
Make HHI types match runtime types

lexidor:
In the face-sql implementation from slack, we discussed how they were casting NULL to string. I presumed that the return value should have a nullable string, instead of a string. `mapRows(): Map<string, ?string>` instead of `Map<string, string>`. I checked my HHI, but it says that a string was used. I did not believe that, so I tried the runtime. At runtime you actually get a nullable string.

### DIFF
--- a/hphp/hack/hhi/stdlib/builtins_async_mysql.hhi
+++ b/hphp/hack/hhi/stdlib/builtins_async_mysql.hhi
@@ -198,10 +198,10 @@ class AsyncMysqlQueryResult extends AsyncMysqlResult {
   public function numRowsAffected(): int { }
   public function lastInsertId(): int { }
   public function numRows(): int { }
-  public function mapRows(): Vector<Map<string, string>>{ }
-  public function vectorRows() { }
+  public function mapRows(): Vector<Map<string, ?string>> { }
+  public function vectorRows() Vector<?string> { }
   public function mapRowsTyped(): Vector<Map<string, mixed>> { }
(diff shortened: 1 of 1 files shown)


hphp/hack/hhi/stdlib/builtins_async_mysql.hhi
```diff
@@ -198,10 +198,10 @@ class AsyncMysqlQueryResult extends AsyncMysqlResult {
   public function numRowsAffected(): int { }
   public function lastInsertId(): int { }
   public function numRows(): int { }
-  public function mapRows(): Vector<Map<string, string>>{ }
-  public function vectorRows() { }
+  public function mapRows(): Vector<Map<string, ?string>> { }
+  public function vectorRows() Vector<?string> { }
```
hgoldstein: You need a colon here between the type and the method name / args.

hphp/hack/hhi/stdlib/builtins_async_mysql.hhi
```diff
@@ -198,10 +198,10 @@ class AsyncMysqlQueryResult extends AsyncMysqlResult {
...
   public function mapRowsTyped(): Vector<Map<string, mixed>> { }
-  public function vectorRowsTyped() { }
+  public function vectorRowsTyped() Vector<mixed> { }
```
hgoldstein: ditto.

lexidor:
Fixes #8540
Make github close the accompanying issue.

Issue #8322, opened 2018-09-18
hh_client hangs on OSX, but works if killed and retried


ssandler:
### HHVM Version

```
hhvm --version
HipHop VM 3.28.1-0-slack (rel)
Compiler: 1536019668_N
Repo schema: 67f5fcde7bddd6a0837c7f2e1ec04a5e6fee91f2
```

### Operating System and Version

MacOS High Sierra

### Standalone code, or other way to reproduce the problem

This only reproduces when running `hh_client` on a large repository. Small repositories work fine. Ours is much smaller than FB's www I'm sure, about 2M LOC.

### Actual result

hh_client hangs until it runs out of retries or you ctrl+c it. Subsequent runs of hh_client work just fine, and report the expected result. It takes several minutes to run out of retries, far longer than the 22 seconds the server spends getting ready from these logs.

```
$ hh_client
For more detailed logs, try `tail -f $(hh_client --monitor-logname) $(hh_client --logname)`
Server launched with the following command:
	'/usr/local/Cellar/hhvm@3.28/3.28.1/bin/hh_server' '-d' '/Users/ssandler/Documents/dev/webapp' '--waiting-client' '5'
Spawned typechecker (child pid=96900)
Logs will go to /private/tmp/hh_server/monitor_logs/zSUserszSssandlerzSDocumentszSdevzSwebapp-2018-09-18-14-15-02.monitor_log
hh_server is busy: [processing] /
Error: Ran out of retries, giving up!


$ cat $(hh_client --monitor-logname)
[2018-09-18 14:15:02.264] Loading config exception: (Sys_error "/usr/local/etc/hh.conf: No such file or directory")
[2018-09-18 14:15:02.264] Could not load config at /usr/local/etc/hh.conf, using defaults
Informant using dummy - resigning
[2018-09-18 14:15:02.331] Starting first server
[2018-09-18 14:15:02.332] About to spawn typechecker daemon. Logs will go to /private/tmp/hh_server/zSUserszSssandlerzSDocumentszSdevzSwebapp.log

[2018-09-18 14:15:02.334] Just started typechecker server with pid: 96902.
[2018-09-18 14:15:03.336] Got default request for typechecker. Prior request 1.0 seconds ago
```

```
$ cat $(hh_client --logname)
[2018-09-18 14:15:02.472] Loading config exception: (Sys_error "/usr/local/etc/hh.conf: No such file or directory")
[...]

PR #8546, opened 2019-07-30
workaround for MacOS hh_client hang

jeffomatic:
This change introduces a workaround for issues raised in #8322.

On MacOS, hh_client will hang in situations where the server process takes a long time to get to its "hello" response. The underlying cause appears to be that fds handed-off via Unix domain sockets can go stale on the receiving side if the equivalent fds are closed from the process that initiated the handoff. 72ab5dd addressed the same issue, but works only when server startup time is relatively fast.

One workaround for longer startup times is having the client re-establish a connection to the monitor in cases where it took a long time to get to the server hello. This is a little clumsy but better than the alternative, which is to manually kill hh_client. A true fix would require a bit more research, and may be moot given that [there are plans](https://github.com/facebook/hhvm/issues/8322#issuecomment-515128629) to re-architect hh_client/hh_server IPC.

### DIFF
--- a/hphp/hack/src/client/clientConnect.ml
+++ b/hphp/hack/src/client/clientConnect.ml
@@ -191,6 +191,7 @@ let with_server_hung_up (f : unit -> 'a Lwt.t) : 'a Lwt.t =
 
 let rec connect
     ?(first_attempt=false)
+    ?(allow_macos_hack=true)
     (env : env)
     (start_time : float)
   : conn Lwt.t =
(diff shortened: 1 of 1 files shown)


jeffomatic:
"MacOS has subtle POSIX bug" isn't that satisfying an explanation, so when there's time I'm hoping to see if this might be a configuration issue in libancillary or elsewhere, but hopefully this diff is a reasonable hack for the time being.

PR #8571, opened 2019-09-21
Make `%d` take `num` instead of mixed

lexidor:
I did `invariant($some_condition, 'Id must be greater than zero %d', $array_key);`
Scratching my head why this was allowed and figuring out that `%d` takes `mixed` in `PlainSprintf`, but `int` in `HH\Lib\Str\SprintfFormat`.

The docblock specifies `?int`

> function format_d(?int $s) : string;

But then again, another comment also specifies that it would be nice if this was `number` (`num`) instead.

> // It's common to pass floats; would be nice to type this as
> // 'number' once that type becomes available in userland.

It might be nice to make both formatters agree and end this confusion about what formatting interface what function is using. Either make `Str\format` weaker to take `num` too.
Or use `num` in `PlainSprintf` as an intermediate step to migrate to `int` later.


lexidor:
This will most definitely introduce typecheck failures in a codebase as big as Facebook's. I think using `%d` to format a numeric string or something nullable would be the most common. It is up to you if you want that loosely typed behavior to be retained.

If this is rejected for causing too many test failures, maybe we could make `invariant` and friends take `HH\Lib\Str\SprintfFormat` instead, since these functions are not replaceable in userland, because of their special meaning to the typechecker.

lexidor:
Refer to #8575 instead, since this PR has a merge commit in the history.

Issue #8570, opened 2019-09-20
[Typechecker] Error message for object casts doesn't apply in HHVM4.23 (4.24-dev)


lexidor:
### HHVM Version

```
HipHop VM 4.24.0-dev (rel)
Compiler: 1568778476_831641667
Repo schema: 557dc2470ba5f41e75c58f5b1ea62ff53c729c14
hh_server --version
hackc-2b00aad2b355c924e392ce3d004894215f1ad9ba-4.24.0-dev
```

### Operating System and Version
Ubuntu 18.04

### Standalone code, or other way to reproduce the problem

```HACK
<?hh // partial
/* Was previously already illegal in `// strict` */
function some_json_api(): stdClass {
    // $errorCode is defined, but skipping for brevity.
    return (object)["success" => false, "error" => "Curl Error: ".$errorCode];
}
```

### Actual result

```
Naming[2055] Object casts are unsupported. Try 'if ($var instanceof object)' or 'invariant($var instanceof object, ...)'.
   --> file.php
100 |             return (object)["success" => false, "error" => "Curl Error: ".$errorCode];
    |                     ^^^^^^
Naming[2049] Unbound name: object (an object type)
   --> file.php
100 |             return (object)["success" => false, "error" => "Curl Error: ".$errorCode];
    |                     ^^^^^^

```

### Expected result
 - Don't mention `instanceof`.
 - There is no typechecker type `object`, so this doesn't work with `is` either.

Wilfred:
T61626899

lexidor:
Fixed:
```
Naming[2049] Unbound name: object
   --> file.hack
558 |     (object)varray[];
    |      ^^^^^^

Naming[2055] Casts are only supported for bool, int, float and string.
   --> file.hack
558 |     (object)varray[];
    |      ^^^^^^

2 errors found.
```

PR #8572, opened 2019-09-21
Make error message for object casts compatible with newer HHVM versions

lexidor:
`instanceof` is dead and should therefore not be mentioned.
`object` is not a type that the typechecker knows about.

_Disclaimer: These are my very first lines of OCaml. I don't know if unused parameters prevent compilation. If just tried to make this look like `unset_cast`. I might need to pick up some OCaml to better be able to contribute to this project._


lexidor:
Fixes #8570

lexidor:
Changing the error message causes more issues than it helps atm.

PR #8575, opened 2019-09-25
Change the type of format `%d` from `mixed` to `num`.

lexidor:
This is a remake of #8571, but without the merge commit.

---

I did `invariant($some_condition, 'Id must be greater than zero %d', $array_key);`
Scratching my head why this was allowed and figuring out that `%d` takes `mixed` in `PlainSprintf`, but `int` in `HH\Lib\Str\SprintfFormat`.

The docblock specifies `?int`

> function format_d(?int $s) : string;

But then again, another comment also specifies that it would be nice if this was `number` (`num`) instead.

> // It's common to pass floats; would be nice to type this as
> // 'number' once that type becomes available in userland.

It might be nice to make both formatters agree and end this confusion about what formatting interface what function is using. Either make `Str\format` weaker to take `num` too.
Or use `num` in `PlainSprintf` as an intermediate step to migrate to `int` later.

---

This will most definitely introduce typecheck failures in a codebase as big as Facebook's. I think using `%d` to format a numeric string or something nullable would be the most common. It is up to you if you want that loosely typed behavior to be retained.

If this is rejected for causing too many test failures, maybe we could make `invariant` and friends take `HH\Lib\Str\SprintfFormat` instead, since these functions are not replaceable in userland, because of their special meaning to the typechecker.

### DIFF
--- a/hphp/hack/hhi/printf.hhi
+++ b/hphp/hack/hhi/printf.hhi
@@ -20,7 +20,7 @@ namespace {
  * causes the corresponding method in T to be looked up. For example,
  * '%d' will "call" the method
  *
- *   function format_d(?int $s) : string;
+ *   function format_d(num $s) : string;
  *
  * and consume an 'int' from the argument list.
  *
@@ -40,10 +40,8 @@ namespace {
  */
 
 interface PlainSprintf {
-  // It's common to pass floats; would be nice to type this as
-  // 'number' once that type becomes available in userland.
  <<__Rx>>
-  public function format_d(mixed $s) : string;
+  public function format_d(num $s) : string;
  <<__Rx>>
   public function format_s(mixed $s) : string;
  <<__Rx>>
--- a/hphp/hack/hhi/printf.hhi
+++ b/hphp/hack/hhi/printf.hhi
@@ -22,7 +22,7 @@ namespace {
  *
  *   function format_d(num $s) : string;
  *
- * and consume an 'int' from the argument list.
+ * and consume a 'num' from the argument list.
  *
  * Hex escapes are used for non-alphabetic characters. The '%%'
  * pseudo-specifier consumes nothing and appears as


jjergus:
> It's common to pass floats

Really, for `%d`? That sounds dangerous, since you lose all the decimal places. IMHO `%d` should be typed as `int` -- if someone wants to pass in a float (and lose all the decimals), they should explicitly cast it, like `Str\format('Pi is %d', (int)$pi)`.

`%f` is the one that should be typed as `num`, since passing an `int` value when a `float` is expected is not dangerous, we don't silently lose any information there.

fredemmott:
@jjergus for the more modern interfaces (like in the HSL), we define '%d' as int; this is a different format string definition, largely defined for PHP compatibility, which currently defines it as mixed

I'd also be fine with moving to int, but I'm unsure how practical it would be in FB www. num seems like a good first step

jjergus:
Oh, I see. Well if this is for PHP compatiblity and we want to keep that, then we should keep it as mixed and avoid using it; update invariant etc. to use Str\SprintfFormat like @lexidor suggested.

If we are OK with breaking the PHP compatibility, then codemod from mixed -> int sounds more reasonable to me than codemod from mixed -> num, which still breaks compatibility and is also still inaccurate/dangerous (I'd expect both codemods to be about the same amount of work anyway).

Actually if we don't need the PHP compatiblity anymore, maybe we could just delete PlainSprintf in favor of Str\SprintfFormat...? (so we don't have 2 similar interfaces, one of which should not be used)

lexidor:
@jjergus I am okay with the suggestion that format_f should take `num`. In the HSL it takes float. This has caused many a `Str\format('Power level: %f', (float) $num)`. This is not very nice for many a reason. This just begs for someone not noticing the dooming type error when $num is not a number. The cast will silently convert it to null and no one will be any the wiser. I'll make a PR on the [HSL](/hhvm/hsl).

lexidor:
@jjergus I think it is better to close this PR. We can always bring it up later. This interface is depended upon by every hack program in existence, so we can't roll this out without an automated hhast migration.

PR #8576, opened 2019-09-25
Allow for the use of disallow_array_literal outside of fb

lexidor:
I applied the typechecker setting `disallow_array_literal` and made my code use `darray[...]` and `varray[...]` instead of `[...]` and `array(...)`. HHVM's built-in hhi's don't typecheck with this setting turned on. The PR attempts to mitigate that.

As far as I understand the default values in hhi's are not used at runtime. I did check for PHP examples of each function, to make sure that the `d/v`array type was possible. I did not try to rule out if a function could also take the other type.


lexidor:
These changes will not be adequately tested by facebook's infra, since you use the setting `deregister_php_stdlib`. All these functions are marked with `__PHPStdLib`.

PR #8578, opened 2019-09-30
XHP namespace support in parser and typechecker

johanoskarsson:
Prototype for XHP namespace support. This only covers the parser and type checker, not runtime yet. Before taking this any further I would love some feedback on the general direction. Is this PR doing the right things at the right level?

For a more in depth writeup and syntax discussion see https://docs.google.com/document/d/1zWf5mQBdVmuDQ65wz93DPC7eCyWLTw_ngeJqIzA65f4/edit#heading=h.cc3sepimajn8

This PR supports the following:
Defining XHP classes within namespaces, same as regular classes
use type foo\:bar
use type foo\:bar as :baz
use namespace foo
use namespace foo as bar
Fully qualified XHP element names, <:foo:bar> will be resolved as foo\:bar

This PR does not support the following:
Automatically setting an XHP class inside of a namespace based on name. 
`class :foo:bar {}`
will not end up as this automatically 
```
namespace foo { 
  class :bar{}
}
```
This was suggested as a temporary migration path but I've left it out for now to keep the PR simpler.

Code changes in this PR:

declaration_parser.rs:
change is to allow an xhp class name to be used in a namespace alias 

parser_trait.rs:
The function for finding a qualified name is extended to support XHPClassName beyond just Name. 

namespaces.ml:
The meat of the changes. From the elaborate_id_impl function doc:
"Resolves an identifier in a given namespace environment. For example, if we are in the namespace "N\O", the identifier "P\Q" is resolved to "\N\O\P\Q"."

We extend that function to do something similar for XHP elements. An element like :foo:bar is split into the namespace portion (foo) and the element (bar). Think of it like \foo\:bar.


fredemmott:
Current discussion at https://gist.github.com/fredemmott/7316287e3fb9cc43b312b3e51c155ba5

fredemmott:
Should this be closed in favor of #8595 ?

johanoskarsson:
Once #8595 is ready for prime time I'll update this branch to work with that, so we can close this and I'll open another one with the new approach.

PR #8585, opened 2019-10-10
optional integer overflow logging/exception

zuercher:
Adds a configuration setting hhvm.hack.lang.check_int_overflow. When enabled,
compilation generates overflow variants of the Add, MinusEqual, Mul, MulEqual,
PlusEqual, PostDec, PostInc, PreDec, PreInc, and Sub instructions (with O
suffixes). During execution, if an integer overflow occurs, the variant
instructions raise an E_WARNING (if check_int_overflow=1) or throws an
ArithmeticError (if check_int_overflow=2).


fredemmott:
Thank you so much! This looks great to me - importing and getting better reviewers :)

fredemmott:
Several tests e.g. hphp/test/slow/int_overflow/inc_dec_1.php (and several others in that directory) are failing in repo-authoritative mode

zuercher:
Hm. Ok. I'll take a look.

zuercher:
Ah, I see. The Hack.Lang.CheckIntOverflow flag needs to be set at compile time and I failed to add it to the .hphp_opts files. I re-ran these tests locally with `test/run -m interp -r` and `test/run -r` flags in addition to the default.

zuercher:
@fredemmott any word on this?

fredemmott:
sorry, think I missed the last update given the timing :/

fredemmott:
Main test runs are clean, a few of the unusual ones are taking longer than expected.

fredemmott:
reran tests, looks bogus. Poking people for review.

fredemmott:
Approved internally, waiting for clean test run.

If run is clean, we'll be landing this with https://gist.github.com/fredemmott/7ecacbd1ba3ef220a6664d9443d5b976

If we find issues, we'd appreciate it if you could merge in those - but don't update this PR just for those (updating the PR resets a bunch of review/test state)

fredemmott:
One that we need timed out :( rebasing and re-running

fredemmott:
Had the specific one I needed clean, but have other failures; confident they're unrelated so landing, but chance of a revert if I'm wrong.

PR #8593, opened 2019-11-07
Add build target for run_generate_full_fidelity

johanoskarsson:


hphp/hack/src/dune
```diff
@@ -130,13 +130,24 @@
...
   (name generate_full_fidelity)
+  (modules generate_full_fidelity)
+  (libraries
+    core_kernel
+    html_entities
+    str
+    parser_schema
+    collections))
+
+(executable
```
fredemmott: why is this split needed?

PR #8595, opened 2019-11-09
XHP Class modifier

johanoskarsson:
As a first step towards XHP namespace support this PR introduces the xhp class modifier. This will be used to declare an xhp class:
`xhp class name {}`
compared to the old style
`class :name {}`

There are two options in support of this change as requested.
enable_xhp_class_modifier: with this enable the xhp modifier is allowed, otherwise an error is thrown.
disable_xhp_element_mangling: in the old style xhp if an xhp element <name/> is used the name is assumed to have colon prepended and is also mangled by under the hood be represented as xhp_name. This clashes with the new style since the resolve won't find it.

I'd love some feedback on this general direction and any tips.

TODOs still left:
* copy xhp*.php tests and modify to new style and enable the flags
* I'm struggling to avoid the redundant reparsing in `parse_classish_declaration`, any help is welcome
* the regen.sh is tricky to use for open source folks, Fred is helping make this easier though
* blow up if we find a - in the class name

For more context and discussion see #xhp in Slack and https://gist.github.com/fredemmott/de7eeb91765b0c98fafac9c591263aca


fredemmott:
FB: D18531883 and D18519816 will make src/oxidized/regen.sh fully supported, and make it produce signed source in open source builds

Wilfred:
@johanoskarsson this is looking like a great start!

A few suggestions:

(1) You've added a ton of test files -- do you really need all of them? It looks to me like some are just copy-pasted from typechecker tests that aren't really exercising the new parsing logic you've added.

(2) I wouldn't bother with `foo.like_types.exp` or `foo.legacy_decl.exp` at least. If you don't have a `.like_types.exp` or a `legacy_decl.exp` file, we don't run the like type tests or legacy decl tests on the relevant `.php` file, which seems appropriate here.

(3) Looks like you're adding `xhp` as a class modifier, allowing both `xhp final class` and `final xhp class`. I suggest you require it in a consistent position rather than adding it to modifiers in order. We've been stricter with other newer modifiers too, like `async`.

I think this would make life easier for readers and for your parsing logic too. 

I suggest you make the syntax `xhp final class Foo {}`, always requiring the `xhp` modifier first. (Always last: `final xhp class` is fine by me too.) It's easier to parse both in code and for readers.

This would let you add a `self.parse_xhp_keyword()` before/after the call to `self.parse_classish_modifiers()` in `parse_classish_declaration`.

(4) Since the parsing logic deal with a 'classish', a class/interface/trait/enum, please add tests to make sure that `xhp trait Foo {}`, `xhp interface Foo {}`, `xhp enum Foo {}` are all parse errors.

I hope that helps! If you have any specific questions about files or tooling, do let me know :)

ecreeth:
This is awesome! :100:

fredemmott:
> (1) You've added a ton of test files -- do you really need all of them? It looks to me like some are just copy-pasted from typechecker tests that aren't really exercising the new parsing logic you've added.

Yeah, this needs revisiting; at the time, the parser tests weren't runnable externally as the s-expr dumper was fb-only; fixed this last week.

fredemmott:
> I suggest you make the syntax xhp final class Foo {}, always requiring the xhp modifier first. (Always last: final xhp class is fine by me too.) It's easier to parse both in code and for readers.

I'd strongly prefer `xhp class` - grepping for `xhp class foo` seems likely to be much more frequent than grepping for `final class foo`

fredemmott:
That said, yeah, I think the order should be specified, and it should be `xhp class`; I'm not really thinking of `xhp` as a modifier: I think of `xhp class` as a distinct thing (similar how to `element` was in the original XHP proposal a decade ago :p ), and it should stay together

johanoskarsson:
Thanks for the review @Wilfred, much appreciated. I was hoping to put up a new version today but looks like I won't get it finished until tomorrow so I'll write a short update here to make sure I'm at to make sure I'm heading in the right direction.

1) I did indeed copy the old tests, mostly so I could be confident in my change. I can reduce the number of tests to a more sensible number.
2) Removed
3) Looks like Fred and Scott are in favor of `<modifiers> xhp class` so I'll move in that direction. My understanding is that this would then be a new parameter to ClassishDeclaration, separate from the current modifiers?

I have one test related question. One test, pocket_universes_compile_good, uses hh_single_compile under the hood. It fails because Parser_options_provider: global parser options not set. That makes sense since I don't see it set anywhere and I use that to check if the xhp modifier flag is enabled. I tried to modify hh_single_compile to set it, but then I get an oom. Seems to be due to SharedMem.init ~num_workers:0 GlobalConfig.empty_sharedmem_config setting the shared memory to 0. So I'm in a bit of a pickle. Any ideas for how to solve that catch 22 so I can access the flags? Is there a better way?

I'll keep hammering at this tomorrow and hopefully have this PR up for review again.

johanoskarsson:
After some fighting with the latest master I think I've got this in good shape.
* Update to master
* Reduced the number of tests and added the ones suggested
* Changed the xhp keyword to not be part of the any-order-modifiers list but separate just before class
* I still have that one issue with Parser_options_provider that is failing a test. I'll try again tomorrow but would love a hand in figuring it out.

This PR should either way be ready for a re-review while I work on that.

hphp/hack/src/parser/full_fidelity_ast.ml
```diff
@@ -3587,6 +3591,7 @@ if there already is one, since that one will likely be better than this one. *)
         let is_abs = List.mem kinds Abstract ~equal:( = ) in
         match token_kind kw with
         | Some TK.Class when is_abs -> Cabstract
+        | Some TK.Class when c_is_xhp -> Cxhp
```
Wilfred: What about `abstract xhp class Foo {}`?

hphp/hack/src/parser/rust_parser_errors.rs
```diff
@@ -4196,6 +4196,20 @@ where
...
+                match Self::token_kind(&cd.classish_keyword) {
+                    Some(TokenKind::Interface) => self.errors.push(Self::make_error_from_node(
+                        node,
+                        errors::invalid_xhp_classish("Interfaces"),
+                    )),
+                    Some(TokenKind::Trait) => self.errors.push(Self::make_error_from_node(
+                        node,
+                        errors::invalid_xhp_classish("Traits"),
+                    )),
+                    _ => (),
```
Wilfred: Does this need to handle `xhp enum Foo {}`?

hphp/hack/test/mro/xhp.php.exp
```diff
@@ -1,6 +1,6 @@
 Linearization for class \:x:
-[\:x]
+[\:x(PAT)]
```
Wilfred: Do you know why this has changed? It's not obvious to me.

Wilfred:
I've left a few comments, but this is looking in pretty good shape now. This is definitely a sprawling part of our codebase, so I'm super impressed you've made this work :)

Are you still suffering with the pocket universe test issue? If so, could you point at the lines of code you're referring to?

johanoskarsson:
@Wilfred:
Some more details on the test failure I'm seeing. When I run `make hack_test` the test that fails is this one:
`hphp/hack/test/pocket_universes/compile/translate.good.php`

When I review the output it gives me this error:
```
Parser_options_provider: Attempt to read global parser options before they were set
Parser_options_provider: global parser options not set
```

This test uses hh_single_compile under the hood so that's where I went looking.
I've seen a similar error in hh_parse where the solution seems to be to `Parser_options_provider.set popt`.

Perhaps the real solution is that I should be doing something differently when I read the option in declaration_parsers and beyond?

Either way, I attempted the `Parser_options_provider.set` solution in hh_single_compile by adding it to `parse_file`.
Unfortunately this leads to a different error.

```
Raised by primitive operation at file "src/heap/sharedMem.ml", line 465, characters 60-70
Called from file "src/heap/workerCancel.ml", line 37, characters 6-10
Called from file "src/heap/sharedMem.ml", line 505, characters 43-59
Called from file "src/heap/sharedMem.ml", line 1375, characters 4-18
Called from file "src/hh_single_compile.ml", line 316, characters 2-34
Called from file "src/hh_single_compile.ml", line 380, characters 4-115
Called from file "src/hh_single_compile.ml", line 498, characters 8-155

Error in file translate.good.php: SharedMem.Heap_full
```

I assume this in turn is due to the shared memory being set to 0 at the beginning of hh_single_compile:

```
    let handle =
      SharedMem.init ~num_workers:0 GlobalConfig.empty_sharedmem_config
    in
    ignore (handle : SharedMem.handle);
```

Any suggestions on how to solve this would be welcome.

hphp/hack/test/mro/xhp.php.exp
johanoskarsson: Fixed in the last commit. Maybe the check should be flipped but `mro_passthrough_abstract_typeconst` was comparing the class kind to just Cnormal

hphp/hack/src/parser/rust_parser_errors.rs
johanoskarsson: Fixed! Good catch

hphp/hack/src/parser/full_fidelity_ast.ml
johanoskarsson: Good point. I added a case for when a class is abstract and xhp to make it xhp.

johanoskarsson:
I've fixed the merge conflicts again. I also fixed the final test that relied on hh_single_compile to use GlobalConfig.default_sharedmem_config instead of GlobalConfig.empty_sharedmem_config so I can use Parser_options_provider.set. I'm not entirely sure of the implications of this, but it seems to work at least.
This PR should be ready for another review now.

hphp/hack/test/typecheck/xhp_modifier/xhp_modifier_no_dashes.php
```diff
@@ -0,0 +1,4 @@
+<?hh // strict
+
+# this is not allowed
+xhp class Foo-bar {}
```
Wilfred: Great test! Minor nit: I'd prefer `//` style comments for consistency with our other tests.

hphp/hack/test/typecheck/xhp_modifier/xhp_modifier_not_on_enum.php
```diff
@@ -0,0 +1,4 @@
+<?hh // strict
+
+# this is not allowed
```
Wilfred: Could you make your comments more explicit? I can see it's not allowed by looking at the `.exp` file, but something like "the xhp keyword only applies to classes" would be more helpful here.

hphp/hack/test/typecheck/xhp_modifier/optional_xhp.php
```diff
@@ -0,0 +1,20 @@
...
+
+class A {
+  public ?:x:base $xhp;
+
+  public function foo(?:x:base $x): ?:x:base {
+    return null;
+  }
+}
+
+xhp class x:base {}
```
Wilfred: Most of your tests make sense to me, but I'm not sure what this is testing. Could you add a comment? Something to do with mangling names when they're nullable?

hphp/hack/test/typecheck/xhp_modifier/xhp_modifier_no_dashes.php
johanoskarsson: Will fix. This our internal style and old habits die hard :)

Wilfred:
I've left some minor nits, but overall this is almost complete I think.

I do have a concern about using `class_kind` here. Since we have `Cabstract` in there, we're not able to represent a class being both abstract and xhp.

(Really we shouldn't have `Cabstract` either I think, but it's existed for a long time, since at least a630d46af46de08e7af80aa187e13918128ac1b5.)

I suggest you add a `c_xhp_keyword: bool` to `class_` in `ast.ml` rather than using `c_kind`. There's already `c_is_xhp` there, so we should probably add a comment to distinguish the two. This should make the PR smaller and it should be easier to reason about.

johanoskarsson:
I've fixed the comments and tests. I removed xhp from the c_kind and it's a boolean same as is_xhp now. This miraculously passed hack_test locally. I'll do some more testing and see how how it fares at runtime tomorrow, but figured it's worth another review to see that this is what you were thinking.

Wilfred:
Sorry for the delay in reviewing this. I ran out of time today, but I've not forgotten! I'll have a look tomorrow and investigate the parser issues.

johanoskarsson:
I updated this PR to where my local branch is. The changes include:
* Merging against master to resolve conflicts
* Since I couldn't get the previous config options to work I have pulled it into the ns env. This allows me to set it without having to read the parser options where i need the flag check in the parser.

TODOs:
* While the hack_tests pass on my machine i'm still struggling to run hhvm against a php file and passing the two flags to enable the xhp modifier functionality. It might just be that i've missed threading it through somewhere as it doesn't arrive at the hh_single_compile. It's also entirely possible that I don't quite understand how the configs work (they're quite sprawling). Any help would be welcome, but i'll keep slogging away to get this to work.

Wilfred:
I grabbed the changes and played with them yesterday. I didn't have a chance to try to execute on top of HHVM, but I did notice a few things:

(1) Reading the code, I think it would make more sense to rename `is_xhp_keyword` to `has_xhp_keyword`. I know the previous field name was my suggestion, sorry about that :)

(2) Your code changes aren't formatted: we use `rustfmt` and `ocamlformat`. Both should just work here. No pressure to fix this (I'm happy to handle it), but you might find fewer conflicts when rebasing if your style matches.

(3) The Rust code wasn't compiling when I tried it (I was using 39be5cbe8fc454526a8d06b3e49712b73443d0cd I think). I haven't tried the changes you've just pushed today yet (it's getting late here), but I fixed it with:

```diff
summary:     Fix direct decl parser

diff --git a/fbcode/hphp/hack/src/decl/direct_decl_smart_constructors.rs b/fbcode/hphp/hack/src/decl/direct_decl_smart_constructors.rs
--- a/fbcode/hphp/hack/src/decl/direct_decl_smart_constructors.rs
+++ b/fbcode/hphp/hack/src/decl/direct_decl_smart_constructors.rs
@@ -365,6 +365,7 @@
     Static,
     Super,
     Trait,
+    XHP,
     Yield,
 }

@@ -923,6 +924,7 @@
             TokenKind::DotDotDot => Node_::DotDotDot,
             TokenKind::Final => Node_::Final,
             TokenKind::Interface => Node_::Interface,
+            TokenKind::XHP => Node_::XHP,
             TokenKind::Yield => Node_::Yield,
             TokenKind::Namespace => {
                 self.state.namespace_builder.to_mut().is_building_namespace = true;
@@ -1260,6 +1262,7 @@
         &mut self,
         _arg0: Self::R,
         modifiers: Self::R,
+        xhp_keyword: Self::R,
         class_keyword: Self::R,
         name: Self::R,
         _arg4: Self::R,
@@ -1297,6 +1300,10 @@
             },
             final_: false,
             is_xhp: false,
+            is_xhp_keyword: match xhp_keyword? {
+                Node_::XHP => true,
+                _ => false,
+            },
             kind: match class_keyword? {
                 Node_::Interface => ClassKind::Cinterface,
                 Node_::Trait => ClassKind::Ctrait,
```

Wilfred:
Language changes are always slow to land, so thanks for pressing on.

I suggest the next thing we do is write some docs. This can be the PR description, and we can recycle the content for the official hacklang.org docs and perhaps a blog post too.

I think this will help us check we've covered all the cases. It's really useful to write sample code thinking about compelling examples. This tends to produce different code to test programs.

We'd probably want a structure like:

```
# Declaring classes with `xhp`

It's also possible to declare Hack XHP classes using the `xhp` keyword:

... minmal example

## XHP Inside Namespaces

... how XHP classes behave inside namespaces, how to access things like p, migration

## XHP Name Rules

... Use of - in XHP names, use of _, contrasting with :foo rules

## XHP Mangling

... how XHP is mangled, the flags available, the differences
```

I'm happy to help or to have a go at writing this if you need :)

fredemmott:
FYI we need the merge commit rebasing out before we can pull this in again

johanoskarsson:
Thanks for the review!

I've pushed a version now with the following:
* Removed the merge commit (I will keep forgetting this since it's not my usual workflow)
* Applied the direct_decl_smart_constructors.rs patch you suggested. However even without the patch I don't get a build error. Is this another buck vs open source build system difference? I haven't had a chance to verify if the file is built with make yet.
* I also attempted to run rustfmt on the files but I get a lot of unrelated format changes. As in, if I run it against an untouched master it changes the formatting unexpectedly. Is there a config or args that I should be using? Or is it a version difference? I have `rustfmt 1.4.8-stable` apparently.
If you want to run it that's fine too, whatever is easy.

I can write some basic docs next week to get us started. However I don't know if I consider this work done before the namespacing support is in place since that's why we're doing this in the first place.

fredemmott:
@Wilfred based on https://our.internmc.facebook.com/intern/diff/D18428579/?dest_number=95433289 those build issues appear to be a dune/cargo vs buck issue - the -mac and -cmake jobs pass, the buck builds fail; I'm not sure if that's that are cargo/dune config needs to be modified to closer match buck, or if there's fb-only changes needed to be made to buck

fredemmott:
It looks like the OSS build doesn't depend on that as the only consumer is `hh_decl`, which is currently in src/facebook/ , and only has a buck build config.

Wilfred:
We have an `.ocamlformat` here: https://github.com/facebook/hhvm/blob/master/hphp/hack/.ocamlformat which we use ocamlformat 0.12 (latest release) with. For rust, we have the following in our configuration TOML:

```
edition = "2018"
merge_derives = false
reorder_imports = true
reorder_modules = true
tab_spaces = 4
use_field_init_shorthand = true
use_try_shorthand = true
```

I've pushed two commits: one formatting, and one fixing a compilation issue. I suggest you just squash them into your commits :)

hphp/hack/src/hh_single_type_check.ml
```diff
@@ -533,6 +535,14 @@ let parse_options () =
       ( "--disallow-discarded-nullable-awaitables",
         Arg.Set disallow_discarded_nullable_awaitables,
         "Error on using discarded nullable awaitables" );
+      ( "--disable-xhp-element-mangling",
+        Arg.Set disable_xhp_element_mangling,
+        "Disable mangling of XHP elements. With the mangling enabled the new style xhp class will not resolve due to naming mismatch."
```
Wilfred: "new style" isn't as helpful for new maintainers. How about something like this?

```
Disable mangling of XHP elements :foo. This is necessary for classes using xhp class modifiers.
```

Wilfred:
OK, this logic is looking sensible to me. We should test it on HHVM, but it's a fairly small change to bytecode emission so I'm not very concerned.

We definitely need to ban `-` in XHP class names though. I'm guessing that it will just work as the parser doesn't see a `:`. I suggest you write a test `xhp class Foo-Bar`. Also, we should have a test that `xhp class :foo` is banned.

fredemmott:
> We definitely need to ban - in XHP class names though

Just to check we're on the same page - for now:
- `class :foo:bar-baz` is fine
- `xhp class foo:bar` is an error
- `xhp class foo-bar` is an error
- `class foo:bar` and `class foo-bar` are always errors
- additionally, either `xhp class` or `class :` is conditionally banned, depending on the global options?

johanoskarsson:
Merged in the formatting and compile error patches, thanks Wilfred! I also updated the argument description as suggested. hack_test passes on my machine so it should be in a good place now.

Just to make sure I also went through the tests for the scenarios you both mentioned and linked them below. I think we're in a good state here, but let me know if I've misunderstood anything.

class :foo:bar-baz is fine
This isn't explicitly covered by my tests, but ought to be by others.

> xhp class foo:bar is an error

https://github.com/facebook/hhvm/blob/1042e3dd81f237f564b80e77680488d505218606/hphp/hack/test/typecheck/xhp_modifier/xhp_modifier_no_colon.php

> xhp class foo-bar is an error

https://github.com/facebook/hhvm/blob/1042e3dd81f237f564b80e77680488d505218606/hphp/hack/test/typecheck/xhp_modifier/xhp_modifier_no_dashes.php

> class foo:bar and class foo-bar are always errors

https://github.com/facebook/hhvm/blob/1042e3dd81f237f564b80e77680488d505218606/hphp/hack/test/typecheck/xhp_modifier/regular_class_names_colon.php
https://github.com/facebook/hhvm/blob/1042e3dd81f237f564b80e77680488d505218606/hphp/hack/test/typecheck/xhp_modifier/regular_class_names_dash.php

> additionally, either xhp class or class : is conditionally banned, depending on the global options?

With the flags disabled the XHP keyword is explicitly banned:
https://github.com/facebook/hhvm/blob/1042e3dd81f237f564b80e77680488d505218606/hphp/hack/test/typecheck/xhp/xhp_modifier_banned.php.exp
However I do not ban `class :` if the flag is enabled. I think this is fine?

johanoskarsson:
I've written a little bit of documentation as requested. This version does assume that the namespace support is in place, so it's not entirely true as of yet.
I didn't dive too much into the details since I assume it's for the end user's benefit. Let me know what you think.
https://gist.github.com/johanoskarsson/1228ae1d2370b7e6244302e095a8fd37

Wilfred:
Awesome, those docs are really helpful!

I really like your split up pull request too. Phabricator squashes and rebases commits, so merging a pull request will create a single commit here. By splitting it up, we can see that your tests are exercising the new parser, not the new mangling.

Wilfred:
Landed in 432343666317cbdeb82a6ec5c369c992f5bd79c6.

@johanoskarsson do you have a PR up for the mangling changes?

PR #8603, opened 2019-11-15
Directly submodule most of third-party

fredemmott:
Keep forks in third-party/forks/


.gitmodules
```diff
@@ -1,4 +1,165 @@
...
+[submodule "libsodium/libsodium"]
+  path = third-party/libsodium/libsodium
+  url = https://github.com/jedisct1/libsodium.git
+[submodule "rsocket-cpp/roscket-cpp"]
+  path = third-party/rsocket-cpp/rsocket-cpp
+  url = https://github.com/rsocket/rsocket-cpp.git
+[submodule "fb303/src"]
+  path = third-party/fb303/src
+  url = https://github.com/facebookincubator/fb303.git
+[submodule "third-party/forks"]
```
jjergus: should this be here again? (same path + url is at the top)

PR #8613, opened 2019-11-25
Fix the dune build file to properly create a generate_full_fidelity executable

johanoskarsson:
Discussed this with Fred: turns out the executable created by dune here would run generate_full_fidelity that does not have a 'main' function so nothing would happen.


fredemmott:
i'll see if I can get this in today; in case, heads up for @jjergus  - this will need buck changes

fredemmott:
please change _lib to _data (to match our internal targets)

PR #8614, opened 2019-11-26
build_and_run.sh needs the hack root in the path

johanoskarsson:
It seems like the .ml file (hh_oxidize.ml in my case) is never found so that path isn't run. This fixes it for me, but perhaps there's another solution


fredemmott:
Thanks - was fixed by https://github.com/facebook/hhvm/commit/04642519fb6e25088803bab2dc48f3c5705bcd27

PR #8615, opened 2019-11-30
Expose __FUNCTION_CREDENTIAL__ to hack type checker

lexidor:
Summary: exposes FunctionCredential class to hack type checker

The type was added in 78ad5e1d, but the constant was not yet defined.

### DIFF
--- a/hphp/hack/hhi/constants.hhi
+++ b/hphp/hack/hhi/constants.hhi
@@ -59,6 +59,5 @@ const string __FUNCTION__ = '';
 const string __METHOD__ = '';
 const string __NAMESPACE__ = '';
 const string __COMPILER_FRONTEND__ = '';
-/*HH_IGNORE_ERROR[1002] The `new` operator can not make a constant*/
-/*HH_IGNORE_ERROR[2023] Illegal constant value (object type)*/
-const FunctionCredential __FUNCTION_CREDENTIAL__ = new FunctionCredential();
+/*HH_FIXME[4110] It is not possible to declare a constant of this type*/
+const FunctionCredential __FUNCTION_CREDENTIAL__ = 0;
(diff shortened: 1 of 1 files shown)


lexidor:
https://github.com/facebook/hhvm/commit/78ad5e1d70f4d42fef98f7632c218f0c6c38d73a#commitcomment-36204906

fredemmott:
Hmm, AIUI these should all be special-cased, as they're not namespaced; there at least should also need to be an autoimport list updated

lexidor:
Unable to locate the auto import map, but it is indeed special cased like `__METHOD__` and `__FUNCTION__` are.
https://github.com/facebook/hhvm/blob/eab6e2815200eeec2db6f3fb31c7fa6b39ccb885/hphp/hack/src/parser/hh_autoimport.ml#L109

https://github.com/facebook/hhvm/blob/eab6e2815200eeec2db6f3fb31c7fa6b39ccb885/hphp/hack/src/naming/naming_special_names.rs#L634

https://github.com/facebook/hhvm/blob/eab6e2815200eeec2db6f3fb31c7fa6b39ccb885/hphp/hack/src/naming/naming_special_names.ml#L606

fredemmott:
Also, I'm a little surprised that we're actually defining it as a class; it feels like just using an opaque type alias to a bogus value would be better, and avoid the fixme - @paulbiss ?

lexidor:
Going through my stale PRs.
Bump @paulbiss 
Can we add `__FUNCTION_CREDENTIAL__` to the typechecker?

lexidor:
Superceeded by https://github.com/facebook/hhvm/commit/c374ede3c7bbd3f4622ca3b19fe0bb8a0ce8a320

PR #8629, opened 2019-12-14
Put stringish_cast into hhvm

lexidor:


lexidor:
@fredemmott This failed the tests. Did I not correctly update all sites that mention the `stringish_cast` to `HH\stringish_cast`? The linter failed. I am presuming this is whitespace related. I'll check for whitespace errors.

fredemmott:
> fbcode/hphp/system/php/stringish_cast.php

Does not end in newline

All builds are broken, including the ones without anything FB-specific:

```
An error has been introduced into the systemlib, but we cannot give you a file and line number right now.
Check all of your changes to hphp/system/php
HipHop Parse Error: Cannot mix bracketed namespace declarations with unbracketed namespace declarations 1476
```

hphp/system/php/stringish_cast.php
```diff
@@ -0,0 +1,14 @@
+<?hh // strict
```
fredemmott: compare to other .php files in system/php using namespaces:

- all have .ns.php extension
- all have `namespace Foo {}` syntax instead of `namespace Foo;`

They're concatenated in a rather straightforward way at present, which depends on these conventions

lexidor: I was not aware of this convention. Should be fixed in a jiffy.

lexidor:
I'll get some more ram so I can build (and test) hhvm locally. 8 GB isn't enough for the compiler and it fails with a hard crash.

Not being able to see the test failures (even for tiny PR's like this) makes it hard to fix what's wrong.

@fredemmott can you tell me one last time what failed, so I can try and fix it? Next PR I make will have been actually built and tested locally.

PR #8633, opened 2019-12-23
Add return types to html encoding functions

lexidor:
These functions have signatures like `String HHVM_FUNCTION(function_name_here, ...`.
So they always return strings.

https://github.com/facebook/hhvm/blob/master/hphp/runtime/ext/string/ext_string.cpp#L1698
https://github.com/facebook/hhvm/blob/master/hphp/runtime/ext/string/ext_string.cpp#L1708
https://github.com/facebook/hhvm/blob/master/hphp/runtime/ext/string/ext_string.cpp#L1720
https://github.com/facebook/hhvm/blob/master/hphp/runtime/ext/string/ext_string.cpp#L1727

If it wasn't for the TAny in the HHI I would have never gone to the PHP docs and read the following for htmlspecialchars:
`If the input string contains an invalid code unit sequence within the given encoding an empty string will be returned, unless either the ENT_IGNORE or ENT_SUBSTITUTE flags are set. `
I assumed false on failure since that is the PHP-way.

### DIFF
--- a/hphp/hack/hhi/stdlib/builtins_string.hhi
+++ b/hphp/hack/hhi/stdlib/builtins_string.hhi
@@ -110,13 +110,13 @@ function str_repeat(string $input, int $multiplier);
 <<__PHPStdLib>>
 function wordwrap(string $str, int $width = 75, string $wordbreak = "\n", bool $cut = false);
 <<__PHPStdLib>>
-function html_entity_decode(string $str, int $quote_style = ENT_COMPAT, string $charset = "ISO-8859-1");
+function html_entity_decode(string $str, int $quote_style = ENT_COMPAT, string $charset = "ISO-8859-1"): string;
 <<__PHPStdLib>>
-function htmlentities(string $str, int $quote_style = ENT_COMPAT, string $charset = "ISO-8859-1", bool $double_encode = true);
+function htmlentities(string $str, int $quote_style = ENT_COMPAT, string $charset = "ISO-8859-1", bool $double_encode = true): string;
 <<__PHPStdLib>>
-function htmlspecialchars_decode(string $str, int $quote_style = ENT_COMPAT);
+function htmlspecialchars_decode(string $str, int $quote_style = ENT_COMPAT): string;
 <<__PHPStdLib>>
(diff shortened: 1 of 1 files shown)


WizKid:
I could be wrong but I have some memory of the C++ class String also could contain null. Or maybe that is only true for Array.

lexidor:
Thanks a lot for that info. This will most certainly prevent a wrong transcription in the future.
These functions don't return anything other than strings, according to the PHP manual, but that is a secondary source. I'll dive deep into the cpp and try to figure out if a null could sneak out here.

If anyone known that this null mapping does not apply to `String`, please tell me.

lexidor:
With my limited cpp knowledge, I was unable to find a way to sneak a null in here. I can't find a way for a null literal to reach the return. However, sine cpp typehints (:roll_eyes: limited vocabulary) are implicitly nullable, I can never be 100% sure.

lexidor:
@fredemmott Does `PHP\htmlspecialchars()` have a `string` return type?
If yes, we can safely update these too.

fredemmott:
Welll.... ` \DisableRuntimeTypecheck<string> {` :D

`lies_we_tell_hack/disable_runtime_typecheck.hhi` contains `type DisableRuntimeTypecheck<T> = T;`
`truths_we_hide_from_hack/disable_runtime_typecheck.php` contains `type DisableRuntimeTypecheck<T> = mixed;`

lexidor:
That is the most Facebook thing ever.
Facebook: let's make our own cpp implementation of string to make everything a little bit faster. `fb::string`.
Also Facebook: let's make a type hint that vanishes at runtime to make something slightly faster.

From the point of view of this PR, the types are safe then, right? Everything I know about these functions seems to indicate that they must return `?string` and it never mentions null anywhere in the php docs.

fredemmott:
> Also Facebook: let's make a type hint that vanishes at runtime to make something slightly faster.

Nope - let's make a type hint that vanishes at runtime so we can sed everything to it, and not have runtime exceptions for stuff the typechecker can't find

fredemmott:
> the types are safe then, right?

Can't infer that from this definition; going to pull it in for review, but this is going to depend on analyzing the C++ and trying to think through edge cases

PR #8635, opened 2019-12-30
Add getAttributesNamespaced() to ReflectionThing

lexidor:
Where Thing is File, Function, TypeAlias, and Class.

I didn't know this function existed and I have been emulating it for some time using tens of `getAttributeClass()` calls.

The signatures came from [here](https://github.com/facebook/hhvm/blob/7f4b187f5e39de84ea4c559175d3a96d03461cb8/hphp/runtime/ext/reflection/ext_reflection_hni.php#L2524).

If there is no reason to hide this, I would really love to have this. :sweat_smile: It would save me from programming backwards.

### DIFF
--- a/hphp/hack/hhi/reflection.hhi
+++ b/hphp/hack/hhi/reflection.hhi
@@ -17,7 +17,7 @@ class Reflection  {
   <<__Rx>>
   public static function getModifierNames($modifiers);
   public static function export(Reflector $reflector, $return = false);
-}
+}Reflection
 
 class ReflectionClass implements Reflector {
 
@@ -396,6 +396,8 @@ class ReflectionProperty implements Reflector {
   <<__Rx, __MaybeMutable>>
   public function getTypeText();
   public function getAttributes(): darray<string, varray<mixed>>;
+  <<__Rx, __MaybeMutable>>
+  final public function getAttributesNamespaced(): dict<arraykey, varray<mixed>>;
   public function getAttribute(string $name): ?varray<mixed>;
 }
 
--- a/hphp/hack/hhi/reflection.hhi
+++ b/hphp/hack/hhi/reflection.hhi
@@ -17,7 +17,7 @@ class Reflection  {
   <<__Rx>>
   public static function getModifierNames($modifiers);
   public static function export(Reflector $reflector, $return = false);
-}Reflection
+}
 
 class ReflectionClass implements Reflector {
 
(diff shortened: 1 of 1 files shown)


lexidor:
Whoops, markdown doesn't work in a commit message. Let me copy it here then:

I could not find the PHP class for ReflectionProperty, so I was afraid I couldn't do this.
However, there is a [C++ implementation](https://github.com/facebook/hhvm/blob/ba4219c5a9e1f8e8386711af729b004b134839ba/hphp/runtime/ext/reflection/ext_reflection.cpp#L1969) of this function.
It constructs a *dict* instead of a darray.
I was surprised by that, since all its friends construct darray's.

lexidor:
There may be many more, but I will need some extra knowledge on how to read (and search for) PHP classes that are written in C++. If there is some documentation that is reasonably accurate, I would love to know.

fredemmott:
IIRC basically everything except `getAttributeClass`:
- is discouraged in favor of getAttributeClass on general principle of preferring typed user attributes
- have some pretty bad and subtle bugs (maybe inheritance or trait related? Not sure) which we only realized when looking at migrating to getAttributeClass(), and we didn't investigate fixing them separately given they're OK in `getAttributeClass`

It looks like @vassilmladenov probably was the most involved with it, and I think @oulgen might also have been?

vassilmladenov:
`getAttributesNamespaced` is a rename I did when I changed the compiler to add namespaces to attributes in a file. The original was `getAttributes`, so I re-created it to call the builtin and strip off the namespace to maintain compatibility for legacy code. 

Right now, the attributes are carried as strings in the HHAS and the untyped builtins return arrays. I’d strongly prefer not to expose the HHVM builtin directly in HHIs so that we can change the underlying representation. I’d be more in favor of a `getTypedAttributes` function that returns an array of objects. @lexidor could you add an HHI to your Hack project instead of adding it to the ones that ship with Hack?

lexidor:
That would probably be for the better. I too strongly prefer typed data. If we are okay with adding new functions, I like your suggestion. I presume that `getTypedAttributes` will be a `dict<classname<object>, object>` function.

I'll make a helper function in my own projects that uses `getNamespacedAttributes`, but erases the values of the darray (and turns it into a dict). A `dict<classname<object>, null>` function. When `getTypedAttributes` arrives, I can make the switch to that. I really only needed the keys for quickly `C\contains_key`ing from a list of known attributes.

PR #8637, opened 2020-01-02
update FB submodules

jjergus:
It's split into logical commits, please look at each commit separately.

- "regen thrift & mcrouter": probably doesn't need reviewing, the only suspicious thing is that it generated a bunch of large new files so maybe I did something wrong there
- "remove unneeded patches": all of these are noops (already applied upstream)
- "update folly CMake file": adds `fmt` dependency, removes some unused (hopefully?) files that don't compile (the `experimental` ones seem consistent with what we were already removing, the only questionable one is the `compression` one, but I can't find any references to it anywhere (and it compiles and links fine) so it's probably OK
- "add folly+openssl patches": see referenced diffs for discussion
- "add thrift/mcrouter gcc5 compatibility patch": I'll try to get these committed to fbcode so we can remove them later, but I didn't want to wait for that
- "fix mcrouter ALPN patch": we should probably revert D16970597 since this surpasses it; but it probably doesn't make sense to commit this patch to fbcode (in general we probably should't silently skip parts of mcrouter logic based on openssl capabilities -- it's only OK for HHVM where we know that code is unused -- so it should probably stay a HHVM-specific patch)


jjergus:
I haven't updated MySQL because that needs to deal with the `zstd` dependency, which is no longer optional. I'll work on that separately.

jjergus:
Thanks! I'll try to commit this shortly after the next release, so we have more time to deal with any new issues that show up in nightly builds.

ecreeth:
I can see that this project is using this two archived repositories: [FastLZ](https://github.com/ariya/FastLZ), [libafdt](https://github.com/facebookarchive/libafdt) and [squangle](https://github.com/facebookarchive/squangle) found in the archived Facebook repositories, although it is not marked as such.

jjergus:
> I can see that this project is using this two archived repositories

True; is that a problem? I'm not sure if there's anything we can do about that :(

fredemmott:
If they're completely dead, it's a problem; Facebook also archives projects when the projects are 'alive' as such, but there is no external support; for example, Squangle has always been in Facebook Archive, has internal support within Facebook, but does not offer support (or even a build system) for direct usage by non-Facebook C++ projects - but the Hack API built on top of it is supported by the HHVM project.

jjergus:
Bumped everything to `v2020.01.06.00`, removed some patches that had made it to the upstream projects since then.

fredemmott:
still lgtm, thanks :)

PR #8638, opened 2020-01-05
Add support for XHP class modifier

johanoskarsson:
Add support for `xhp class foo` syntax to Hack.

This diff adds support for "XHP classes", which are classes that use the `xhp` keyword. They are intended to be used with XHP namespaces, which will be done in follow-up work.

```
xhp class MyExample extends :x:element {
  protected function render(): XHPRoot {
    return <strong>Hello!</strong>;
  }
}
```

XHP class names are syntactically more similar to normal class names: `foo-bar` and `:foo` (leading colons) are not permitted, although `foo:bar` is allowed.


hphp/hack/src/parser/parser_trait.rs
```diff
@@ -736,6 +741,10 @@ where
         self.lexer_mut().next_xhp_class_name()
     }
 
+    fn next_xhp_class_name_not_qualified(&mut self) -> S::Token {
```
Wilfred: I find the term 'not qualified' a little confusing here. How about something like `next_xhp_modifier_class_name` to make it clear you're accepting the syntax `xhp class Foo...`?

A comment (perhaps on `next_xhp_class_name_no_qualifier` in lexer.rs) with an example of the expected syntax would be super helpful too.

hphp/hack/test/typecheck/xhp_modifier/xhp_modifier_no_colon.php
```diff
@@ -0,0 +1,6 @@
+<?hh // strict
+
+// The colon in the XHP class name is not allowed.
+xhp class :foo {}
+
+xhp class foo:bar {}
```
Wilfred: This is slightly misleading: we're allowing colons in XHP class names, but not leading colons. How about something like this?

```
// XHP classes do no allow leading colons in class names.
xhp class :foo {}

// Colons elsewhere are fine.
xhp class foo:bar {}
```

Wilfred:
Awesome! I've had a full read through of this diff, and it looks great to me!

I've left two style comments if you get a chance to fix them.

Wilfred:
I've proposed the following summary on the internal diff:

---

Add support for `xhp class foo` syntax to Hack.

This diff adds support for "XHP classes", which are classes that use the `xhp` keyword. They are intended to be used with XHP namespaces, which will be done in follow-up work.

```
xhp class MyExample extends :x:element {
  protected function render(): XHPRoot {
    return <strong>Hello!</strong>;
  }
}
```

XHP class names are syntactically more similar to normal class names: `foo-bar` and `:foo` (leading colons) are not permitted, although `foo:bar` is allowed.

---

I think that's a good summary, but let me know if you think it's unclear/missing anything.

Writing this summary also made me think about attributes. We should probably have a test that the following is accepted:

```
xhp class foo {
    attribute int userid @required;
}
```

fredemmott:
```
diff --git a/fbcode/hphp/hack/src/hhbc/closure_convert.rs b/fbcode/hphp/hack/src/hhbc/closure_convert.rs
--- a/fbcode/hphp/hack/src/hhbc/closure_convert.rs
+++ b/fbcode/hphp/hack/src/hhbc/closure_convert.rs
@@ -395,6 +395,7 @@
         file_attributes: vec![],
         final_: false,
         is_xhp: false,
+        has_xhp_keyword: false,
         kind: ClassKind::Cnormal,
         name: Id(p.clone(), make_closure_name(env, st)),
         tparams: class_tparams,
diff --git a/fbcode/hphp/hack/src/options/parserOptions.ml b/fbcode/hphp/hack/src/options/parserOptions.ml
--- a/fbcode/hphp/hack/src/options/parserOptions.ml
+++ b/fbcode/hphp/hack/src/options/parserOptions.ml
@@ -139,17 +139,7 @@

 (* Changes here need to be synchronized with rust_parser_errors_ffi.rs *)
 type ffi_t =
-  bool
-  * bool
-  * bool
-  * bool
-  * bool
-  * bool
-  * bool
-  * bool
-  * bool
-  * bool
-  * bool
+  bool * bool * bool * bool * bool * bool * bool * bool * bool * bool * bool
```

Additionally, please confirm behavior and add tests for these - IMO they should all fail with the option off, and all pass wit hthe option on:

- `xhp class xhp {}`
- `children (xhp)`
- `children (foo)`
- `children (foo:bar)`
- `children (:foo:bar)` < this one should always pass

fredemmott:
Additionally, we have a non-XHP class called `xhp` containing related helpers; this should continue to work, and be in the unit tests - including as a parameter/return type

johanoskarsson:
This latest version should have addressed the comments above:
* Renamed functions as requested
* Clarified comments
* Applied the diff from fred re formatting and another file I didn't have

Few more things we talked about on Slack
* Fixed it so you can name a class `xhp` still so hopefully the FB internal helper still works
* Added the suggested tests for children() and in the case of the flag being off banned children(foo) as mentioned.

Wilfred: that summary looks good, thanks!

See more of the discussion here for background https://hacklang.slack.com/archives/CAYUS1HUZ/p1578516283001000

fredemmott:
There's a few small issues with this and internal failures; these are being resolved internally. No action needed on this PR at present.

PR #8645, opened 2020-01-17
Add support for XHP namespaces

johanoskarsson:
This contains the second half of the split #8595 PR. It adds namespace support for XHP. Note that this is only supported if the xhp class has been defined using the xhp modifier style. For example `xhp class name {}`.

Examples of supported syntax:
<:foo:bar> finds the xhp class \foo\bar
Use namespace
Use namespace as
Use type
Use type group

Currently not supported in this PR:
Declaring a class inside of a namespace using the xhp syntax. For example: `xhp class foo:bar {}`
I've struggled to get this to work. Will try again in a follow up PR when there are fewer moving parts to take care of. I think that's free standing enough that it's not a requirement to for this PR (that is already getting quite big).

See a more detailed list of syntax here: https://gist.github.com/fredemmott/de7eeb91765b0c98fafac9c591263aca

The meat of this PR is in `namespaces.ml` where we change an incoming :foo:bar into \foo\bar to match regular namespaces. `parser_trait.rs` contains the other big chunk where we change xhp elements to allow a colon prefix. This is to allow `<:foo:bar>` to be looked up as `\foo\bar` (as opposed to `<foo:bar>` that will look in `currentnamespace\foo\bar`.

Most of this functionality is hidden behind the flag `disable_xhp_element_mangling`. Perhaps that should be called enable_xhp_namespaces instead since it's not truly just about mangling anymore.

This also contains the following bug fix PR but that one should probably go in sooner rather than later:
#8644


johanoskarsson:
Ran into a merge conflict, rebased and now something broke. Will investigate.

fredemmott:
> Declaring a class inside of a namespace using the xhp syntax. For example: xhp class foo:bar {}

Does `namespace Foo; xhp class bar extends :x:element {}` work (with no `:` in the defined class name) work?

fredemmott:
Failures so far are need for changes to the FB-only buck files; not got useful signal on this yet, waiting for a mac job to finish - which uses the open source build system and runs hacks' tests

fredemmott:
Linters:
- hphp/hack/src/oxidized/gen/*.rs have invalid checksums - perhaps you ran rustfmt on them after generating?
- xhp_modifier/ test files contain tab literals - they should be two spaces instead

fredemmott:
These changes are also needed:

```
diff --git a/fbcode/hphp/hack/src/hh_compile.rs b/fbcode/hphp/hack/src/hh_compile.rs
--- a/fbcode/hphp/hack/src/hh_compile.rs
+++ b/fbcode/hphp/hack/src/hh_compile.rs
@@ -92,6 +92,7 @@
             const_uses: SMap::new(),
             name: None,
             auto_ns_map: vec![],
+            disable_xhp_element_mangling: false,
         },
         config_jsons: vec![],
         config_list: vec![],
diff --git a/fbcode/hphp/hack/src/hhbc/closure_convert.rs b/fbcode/hphp/hack/src/hhbc/closure_convert.rs
--- a/fbcode/hphp/hack/src/hhbc/closure_convert.rs
+++ b/fbcode/hphp/hack/src/hhbc/closure_convert.rs
@@ -805,7 +805,7 @@
 }

 pub fn convert_toplevel_prog(e: &mut Emitter, defs: &mut Program) {
-    let empty_namespace = namespace_env::Env::empty(vec![], false);
+    let empty_namespace = namespace_env::Env::empty(vec![], false, false);
     if e.options()
         .hack_compiler_flags
         .contains(CompilerFlags::CONSTANT_FOLDING)
```

FB-only changes: P125318614

fredemmott:
I've updated this internally with those fixes to get a full test run; I've not fixed the lint issues I mentioned

fredemmott:
Failing public tests:
- hphp/hack/test/full_fidelity:full_fidelity_sexp_test
- hphp/hack/test/nast:nast
- hphp/hack/test/tast:tast

These mostly need re-recording by the looks of things

johanoskarsson:
Updated this to fix `namespace Foo; xhp class bar extends :x:element {}` to work and added test for it.
Fixed the tests to have spaces instead of tabs.

Two remaining known issues (as discussed via Slack):
* Regen.sh doesn't seem to work on master, something due to the new typing_defs_core file from https://github.com/facebook/hhvm/commit/c56edbbe409ae4e00659e5323826da4969d6edcd
* Missing dune build files in master so I didn't see those test failures. I'll try to run them manually and push the fixes

fredemmott:
- hphp‎/hack‎/test‎/typecheck‎/xhp_modifier‎/xhp_helper_class.php‎ contains tabs
- hphp‎/hack‎/src‎/parser‎/lowerer.rs‎ needs rustfmt
- oxidized checksums are invalid; shoudl be fixed by a rebase and regen

Fixing these locally, please including them if other changes are needed, otherwise leave as-is :)

fredemmott:
Public test fixes: https://gist.github.com/fredemmott/4b2cfcd0dc6e82ce3b21f1a9b4c1944a

fredemmott:
FB-only test fixes: P125416325

fredemmott:
FB update steps:
1. re-import; wait for bundle to be regenerated by bot
2. apply to devserver
3. rebase
4. hphp/hack/src/oxidized/regen.sh (hopefully unneeded after next rebase/update of this PR)
5. `pastry P125318614 | patch -p2` - fb-only build fixes
6. `pastry P125416325 | patch -p2` - fb-only test fixes
7. submit changes

hphp/hack/src/parser/core/parser_trait.rs
```diff
@@ -421,6 +425,13 @@ where
                     name_opt = Some(token);
                     has_backslash = false;
                 }
+                (false, TokenKind::XHPClassName) => {
+                    // found a xhp class name, recurse to look for backslash
```
Wilfred: Could we clarify here if we've seen `xhp class foo` or `class :foo`?

hphp/hack/test/typecheck/xhp_modifier/xhp_xhp_name.php.exp
```diff
@@ -1 +1,2 @@
-No errors
+File "xhp_xhp_name.php", line 22, characters 13-19:
+Unbound name: foo\bar (an object type) (Naming[2049])
```
Wilfred: Why is this now an error? Looking at the program it's exercising:

```
xhp class foo:bar {}

xhp class foobarchildren {
  children (foo:bar)*;
}
```

hphp/hack/src/options/globalOptions.mli
```diff
@@ -224,6 +224,12 @@ type t = {
    * New style: xhp class name {}
    *)
   po_enable_xhp_class_modifier: bool;
+  (*
+   * Flag to disable the old stype xhp element mangling. <something/> would otherwise be resolved as xhp_something
+   * The new style xhp class something {}; does not do this style of mangling, thus we need a way to disable it on the
+   * 'lookup side'.
```
Wilfred: This is hard to read without code delimiters. How about something like this?

```
   * Flag to disable the old stype xhp element mangling. `<something/>` would otherwise be resolved as `xhp_something`
   * The new style `xhp class something {}` does not do this style of mangling, thus we need a way to disable it on the
```

hphp/hack/test/typecheck/xhp_modifier/xhp_xhp_name.php.exp
johanoskarsson: Hmm yeah this is a bit odd. The spec does allow for `xhp class foo:bar {}` which will basically be the same as:
```
namespace foo {
  xhp class bar {}
}
```
However this is not yet implemented. I attempted to add it to this PR but I ran into some issues. Since it's already quite big I think it would be better to not grow this one any further and instead take it as a follow on task. So this test does (correctly) fail, but will pass once that is implemented. For now what I'll do is remove that specific class and re-add it in the follow up.

fredemmott:
```
namespace foo {
  xhp class bar {}
}
```

not quite - it would be the same as:

```
xhp class foo\bar {
/ * with no change to resolution here; no new scope for `use` statements etc */
}
```

In past versions of hackc, `class foo\bar {}` was accidentally supported, with this behavior

johanoskarsson: Ah yes, I always get this wrong. Thanks for the clarification.

hphp/hack/test/typecheck/xhp_modifier/xhp_xhp_name.php
```diff
@@ -8,16 +8,10 @@
 
 xhp class foo {}
 
-xhp class foo:bar {}
```
fredemmott: I'd like to have another PR up resolving this before we take this in: we're not going to be able to start migration. We're not going to be able to start any migration until `xhp class foo:bar` is valid

What class does the typechecker consider to be defined here?
What class does hh_single_compile emit?

Wilfred: Wait, so we're going to support all of:

```
xhp class foo\bar {}
xhp class foo:bar {}
```

What's the benefit of changing `class :foo:bar` to `xhp class foo:bar`? Presumably there's an incremental migration flow you have in mind?

fredemmott: `:` needs to be explicitly a namespace separator, not a sigil as well

at least one out of `xhp class foo:bar` and `xhp class foo\bar` is wanted to make it so we don't need to namespace everything

fredemmott:
```
namespace MyNS {

class :foo:bar {}

$_  = <:foo:bar />; // this doesn't refer to the class above

}
```

fredemmott: `xhp class foo\bar {}` is not supported  - this need to be `xhp class foo:bar`

fredemmott:
The PR as-is has emitter issues with this test file:

```hack
namespace {
class IExistInTheRootNamespace {}

final xhp class foo:bar {
  public function __construct(mixed ...$args) {
    \var_dump(vec[__LINE__, new IExistInTheRootNamespace()]);
    \var_dump(vec[__LINE__, $this, \get_class($this)]);
  }
}
} // namespace {}
namespace HerpDerp {

final xhp class foo:bar {
  public function __construct(mixed ...$args) {
    \var_dump(vec[__LINE__, new \IExistInTheRootNamespace()]);
    \var_dump(vec[__LINE__, $this, \get_class($this)]);
  }
}

<<__EntryPoint>>
function test_instantiation(): void {
 $_ = <foo:bar />;
 $_ = <:foo:bar />;
}

} // namespace HerpDerp {}
```

https://gist.github.com/johanoskarsson/e8cf425531a3e769d0198252789bf5b7 and https://gist.github.com/fredemmott/a0ab8b772adc2fcfdd79c2fc864c09f6 both resolve this in different ways; I'm investigating if the complexity of mine is required

hphp/test/quick/xhp_modifier/config.ini
```diff
@@ -0,0 +1,45 @@
...
+hhvm.scribe.tables.hhvm_rfh.hive = false
+hhvm.scribe.tables.hhvm_serdes.hive = false
+hhvm.scribe.tables.hhvm_swayze.hive = false
+hhvm.scribe.tables.hhvm_tc_branches.hive = false
+hhvm.scribe.tables.hhvm_tc_func_sizes.hive = false
+hhvm.scribe.tables.hhvm_tzdata.hive = false
+hhvm.scribe.tables.hhvm_unit_cache.hive = false
+hhvm.scribe.tables.webserver_shutdown_timing.hive = false
+hhvm.hack.lang.disable_xhp_element_mangling=true
+hhvm.hack.lang.enable_xhp_class_modifier=true
```
Wilfred: Whitespace pedantry: please use spaces around `=` for consistency.

hphp/hack/src/hh_single_type_check.ml
```diff
@@ -506,6 +507,10 @@ let parse_options () =
       ( "--disallow-discarded-nullable-awaitables",
         Arg.Set disallow_discarded_nullable_awaitables,
         "Error on using discarded nullable awaitables" );
+      ( "--disable-xhp-element-mangling",
+        Arg.Set disable_xhp_element_mangling,
+        "Disable mangling of XHP elements :foo. This is necessary for classes using xhp class modifiers."
```
Wilfred: I'm not sure this is a great comment. hh_single_type_check doesn't run code, so it's not immediately clear what it means to change mangling.

I suppose it means whether or not `<foo/>` maps to `class :foo` or `xhp class foo`.

How about something like this? "Disable mangling of XHP class names :foo. This enables use of `xhp class foo` in XHP expressions `<foo/>`".

fredemmott: Agreed on the lack of clarity, though this doesn't enable it: that's `--enable-xhp-class-modifier`

this makes it so that `:foo:bar` is `\foo\bar`, not `xhp_foo__bar`, and bans `-` in XHP class names

Wilfred:
This is looking really good! You've done a huge amount of plumbing too. (Given there's so much boilerplate and resulting .exp file changes, perhaps we should split that into a separate commit or PR so the logic changes are distinct?)

It's great to see an HHVM test too, that's definitely appropriate here. You'll need a .exp file to say what you expect to happen when the test runs (i.e. the stdout written).

fredemmott:
It’s probably worth noting that this doesn’t change how “:foo:bar” is lexed - we still have an XHP Name token, but we’re changing how that maps to an actual type.

We might want to change that eventually - but I strongly feel that this is the right approach while we have these options

PR #8648, opened 2020-01-19
Move libmagic.patch to /patches

ecreeth:
This PR move `libmagic.patch` to `/patches`. This file was created 6 years ago.
The intention is to move this patch to the directory where they are stored. This will make patches taken into account and make it easier to identify.


ecreeth:
> All files in patches/ are automatically applied by the build system, and all paths must exist relative to the root of the repository; this change would break the build as $HHVM_REPO_ROOT/libmagic/ does not exist

PR #8652, opened 2020-01-29
HSL IO: add lower-level resource-free native functions

fredemmott:
This adds:
- `HH\Lib\OS\FileDescriptor`, intended to be an opaque type, without observable refcounting behavior
- `HH\Lib\OS\O_*` constants
- `HH\Lib\OS\_Private\ErrnoException`
- `HH\Lib\OS\_Private\{open,write,close}()` functions, intended to be an extremely thin wrapper around POSIX functions

I plan to implement `HH\Lib\OS\{open,write,close}()` in the userspace HSL library; these functions will be something like:

```hack
namespace HH\Lib\OS;
function open(string $path, int flags, ?int $mode_if_o_create = null): FileDescriptor {
  try {
    return _Private\open($path, $flags, $mode_if_o_create);
  } catch (_Private\ErrnoException $e) {
    throw_user_friendly_exception($e);
  }
}
```

HSL IO will continue to be a substantially higher-level, async-first API, but will be rebuilt on top of HH\Lib\OS, instead of PHP's resource-based functions.

I expect HH\Lib\OS to contain many POSIX functions in the future - not just those needed by HSL IO

refs hhvm/hsl-experimental#77

### DIFF
--- a/hphp/hack/hhi/hsl/ext_hsl_os.hhi
+++ b/hphp/hack/hhi/hsl/ext_hsl_os.hhi
@@ -1,0 +1,27 @@
+<?hh
+/**
+ * Copyright (c) 2018, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the "hack" directory of this source tree.
+ *
+ */
+
+namespace HH\Lib\OS;
+
+newtype FileDescriptor = mixed;
+
+const int O_RDONLY = 0;
+const int O_WRONLY = 0;
+const int O_RDWR= 0;
+const int O_NONBLOCK = 0;
+const int O_APPEND = 0;
+const int O_CREAT/*E < you dropped this*/ = 0;
+const int O_TRUNC = 0;
+const int O_EXCL = 0;
+const int O_SHLOCK = 0;
+const int O_EXLOCK = 0;
+const int O_NOFOLLOW = 0;
+const int O_SYMLINK = 0;
+const int O_CLOEXEC = 0;
--- a/hphp/runtime/ext/hsl/config.cmake
+++ b/hphp/runtime/ext/hsl/config.cmake
@@ -5,6 +5,13 @@ HHVM_DEFINE_EXTENSION("hsl_io"
     ext_hsl_io.php
 )
 
+HHVM_DEFINE_EXTENSION("hsl_os"
+  SOURCES
+    ext_hsl_os.cpp
+  SYSTEMLIB
+    ext_hsl_os.php
+)
+
 HHVM_DEFINE_EXTENSION("hsl_random"
   SOURCES
     ext_hsl_random.cpp
--- a/hphp/runtime/ext/hsl/ext_hsl_os.cpp
+++ b/hphp/runtime/ext/hsl/ext_hsl_os.cpp
@@ -1,0 +1,179 @@
+/*
+   +----------------------------------------------------------------------+
+   | HipHop for PHP                                                       |
+   +----------------------------------------------------------------------+
+   | Copyright (c) 2010-present Facebook, Inc. (http://www.facebook.com)  |
+   +----------------------------------------------------------------------+
+   | This source file is subject to version 3.01 of the PHP license,      |
+   | that is bundled with this package in the file LICENSE, and is        |
+   | available through the world-wide-web at the following url:           |
+   | http://www.php.net/license/3_01.txt                                  |
+   | If you did not receive a copy of the PHP license and are unable to   |
+   | obtain it through the world-wide-web, please send a note to          |
+   | license@php.net so we can mail you a copy immediately.               |
+   +----------------------------------------------------------------------+
+*/
+
+#include "hphp/runtime/base/array-init.h"
+#include "hphp/runtime/base/builtin-functions.h"
+#include "hphp/runtime/ext/extension.h"
+#include "hphp/runtime/vm/native.h"
+#include "hphp/runtime/vm/native-data.h"
+#include "hphp/system/systemlib.h"
+
+#include <fcntl.h>
+#include <stdio.h>
+
+#define THROW_ERRNO_IF_MINUS_ONE(var) \
+  if ((var) == -1) { throw_errno_exception(errno); }
+
+namespace HPHP {
+namespace {
+
+const StaticString s_HSLFileDescriptor("HSLFileDescriptor");
+const StaticString s_fd("fd");
+const StaticString s_valid("valid");
+const StaticString s_source("source");
+const StaticString s_ErrnoException("HH\\Lib\\_Private\\_OS\\ErrnoException");
+
+IMPLEMENT_REQUEST_LOCAL(std::set<int>, s_fds_to_close);
+
+[[noreturn]]
+void throw_errno_exception(int number) {
+  Array params;
+  params.append(folly::sformat("Errno {}: {}", number, ::strerror(number)));
+  params.append(number);
+  throw_object(s_ErrnoException, params);
+}
+
+} // namespace
+
+struct HSLFileDescriptor {
+  static Object newInstance(int fd, const String& source) {
+    if (s_class == nullptr) {
+      s_class = Unit::lookupClass(s_classname.get());
+    }
+    assertx(s_class);
+    Object obj { s_class };
+
+    auto* data = Native::data<HSLFileDescriptor>(obj);
+    data->m_fd = fd;
+    data->m_valid = true;
+    data->m_source = source;
+
+    s_fds_to_close->insert(fd);
+    return obj;
+  }
+
+  int fd() const {
+    if (!m_valid) {
+      throw_errno_exception(EBADF);
+    }
+    return m_fd;
+  }
+
+  void close() {
+    if (!m_valid) {
+      throw_errno_exception(EBADF);
+    }
+    int result = ::close(m_fd);
+    THROW_ERRNO_IF_MINUS_ONE(result);
+    s_fds_to_close->erase(m_fd);
+    m_valid = false;
+  }
+
+  Array __debugInfo() const {
+    Array ret = Array::CreateDArray();
+    ret.set(s_fd, m_fd);
+    ret.set(s_valid, m_valid);
+    ret.set(s_source, m_source);
+    return ret;
+  }
+
+  private:
+    static Class* s_class;
+    static StaticString s_classname;
+    int m_fd;
+    bool m_valid;
+    String m_source;
+};
+
+Class* HSLFileDescriptor::s_class = nullptr;
+StaticString HSLFileDescriptor::s_classname("HH\\Lib\\OS\\FileDescriptor");
+
+Array HHVM_METHOD(HSLFileDescriptor, __debugInfo) {
+  return Native::data<HSLFileDescriptor>(this_)->__debugInfo();
+}
+
+namespace {
+
+Object HHVM_FUNCTION(HSL_os_open, const String& path, int64_t flags, const Variant& mode) {
+  int fd;
+  if (flags & O_CREAT) {
+    // TODO: throw ??? if mode is not an int
+    fd = ::open(path.c_str(), flags, mode.asInt64Val());
+  } else {
+    fd = ::open(path.c_str(), flags);
+  }
+  THROW_ERRNO_IF_MINUS_ONE(fd);
+  if (flags & O_CREAT) {
+    return HSLFileDescriptor::newInstance(fd, folly::sformat("open('{}', {}, {})", path.c_str(), flags, mode.asInt64Val()));
+  }
+  return HSLFileDescriptor::newInstance(fd, folly::sformat("open('{}', {})", path.c_str(), flags));
+}
+
+int64_t HHVM_FUNCTION(HSL_os_write, const Object& obj, const String& data) {
+  auto fd = Native::data<HSLFileDescriptor>(obj)->fd();
+  ssize_t written = ::write(fd, data.data(), data.length());
+  THROW_ERRNO_IF_MINUS_ONE(written);
+  return written;
+}
+
+void HHVM_FUNCTION(HSL_os_close, const Object& obj) {
+   Native::data<HSLFileDescriptor>(obj)->close();
+}
+
+struct OSExtension final : Extension {
+
+  OSExtension() : Extension("hsl_os", "0.1") {}
+
+  void moduleInit() override {
+    Native::registerNativeDataInfo<HSLFileDescriptor>(s_HSLFileDescriptor.get());
+
+    // Remember to update the HHI :)
+    // open() flags ----------
+    // The preprocessor doesn't like "\" immediately before a ##
+#define O_(name) HHVM_RC_INT(HH\\Lib\\OS\\O_##name, O_##name)
+    O_(RDONLY);
+    O_(WRONLY);
+    O_(RDWR);
+    O_(NONBLOCK);
+    O_(APPEND);
+    O_(CREAT);
+    O_(TRUNC);
+    O_(EXCL);
+    O_(NOFOLLOW);
+    O_(CLOEXEC);
+#undef O_
+    // MacOS: O_EVTONLY, O_SHLOCK, O_EXLOCK, O_SYMLINK
+    // Linux: ... lots ...
+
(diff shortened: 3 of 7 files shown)


fredemmott:
One key question is how HSL IO should be given access to things that don't have an FD - for example:
- POST data and response body for HTTP requests
- SSL sockets
- curl responses.

Some approaches are (in order of my personal preference):
1. create new native classes similar to [ReadHandle](https://github.com/hhvm/hsl-experimental/blob/bd4422de0c578a5d3f7210157b10a642edc140ce/src/io/ReadHandle.php), [WriteHandle](https://github.com/hhvm/hsl-experimental/blob/bd4422de0c578a5d3f7210157b10a642edc140ce/src/io/WriteHandle.php), [SeekableHandle](https://github.com/hhvm/hsl-experimental/blob/bd4422de0c578a5d3f7210157b10a642edc140ce/src/io/SeekableHandle.php), and implement them for these 'stream-like' things
2. expose the underlying C/C++ APIs for each implementation.
3. use pipe() to expose an FD interface
4. create fake FD native objects and make `OS\write()` etc handle them.

I believe that (2) will be unavoidable, but should be used in combination with (1)

## 1. Native classes

These should not be clones of the HSL IO interfaces; they should provide the bare minimum functionality. For example, a readable class should provide rawRead (non-async), and if non-blocking, a way to await it. Higher-level functions (e.g. `readLineAsync()`, combining an await and potentially multiple raw reads) should be left for the HSL.

### Why not do this for FDs too?

There are *a lot* of quirks and non-portable behavior with posix async IO; a native implementation would need to directly provide the full async implementation, or the FDs and HSL IO would have to use a mix. Treating "has FD" vs "no FD" completely seems cleaner than trying to mix them

## 2. pipe()

Conceptually clean: allows an FD to be created for any non-FD API. I'm worried about:
- performance
- it feels like a potential bug magnet

## 3. expose the underlying C/C++ calls

This would be doable for most forms, e.g. openssl; I'm not sure if it's viable for POST data/response body, especially e.g. in fastcgi mode.

## 4. create fake HD objects and make `OS\write()` and friends work with them

This feels like it's going to get us back to the complexity of the HPHP::File object and hierarchy; additionally, /some/ posix calls will work on these fake FDs, but many won't. I don't think we'll end up with a clear, user-understandable definition - it'll be "whatever HSL IO neeeds"

fredemmott:
Overall structure:

- `HH\Lib\_Private\OS` is meant for internal HSL use only, and contains native functions, such as `_Private\OS\write()`; these may throw a `_Private\OS\ErrnoException` (or `HerrorException` etc)
- `HH\Lib\OS\write()` would be a thin, Hack wrapper around this (see below), and part of the Hack Standard Library. Strongly discourage use when a higher-level alternative exists
- `HH\Lib\IO` is a higher-level IO library; for actual files, pipes, sockets etc, this would be built on top of `HH\Lib\OS`

Wrappers:

```Hack
namespace HH\Lib\OS;

function write(FileDescriptor $fd, string $bytes): int {
  try {
    return \HH\Lib\_Private\OS\write($fd, $bytes);
  } catch (\HH\Lib\_Private\OS\ErrnoException $e) {
   // actually in a helper function shared by all OS\ functions
    switch ($e->getErrno()) {
      case ENOENT:
        throw new HH\Lib\OS\NotFoundException();
      // ...
      default:
        throw new HH\Lib\OS\Exception($e->getErrno());
    }
  }
}
```

- All `OS\Exception`s will provide an error or herrno
- `OS\FileNotFoundException` extends `OS\Exception`
- this is largely based on Python3's syscall exceptions (OSError and subclasses)

fredemmott:
# Should `FILE*` be used instead of int FDs?

I don't believe so:
- FDs are more generic - native implementation of sockets etc
- we need FDs for native async on these
- given HSL IO /always/ uses non-blocking operations, the caching etc provided by `fread(FILE*)` and friends over `read(int fd)` is less useful

wez:
> # Should `FILE*` be used instead of int FDs?

This may not be relevant to hhvm's target systems, but additional reasons against `FILE*`:

* In Solaris-derived libc, `FILE*` uses a `uint8` to hold the file descriptor, which poses some problems for anything beyond trivial toy use cases (this is the primary reason why PHP doesn't use `FILE*`)
* In MS libc, `FILE*` layers on top of its awful "posix" fd layer, so you have at least double the trouble.  For Windows I would recommend not using its "posix" style APIs and instead tracking the underlying system HANDLE or SOCKET type (SOCKET is really a HANDLE, it just has a different type in the C headers).

# Polling vs Completion

Do you (eventually?) care about Windows?  Some things to consider:

* Windows SOCKET handles are non-blocking compatible (provided that you use the winsock functions on them instead of plain `ReadFile` etc.!), but no other HANDLEs are.  When considering exposing the underlying "fd" for polling purposes you may want keep that in mind; the "fd" returned by the mslibc is totally useless for this because it has nothing at all to do with the kernel.  In terms of API, I would suggest naming the method that exposes this something like `get_pollable_descriptor` rather than simply `get_fd` to make this distinction clearer.
* A number of HANDLE types can be use with IO Completion ports for completion based IO.  `iouring` is a newish linux feature that allows linux to efficiently perform completion based IO.  The async/await syntax exposed to the user in most languages is completion based.  There may be some merit in thinking a little bit about whether or when hhvm wants to evaluate integrating that (note that folly's EventBase class recently grew support for this), and how that might influence some of this design now... which may legitimately be: don't care right now.

wez:
re: pipes, rather than a `pipe(2)`, I'd recommend using `socketpair(2)`.  pipes have slightly different semantics and are not compatible with non-blocking IO on Windows.

While windows doesn't offer a `socketpair` function, you can implement one fairly easily by connecting to yourself.  Here's one of mine in rust:
https://github.com/wez/wezterm/blob/master/filedescriptor/src/windows.rs#L433-L499

(I'm not advocating using pipes as a solution, but if you did choose them, I would recommend `socketpair` instead)

wez:
I'm a bit out of touch with the current state of hhvm and your hsl work, but based on my general frustrations with windows vs. posix, I would suggest the following.  Keep in mind that I don't have all the context that you have:

* I would probably put `OS\open` in a `POSIX\open` namespace because that function really is posix specific
* Consider also placing `FileDescriptor` in the posix namespace
* Have it implement the `ReadHandle`, `WriteHandle`, `SeekHandle` interfaces
* Have the standard library operate in terms of things that implement those interfaces
* Consider introducing a way to expose a pollable descriptor as mentioned above (not sure if that is too low level for an interface, but perhaps it could simply be that)
* You may want to expose the difference between file descriptor and socket as different types; eg: `NET\SocketDescriptor` (you likely have more thoughts on better naming!) that looks similar to `POSIX\FileDescriptor` on posix systems, but that goes directly to the winsock functions on Windows.  Implement `ReadHandle` and `WriteHandle` for this type.  Bonus: because this doesn't implement `SeekHandle`, attempting to seek one of these might now be catchable via type checking
* As a general solution for performing IO for things that don't offer a pollable descriptor, offer a way for those streamy things to perform the underlying blocking operation in another thread.  I'm not sure how difficult this is in the hhvm architecture today; eg: it might require a little bit of native code to be glued in.

fredemmott:
Thanks @wez; I think that a bunch of these need splitting in two: "What's best for HSL IO?" and "What's best for Hack/POSIX?"

> Do you (eventually?) care about Windows? Some things to consider:

Yep - but I think the posix functions should consistently return FileDescriptors; if winsock means that sockets need to be different types, then we probably need separate WinSock functions.

As far as HSL IO goes ,this should all be hidden - on Linux/Mac, a TCP\Socket should be a bunch of wrappers around a OS\FileDescriptor, on Windows, it would act 'the same' but be a bunch of wrappers around a WinSock\Handle or something like that.

> Polling, completion

Hack/POSIX: expose as-is

HSL IO:

In general, we're using libevent's support for waiting on an FD, via folly.

If folly has existing support for 'use posix async IO on most platforms, `iouring` on linux' then we might use that  - as long as we can get at the underlying FD. Hack should eventually be a general purpose language, and being able to do all the obscure stuff rather than 'just' the 90% stuff is important.

> pipes/socketpair

Hack/POSIX: provide pipe() where available, no substitute on other platforms.

HSL IO user API: currently has pipes because I need to add /something/ (stream_socket_pair() resources don't work with all the functions i'd like it to be) and it had the simpler API :) not stuck to this. perhaps both.Not a factor when all reimplemented natively so may change.

HSL IO implementation for non-FDs: no opinion, will definitely consider socketpairs - thanks :)
I'm a bit out of touch with the current state of hhvm and your hsl work, but based on my general frustrations with windows vs. posix, I would suggest the following.  Keep in mind that I don't have all the context that you have:

> * I would probably put `OS\open` in a `POSIX\open` namespace because that function really is posix specific
> * Consider also placing `FileDescriptor` in the posix namespace

I'd like this; I'll be aiming to rename before release, but we might be stuck with OS\ or similar for non-technical reasons.

>  Have it implement the `ReadHandle`, `WriteHandle`, `SeekHandle` interfaces

I'm not set against this, but not currently doing it as:
- I like it being treated as an opaque handle, which is largely how it should be treated in C
- it would **require** that HSL IO be fully implemented as HHVM builtins, rather than just the low-level native functions

> Have the standard library operate in terms of things that implement those interfaces

HSL IO currently does not expose or accept the raw resource; it's an implementation detail.

I'd prefer for this to be the common case, but once it's no longer dealing with `resource`s, this should be a leaky abstraction to allow using the POSIX stuff without having to fully reinvent HSL IO

> Consider introducing a way to expose a pollable descriptor as mentioned above (not sure if that is too low level for an interface, but perhaps it could simply be that)

Definitely going to expose a way to async poll on a file descriptor (like `\stream_await()` now)

Where something is implemented as a builtin rather than an FD (eg POST data), I was thinking of exposing `pollAsync(READ)` instead:
- allows use of conditionwaithandles for callback-based things with no real FD (maybe curl historically, but I think there's an FD api now)
- can trivially back on to libevent/folly_eventbase for FD-based stuff

> * You may want to expose the difference between file descriptor and socket as different types; eg: `NET\SocketDescriptor` (you likely have more thoughts on better naming!) that looks similar to `POSIX\FileDescriptor` on posix systems, but that goes directly to the winsock functions on Windows.  Implement `ReadHandle` and `WriteHandle` for this type.  Bonus: because this doesn't implement `SeekHandle`, attempting to seek one of these might now be catchable via type checking

Hack/POSIX: I want to match the native types
HSL IO: definitely - additionally, read/write is represented in the type system. E.g. there is TCP\ReadOnlySocket, File\ReadOnlyHandle, etc

> * As a general solution for performing IO for things that don't offer a pollable descriptor, offer a way for those streamy things to perform the underlying blocking operation in another thread.  I'm not sure how difficult this is in the hhvm architecture today; eg: it might require a little bit of native code to be glued in.

We do this for curl at present; it's pretty ugly, has performance issues, and I'm not sure if it's needed for curl any more.

Perhaps C++ coroutines might enable this, or we could expose a folly::semifuture-based internal API

fredemmott:
@javierhonduco raised some concerns re interactions between file descriptors from `open` and the garbage collector in Python; I don't think it's an issue here, and I need to clarify the behavior of the `FileDescriptor` class:

Destroying a `FileDescriptor` does not `close()` the FD. FDs are only closed by an explicit call to `OS\close()`, or at the end of the request. It does not matter if the Hack object is deleted by refcounting or by the garbage collector.

This is an intentional decision to make it so that refcounting/destructors are not observable behavior, which will allow us to explore non-refcounting garbage collectors.

This is implemented by:
- storing a request-local set of FDs that have been opened
- when an FD is explicitly close()ed, it is removed from that set
- at the end of the request, every FD in the list is closed, and the set is cleared

When using HSL IO, use of `Disposable`'s is **strongly** encouraged - in these cases, the `__dispose()` function will explicitly close the FD

fredemmott:
> iouring

My main concern here is reducing the number of native implementations: MacOS support hugely depends on us sharing as much code as possible between MacOS and Linux. If folly provides an abstraction that lets us use this but falls back to POSIX async IO (And ideally some windows thing), that would be ideal, otherwise it's an area of caution

fredemmott:
@javierhonduco  is also concerned that it might be possible for an interrupt to happen in between the kernel freeing resources in a `close()` and it actually returning, which could lead to EINTR.

This situation may be undetectable, which would mean we would need to chose between potentially leaking FDs, and potentially closing a newly-opened fd that re-used the same number

fredemmott:
https://lwn.net/Articles/576478/ appears to describe that

fredemmott:
Things that currently seem clear:
- use fd, not `FILE*`
- we definitely need to support things that don't have an FD in HSL IO; on windows, this includes several things that POSIX provides an FD for
- remove EINTR handling, at least for close() :(

fredemmott:
- deleted EINTR handling :(
- added new exception class
- added HHI

Changed PR description to be suitable for merge; original here:


Goals for HSL IO include:
- implement as much as possible as Hack code, i.e. to minimize the amount of native code.
- do not expose destructor-like behavior (to unblock GC work)
- handle errors in a reliable and machine-friendly way
  - do not use the PHP error handling system
  - do not suppress the PHP error handling system
- avoid common mistakes

I'm aiming to address these by making very thin wrappers to POSIX
functions directly available to Hack code.

- These are in the _Private namespace to allow more behavior to be
  defined in Hack code in the HSL; for example, `OS\open()` might be
  defined as `return convert_exception(() ==>
  _Private\open(...$args));`. The current plan is for the public facing
  API to throw exceptions similar to Python3. This is implemented at
  https://github.com/hhvm/hsl-experimental/blob/bd4422de0c578a5d3f7210157b10a642edc140ce/src/os/_Private/Errno.php#L202
- Constants are not in a private namespace as they do not need wrapping
- Automatically handle EINTR (a.k.a. EAGAIN)
- Automatically throw for any other errno without leaving C++
  - avoids race conditions that exist when calling `posix_get_errno()`
    after the call
  - avoids "Do I want CLI client or CLI server or lightprocess master
    errno?" issues in the same situation
  - will need to be extended to do the same for herror when socket
    function are introduced
- wrap FDs:
  - don't allow an FD to be used after being closed; this also prevents
    hard-to-debug issues where an fd may be reused
  - close all open FDs *at the end of the request*
  - don't allow arbitrary int FDs to be used, avoiding security issues
    when in CLI server mode
  - similar wrapping will be needed for other 'resource'-like handles
- no destructor-based behavior

I'm planning on builting HSL IO on `open`, `write` and friends instead of
`fopen`/`fwrite` etc:
- HSL IO exclusively uses non-blocking operations; buffer is managed in
Hack
- FDs are needed for async polling
- HSL IO also deals with sockets and pipes; this allows consistent
  behavior without the need to introduce a heavier-weight abstraction

Not ready for commit. Proof of concept:

```hack
use namespace HH\Lib\OS;
use namespace HH\Lib\_Private\OS as _OS;
<<__EntryPoint>>
function main(): void {
  $fd = _OS\open('/tmp/myfile', OS\O_CREAT | OS\O_WRONLY, 0644);
  \var_dump($fd);
  _OS\write($fd, "Hello, world\n");
  _OS\close($fd);
  \var_dump(\file_get_contents('/tmp/myfile'));
  \var_dump($fd);
  printf("Expecting exception EBADF\n");
  try {
    _OS\write($fd, "Foo bar\n");
  } catch (\Throwable $e) {
    printf("> Caught exception: %s\n", $e->getMessage());
  }
  printf("Expecting exception EEXIST\n");
  try {
    $fd = _OS\open('/tmp/myfile', OS\O_CREAT | OS\O_EXCL | OS\O_WRONLY, 0644);
  } catch (\Throwable $e) {
    printf("> Caught exception: %s\n", $e->getMessage());
  }
}
```

```
./hhvm ~/iotest.hack
object(HH\Lib\OS\FileDescriptor) (3) {
  ["fd"]=>
  int(13)
  ["valid"]=>
  bool(true)
  ["source"]=>
  string(24) "open('/tmp/myfile', 513)"
}
string(13) "Hello, world
"
object(HH\Lib\OS\FileDescriptor) (3) {
  ["fd"]=>
  int(13)
  ["valid"]=>
  bool(false)
  ["source"]=>
  string(24) "open('/tmp/myfile', 513)"
}
Expecting exception EBADF
> Caught exception: Errno 9 (Bad file descriptor)
Expecting exception EEXIST
> Caught exception: Errno 17 (File exists)
```

PR #8654, opened 2020-01-31
Make it possible for extensions to register new CLI client callbacks

fredemmott:
** review second commit only **

Test Plan:

Start hhvm CLI server as root, then run this as me...

```hack
<<__EntryPoint>>
function main(): void {
  var_dump(getmypid());
  $fda=\HH\Lib\_Private\OS\open('/tmp/mytestfile', HH\Lib\OS\O_CREAT, 0644);
  $fdb=\HH\Lib\_Private\OS\open('/var/root/mytestfile', HH\Lib\OS\O_CREAT, 0644);
}
```

- check the dumped PID is the server PID to make sure we're actually
using the CLI server
- check that /tmp/mytestfile is owned by me, not root
- check that a permission denied exception was thrown for the attempt to
write to root's home dir (/var/root on mac)

### DIFF
--- a/hphp/hack/hhi/hsl/ext_hsl_os.hhi
+++ b/hphp/hack/hhi/hsl/ext_hsl_os.hhi
@@ -1,0 +1,27 @@
+<?hh
+/**
+ * Copyright (c) 2018, Facebook, Inc.
+ * All rights reserved.
+ *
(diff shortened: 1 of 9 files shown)


fredemmott:
Manually imported as D19890878

PR #8663, opened 2020-02-22
update libzip submodule to v1.6.1

jjergus:
The CMakeLists.txt in the new libzip version has flags for everything, so I think we don't need a custom CMakeLists.txt now and we can just use ExternalProject_Add (with the right flags passed in)?

I am not sure how to correctly set it up with target_link_libraries and target_include_directories though.


CMake/HPHPFindLibs.cmake
```diff
@@ -510,7 +510,11 @@ macro(hphp_link target)
   if (LIBZIP_LIBRARY)
     target_link_libraries(${target} ${LIBZIP_LIBRARY})
   else()
-    target_link_libraries(${target} zip_static)
+    add_dependencies(${target} zip_static)
+    target_link_libraries(${target} "${CMAKE_CURRENT_SOURCE_DIR}/third-party/libzip/src/lib/libzip.a")
```
jjergus: This doesn't work (I'm probably using `${CMAKE_CURRENT_SOURCE_DIR}` incorrectly here) but I wanted to see if you had any better suggestions for how to correctly link the library before I spend too much time fixing the exact paths...

fredemmott:
It needs to be CMAKE_CURRENT_BINARY_DIR; Do we need to patch this? If not it might be worth trying to use the externalproject_add URL support instead like we’re doing for some of the more recent deps

fredemmott:
Screen Shot 2020-02-24 at 9 35 27 AM


I'm guessing there's a test build in progress with the new fancy patched build thing? :)

jjergus:
I'm running a test in-source build + out-of-source build. Is it worth testing on all distros or should I just wait for the next nightly build?

fredemmott:
I'd like to see oldest + newest + ideally macos - oldest > newest > macos though

jjergus:
OK I finally got the CMake configs to work (the previous version didn't work because it used a variable before it was set...I think), can you check if there's a better way to do any of it? I know you were planning to clean up our CMake configs at some point and I'm probably making it worse :)

As far as I can tell this should set up the dependency correctly for linking, but I'm not sure about the include paths (there might be a race condition if we try to build something that #includes libzip before libzip?).

PR #8664, opened 2020-02-26
fix libzip build on latest ubuntu/debian

jjergus:
Some systems call cmake with explicitly overridden CMAKE_INSTALL_LIBDIR,
so we have to pass that through.


fredemmott:
`make install` of HHVM should not be installing our libzip, and our lipzip should not be part of the main HHVM package - possibly in the `-dev` package, but not in a standard location; what's being installed/depended on here?

fredemmott:
In particular, if I do `cmake -DCMAKE_INSTALL_PREFIX=/usr; make install`, it should not overwrite my system libzip, and I think this would

jjergus:
We don't actually install libzip, but we "install" it into `${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_INSTALL_LIBDIR}`

This becomes a problem if HHVM's cmake and libzip's cmake disagree on `${CMAKE_INSTALL_LIBDIR}`

fredemmott:
is there a DESTDIR= on the make install?

LIBDIR defaults to PREFIX/lib; it is not relative to prefix, so specdifying an absolute path there should make it go *exactly* there, not relative to the build dir

third-party/libzip/CMakeLists.txt
```diff
@@ -44,6 +44,7 @@ if (NOT LIBZIP_FOUND)
     -DCMAKE_C_FLAGS=${CMAKE_C_FLAGS}
     -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
     -DCMAKE_INSTALL_PREFIX=${LIBZIP_PREFIX}
```
jjergus: ^ here we override the prefix

fredemmott:
> CMAKE_INSTALL_LIBDIR is just the part that gets appended to the prefix (usually "lib" but sometimes things like "lib/x86_64-linux-gnu")


If it's a relative path, yep, but it can also be an absolute path IIRC and according to the docs

jjergus:
https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html says there's CMAKE_INSTALL_FULL_LIBDIR for that

jjergus:
But if it's safer I guess I could pass "lib" as the value instead of passing through the parent cmake's variable. It actually doesn't matter at all what the value is, we just need third-party/libzip/CMakeLists.txt  and libzip's CMakeLists.txt to agree on it.

fredemmott:
AIUI _FULL_LIBDIR is /always/ absolute and generated if needed - _LIBDIR can be either

fredemmott:
yeah, an explict 'lib' would b esafer

PR #8665, opened 2020-02-26
update FB submodules

jjergus:
The part worth reviewing is the last commit.

We were previously (intentionally?) omitting these generated files, but now they are needed because of https://github.com/facebook/proxygen/commit/2144bfb79ca36dd9a2ebef9a01439323575c9640


third-party/proxygen/CMakeLists.txt
```diff
@@ -2,6 +2,38 @@ set(PROXYGEN_LIB_DIR "${CMAKE_CURRENT_SOURCE_DIR}/src/proxygen/lib")
 set(PROXYGEN_HTTPSERVER_DIR "${CMAKE_CURRENT_SOURCE_DIR}/src/proxygen/httpserver")
 set(PROXYGEN_EXTERNAL_DIR "${CMAKE_CURRENT_SOURCE_DIR}/src/proxygen/external")
 
+# ported over from src/CMakeLists.txt
+set(PROXYGEN_FBCODE_ROOT ${CMAKE_CURRENT_SOURCE_DIR}/src)
+set(PROXYGEN_GENERATED_ROOT ${CMAKE_CURRENT_BINARY_DIR}/generated)
+file(MAKE_DIRECTORY ${PROXYGEN_GENERATED_ROOT})
+
+find_program(PROXYGEN_PYTHON python)
```
fredemmott: If there's issues on some builds, likely to need to add python3 to build-depends in debian control files, and try `python3` first, falling back to `python` as the executable

third-party/proxygen/CMakeLists.txt
```diff
@@ -2,6 +2,38 @@ set(PROXYGEN_LIB_DIR "${CMAKE_CURRENT_SOURCE_DIR}/src/proxygen/lib")
 set(PROXYGEN_HTTPSERVER_DIR "${CMAKE_CURRENT_SOURCE_DIR}/src/proxygen/httpserver")
 set(PROXYGEN_EXTERNAL_DIR "${CMAKE_CURRENT_SOURCE_DIR}/src/proxygen/external")
 
+# ported over from src/CMakeLists.txt
+set(PROXYGEN_FBCODE_ROOT ${CMAKE_CURRENT_SOURCE_DIR}/src)
+set(PROXYGEN_GENERATED_ROOT ${CMAKE_CURRENT_BINARY_DIR}/generated)
```
fredemmott: <3

fredemmott:
Hopefully this isnt' relevant, but if you see errors about `'''` in the python code, that's that the system python is too old (which is why we currently don't run the hack tests on our internal ubuntu 16.04 test jobs)

PR #8679, opened 2020-04-01
Add `_OS/mkostemps` primitive

fredemmott:
Summary:
Not actually a POSIX function, however:
- can be used to implement the related POSIX mkstemp function, and other related
  functions (covered in unit test)
- it is available on both platforms we support (Linux, MacOS)
- while we may support Windows in the future, the related POSIX function
  (mkstemp) is not available there either

In this diff, I'm
- adding the function
- adding specialization for std::tuple to CLI server extensibility
- while I'm here, standardize on `class` instead of `typename` in the
cli-server-ext-detail.h file.
- while I'm here, removing all the assertx(fd >= 0), instead putting that in the HSLFileDescriptor constructor, and throwing an exception instead of asserting so that the CLI client can't intentionally crash the CLI server by returning a bad fd

Reviewed By: paulbiss

Differential Revision: D20771508

### DIFF
--- a/hphp/hack/hhi/hsl/ext_hsl_os_private.hhi
+++ b/hphp/hack/hhi/hsl/ext_hsl_os_private.hhi
@@ -168,6 +168,11 @@ const int F_SETOWN = 0;
 
 const int FD_CLOEXEC = 0;
 
+const int LOCK_SH = 0;
+const int LOCK_EX = 0;
+const int LOCK_NB = 0;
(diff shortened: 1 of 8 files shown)


fredemmott:
Abandoning for a month, until we drop GCC5 support; we need real C++17 for the current approach to work, http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/n4387.html in particular

PR #8680, opened 2020-04-02
Reuse Zstd DContext to avoid frequent memory allocations

efiks:
Summary: Avoid high volume of calls to malloc during memcache client decompression.

Differential Revision: D20802366


jthemphill:
In April 2020, we ran experiments testing how this would improve/regress perf in production. Unfortunately, the experiments showed a regression...

PR #8685, opened 2020-04-21
Update the idx() docblock for 2020 hhvm behavior

### LINKED ISSUE #8684: Rewrite docblock of HH\idx() to be more friendly
/assign @lexidor

https://github.com/facebook/hhvm/blob/master/hphp/system/php/misc/idx.php#L5-L56

This section contains a bit of sass, for the lack of a better term.
It says things like: `Instead, just access it normally like a sensible human being`
I feel like we can do better.

It might also be worth noting the differences between these two cases here:
```HACK
$dict = dict['a' => null];
$dict['a'] ?? 1; // 1
idx($dict, 'a', 1); // null
```

The docblock also states: `This will give you a helpful warning if the index doesn't exist, allowing you to identify a bug in your program and fix it.`
This is not true anymore, since every KeyedContainer now consistently throws an OutOfBoundsException when indexed out of bounds.

lexidor:
closes #8684

### DIFF
--- a/hphp/system/php/misc/idx.php
+++ b/hphp/system/php/misc/idx.php
@@ -4,7 +4,7 @@ namespace HH {
 
 /**
  * Returns the value at a key of a KeyedContainer, if this key is present.
- * This function simplifies the common pattern of checking for an key in
+ * This function simplifies the common pattern of checking for a key in
  * a KeyedContainer and using a default value if the key is not present.
  * You should NOT use `idx` as a replacement for accessing elements
  * of KeyedContainers, since this makes the code harder to reason about.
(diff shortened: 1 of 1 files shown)


lexidor:
My motivations to update these docs were the tone at which the docblock was written and the incorrect behavior of php notices.

I understand why you'd write a docblock this firmly for internal use, but this does show up on the official hacklang.org website, so...

I removed the text about accessing arrays backwards by providing a negative offset `idx($vec, -1)`.
This note was already present when this code was first open sourced back in 2015.
I don't think this notice helps people that remember the old behavior, since they have at least 5 years of experience with `idx()` and are unlikely to read the docblock.
For newer users, accessing with negative integers is not intuitive and might actually end up confusing them.

lexidor:
image

PR #8691, opened 2020-04-29
Update the printf hhi

lexidor:


hphp/hack/hhi/printf.hhi
```diff
@@ -37,11 +37,13 @@ namespace {
...
+ * For another example on how to implement your own format string interface,
+ * see \HH\Lib\Str\SprintfFormat in the HSL.
  */
 
 interface PlainSprintf {
-  // It's common to pass floats; would be nice to type this as
-  // 'number' once that type becomes available in userland.
+  // Technically %d should only take ints, but we don't.
  <<__Rx>>
   public function format_d(mixed $s) : string;
```
azjezz, line 48: `num`? :thinking:

lexidor: https://github.com/facebook/hhvm/pull/8575

azjezz: I think #8575 should be reopened, the change to the comment isn't clear. and `num` is the semi-correct type to put here. ( `int` would be better, but I think `num` is more suitable since this isn't HSL ) 

/cc @fredemmott

lexidor: This was closed because `num` give a false sense of security.
You lose precision on floats.
This is also why `%f` does not take `num` in the hsl (it takes float), since you will lose precision on ints above 2^53.

azjezz: there's a difference between something that gives a false impression and something that just doesn't work. passing a resource to printf for `%d` is now allowed, it shouldn't, passing a `float` *might* ( or might not be ) wrong, this is why `num` is prefered IMO.

jthemphill (approved):
Looks like @jjergus accepted this PR internally in May of, uh... 2020. Then we needed to rebase the PR and it fell off our radar. We definitely approve of this PR though!

jthemphill:
Landed in #8822. This should be resolved now. @fredemmott, could you close this?

Issue #5823, opened 2015-07-30
SplFileObject::fgetcsv does not skip last line in CSV mode


jjanvier:
I'm reading CSV files using SplFileObject class and `fgetcsv` method. Below code does not work as intended in HHVM 3.8.0 meaning `fgetcsv` does return `false` on the LAST line. In PHP 5.6.7-1 the last line returns `null`.

With the following CSV file
```csv
row11;;row13
row21;row22;row23

```

and the following code
```php
$aFile = new SplFileObject ($aPath);
$aFile->setFlags(SplFileObject::SKIP_EMPTY | SplFileObject::READ_AHEAD | SplFileObject::DROP_NEW_LINE | SplFileObject::READ_CSV);
$aFile->setCsvControl(';');

for($i = 0; $i < 3; $i++) {
    $row = $aFile->fgetcsv();
    var_dump($row);
}
```

Result for PHP
```
array(3) {
  [0] =>
  string(5) "row11"
  [1] =>
  string(0) ""
  [2] =>
  string(5) "row13"
}
array(3) {
  [0] =>
  string(5) "row21"
  [1] =>
  string(5) "row22"
  [2] =>
  string(5) "row23"
}
NULL
```

Result for HHVM
```
array(3) {
  [0]=>
  string(5) "row11"
  [1]=>
  string(0) ""
  [2]=>
  string(5) "row13"
}
array(3) {
  [0]=>
  string(5) "row21"
  [1]=>
  string(5) "row22"
  [2]=>
  string(5) "row23"
}
bool(false)
```

SiebelsTim:
`DROP_NEW_LINE` and `SKIP_EMPTY` produce this result. We match php when you omit both of these flags.

nyamsprod:
Hope this little snippet will help you fix this PHP incompatibility

https://3v4l.org/5ZFKe

PR #8708, opened 2020-05-23
Widen SplFileObject::fgetcsv()'s return type to mixed

lexidor:
The Hack code that implements this method has a literal `return false;`
See https://github.com/facebook/hhvm/blob/a20c7945409cb0b10cf901abb029a2ca6035897c/hphp/system/php/spl/file_handling/SplFileObject.php#L172-L174

This also has the side-effect of removing a HH_FIXME for a genericless varray.
A vscode search for `>\s*?fgetcsv` found three callsites.

An assignment to an untyped member variable.
https://github.com/facebook/hhvm/blob/a20c7945409cb0b10cf901abb029a2ca6035897c/hphp/system/php/spl/file_handling/SplFileObject.php#L567-L571

Used as an immediate to var_dump().
https://github.com/facebook/hhvm/blob/a20c7945409cb0b10cf901abb029a2ca6035897c/hphp/test/zend/good/ext/spl/tests/SplFileObject_fgetcsv_escape_error.php#L7

Used as an immediate to var_dump().
https://github.com/facebook/hhvm/blob/a20c7945409cb0b10cf901abb029a2ca6035897c/hphp/test/zend/good/ext/spl/tests/SplFileObject_fgetcsv_delimiter_error.php#L12

refs https://github.com/facebook/hhvm/issues/5823 PHP returns `NULL` when reading the line after the last line, while we return `false`. The documentation on PHP.net says that we should return false.
_Returns an indexed array containing the fields read, or FALSE on error._

We could also match PHP's behavior here and return NULL (deviating from the docs).
However, with the low test coverage we have, I don't dare changing this over to `return null;`.

### DIFF
--- a/hphp/hack/hhi/stdlib/builtins_spl.hhi
+++ b/hphp/hack/hhi/stdlib/builtins_spl.hhi
@@ -121,12 +121,12 @@ class SplFileObject extends SplFileInfo
   public function eof(): bool;
   public function fflush(): bool;
   public function fgetc(): string;
+  /** @return varray|false*/
   public function fgetcsv(
     string $delimiter = ",",
     string $enclosure = "\"",
     string $escape = "\\",
-    /* HH_IGNORE_ERROR[2071] */
-  ): varray;
+  ): mixed;
   public function fgets();
   public function fgetss(?string $allowable_tags = null): string;
   public function flock(int $operation, inout mixed $wouldblock): bool;


lexidor:
Given the following code
```PHP
<?php
// <?hh // partial

function main(): void {
    $aFile = new SplFileObject(__DIR__.'/empty_file.csv');
    $aFile->setFlags(SplFileObject::READ_CSV);
    $aFile->setCsvControl(';');
    \var_dump($aFile->fgetcsv());
    \var_dump($aFile->fgetcsv());
}

main();
```

PHP returns
```
array(1) {
  [0]=>
  NULL
}
NULL
```

HHVM returns
```
bool(false)
bool(false)
```

alshamiri5 (approved):

PR #8722, opened 2020-06-23
Update cmake, simplify CXX version handling

fredemmott:
- We currently manually set the C++ version flags for each compiler, let
  CMake do this instead
- That requires updating the minimum CMake version
- that makes the syntax for target_link_libraries slightly stricter; update folly build
- it also makes Apple Clang start being considered it's own thing ('AppleClang' instead of just 'Clang');
  opt out of the behavior for now, as we do `if (clang)` a bunch.

Seen a few diffs recently where people can't use some specific part of C++17 - hopefully this fixes it,
as well as making future updates simpler.

I set the cmake minimum version to the minimum version that has CMAKE_CXX_STANDARD in it, which is older
than anything we support, but given that updating that changes cmake's behavior, trying to keep it as small
as possible for this diff.

### DIFF
--- a/CMake/FindLibDwarf.cmake
+++ b/CMake/FindLibDwarf.cmake
@@ -72,12 +72,6 @@ if (LIBDWARF_LIBRARIES AND LIBDWARF_INCLUDE_DIRS)
       # Check to see if we can use a const name.
       unset(DW_CONST CACHE)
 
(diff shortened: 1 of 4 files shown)


bosszhit4l69 (approved):

bosszhit4l69 (approved):
[]()

PR #8723, opened 2020-07-02
update FB submodules (incl. fb-mysql, rsocket-cpp)

jjergus:
- updates the usual FB submodules
- for the first time, also updates fb-mysql (5.6 -> 8.0)
- removes rsocket-cpp (D22342884)

I got a successful build (and test/quick pass) on Ubuntu 18.04, still need to test

- all other platforms
- internal (sandcastle) builds
- manually test async mysql since it doesn't look like it's covered by tests

but wanted to get this out now so you can take a look and suggest improvements.

Look at individual commits when reviewing, interesting ones are (some of these are pretty hacky so please suggest better solutions):

- https://github.com/facebook/hhvm/commit/3e243605183e829889a5bf9a3eb05127fd18b6f3 fb-mysql needs an extra flag to allow insource build. I'm not sure why we're building insource originally, but I didn't want to change that here so as to not do too many changes at once. I can try switching to out-of-source in a separate PR later.
- https://github.com/facebook/hhvm/commit/6b7d7cf568209ca5b93b711ecac2103aa03190dc (see comment there) super hacky but I couldn't think of a better way
- https://github.com/facebook/hhvm/commit/c5cb19a09da574d0722d4960049a32323e6af1e7 new fb-mysql puts the library elsewhere -- probably OK but better sanity-check that I'm not missing anything (also the new library name doesn't have the _r suffix, was that important?)
- https://github.com/facebook/hhvm/commit/b37abf15913e64e1a53b882b1d5749bb241c9dae new fb-mysql uses stuff from lz4frame.c/h


jjergus:
#8725

PR #8724, opened 2020-07-10
use ExternalProject for lz4

jjergus:
- remove submodule, use ExternalProject to download the lz4 sources
- lz4 doesn't use cmake, so override the INSTALL_COMMAND
- other than that this is the same as what we did for libzip a while ago

I was planning to do this after https://github.com/facebook/hhvm/pull/8723 but I keep having issues because the dependency between the new fb-mysql and lz4 is not properly handled there -- so doing this first.

### DIFF
--- a/CMake/HPHPFindLibs.cmake
+++ b/CMake/HPHPFindLibs.cmake
@@ -126,12 +126,6 @@ if (DOUBLE_CONVERSION_INCLUDE_DIR)
   include_directories(${DOUBLE_CONVERSION_INCLUDE_DIR})
 endif ()
 
-# liblz4
-find_package(LZ4)
-if (LZ4_FOUND)
-  include_directories(${LZ4_INCLUDE_DIR})
-endif()
-
 # fastlz
 find_package(FastLZ)
 if (FASTLZ_INCLUDE_DIR)
@@ -464,14 +458,7 @@ macro(hphp_link target)
     target_link_libraries(${target} double-conversion)
   endif()
 
-  if (LZ4_FOUND)
-    target_link_libraries(${target} ${LZ4_LIBRARY})
-  else()
-    target_link_libraries(${target} lz4)
-  endif()
-  # The syntax used for these warnings is unparsable by Apple's Clang
-  add_definitions("-DLZ4_DISABLE_DEPRECATE_WARNINGS=1")
-
+  target_link_libraries(${target} lz4)
   target_link_libraries(${target} libzip)
 
   if (PCRE_LIBRARY)
--- a/hphp/hack/CMakeLists.txt
+++ b/hphp/hack/CMakeLists.txt
@@ -66,13 +66,14 @@ if(LZ4_FOUND)
   list(APPEND extra_lib_paths ${pth})
   list(APPEND extra_native_libraries "lz4")
 else()
-  list(APPEND extra_include_paths "${TP_DIR}/lz4/src/lib")
+  get_target_property(LZ4_INCLUDE_DIRS lz4 INTERFACE_INCLUDE_DIRECTORIES)
+  list(APPEND extra_include_paths ${LZ4_INCLUDE_DIRS})
   # If LZ4_FOUND is false either we didn't find lz4 or we found it but it's the
   # wrong version.  We can't just add the new path and a native_lib because we
   # can't control the order (and -l won't accept the raw path to the lib).  By
   # doing it this way we specify the path explicitly.
-  get_target_property(LZ4_DIR lz4 BINARY_DIR)
-  list(APPEND extra_link_opts "${LZ4_DIR}/${CMAKE_STATIC_LIBRARY_PREFIX}lz4${CMAKE_STATIC_LIBRARY_SUFFIX}")
+  get_target_property(LZ4_LIBS lz4 INTERFACE_LINK_LIBRARIES)
+  list(APPEND extra_link_opts ${LZ4_LIBS})
 endif()
 
 if(PC_SQLITE3_FOUND)
--- a/third-party/lz4/CMakeLists.txt
+++ b/third-party/lz4/CMakeLists.txt
@@ -10,21 +10,20 @@ else (LZ4_FOUND)
   cmake_minimum_required(VERSION 2.8.0)
   include(ExternalProject)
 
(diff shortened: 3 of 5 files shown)


CMake/HPHPFindLibs.cmake
```diff
@@ -464,14 +458,7 @@ macro(hphp_link target)
     target_link_libraries(${target} double-conversion)
   endif()
 
-  if (LZ4_FOUND)
-    target_link_libraries(${target} ${LZ4_LIBRARY})
-  else()
-    target_link_libraries(${target} lz4)
-  endif()
-  # The syntax used for these warnings is unparsable by Apple's Clang
-  add_definitions("-DLZ4_DISABLE_DEPRECATE_WARNINGS=1")
```
jjergus, line 473: Mac build failed for an unrelated reason, I'll re-run it to test whether this is still needed.

hphp/hack/CMakeLists.txt
```diff
@@ -66,13 +66,14 @@ if(LZ4_FOUND)
...
+  list(APPEND extra_include_paths ${LZ4_INCLUDE_DIRS})
   # If LZ4_FOUND is false either we didn't find lz4 or we found it but it's the
   # wrong version.  We can't just add the new path and a native_lib because we
   # can't control the order (and -l won't accept the raw path to the lib).  By
   # doing it this way we specify the path explicitly.
-  get_target_property(LZ4_DIR lz4 BINARY_DIR)
-  list(APPEND extra_link_opts "${LZ4_DIR}/${CMAKE_STATIC_LIBRARY_PREFIX}lz4${CMAKE_STATIC_LIBRARY_SUFFIX}")
+  get_target_property(LZ4_LIBS lz4 INTERFACE_LINK_LIBRARIES)
+  list(APPEND extra_link_opts ${LZ4_LIBS})
 endif()
```
jjergus, line 77: I couldn't find an easy way to replace the custom logic here (especially the `if(LZ4_FOUND)` branch above) with `target_link_libraries`, I might look into it some more, or let me know if you have suggestions.

third-party/lz4/CMakeLists.txt
```diff
@@ -1,2 +1,30 @@
...
+  set(LZ4_PREFIX "${CMAKE_CURRENT_BINARY_DIR}/lz4-install")
+
+  ExternalProject_Add(
+    lz4Build
+    URL "https://github.com/lz4/lz4/archive/v1.9.2.tar.gz"
+    URL_HASH SHA256=658ba6191fa44c92280d4aa2c271b0f4fbc0e34d249578dd05e50e76d0e5efcc
+    PREFIX ${LZ4_PREFIX}
+    BUILD_IN_SOURCE true
+    CONFIGURE_COMMAND ""
+    BUILD_COMMAND ""
```
fredemmott, line 22: Don't want an explicit `make` before the install?

hphp/hack/CMakeLists.txt
fredemmott: in general this should be refactored so that `target_link_libraries(foo lz4)` is called, instead of using extra_link_opts; the current construction of custom linker flags is something we should try to move away from

fredemmott: :/ sorry, I'm doing really badly at reading today. Time for more coffee. I don't think there's an easy way, it's going to need a thorough (painful) revisit

third-party/lz4/CMakeLists.txt
jjergus: I had `make liblz4.a` here at first, but then `make install` would just build a bunch of other stuff anyway, so I figured it's easier to just let `make install` build exactly what it installs...

third-party/lz4/CMakeLists.txt
```diff
@@ -1,2 +1,30 @@
...
+  cmake_minimum_required(VERSION 2.8.0)
+  include(ExternalProject)
+
+  set(LZ4_PREFIX "${CMAKE_CURRENT_BINARY_DIR}/lz4-install")
+
+  ExternalProject_Add(
+    lz4Build
+    URL "https://github.com/lz4/lz4/archive/v1.9.2.tar.gz"
+    URL_HASH SHA256=658ba6191fa44c92280d4aa2c271b0f4fbc0e34d249578dd05e50e76d0e5efcc
+    PREFIX ${LZ4_PREFIX}
```
fredemmott, line 19: I think my previous comment was lost by github - this should be `<INSTALL_DIR>` rather than specified, then fetched with ExternalTarget_get_property

hphp/hack/CMakeLists.txt
jjergus: As far as I can tell, these get passed to Dune/Ocaml. Is it even possible to make that work with CMake?

third-party/lz4/CMakeLists.txt
fredemmott: Example from one of my personal projects: https://github.com/fredemmott/StreamDeck-CPPSDK/blob/fdd50d9972d92b6486d1aa9b3fbe0a3cd7572d6b/Vendor/fmt.cmake

Uses cmake, but the placeholder works regardless

hphp/hack/CMakeLists.txt
fredemmott: more reading fail on my end. I thought this was hphp/CMake/ - this is absolutely fine in hphp/hack/CMake/

third-party/lz4/CMakeLists.txt
```diff
@@ -1,2 +1,29 @@
...
+
+else (LZ4_FOUND)
+  cmake_minimum_required(VERSION 2.8.0)
+  include(ExternalProject)
+
+  ExternalProject_Add(
+    lz4Build
+    URL "https://github.com/lz4/lz4/archive/v1.9.2.tar.gz"
+    URL_HASH SHA256=658ba6191fa44c92280d4aa2c271b0f4fbc0e34d249578dd05e50e76d0e5efcc
+    BUILD_IN_SOURCE true
```
fredemmott, line 17: Sorry, now that INSTALL_DIR is being used, can you try turning this off?

jjergus: That creates an empty build directory and runs `make` in it. I think an out-of-source build requires a cmake-style configuration step that creates makefiles in the build directory.

But this should be fine, because "in source" here means inside the downloaded source, which itself is in HHVM's build directory.

PR #8725, opened 2020-07-14
update FB submodules (incl. mysql, rsocket) -- final version

jjergus:
Making a new PR since the old one's a mess and there are some significant changes worth reviewing again:

- most importantly: https://github.com/facebook/hhvm/commit/6a8558193893ad97d23acfa3cebc034ce7fb5dba rewrites every cmake reference to mysql (I'll leave some comments inline)
- updated to this week's (2020-07-13) version of everything, mainly to avoid adding patches for the 4 upstream fixes from last week
- https://github.com/facebook/hhvm/commit/86d485ebb917eb68b8b5a791cdba0b22b40a7d33 removes some special handling of `FOLLY_HAVE_WEAK_SYMBOLS` for MacOS which turned out to be wrong (some discussion on D21535034 and D22420581)
- https://github.com/facebook/hhvm/commit/44256ca91c9d5f8d4d450080996b4436dd3e0f53 removed one part of the thrift codegen instructions, since that only contained rsocket-cpp files that no longer exist
- includes #8724

Closes #8724


CMake/HHVMExtensionConfig.cmake
```diff
@@ -605,6 +605,7 @@ function (HHVM_EXTENSION_INTERNAL_HANDLE_LIBRARY_DEPENDENCY extensionID dependen
     ${libraryName} STREQUAL "lz4" OR
     ${libraryName} STREQUAL "mbfl" OR
     ${libraryName} STREQUAL "mcrouter" OR
+    ${libraryName} STREQUAL "mysql" OR
```
jjergus, line 608: mysql is handled in hphp_link() in the exact same way as everything above and below, so I think it belongs here too?

The only part of the mysql branch below that was not already handled in HPHPFindLibs is setting -DENABLE_ASYNC_MYSQL=1 so I added it there, to have everything in one place.

Alternatively, should we remove all the mysql references from HPHPFindLibs and instead keep the mysql branch here (and only here)? That might be better if mysql is not referenced from anywhere else except these extensions...

hphp/runtime/ext/CMakeLists.txt
```diff
@@ -43,9 +43,6 @@ add_object_library(hphp_runtime_ext
 auto_source_group("hphp_runtime_ext" "${CMAKE_CURRENT_SOURCE_DIR}"
   ${ASM_SOURCES} ${C_SOURCES} ${CXX_SOURCES} ${HEADER_SOURCES})
 add_dependencies(hphp_runtime_ext hphp_runtime_static)
-if (ENABLE_ASYNC_MYSQL)
-  add_dependencies(hphp_runtime_ext fbmysqlclient squangle)
-endif ()
```
jjergus, line 48: hphp_runtime_ext goes through hphp_link so I think this was redundant here?

hphp/system/CMakeLists.txt
```diff
@@ -1,9 +1,6 @@
 set(CXX_SOURCES "systemlib.cpp")
 
 add_library(hphp_system STATIC ${CXX_SOURCES})
-if (ENABLE_ASYNC_MYSQL)
-  add_dependencies(hphp_system fbmysqlclient)
-endif ()
```
jjergus, line 6: Not sure why this was needed here, I don't see any references to mysql in hphp_system.

hphp/util/CMakeLists.txt
```diff
@@ -71,6 +71,3 @@ endif()
 if (LIBNUMA_LIBRARIES)
   target_link_libraries(hphp_util ${LIBNUMA_LIBRARIES})
 endif()
-if (ENABLE_ASYNC_MYSQL)
-  add_dependencies(hphp_util fbmysqlclient)
-endif ()
```
jjergus, line 76: Also not sure why this was needed here, I don't see any references to mysql in hphp_util.

third-party/fb-mysql/CMakeLists.txt
```diff
@@ -1,26 +1,62 @@
...
+  set(DEPS lz4)
+  add_dependencies(mysqlclientBuild ${DEPS})
+
+  ExternalProject_Get_Property(mysqlclientBuild INSTALL_DIR)
+
+  add_dependencies(mysqlclient mysqlclientBuild)
+  target_include_directories(mysqlclient INTERFACE "${INSTALL_DIR}/include")
+  target_link_libraries(mysqlclient INTERFACE
+    "${INSTALL_DIR}/lib/${CMAKE_STATIC_LIBRARY_PREFIX}mysqlclient${CMAKE_STATIC_LIBRARY_SUFFIX}"
+    ${DEPS})
```
jjergus, line 54: I think we need to duplicate the `${DEPS}` here because libmysqlclient.a does *not* have the dependencies statically linked inside it -- so anything that links with libmysqlclient.a will also need to include them as part of the linker invocation... but I might be misunderstanding something, so let me know if this makes sense. The build seems to succeed with or without the ${DEPS} here but that might just be luck.

third-party/re2/CMakeLists.txt
```diff
@@ -12,3 +12,4 @@ set(RE2_INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/src" CACHE STRING "RE2 include
 # util/arena.h
 SET_PROPERTY(DIRECTORY PROPERTY INCLUDE_DIRECTORIES "${RE2_INCLUDE_DIR}")
```
jjergus, line 13: this would normally be redundant after the change below, but probably still needed because of what the comment says

third-party/re2/CMakeLists.txt
```diff
@@ -12,3 +12,4 @@ set(RE2_INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/src" CACHE STRING "RE2 include
...
 add_library(re2 STATIC ${util_sources} ${sources})
+target_include_directories(re2 PUBLIC ${RE2_INCLUDE_DIR})
```
jjergus, line 15: this should add the include path for the current project + all dependencies (PUBLIC = PRIVATE + INTERFACE)

CMake/HHVMExtensionConfig.cmake
fredemmott: > The only part of the mysql branch below that was not already handled in HPHPFindLibs is setting -DENABLE_ASYNC_MYSQL=1 so I added it there, to have everything in one place.

We should probably just remove the option; the async extension is by far the best way to use MySQL with Hack, and I don't think we're testing (even building) with it disabled at all.

Either way, yeah, moving this to the specific extension seems like a good idea.

third-party/fb-mysql/CMakeLists.txt
fredemmott: Yeah, this should be here - the mysqlclient library should be the source of truth.

It could also be that everything that needs mysql explicitly depends on the recursive dependencies directly - this would make sense with how the cmake was originally written (and what cmake supported however many years ago), but isn't great nowadays

CMake/HHVMExtensionConfig.cmake
jjergus: I tested it, it builds and successfully runs a MySQL query :) But it seems to only support PHP's old `mysql_*` API which is deprecated so you're probably right, we should just kill it.

Curiously, ENABLE_ASYNC_MYSQL is only referenced from the *non*-async mysql extension.

The socket path (PHP_MYSQL_UNIX_SOCK_ADDR) is also only referenced from the non-async extension, as far as I can tell the async extension ignores it and uses the wrong path :( There aren't any references to any sockets in the async extension's code, I think it just delegates to squangle/libmysqlclient. Perhaps we need to pass the socket path to libmysqlclient instead?

> Either way, yeah, moving this to the specific extension seems like a good idea.

Alright, I'll try removing all the mysql stuff from HPHPFindLibs and leave it only in the extension code. Should I kill non-async mysql while I'm at it, or would we want to announce that first so people can prepare?

fredemmott: > Curiously, ENABLE_ASYNC_MYSQL is only referenced from the non-async mysql extension.

> should I kill non-async mysql

This is a bit misleadingly named: its' primary purpose is/was to control if we build fbmysqlclient (or previously, webscalesqlclient)_

If we build fbmysqlclient:
- we can build the async extension
- the non-async extension also links against fbmysqlclient

If we don't:
- we can never build the async extension
- the non-async extension should link against a real libmysqlclient

I don't think we need to warn about removing support for linking against libmysqlclient if we keep the extension for now; are you also thinking of removing the non-async extension?

jjergus: Oh I think you meant removing the whole extension, if it's just the build option then yeah that should be fine. I'll test that + moving all the cmake stuff from HPHPFindLibs to the extension cmake config.

jjergus:
OK I think I'll land this and make a separate PR for the remaining MySQL changes, that should make it easier to test and review.

jjergus:
I'll include #8727 here since both are reviewed now, so that I don't have to wait on too many builds.

Closes #8727

PR #8727, opened 2020-07-15
always build fbmysqlclient, link only in hphp/runtime/ext 

jjergus:
(only review last commit, everything else is identical to #8725)


CMake/HHVMExtensionConfig.cmake
```diff
@@ -906,6 +903,10 @@ function (HHVM_EXTENSION_INTERNAL_HANDLE_LIBRARY_DEPENDENCY extensionID dependen
       HHVM_EXTENSION_INTERNAL_ADD_LINK_LIBRARIES(${LIBMEMCACHED_LIBRARY})
       HHVM_EXTENSION_INTERNAL_ADD_DEFINES("-DHAVE_LIBMEMCACHED")
     endif()
+  elseif (${libraryName} STREQUAL "mysql")
+    HHVM_EXTENSION_INTERNAL_ADD_LINK_LIBRARIES(fbmysqlclient)
+    MYSQL_SOCKET_SEARCH()
```
jjergus, line 908: the check is done below, no need to repeat it here

third-party/fb-mysql/CMakeLists.txt
```diff
@@ -1,62 +1,64 @@
...
+  -DWITHOUT_SERVER=TRUE
+  -DDISABLE_SHARED=TRUE
+  -DCMAKE_C_FLAGS=${CMAKE_C_FLAGS}
+  -DCMAKE_CXX_FLAGS=${CMAKE_CXX_FLAGS}
+  -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
+  -DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
+  -DOPENSSL_INCLUDE_DIR=${OPENSSL_INCLUDE_DIR}
+  -DOPENSSL_LIBRARY=${OPENSSL_SSL_LIBRARY}
+  -DCRYPTO_LIBRARY=${OPENSSL_CRYPTO_LIBRARY}
+  -DMYSQL_UNIX_ADDR=${MYSQL_UNIX_SOCK_ADDR}
```
jjergus, line 38: This fixes 'localhost' support in the async_mysql extension (I'm not sure if it was broken by my fb-mysql update or something else, but anyway it works now).

third-party/fb-mysql/CMakeLists.txt
```diff
@@ -1,62 +1,64 @@
...
+# TODO: re2, zstd are also dependencies, but first they need to be split into
+# re2 + re2Build, zstd + zstdBuild -- otherwise putting them here would force
+# them to always be built, instead of using the system ones where available.
+set(DEPS lz4)
+add_dependencies(fbmysqlclientBuild ${DEPS})
+
+# This extra indirection is needed because if the fbmysqlclient target links
+# ${DEPS} directly, the correct linking order is not guaranteed.
+add_library(fbmysqlclientDeps INTERFACE)
+target_link_libraries(fbmysqlclientDeps INTERFACE ${DEPS})
```
jjergus, line 55: I think this is not a new problem, but I got lucky in my previous PR that the linking order happened to be correct without this.

third-party/fb-mysql/CMakeLists.txt
```diff
@@ -1,62 +1,64 @@
...
+
+ExternalProject_Get_Property(fbmysqlclientBuild INSTALL_DIR)
+
+add_dependencies(fbmysqlclient fbmysqlclientBuild)
+target_include_directories(fbmysqlclient INTERFACE "${INSTALL_DIR}/include")
+target_link_libraries(fbmysqlclient INTERFACE
+  "${INSTALL_DIR}/lib/${CMAKE_STATIC_LIBRARY_PREFIX}mysqlclient${CMAKE_STATIC_LIBRARY_SUFFIX}"
+  fbmysqlclientDeps
```
jjergus, line 63: Basically, the ordering here doesn't guarantee anything. But with this, it registers fbmysqlclientDeps as a dependency of fbmysqlclient and ensures the correct linking order. I'm not sure why using lz4 here directly doesn't register that as a dependency :/

jjergus: Also changing the order here breaks it too. For some reason, it only works if both:

1. there is the extra indirection (fbmysqlclient -> fbmysqlclientDeps -> lz4)
2. the order here is also correct

jjergus:
> The need for indirection sounds strange, but sure; ordering being picky is not surprising

I have a simple repro in my cmake playground, in case you want to dig into it, but as far as I can tell, this is the explanation:

1. hhvm depends on lz4, mysql
2. mysql depends on lz4

(i.e., hhvm depends on lz4 both directly and indirectly through mysql)

Because of (1), we call (a) `target_link_libaries(hphp_something lz4)` at some point, and also (b) `target_link_libraries(hphp_something fbmysqlclient)` at some point, which happens to be later (which shouldn't matter but does).

Because of (2), we call `target_link_libraries(fbmysqlclient "path/to/libmysqlclient.a" lz4)` at some point.

(a) adds lz4 to the linker invocation for hphp_something.
(b) causes cmake to look at fbmysqlclient's "interface link libraries". It finds "path/to/libmysqlclient.a" and adds it to the linker invocation for hphp_something. Then it finds lz4 but *sees that it's already included in the linker invocation, so ignores it*.

The problem is, if I understand it correctly, that cmake doesn't know there is a dependency of "path/to/libmysqlclient.a" on lz4, so it thinks it is correct to "deduplicate" the lz4 part of the linker invocation. In other words, listing multiple libraries in `target_link_libraries` doesn't register any dependencies between them, which is probably reasonable, but makes it hard to handle our situation without ugly hacks.

Linking lz4 indirectly through fbmysqlclientDeps makes (I think) cmake realize there is a dependency, so it ensures fbmysqlclientDeps' link libraries correctly appear *after* fbmysqlclient's "path/to/libmysqlclient.a".

This doesn't explain why changing the order to `target_link_libraries(fbmysqlclient fbmysqlclientDeps "path/to/libmysqlclient.a")` breaks it though, I'm still not sure about that part.

jjergus:
Included in https://github.com/facebook/hhvm/commit/df26d44ce886773b34d414db195bcad7479d9ccd

PR #8728, opened 2020-07-15
unbreak repomode + symlinks + namespaces

jjergus:
It looks like the fix from https://github.com/facebook/hhvm/commit/45adf85b640450b5d90b7b0640e180c434652414 (D22168683) doesn't work if namespaces are involved :(


jjergus:
I should probably also cherry-pick this into the broken release branches (4.64, 4.65, 4.66) and build a new release for each.

Issue #8716, opened 2020-05-28
_OS\poll_async() on non-blocking server socket blocks forever after _OS\accept() and _OS\close()


fredemmott:
**Describe the bug**

Best done by example code below. Also applies to resources/stream_await()

**Standalone code, or other way to reproduce the problem**

```hack
use namespace HH\Lib\_Private\_OS;
<<__EntryPoint>>
async function main(): Awaitable<void> {
  $sin = new _OS\sockaddr_in(0, _OS\INADDR_LOOPBACK);
  $server = _OS\socket($sin->sa_family, _OS\SOCK_STREAM, 0);
  _OS\bind($server, $sin);
  _OS\listen($server, 128);

  _OS\fcntl($server, _OS\F_SETFL, (_OS\fcntl($server ,_OS\F_GETFL) as int )| _OS\O_NONBLOCK);
  try {
    _OS\accept($server);
  } catch (_OS\ErrnoException $e) {
    invariant($e->getCode() === _OS\EAGAIN, 'unexpected errno');
  }


  $poll = _OS\poll_async($server, \STREAM_AWAIT_READ, 0);
  _OS\close($server);
  await $poll; // blocks forever
}
```

**Expected behavior**

`poll_async()` fails with` \STREAM_AWAIT_CLOSED`

**Actual behavior**

Blocks forever

**Environment**

 - Operating system: MacOS Catalina, Ubuntu 20.04
 - Installation method: homebrew, FB packages
 - HHVM Version: 4.59.0

**Additional context**

Originally filed as https://github.com/hhvm/hsl-experimental/issues/131

@jano : it looks like this is basically undefined behavior in the C++ side given the threading; I think that HHVM should track the `FileAwait`'s associated with each FD/resource, and either:
1. make the close call fail if any are pending, and provide a way to wait for them all
2. make the close() call cancel the fileawaits  - the stream_await()\poll_async() calls should fail with STREAM_AWAIT_CLOSED

IMO (2) is the correct solution - does it sound sane? I think this would be similar to the timeout cancellations - something like `runInEventThreadBaseAndWait(for(const auto& fa : m_fileawaits): fa.setFinished(FileAwait::CLOSED) }` ?

jjergus:
The workaround seems to introduce another ugly edge case. To reproduce, in hsl-experimental:

```
vendor/bin/hacktest tests/tcp/HSLTCPTest.php tests/unix/HSLUnixSocketTest.php
[...]

PR #8733, opened 2020-07-23
fix for `_OS\poll_async` blocked forever after close

jjergus:
This is my attempt at a fix for #8716 implemented fully inside HHVM -- hsl-experimental tests now pass even after deleting `CancelablePoller` ([hsl-experimental diff here](https://gist.github.com/jjergus/bcbeb535e1945d82c1a72e503374afba)).

@fredemmott I'm adding you as a reviewer first so you can check that

- this looks like the approach we want (compared to CancelablePoller or other possible solutions)
- I'm not doing anything obviously wrong (e.g. memory management -- I couldn't find any good documentation about how to properly manage memory in HHVM code. As far as I can tell C++'s smart pointers are generally not used and Object is the closest equivalent I could find (it seems roughly equivalent to shared_ptr?), but I'm not sure if Object was meant to be used like that...)

I'll import it to Phabricator and add more reviewers after that.

### DIFF
--- a/hphp/runtime/ext/hsl/ext_hsl_os.cpp
+++ b/hphp/runtime/ext/hsl/ext_hsl_os.cpp
@@ -392,6 +392,7 @@ struct HSLFileDescriptor {
   void close() {
     switch (m_type) {
       case Type::FD:
+        FileAwait::closeAllForFD(m_fd);
         throw_errno_if_minus_one(::close(fd()));
         s_fds_to_close->erase(m_fd);
         m_fd = -1;
--- a/hphp/runtime/base/file-await.cpp
+++ b/hphp/runtime/base/file-await.cpp
@@ -12,11 +12,13 @@ namespace HPHP {
 /////////////////////////////////////////////////////////////////////////////
 
 void FileTimeoutHandler::timeoutExpired() noexcept {
-  m_fileAwait.setFinished(FileAwait::TIMEOUT);
+  m_fileAwait.m_eventHandler.unregisterHandler();
+  m_fileAwait.setFinished(FileAwait::TIMEOUT, true);
 }
 
 void FileEventHandler::handlerReady(uint16_t events) noexcept {
-  m_fileAwait.setFinished(events ? FileAwait::READY : FileAwait::CLOSED);
+  m_fileAwait.m_timeoutHandler.cancelTimeout();
+  m_fileAwait.setFinished(events ? FileAwait::READY : FileAwait::CLOSED, true);
 }
 
 /////////////////////////////////////////////////////////////////////////////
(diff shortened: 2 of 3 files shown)


hphp/runtime/ext/hsl/ext_hsl_os.cpp
```diff
@@ -390,6 +390,17 @@ struct HSLFileDescriptor {
   }
 
   void close() {
+    for (auto obj : m_async_polls) {
+      auto wh = wait_handle<c_ExternalThreadEventWaitHandle>(obj.get());
+      if (!wh->isFinished()) {
+        auto file_await = static_cast<FileAwait*>(wh->getEvent());
+        file_await->unregisterHandler();
+        wh->cancel(create_object(s_ErrnoException, make_vec_array(
+          "Operation cancelled because FileDescriptor was closed.", ECONNABORTED)));
```
fredemmott, line 399: - I don't think this should be `ECONNABORTED` - the stream_await/poll_async should largely be thought of as 'select, but async', and I think ECONNABORTED will be unexpected there.  Perhaps whatever's done here should be translated to ECONNABORTED by Hack code in the HSL though in accept_async
- I *think* EBADF is the most reasonable thing to happen here; AIUI it could happen anyway if a close hapened before the stream_await; otherwise EINTR seems like a good candidate

hphp/runtime/ext/hsl/ext_hsl_os.cpp
```diff
@@ -408,6 +419,23 @@ struct HSLFileDescriptor {
     }
   }
 
+  // Called from HSL_os_poll_async with the WaitHandle it creates.
+  void addAsyncPoll(Object obj) {
+    // Remove any already finished ones. We could do this elsewhere but there
```
fredemmott, line 424: probably good to talk to @jano or @alexeyt  about this; I think we need these to end up being cleaned up even if there isn't another one added.

This could end up meaning that FileAwaits need to be aware of FileDescriptors, or perhaps poll_async makes a distinct wrapper awaitable that's basically a try/finally around the FileAwait

This also ends up keeping a bunch of stack data (especially if an exception is thrown) until the next one

jjergus: My other idea was to pass an "on_destruct" callback to FileAwait. By "wrapper" do you mean a new type of WaitHandle?

fredemmott: Yep, I was thinking a new kind of WaitHandle.

jjergus:
This new version correctly handles cleanup in both directions:

- `HSLFileDescriptor` has a set of pending `FileAwait`s (like the original version, but I was able to simplify it to store pointers to `FileAwait` directly instead of its wait handles -- so 1 less layer of indirection, and we don't need to expose `wait_handle->getEvent()` anymore)
- `FileAwait` has a `weak_ptr` to the relevant `HSLFileDescriptor` stuff (not directly but captured in a lambda, to make the code generic and not tied to HSL) -- I don't see a way to avoid `weak_ptr` or some similar fancy pointer shenanigans, since the dependencies here are by definition cyclical (we need the objects to be connected in both directions and either of them can be freed first -- in most normal cases the `FileAwait` is freed first, but the opposite can happen e.g. using the code below)

Also:

- I think this also terminates the wait handle more correctly (no longer calling `cancel()`), and from the correct thread (probably)
- I can create a wrapper around `FileAwait` instead of adding `setOnDestruct()` if people prefer that, but it would likely add a bunch more boilerplate code

This test exercises the "dangling weak_ptr" branch of the code:

```hack
public async function testWeakPtr(): Awaitable<void> {
  $s = await TCP\Server::createAsync(
    IPProtocolVersion::IPV4,
    '127.0.0.1',
    0,
  );
  $aw = _OS\poll_async($s->impl, \STREAM_AWAIT_READ, /* timeout = */ 0);
  _OS\close($s->impl);
  echo "\n~~~ should release soon ~~~\n";
  $s = null;
  echo "\n~~~ should be released now ~~~\n";
  $result = await $aw;
  echo "\n~~~ result: $result ~~~\n";
}
```

The HSLFileDescriptor is freed at `$s = null` as expected, with the async poll still pending, and later the callback is invoked and correctly detects the expired weak_ptr.

fredemmott:
Looks reasonable to me ; I think you want to import this and get Alexey + Jan Oravec to look at this next

alshamiri5 (approved):

PR #8754, opened 2020-09-23
update FB submodules (v2020.10.19.00)

jjergus:
A few interesting things here:

- https://github.com/facebook/hhvm/pull/8754/commits/0be80f4342c9c06a31a35013aa5145418d5b0069 -- we have been hardcoding `FOLLY_HAS_LIBUNWIND=1` but we never actually linked that library, so I think doing so is the correct fix (other options might be setting `FOLLY_HAS_LIBUNWIND=0` or finding a way to avoid any code that actually needs it, which is what we've been doing before, probably unintentionally?)
- https://github.com/facebook/hhvm/pull/8754/commits/58ad3703db51f0f1d007203f28f5d888276cd78e -- not sure if this is a good fix but the same combination of conditions is used elsewhere in HHVM code (other option might be figuring out why _GNU_SOURCE ends up being defined on MacOS, but that macro is defined and used in many different places which makes it hard to reason about...)


jjergus:
working on a new version; will open a new PR since this one's a mess now

PR #8801, opened 2021-02-18
update FB dependencies (v2021.02.15.00)

jjergus:


jjergus:
fbthrift needs a fix for GCC 8 compatibility; I'll open a new PR when that's ready

PR #8802, opened 2021-02-24
use updated timelib

jjergus:


fredemmott:
#8870

fredemmott:
Actually, re-opening: the simplification here is good and not included in that PR

fredemmott:
Closing for now: this needs a lot of internal work before we can remove these #ifs :(

PR #8847, opened 2021-06-16
RFC: Replacement for `Str\search_last()`, replacing `$offset` with `$length`

fredemmott:
Summary: Explanation in code

Implementation is just placeholder that is clear about behavior. Not for landing.

Current behavior does not match the documentation in PHP or HHVM

Related: https://bugs.php.net/bug.php?id=73913
https://www.php.net/manual/en/function.strrpos.php

Differential Revision: D29153162

### DIFF
--- a/hphp/hsl/src/str/introspect.php
+++ b/hphp/hsl/src/str/introspect.php
@@ -237,6 +237,50 @@ function search_last(
   return $position;
 }
 
+/** RFC: Future behavior/signature for search_last()
+ *
+ * The `search_last()` behavior is unintuitive:
+ * - if a positive offset is supplied, it's effectively a lower bound; it can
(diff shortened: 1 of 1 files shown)


fredemmott:
# Positive offsets

Rare, can be emulated with:

```
$pos = Str\search_last($a, $b);
if ($pos < $offset) {
  $pos = null;
}
```

# Negative offsets

Extremely rare. Can be converted to a length with

`$length = strlen($haystack) + min(strlen($needle) + $offset, 0)`

fredemmott:
Another approach would be to accept both offset and length parameters, with offset having no effect other than where to start/stop looking; negative offsets could still be supported, but an offset of `-5` would literally mean 'only look in the last 5 bytes'; a length of `-5` could mean 'don't look at the last 5 bytes'

PR #8855, opened 2021-07-08
Add the optional `opaque` key to TypeStructure

lexidor:
```HACK
newtype A = int;
=\HH\ReifiedGenerics\get_type_structure<A>();
Dict
(
    [kind] => 1
    [opaque] => true
    [alias] => "A"
)
```


lexidor:
This PR is covered with dust. Closing.

PR #8874, opened 2021-08-20
Declare SQL format string interfaces at runtime

### LINKED ISSUE #8873: [ Runtime ] Missing definition for HH\Lib\SQL\QueryFormat and friends
**Describe the bug**
The typechecker allows you to implement the interface, but hhvm screams fatally with an undefined interface error.

**Standalone code, or other way to reproduce the problem**
```HACK
final class QueryFormat implements HH\Lib\SQL\QueryFormat {
  // methods
}
```

1. Typecheck, will error on missing methods. When methods are added, no errors
2. Invoke program with hhvm, Fatal error: Undefined interface HH\Lib\SQL\QueryFormat

**Expected behavior**

This code should not fatal.

**Actual behavior**

Invoke program with hhvm, Fatal error: Undefined interface HH\Lib\SQL\QueryFormat

**Environment**
 - Operating system
> Ubuntu 20.04
 - Installation method
> 'apt-get with dl.hhvm.com repository'
 - HHVM Version
> 4.123

**Additional context**
Implemented in [slackhq/hack-sql-fake](https://github.com/slackhq/hack-sql-fake/blob/73481a4e41309f109f478191c72de61dca4a6676/src/QueryAsync/implementations_of_query_format.php)

lexidor:
Fixes #8873

### DIFF
--- a/hphp/hack/hhi/stdlib/builtins_async_mysql.hhi
+++ b/hphp/hack/hhi/stdlib/builtins_async_mysql.hhi
@@ -309,6 +309,8 @@ namespace HH {
   }
 }
 
+// Please update hphp/runtime/ext/async_mysql/ext_async_mysql.php
+// when changes are made to these interfaces.
 namespace HH\Lib\SQL {
   interface ScalarFormat {
     public function format_f(?float $s): string;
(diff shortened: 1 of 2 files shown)


lexidor:
This PR has lost its relevancy. It still addresses a problem (an incoherency between the HHI and the runtime available types), but that is better addressed with a sweeping tool instead of a pinpoint fix. The code in hack-sql-fake comments out the implements clause in order to get the code to run.

PR #8879, opened 2021-08-30
[hh_codesynthesis] Changes to run on external machine

tobielf:
Made a few changes in the code, so that it can be run on the external machine. To be compatible with both internal and external ways of running. We need to find a way to detect the code is running internal environment or external.

We can start a discussion on how to detect in this P/R thread, I will adjust my code later.


jamesjwu:
I don't think you'll need to change hphp here: I tried this and here's the workflow I used to get it to work. 
- Setup a virtualenv with clingo installed (or just have clingo installed on your base python, if you want
- Remove libfb from the implementation
- Run:
```
python -m hphp.hack.src.hh_codesynthesis.hh_codesynthesis --target_lang=hack --n=12 --avg_width=3 --min_classes=3 --min_interfaces=4 --lower_bound=1 --higher_bound=5 --min_depth=0 --output_file=tmp --log log
```
I had to change a few minor things which I'll comment on. But if you then pass in no dependencies, it will output:

```
<?hh
class S9 extends S11  {}
class S10   {}
class S11   {}
interface S0  {}
interface S1 extends S3,S4,S5 {}
interface S2  {}
interface S3  {}
interface S4  {}
interface S5  {}
interface S6  {}
interface S7  {}
interface S8  {}
```

Which looks okay!

tobielf:
Sounds good! Similarly, to run unittest, we can do
```
python3 -m unittest discover -s './hphp/hack/test/hh_codesynthesis' -p '*_test.py'
```

hphp/hack/src/hh_codesynthesis/hh_codesynthesis.py
```diff
@@ -225,9 +231,16 @@ def extract_logic_rules(lines: List[str]) -> List[str]:
 # Take in a dependency graph and a code generator to emit code.
 def do_reasoning(additional_programs: List[str], generator: CodeGenerator) -> None:
     # Logic programs for code synthesis.
-    asp_files = os.path.join(
-        parutil.get_dir_path("hphp/hack/src/hh_codesynthesis/"), "asp_code"
-    )
+
+    try:
+        # Check if we are running in the internal environment.
+        importlib.util.find_spec("libfb.py")
```
jamesjwu, line 237: Can we not duplicate this code? Just set a variable on whether we're internal and use that to decide asp_files.

tobielf: Sure, currently, it only shows in one place. I think the ultimate solution will be a mock module to mock out whatever internal functions/packages. So we reduce the usage of global variables and if checking.
```
if g_internal_run:
    do_something()
else:
    do_otherthing()
```

PR #8882, opened 2021-09-02
RFC: Add Str\foo_l UTF8-capable and locale-aware functions.

fredemmott:
Summary:
We have a single object that represents encoding + locale, similar to the `en_US.UTF-8` strings being used with `setlocale()`.

For now, we only support single-byte and UTF-8, as we require that all pairs are potentially valid thread/process locales - UTF-16 isn't because C locales can't have null bytes except as terminators. We could (and probably should) support UTF-16 in the future, but we would have to make Locale\set_native() throw an exception when a UTF-16 is passed.

As for design:
- if `Str\foo()` exists, so does `Str\foo_l()`, and `Str\foo_l()` will always take a `Locale\Locale` object as the first parameter.
- in most cases, the rest of the parameters match; in some cases, the locale defines previously optional parameters, so I've removed the parameter in the `_l` version

`_Private\_Str`: not a supported API. That said, it follows the BSD libc conventions; most glibc _l functions also exist in BSD libc, and BSD libc has many more.

I went with that convention for `_Str` mostly to unblock implementation. I've not used it for the public API as:
- the parameter convention is 'the last non-variadic non-format-string parameter'; this isn't super clear and feels inconsistent, but makes sense for C. Given that past work shows edge cases like `vsprintf_l(char* str, locale_t loc, const char* format, va_list ap)`, we can avoid them with a more consistent design now.
- unintuitive naming: the case-sensitive version of `strcasecmp()` is `strcoll()`, not `strcmp()`; `strcasecmp()`  and `strcoll()` are locale-sensitive, `strcmp()` isn't.
- having `foo` and `foo_l` with similar args may raise awareness of the need for locale-sensitivity

Manual test/example:

```hack
use namespace HH\Lib\{Locale, Str};
<<__EntryPoint>>
function main(): void {
  $en_us_utf8 = Locale\create('en_US.UTF8');
  $fr_fr = Locale\create('fr_FR');
  var_dump(Str\format_l($en_us_utf8, '%f', 1.23)); // 1.230000
  var_dump(Str\format_l($fr_fr, '%f', 1.23));      // 1,230000
  var_dump(Str\length_l($en_us_utf8, '😀'));       // 1
  var_dump(Str\length_l($fr_fr, '😀'));            // 4
}
```

Differential Revision: D30458827

### DIFF
--- a/hphp/hsl/src/str/introspect_l.php
+++ b/hphp/hsl/src/str/introspect_l.php
@@ -1,0 +1,296 @@
+<?hh
+/*
+ *  Copyright (c) 2004-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the MIT license found in the
+ *  LICENSE file in the hphp/hsl/ subdirectory of this source tree.
+ *
+ */
+
+namespace HH\Lib\Str;
+
+use namespace HH\Lib\{Locale, _Private, _Private\_Str};
+
+/**
+ * Returns < 0 if `$string1` is less than `$string2`, > 0 if `$string1` is
+ * greater than `$string2`, and 0 if they are equal.
+ *
+ * For a case-insensitive comparison, see `Str\compare_ci_l()`.
+ *
+ * Locale-specific collation rules will be followed, and strings will be
+ * normalized in encodings that support multiple representations of the same
+ * characters, such as UTF8.
+ */
+function compare_l(
+  Locale\Locale $locale,
+  string $string1,
+  string $string2,
+)[]: int {
+  /* HH_FIXME[4390] missing [] */
+  return _Str\strcoll_l($string1, $string2, $locale);
+}
+
+/**
+ * Returns < 0 if `$string1` is less than `$string2`, > 0 if `$string1` is
+ * greater than `$string2`, and 0 if they are equal (case-insensitive).
+ *
+ * For a case-sensitive comparison, see `Str\compare_l()`.
+ *
+ * Locale-specific collation and case-sensitivity rules will be used. For
+ * example, case-insensitive comparisons between `i`, `I`, `ı`, and `İ` vary
+ * by locale.
+ */
+function compare_ci_l(
+  Locale\Locale $locale,
+  string $string1,
+  string $string2,
+)[]: int {
+  /* HH_FIXME[4390] missing [] */
+  return _Str\strcasecmp_l($string1, $string2, $locale);
+}
+
+/**
+ * Returns whether the "haystack" string contains the "needle" string.
+ *
+ * An optional offset determines where in the haystack the search begins. If the
+ * offset is negative, the search will begin that many characters from the end
+ * of the string. If the offset is out-of-bounds, a ViolationException will be
+ * thrown.
+ *
+ * Strings will be normalized for comparison in encodings that support multiple
+ * representations, such as UTF-8.
+ *
+ * - To get the position of the needle, see `Str\search_l()`.
+ * - To search for the needle case-insensitively, see `Str\contains_ci_l()`.
+ */
+function contains_l(
+  Locale\Locale $locale,
+  string $haystack,
+  string $needle,
+  int $offset = 0,
+)[]: bool {
+  if ($needle === '') {
+    if ($offset === 0) {
+      return true;
+    }
+    $length = length_l($locale, $haystack);
+    if ($offset > $length || $offset < -$length) {
+      throw new \InvalidArgumentException(
+        format('Offset %d out of bounds for length %d', $offset, $length)
+      );
+    }
+    return true;
+  }
+  return search_l($locale, $haystack, $needle, $offset) !== null;
+}
+
+/**
+ * Returns whether the "haystack" string contains the "needle" string
+ * (case-insensitive).
+ *
+ * An optional offset determines where in the haystack the search begins. If the
+ * offset is negative, the search will begin that many characters from the end
+ * of the string. If the offset is out-of-bounds, a ViolationException will be
+ * thrown.
+ *
+ * Locale-specific rules for case-insensitive comparisons will be used, and
+ * strings will be normalized before comparing if the locale specifies an
+ * encoding that supports multiple representations of the same characters, such
+ * as UTF-8.
+ *
+ * - To search for the needle case-sensitively, see `Str\contains_l()`.
+ * - To get the position of the needle case-insensitively, see `Str\search_ci_l()`.
+ */
+function contains_ci_l(
+  Locale\Locale $locale,
+  string $haystack,
+  string $needle,
+  int $offset = 0,
+)[]: bool {
+  if ($needle === '') {
+    if ($offset === 0) {
+      return true;
+    }
+    $length = length_l($locale, $haystack);
+    if ($offset > $length || $offset < -$length) {
+      throw new \InvalidArgumentException(
+        format('Offset %d out of bounds for length %d', $offset, $length)
+      );
+    }
+    return true;
+  }
+  return search_ci_l($locale, $haystack, $needle, $offset) !== null;
+}
+
+/**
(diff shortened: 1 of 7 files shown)


hphp/hsl/src/str/introspect_l.php
```diff
@@ -0,0 +1,296 @@
...
+  Locale\Locale $locale,
+  string $string,
+  string $suffix,
+)[]: bool {
+  /* HH_FIXME[4390] missing [] */
+  return _Str\ends_with_ci_l($string, $suffix, $locale);
+}
+
+/**
+ * Returns the length of the given string in characters.
```
lexidor, line 165: Something that might be important to note here, `Str\length()` is O(1).
Is `Str\length_l()` `O(n)` for every call?

hphp/hsl/src/str/introspect_l.php
```diff
@@ -0,0 +1,296 @@
...
+ *
+ * An optional offset determines where in the haystack (from the beginning) the
+ * search begins. If the offset is negative, the search will begin that many
+ * characters from the end of the string and go backwards. If the offset is
+ * out-of-bounds, a ViolationException will be thrown.
+ *
+ * - To simply check if the haystack contains the needle, see `Str\contains()`.
+ * - To get the first position of the needle, see `Str\search()`.
+ *
+ * Previously known in PHP as `strrpos`.
```
lexidor, line 245: Are these `Previously known in PHP as ...` comments accurate? I would have expected an `mb_` function here.

hphp/hsl/src/str/transform_l.php
```diff
@@ -0,0 +1,314 @@
...
+ * Locale-specific rules for case-insensitive comparisons will be used, and
+ * strings will be normalized before comparing if the locale specifies an
+ * encoding that supports multiple representations of the same characters, such
+ * as UTF-8.
+ *
+ * - For a case-sensitive search/replace, see `Str\replace_l()`.
+ * - For multiple case-sensitive searches/replacements, see `Str\replace_every_l()`.
+ * - For multiple case-insensitive searches/replacements, see `Str\replace_every_ci_l()`.
+ */
+// not pure: str_ireplace uses global locale for capitalization
```
lexidor, line 146: Is this comment about impurity accurate? We're passing in the Locale, so why would it use the global locale?

lexidor (commented):
We've inherited named from the non Locale aware hsl, like starts_with and trim_left. How do these functions operate on ltr text?
---
About that UTF.8 normalization that keeps popping up, does this mean that `Str\ends_with_l()` can return true even with differing final bytes?

hphp/hsl/src/str/introspect_l.php
```diff
@@ -0,0 +1,296 @@
...
+  /* HH_FIXME[4390] missing [] */
+  return _Str\strcasecmp_l($string1, $string2, $locale);
+}
+
+/**
+ * Returns whether the "haystack" string contains the "needle" string.
+ *
+ * An optional offset determines where in the haystack the search begins. If the
+ * offset is negative, the search will begin that many characters from the end
+ * of the string. If the offset is out-of-bounds, a ViolationException will be
```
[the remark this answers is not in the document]
fredemmott: whoops, good catch. We are actually throwing InvalidArgumentException now.

fredemmott:
> We've inherited named from the non Locale aware hsl, like starts_with and trim_left. How do these functions operate on ltr text?

Good question; it's whatever ICU does, which isn't super clear to me. I'll hopefully be able to dig into some examples/tests next week.

If you feel like experimenting, the `HH\Lib\_Private\_Str` functions backing these have been in builds for a while now and are usable

> About that UTF.8 normalization that keeps popping up, does this mean that Str\ends_with_l() can return true even with differing final bytes?

Yes. `strip_suffix_l()` will do the right thing though. I need to document clearly that `Locale\c()` gets you byte-wise operations, and `strip_suffix()` will shortly be equivalent to `strip_suffix_l(Locale\c(), ...)`

I see how this isn't necessarily what people will always expect, but:
- sometimes it will depending on what they're actually doing
- "if you want byte-exact, use the non-_l form" seems to be the most consistent solution to varying expectations


I'm also tempted to add `Locale\bytes()` as an alias for `Local\c()` just for clarity here.

It's also important to note that for all single-byte locales, HHVM already does byte-based operations in many contexts, not C-string operations - for example, `HH\Lib\Str\length("foo\0bar")` is 7, not 3

hphp/hsl/src/str/introspect_l.php
fredemmott: That depends on the locale; for UTF-8 yes, but single-byte locales, no.

UTF-16 would also be O(n), while UCS-2 would be O(1)

fredemmott:
Also ack on all the comments about inaccurate/badly copy-pasted doccomments. You're right on all of them.

Also, why UTF-16 in the future: every UTF-8 op involves an implicit conversion to/from UTF-16 (as ICU's UnicodeString class is UTF-16), which is hugely wasteful if you're doing a bunch of them. ICU is starting to get some native UTF-8 support which we can move to in the future, but it's not enough yet, or widely available.

For things doing lots of unicode string work, they will get *much* better performance converting to/from UTF-16 exactly once, and doing all processing in UTF-16.

lexidor:
> I'm also tempted to add Locale\bytes() as an alias for Locale\c() just for clarity here.

Your temptation is justified, I would have expected `Str\length_l(Locale\c(), "a\0b")` to be 1, because, well, C.

fredemmott:
Maybe I should remove `c()` entirely, and just doccomment it as "This is close to the "C" locale in posix systems"

RE strlen, HHVM already has an assumption that strings are well-formed for purpose, which is why we don't actually call strlen. For now I'm sticking with it for consistency, but I'm not convinced it's the long-term right thing if locale != bytes

fredemmott:
> Maybe I should remove c() entirely, and just doccomment it as "This is close to the "C" locale in posix systems"

The reason for this is even ignoring the bytes/strlen stuff, there's a lot of confusion about whether `"C"` means 'the current libc locale' or 'the special basic locale'

lexidor:
> every UTF-8 op involves an implicit conversion to/from UTF-16

Well, that would indeed consume a lot of time.

> For things doing lots of unicode string work, they will get much better performance converting to/from UTF-16 exactly once, and doing all processing in UTF-16.

I'd need some help from the typesystem here. Having utf-16 strings escape their context and getting mixed with utf-8 (output) doesn't trigger any static analysis errors, no runtime errors. This just corrupts your strings. I might just wrap all these utf-16ed strings into a marker class with `->toUtf8()` and `->toUtf16()` methods on it.

fredemmott:
> I'd need some help from the typesystem here. Having utf-16 strings escape their context and getting mixed with utf-8 (output) doesn't trigger any static analysis errors, no runtime errors. This just corrupts your strings. I might just wrap all these utf-16ed strings into a marker class with ->toUtf8() and ->toUtf16() methods on it.

I don't really see this as a new problem, but it is one it would be good for us to solve; I'm thinking:
- `Str\foo_l()` are good as-is: explicitly specify the locale/encoding, it's on you to do it right, matching the current mb_ functions
- consider "UTF-16 is fast" an implementation detail. Add `UStr\foo()` and `newtype UnicodeString = string;` , and `UnicodeString <-> (Locale, string)` conversions - which would use UTF-16 for the forseeable future, but not be guaranteed to in the future.
  - `UStr\length(UnicodeString): int`
  - `UStr\starts_with(UnicodeString, UnicodeString): int`
  - `UStr\starts_with_ci(UnicodeString, UnicodeString): int` // character folding rules
  - `UStr\starts_with_ci_l(Locale, UnicodeString, UnicodeString): int` // eg turkish i aware
  
  I'm not sure if `UStr\foo_l` should exist in that design, though if it did, it would ignore LC_CTYPE.

PR #8902, opened 2021-09-27
Agent graph generation

tobielf:
Summary:
A Python binary that takes a set of parameters about the agent graph to produce a valid agent graph for the next phase.

The metrics included so far: 
1. number of agents.
2. number of roots.
3. number of leaves.
4. number of infra agents.
5. number of product agents.
6. number of levels and the agent distribution at each level.
7. infra agent profile.
8. product agent profile.

Metric 1, 2, 6 specified in an array of integers as the input.
[2, 4, 4] means we have 10 agents in total, 2 of these are roots. And these agents are distributed in three levels with number of 2, 4, 4 agents at each level accordingly.

The number of leaves in this case is 4 at least. The user can specify a minimum number other than 4. For example, if we specify the leaves to be 5, which means there is one agent in the second level didn't connect to the third level. The sum of infra agents and product agents couldn't exceed the number of total agents.

The profile for each type of the agent, specified how many other agents could related with. For example, infra agent in degree [2, 10] means an infra agent should at least depend on other 2 agents, and couldn't exceed 10.

A set of customizable boundary functions as well as a set of customizable connector functions are provided for easy experimenting on different scenarios in the future.

Test Plan:
Under `GeneratingAgentDistributionTest`, three test cases covered the single level, multiple level, and invalid level inputs.

Under `GeneratingAnAgentGraphTest`, one test case covered the usage of customizable function/parameters to get variety results, the other one test case covered the "unsatisfiable" case.

### DIFF
--- a/hphp/hack/src/hh_codesynthesis/agentGraphGenerator.py
+++ b/hphp/hack/src/hh_codesynthesis/agentGraphGenerator.py
@@ -154,6 +154,21 @@ class AgentGraphGenerator:
     def add_edge(self, left_agent: int, right_agent: int) -> None:
         self.edges[left_agent].add(right_agent)
 
+    def validate_range_in_profile(self, degree: List[int], direction: str) -> bool:
+        for node, degree in enumerate(degree):
+            # Select a right profile to use.
+            agent_profile = self.solving_context.infra_agent_profile
+            if node in self.product_agents:
+                agent_profile = self.solving_context.product_agent_profile
+
+            # Check the degree within the range or not.
+            if (
+                degree < agent_profile[direction][0]
+                or degree > agent_profile[direction][1]
+            ):
+                return False
+        return True
+
     def validate(
         self,
         customize_validator: Optional[
@@ -175,33 +190,13 @@ class AgentGraphGenerator:
             for x in depends_on:
                 out_degrees[x] += 1
 
-        for node, degree in enumerate(in_degrees):
-            agent_profile = self.solving_context.infra_agent_profile
-            if node in self.product_agents:
-                agent_profile = self.solving_context.product_agent_profile
+        # Validate the degrees are in the range specified by the profile.
+        if self.validate_range_in_profile(in_degrees, "in_degree") is False:
+            return False
+        if self.validate_range_in_profile(out_degrees, "out_degree") is False:
+            return False
 
-            if (
-                degree < agent_profile["in_degree"][0]
-                or degree > agent_profile["in_degree"][1]
-            ):
-                raise RuntimeError(
-                    "node: {0}, degree: {1}, models: {2}".format(
-                        node, degree, self._raw_model
-                    )
-                )
-                return False
-
-        for node, degree in enumerate(out_degrees):
-            agent_profile = self.solving_context.infra_agent_profile
-            if node in self.product_agents:
-                agent_profile = self.solving_context.product_agent_profile
-
-            if (
-                degree < agent_profile["out_degree"][0]
-                or degree > agent_profile["out_degree"][1]
-            ):
-                return False
-
+        # Customized range function needs a customized validator.
         if customize_validator is not None:
             return customize_validator(
                 self.infra_agents, self.product_agents, in_degrees, out_degrees
--- a/hphp/hack/src/hh_codesynthesis/agentGraphGenerator.py
+++ b/hphp/hack/src/hh_codesynthesis/agentGraphGenerator.py
@@ -11,7 +11,7 @@
 import argparse
 import sys
 
-from typing import List, Optional, Set, Union
+from typing import Callable, Dict, List, Optional, Set, Union
 
 import clingo
 from clingo.symbol import Number, Symbol
@@ -23,11 +23,27 @@ class AgentGraphClingoContext:
     different settings. Refer to the test case `test_small_agent_graph`
     for more detail usages."""
 
-    number_of_leaves = 30
-    number_of_infra_agents = 20
-    number_of_product_agents = 60
-    infra_agent_profile = {"in_degree": [0, 10], "out_degree": [5, 100]}
-    product_agent_profile = {"in_degree": [5, 20], "out_degree": [0, 5]}
+    def __init__(
+        self,
+        number_of_leaves: int,
+        number_of_infra_agents: int,
+        number_of_product_agents: int,
+        infra_agent_profile: Dict[str, List[int]],
+        product_agent_profile: Dict[str, List[int]],
+    ) -> None:
+        if (
+            number_of_leaves <= 0
+            or number_of_infra_agents <= 0
+            or number_of_product_agents <= 0
+            or len(infra_agent_profile) < 2
+            or len(product_agent_profile) < 2
+        ):
+            raise RuntimeError("Invalid agent graph metrics.")
+        self.number_of_leaves = number_of_leaves
+        self.number_of_infra_agents = number_of_infra_agents
+        self.number_of_product_agents = number_of_product_agents
+        self.infra_agent_profile = infra_agent_profile
+        self.product_agent_profile = product_agent_profile
 
     def nl(self) -> Symbol:
         return Number(self.number_of_leaves)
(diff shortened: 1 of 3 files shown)


hphp/hack/src/hh_codesynthesis/agentGraphGenerator.py
```diff
@@ -0,0 +1,266 @@
...
+
+
+def main() -> int:
+    agent_graph = AgentGraphGenerator()
+
+    agent_graph.on_model = agent_graph.generate_raw
+
+    # Parse the arguments
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--agents", nargs="+", type=int)
```
jamesjwu, line 225: Can we annotate this with a comment on what each number after the first number means?

hphp/hack/src/hh_codesynthesis/agentGraphGenerator.py
```diff
@@ -0,0 +1,266 @@
...
+    def product_agent_out_degree_high(self, agent: Number) -> Symbol:
+        return Number(self.product_agent_profile["out_degree"][1])
+
+    def product_agent_in_degree_low(self, agent: Number) -> Symbol:
+        return Number(self.product_agent_profile["in_degree"][0])
+
+    def product_agent_in_degree_high(self, agent: Number) -> Symbol:
+        return Number(self.product_agent_profile["in_degree"][1])
+
+    def agent_connector(
```
jamesjwu, line 65: Please add a comment on what this function actually does, it's confusing from an outsider to review

hphp/hack/src/hh_codesynthesis/agentGraphGenerator.py
```diff
@@ -0,0 +1,266 @@
...
+            return Number(0)
+
+    def agent_type_connector(
+        self, source_agent: Number, target_agent: Number
+    ) -> Symbol:
+        src_agent_number = source_agent.number
+        tgt_agent_number = target_agent.number
+        # Since clingo doesn't support Boolean return, we are returning
+        # Number(1) for true, and Number(0) for false.
+        if tgt_agent_number * 3 + 200 < src_agent_number:
```
jamesjwu, line 94: Where does this number come from? What does *3 + 200 do?

jamesjwu (changes requested):
Overall, really good; I would like to see a bit better testing and have some minor questions.

hphp/hack/test/hh_codesynthesis/agentgraphgenerator_test.py
```diff
@@ -0,0 +1,175 @@
...
+            if agent.number < 2:
+                return Number(0)
+            else:
+                return Number(2)
+
+        AgentGraphClingoContext.infra_agent_in_degree_low = (
+            customize_infra_agent_in_degree_low
+        )
+
+        exp = (
```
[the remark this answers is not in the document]
tobielf: I was thinking about the graphviz as well, but the graph for hundred nodes looks messy so I use a function to collect duplicated agents and print out the statistic information.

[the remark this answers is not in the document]
tobielf: Sounds good, I think I can add an analysis function in the test suite to evaluate the generated graph is matching the input metrics or not. Or maybe we can put this under a util under the "src"? So we can use it in our "evaluate" method as well.

hphp/hack/src/hh_codesynthesis/agentGraphGenerator.py
tobielf: It's like a hash function I created during the experiment on agent graph generation. Can define any other function to provide a set of agents for the current agent to depend on. Then the solver to choose a subset from this.

hphp/hack/src/hh_codesynthesis/agentGraphGenerator.py
```diff
@@ -0,0 +1,266 @@
...
+def main() -> int:
+    agent_graph = AgentGraphGenerator()
+
+    agent_graph.on_model = agent_graph.generate_raw
+
+    # Parse the arguments
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--agents", nargs="+", type=int)
+
+    parser.add_argument("--number_of_infra_agents", type=int)
```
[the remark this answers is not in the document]
tobielf: I plan to add the json configuration in a higher level file. We are providing a library function in `agentGraphGenerator.py` and the `main` in this file allows user to run this file separately. Like can do dry run, and easier to experiment with different metrics with command line options. (don't have to change the config file back and forth). Plan to turn current `hh_codesynthesis.py` to `agentGenerator.py`, adding a new `hh_codesynthesis.py` with json configuration in it.

hphp/hack/src/hh_codesynthesis/agent_graph_generator.lp
```diff
@@ -0,0 +1,92 @@
...
+:- out_degree_of_agent(X, D), product_agent(X), D >= @product_agent_out_degree_high(X).
+
+:- in_degree_of_agent(X, D), product_agent(X), D < @product_agent_in_degree_low(X).
+:- in_degree_of_agent(X, D), product_agent(X), D >= @product_agent_in_degree_high(X).
+
+% $$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$
+% Edge (Agent relationships) requirement.
+% =============================================================================
+% If X, Y is number, not string, we can simply write without XN, YN.
+{depends_on(Y, X) : @agent_connector(Y, X, YLevel, XLevel) != 0} :-
```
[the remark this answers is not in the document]
tobielf: This is a choice rule, so in Python interface, `@agent_connector` will decide this rule is enable or disable for agent pair <X, Y>. If it's enabled, the solver can choose to form an edge or not. The edge constrained by other requirements. For example, we said agent 2, 3, 4, 5 could all depend on agent 1 (by our `@agent_connector`), then we also have a constraint says, "agent 1 at most can have two other agents depend on it". The solver have to pick one pair of agents from {2, 3}, {2, 4}, {2, 5}, {3, 4}, {3, 5}, {4, 5} combinations.

We can adjust `@agent_connector` to add more randomness.

hphp/hack/src/hh_codesynthesis/agentGraphGenerator.py
tobielf: I am thinking about a "pseudo random number generator" here. With the same seed during unit test, we should get the same output. And the user can change to a seed based on UNIX time if they want variation. And the total candidates we selected for each agent should be in O(C), e.g a constant like 100 or 200.

hphp/hack/src/hh_codesynthesis/agentGraphGenerator.py
```diff
@@ -0,0 +1,266 @@
...
+        "in_degree"
+    ] = args.product_agent_indegrees or [5, 20]
+    AgentGraphClingoContext.product_agent_profile[
+        "out_degree"
+    ] = args.product_agent_outdegrees or [0, 5]
+
+    if args.evaluate:
+        agent_graph.on_model = agent_graph.evaluate
+    generating_an_agent_graph(agent_distribution, agent_graph)
+    print(agent_graph._raw_model)
```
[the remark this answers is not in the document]
tobielf: I was trying to do the "C style" return 0 for success, negative number for errors. But it seems like we are throw an exception on error. Let me change the return value to None.

hphp/hack/src/hh_codesynthesis/agentGraphGenerator.py
```diff
@@ -0,0 +1,266 @@
...
+
+    def generate_raw(self, m: clingo.Model) -> None:
+        self._raw_model = m.__str__()
+
+    def on_model(self, m: clingo.Model) -> None:
+        raise RuntimeError("Must specify a valid method.")
+
+
+# Generate a set of agents using given distribution. The distribution has the
+# critical path requirement (levels) as well as how many roots are in the graph.
```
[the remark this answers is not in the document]
tobielf: Added in the lp file to describe main idea and metrics are used.

hphp/hack/test/hh_codesynthesis/agentgraphgenerator_test.py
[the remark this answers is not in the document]
jamesjwu: Is there no simpler way where we just check it in the unit test? Like just parse the words and check that the number of things depending on A is X number, etc. 

You could then use the same thing for writing a bigger program as well

hphp/hack/src/hh_codesynthesis/agentGraphGenerator.py
```diff
@@ -175,33 +190,13 @@ def validate(
...
-            if node in self.product_agents:
-                agent_profile = self.solving_context.product_agent_profile
-
-            if (
-                degree < agent_profile["out_degree"][0]
-                or degree > agent_profile["out_degree"][1]
-            ):
-                return False
+        # Validate the degrees are in the range specified by the profile.
+        if self.validate_range_in_profile(in_degrees, "in_degree") is False:
```
jamesjwu, line 194: nit: use if not

hphp/hack/src/hh_codesynthesis/agentGraphGenerator.py
```diff
@@ -154,6 +154,21 @@ def add_product_agent(self, agent_number: int) -> None:
     def add_edge(self, left_agent: int, right_agent: int) -> None:
         self.edges[left_agent].add(right_agent)
 
+    def validate_range_in_profile(self, degree: List[int], direction: str) -> bool:
+        for node, degree in enumerate(degree):
+            # Select a right profile to use.
+            agent_profile = self.solving_context.infra_agent_profile
```
jamesjwu, line 160: This would be better written with if else, I think, and would be a lot clearer what's happening

hphp/hack/src/hh_codesynthesis/agentGraphGenerator.py
```diff
@@ -119,46 +135,82 @@ class AgentGraphGenerator:
...
         self.product_agents.append(agent_number)
 
-    def edge_evaluate(self, left_agent: int, right_agent: int) -> None:
+    def add_edge(self, left_agent: int, right_agent: int) -> None:
         self.edges[left_agent].add(right_agent)
 
-    def evaluate(self, m: clingo.Model) -> None:
-        self.infra_agents = []
-        self.product_agents = []
+    def validate(
```
jamesjwu, line 172: This is great! Nit: I think having the validate script use assert instead of returning False might give better test signal if something breaks later

hphp/hack/src/hh_codesynthesis/agentGraphGenerator.py
tobielf: Also added an exception handling.

hphp/hack/test/hh_codesynthesis/agentgraphgenerator_test.py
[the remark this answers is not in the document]
tobielf: Yes, I wrote the parsing and checking under the "validate" method, so we can reuse it outside of unit test. Also can do a more thorough checking.

hphp/hack/src/hh_codesynthesis/agentGraphGenerator.py
tobielf: Also added an exception handling.

hphp/hack/src/hh_codesynthesis/agentGraphGenerator.py
tobielf: assert with error message.

hphp/hack/src/hh_codesynthesis/agentGraphGenerator.py
tobielf: We are using "assert" inside the function, so just invoke the function here.

PR #8913, opened 2021-10-14
Fix invalid inout arguments error message

vulongphan:
This PR does:
- Fix the error message for invalid `inout` type expression
- Include expect test files


vulongphan:
I just revert the changes in `expression_tree_pipe.php.exp` and the test did not pass, so maybe there is an issue with how the tests are run?
Screen Shot 2021-10-17 at 4 51 05 PM

shayne-fletcher:
@vulongphan can i ask you to squash your commits here too please?

Issue #8701, opened 2020-05-15
[ Typechecker error message ] Accessing a static property non statically, creates a suggestion for a static method, not a property


lexidor:
**Describe the bug**

Accessing a static property non statically, creates a suggestion for a static method, not a property

**Standalone code, or other way to reproduce the problem**
```HACK
class X { public static ?string $x; }
function main(): void {
  $x = new X();
  $x->x;
}
```

1. Observe error message

**Expected behavior**

```
No property 'x' in X (did you mean static property 'x'?)
```

**Actual behavior**
```
No property 'x' in X (did you mean static method 'x'?)
```

**Desktop (please complete the following information):**
 - OS: Ubuntu 18.04
 - HHVM Version: 4.56.0 LTS

**Additional context**

Found via:
https://github.com/facebook/hhvm/issues/3678#issuecomment-628972029

lexidor:
The opposite is also true. If the property is an instance property, but you access it using `::`, it still says `method` instead of `property`.

lexidor:
@lexidor (me) I remember seeing something in the commit log that fixes this late last year(?) Repro again and see if the issue has already been resolved.

lexidor:
> @lexidor (me) I remember seeing something in the commit log that fixes this late last year(?) Repro again and see if the issue has already been resolved.

Nope, that was a figment of my imagination. The repro still holds.

lexidor:
Closing, as the repro has been fixed. (Most likely in 05e9747)

PR #8914, opened 2021-10-15
Fix static method used as instance error message

vulongphan:
This PR:
- Update the error message for the case where a static method/property is called as an instance method, the error message should say it is a `static method` or `static property` depending on whether the static method or a static property is being called
- Link to #8701

### DIFF
--- a/hphp/hack/src/errors/errors.ml
+++ b/hphp/hack/src/errors/errors.ml
@@ -3179,10 +3179,13 @@ let unknown_type description pos r =
   let msg = "Was expecting " ^ description ^ " but type is unknown" in
   add_list (Typing.err_code Typing.UnknownType) (pos, msg) r
 
-let not_found_suggestion orig similar =
+let not_found_suggestion orig similar kind =
   match similar with
   | (`static, pos, v) ->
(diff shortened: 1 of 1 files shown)


shayne-fletcher:
@vulongphan can you tell me what this is about?
Screen Shot 2021-10-17 at 6 39 36 PM

i can see that you have force pushed - i assume you were attempting to squash your 9 commits but it doesn't seem to have had that effect.

vulongphan:
Yes, I tried to stash 9 commits together but it did not work out. I will look into how to do that properly

shayne-fletcher:
@vulongphan i suggest the following procedure should do what you want:
```bash
# checkout the branch
git checkout fix-static-method-used-as-instance-error-message

# reset to the commit from which it started
git reset --soft 645617ad24331a7f4d2f668deafc513879af6c89

# merge the most recent changes from facebook
git fetch origin
git merge origin/master

# commit your single change
git add hphp/hack/src/errors/errors.ml
git commit -m "Fix static method error message"

# examine `git log` here: you will find you there is now one commit 

# force push your branch
git push -f fork fix-static-method-used-as-instance-error-message:fix-static-method-used-as-instance-error-message
```

PR #8921, opened 2021-10-20
fix error message when instance method/property called as static method/property

vulongphan:
This PR: 
- Update error message when an instance method/property is used as a static method/property
- Link to this [Issue](https://gist.github.com/shayne-fletcher/4cece69915f546ea68983172aaa7318e)

### DIFF
--- a/hphp/hack/src/errors/errors.ml
+++ b/hphp/hack/src/errors/errors.ml
@@ -3177,10 +3177,19 @@ let not_found_suggestion orig similar kind =
     end
   | (`instance, pos, v) -> suggestion_message orig v pos
 
-let snot_found_suggestion orig similar =
+let snot_found_suggestion orig similar kind =
   match similar with
   | (`instance, pos, v) ->
-    suggestion_message ~modifier:"instance method " orig v pos
+    begin
+      match kind with 
+      | `static_method -> 
+      suggestion_message ~modifier:"instance method " orig v pos
+      | `class_constant -> 
+      suggestion_message ~modifier:"instance property " orig v pos
+      | `class_variable 
+      | `class_typeconst -> 
+      suggestion_message orig v pos
+    end
   | (`static, pos, v) -> suggestion_message orig v pos
 
 let string_of_class_member_kind = function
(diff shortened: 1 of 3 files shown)


vulongphan:
I have made the according changes to `errors.ml` and add test files

shayne-fletcher (commented):
Thanks!

hphp/hack/test/typecheck/instance_property_as_class_constant.php
```diff
@@ -0,0 +1,18 @@
...
+ *
+ */
+
+class X {                                                                                  
+    public ?string $foo;                                                                   
+}                                                                                          
+                                                                                           
+function main(): void {                                                                    
+    var_dump(X::foo);                                                                      
+}
```
Wilfred, line 18: Small nit: you've got trailing whitespace here. Could you remove it?

```suggestion
class X {
    public ?string $foo;
}

function main(): void {
    var_dump(X::foo);
}

```

hphp/hack/src/errors/errors.ml
```diff
@@ -3177,10 +3177,15 @@ let not_found_suggestion orig similar kind =
...
+let snot_found_suggestion orig similar kind =
   match similar with
   | (`instance, pos, v) ->
-    suggestion_message ~modifier:"instance method " orig v pos
+    begin
+      match kind with 
+      | `static_method -> suggestion_message ~modifier:"instance method " orig v pos
+      | `class_constant -> suggestion_message ~modifier:"instance property " orig v pos
+      | `class_variable | `class_typeconst -> suggestion_message orig v pos
+    end
```
Wilfred, line 3188: Please run `ocamlformat` so everything stays nicely formatted. It'll give you this:

```suggestion
    begin
      match kind with
      | `static_method ->
        suggestion_message ~modifier:"instance method " orig v pos
      | `class_constant ->
        suggestion_message ~modifier:"instance property " orig v pos
      | `class_variable
      | `class_typeconst ->
	suggestion_message orig v pos
    end
```

Wilfred (changes requested):
This is looking great! Two tiny formatting fixes, and this is good to merge :)

hphp/hack/src/errors/errors.ml
vulongphan: I ran `ocamlformat hphp/hack/src/errors/errors.ml` which gave me some output but the actual file is not changed at all. Could you please suggest what might have gone wrong here? I read into the documentation [here](https://github.com/ocaml-ppx/ocamlformat) and it seems like there is some styling option I should pass to the command as well? 
For now I am just going to manually format this part of the file

PR #8930, opened 2021-11-01
Add truncate to writable files

lexidor:
Fixes hhvm/hsl-experimental#165

I had great difficulty reading the manpage for ftruncate. I believe there are no errno errors that we need to wrap for the caller. But because I am very inexperienced with this, I have included my reasoning below.
[manpage used](https://www.man7.org/linux/man-pages/man3/ftruncate.3p.html)

_EINTR  A signal was caught during execution._
From src/os/README.md
```
### Performance

- constructing an exception requires constructing the backtrace; it should be
  avoided for 'expected' cases. For example, if it is safe to, calls are
  retried on EINTR 5 times.
  - only do this if it is transparent - i.e. the user can't easily tell if
    the call was retried or was just slow.
```
Would this be a candidate for automatic retry?

_EINVAL The length argument was less than 0._
Caught this with `_OS\arg_assert()`.

_EFBIG or EINVAL The length argument was greater than the maximum file size._
I can't know what the maximum file size is. It might throw EINVAL (which is what `_OS\arg_assert($length <= $MAX_FILE_SIZE, '...')` would have thrown) or `EFBIG`. I would catch EFBIG and throw EINVAL instead, but the next entry would incorrectly have its error code changed.

_EFBIG  The file is a regular file and length is greater than the offset maximum established in the open file description associated with fildes._
See previous entry.

_EIO    An I/O error occurred while reading from or writing to a file system._
This is disks being disks...

_EBADF or EINVAL The fildes argument is not a file descriptor open for writing._
`File\open_write_only()` already checked for write permission.

### DIFF
--- a/hphp/hsl/src/file/Handle.php
+++ b/hphp/hsl/src/file/Handle.php
@@ -49,6 +49,7 @@ interface ReadHandle extends Handle, IO\SeekableReadFDHandle {
 }
 
 interface WriteHandle extends Handle, IO\SeekableWriteFDHandle {
+  public function truncate(?int $length = null): void;
 }
 
 interface ReadWriteHandle extends WriteHandle, ReadHandle, IO\SeekableReadWriteFDHandle {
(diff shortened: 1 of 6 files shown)


fredemmott:
> manpages

For network stuff, it's also worth cross-referencing with netbsd; macos acts the same way, but the macos man pages have some mistakes, especially around sockaddr_t

fredemmott:
> EINVAL The length argument was less than 0.
Caught this with _OS\arg_assert().


The reason this one is special is that on some platforms, it's impossible to pass a negative value to libc's ftruncate, as `off_t` is often an unsigned type. As it's not checked by libc, it needs to be checked in HHVM or in Hack

Issue #8916, opened 2021-10-18
Quickfix for a 'self' that should be 'static'


shayne-fletcher:
# Quickfix for a 'self' that should be 'static'

- the problem in the following program is trying to call `self::childImplements` which is abstract. the fix is to call `static::childImplements` (learn more about this [here](https://www.php.net/manual/en/language.oop5.static.php) )
```php
    <?hh

    abstract class MyClass {
      abstract public static function childImplements(): void;

      public static function foo(): void {
        // autofix: static::childImplements();
        self::childImplements();
      }
    }
```
- by searching on the typechecker's error message in `errors.ml` we find the function of interest that needs improvement is `self_abstract_call`
- in `errors.ml` observe the definition of the type `'pos quickfix`:
```ocaml
    type 'pos quickfix = {
       title: string;
       new_text: string;
       pos: 'pos;
    }
```
- in `errors.ml` observe `add_list` takes an optional list of quickfixes:
```ocaml
    ...
    and add_list code ?(quickfixes = []) (claim : _ message) reasons =
      add_error { code; claim; reasons; quickfixes }
    ...
```
- so, we change `self_abstract_call`:
  - add a `self_pos` parameter representing the position of `self`
  - add a quickfix in the call to `add_list`
```ocaml
        let self_abstract_call meth_name self_pos call_pos decl_pos =
          add_list
            (Typing.err_code Typing.AbstractCall)
            ~quickfixes:[
                  { title = "Change to -> static::" ^ meth_name; new_text = "static::" ^ meth_name; pos = self_pos };
                ]
             ( call_pos,
              "Cannot call "
              ^ Markdown_lite.md_codify ("self::" ^ meth_name ^ "()")
              ^ "; it is abstract. Did you mean "
              ^ Markdown_lite.md_codify ("static::" ^ meth_name ^ "()")
              ^ "?" )
             [(decl_pos, "Declaration is here")]
```
- of course this will no longer compile until
- we update the `self_abstract_call` signature in `error.mli` to account for the newly added position parameter
```ocaml
            val self_abstract_call : string -> Pos.t -> Pos.t -> Pos_or_decl.t -> unit
```
  - we update the call-sites of `self_abstract_call_meth_name` so as to pass through the position for `self`
    - there are two both of which occur in the body of `check_class_get` in `typing.ml`
  - in the first instance, we can pass through `Pos.none` (see 'hphp/hack/src/utils/pos.ml')
    - this will be sufficient to recover a build
[...]

PR #8931, opened 2021-11-03
add quickfix when an abstract static method is called using self

vulongphan:
This PR:
- Pass `quickfixes` argument to `self_abstract_call`
- Find the position of keyword `self` that is stored in `_cid_pos` of `class_get_inner` then pass to `check_class_get` then pass it to `self_abstract_call` 
- Link to #8916


shayne-fletcher (commented):
this looks great @vulongphan. is there any chance we could put in a test for this please?

vulongphan:
@shayne-fletcher I have added a test and an exp file under `/home/ubuntu/hhvm/hphp/hack/test/quickfixes`. Please let me know what you think

shayne-fletcher:
@vulongphan thanks!

PR #8932, opened 2021-11-04
Fix deadlock in WatchmanConnection

jthemphill:
Summary:
I was inspired to look at WatchmanConnection after I saw HHVM deadlock while I tried to run `t QISocialContextParameterProviderTest` on WWW commit `4a918e77aa75`. The deadlock was triggered when running `HH\watchman_run()` and `HH\watchman_unsubscribe()` concurrently, invoked by build systems based on state that's not in source control.

I took a quickstack and did some GDBing to find the source of the deadlock:

* Quickstack: P466298918
* GDB notes on the deadlock: P466298910

The thread calling `HH\watchman_unsubscribe()` is holding `s_sharedDataMutex`:

https://www.internalfb.com/code/fbsource/[d6d884676c67d5863438627e54b552fc3b2b502f]/fbcode/hphp/runtime/ext/watchman/ext_watchman.cpp?lines=791-794

It is stuck trying to close its connection, which is stuck trying to wait for the `WatchmanConnection::eventBase_` thread to run this code:

https://www.internalfb.com/code/fbsource/[ec63276a6af947cd82f75d15efb1c65eb0de905e]/fbcode/watchman/cppclient/WatchmanConnection.cpp?lines=134-140

But the `WatchmanConnection::eventBase_` thread is stuck trying to acquire `s_sharedDataMutex`:

https://www.internalfb.com/code/fbsource/[d6d884676c67d5863438627e54b552fc3b2b502f]/fbcode/hphp/runtime/ext/watchman/ext_watchman.cpp?lines=130-133

So we can't release `s_sharedDataMutex` until the EventBaseThread finishes its callback, and the EventBaseThread can't finish its callback until we release `s_sharedDataMutex`. Classic deadlock.

`HH\watchman_unsubscribe()` doesn't need to block on the `WatchmanClient` actually being destroyed. Instead, we can send our `WatchmanClient` to its EventBase thread, where it will eventually be destroyed there while we're not holding `s_sharedDataMutex`.

Reviewed By: chadaustin

Differential Revision: D32157097

### DIFF
--- a/hphp/runtime/ext/watchman/ext_watchman.cpp
+++ b/hphp/runtime/ext/watchman/ext_watchman.cpp
@@ -399,7 +399,24 @@ struct ActiveSubscription {
     if (!m_alive) { return false; }
     if (m_watchmanClient.get() && m_watchmanClient->isDead()) {
       m_alive = false;
-      m_watchmanClient.reset();
+
+      // Send our WatchmanClient reference to the EventBaseThread to be
+      // destroyed there.
+      //
+      // `m_watchmanClient`'s destructor always runs some work on the EventBase
+      // thread, because every `folly::AsyncSocket` must be destroyed on its
+      // EventBase thread. Trying to destroy `m_watchmanClient` on this thread
+      // will just block us until the EventBase thread's work is done.
+      //
+      // Because we're currently holding `s_sharedDataMutex`, and we sometimes
+      // ask the EventBase thread to acquire `s_sharedDataMutex`, we sometimes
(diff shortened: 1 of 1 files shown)


jthemphill:
Notes on the deadlock fixed by this diff:

Quickstack: https://gist.github.com/jthemphill/487b41c21c35fa026f71fe82939d0413
GDB debugging session: https://gist.github.com/jthemphill/bc3d12cf93143cd66a0e90691ca74d11

PR #8935, opened 2021-11-12
Stop submoduling re2

fredemmott:
- Use release tarball instead of stale submodule
- faster builds
- more consistent with other stuff
- always defines re2 target, conditionally defines bundled_re2 target
- resolve re2 build TODO in fbmysqlclient CMake, which required this split (fooBuild was the old name for bundled_foo targets) by making it depend on re2 (but not bundled_re2)
- clean up fbmysqlclient's cmake to get it up to speed with our current conventions while I'm here

The net effect is that:
- build of re2 should be faster, as we'll be fetching a source tarball instead of gitsubmodule - and in FB infra, it's cached locally
- we won't spend time fetching re2 sources at all if the system version is new enough
- fbmysqlclient will also build faster if system re2 does not exist or is too old: previously HHVM and fbmysqlclient would each build their own copy

Test plan:

*castle and built on AWS worker.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -13,9 +13,6 @@
 [submodule "libafdt"]
 	path = third-party/libafdt/src
 	url = https://github.com/facebook/libafdt
-[submodule "re2"]
-	path = third-party/re2/src
-	url = https://github.com/google/re2.git
 [submodule "xed-mbuild"]
 	path = third-party/xed/mbuild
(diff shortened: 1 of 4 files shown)


fredemmott:
Draft as this depends on (and includes) a similar change for mcrouter, but we can't import stacked PRs.

Issue #8919, opened 2021-10-20
Add a quickfix for missing type annotations on constants


Wilfred:
## What's the problem?

Constants need a type annotation. 

```
const MY_CONST = 123;
```

We currently report this an an error:

```
Naming[2001] Please add a type hint const SomeType MY_CONST [1]

x.php:12:7
    10 |
    11 | 
[1] 12 | const MY_CONST = 123;
```

However, we could definitely provide an autofix, to make this code say this:

```
const mixed MY_CONST = 123;
```

## Implementing it

(1) You'll want to create a quickfix and pass it to `const_without_typehint`.

https://github.com/facebook/hhvm/blob/e7c9b2bf6c448ee8b2cd4ffcc2c0034c85228d87/hphp/hack/src/typing/nast_check/global_const_check.ml#L12-L15

You can use the position of the `gconst` name: `fst gconst.cst_name` to know where to insert the `"mixed "`.

(2) Add a `.php` file that needs this fix:

```
<?hh

const MY_CONST = 123;
```

You can see an example test file here: https://github.com/facebook/hhvm/blob/master/hphp/hack/test/quickfixes/bad_local.php

## Polish

Constants are usually integer or string literals. We should use a more precise type in those cases.

```
const int MY_CONST = 123;
const string MY_OTHER_CONST = "abc";

const mixed YET_ANOTHER_CONST = vec[1, "foo", null];
```

You can match on `gconst.cst_value` to see if the value is a string or integer literal.

PR #8944, opened 2021-12-01
Add type annotation quickfix for global constant variable

vulongphan:
This PR does:
- Provide a quickfix when a global constant is declared without a type annotation, the quickfix currently supports `int`, `float` and `string` literals
- Add test files for respective type annotation
- #8919

### DIFF
--- a/hphp/hack/src/errors/errors.ml
+++ b/hphp/hack/src/errors/errors.ml
@@ -1790,15 +1790,23 @@ let typeconst_concrete_concrete_override pos parent_pos =
       (parent_pos, "Previously defined here");
     ]
 
-let const_without_typehint sid =
+let const_without_typehint sid type_ =
   let (pos, name) = sid in
-  let msg =
-    Printf.sprintf
-      "Please add a type hint `const SomeType %s`"
-      (Utils.strip_all_ns name)
+  let name = Utils.strip_all_ns name in
+  let msg = Printf.sprintf "Please add a type hint `const SomeType %s`" name in
+  let (title, new_text) =
+    match type_ with
+    | "string" -> ("Add string type annotation", "string " ^ name)
+    | "int" -> ("Add integer type annotation", "int " ^ name)
+    | "float" -> ("Add float type annotation", "float " ^ name)
+    | _ -> ("Add mixed type annotation", "mixed " ^ name)
   in
-  add (Naming.err_code Naming.AddATypehint) pos msg
-
+  add_list
+    (Naming.err_code Naming.AddATypehint)
+    ~quickfixes:[Quickfix.make ~title ~new_text pos]
+    (pos, msg)
+    []
+  
 let prop_without_typehint visibility sid =
   let (pos, name) = sid in
   let msg =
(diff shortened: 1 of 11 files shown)


hphp/hack/src/errors/errors.ml
```diff
@@ -13,46 +13,31 @@ open String_utils
...
 
 let get_message_str (msg : 'a message) = snd msg
 
-type phase =
-  | Init
-  | Parsing
-  | Naming
-  | Decl
-  | Typing
+type phase = Init | Parsing | Naming | Decl | Typing
```
Wilfred, line 24: Looks like `ocamlformat` isn't picking up the configuration file, which is here:

https://github.com/facebook/hhvm/blob/master/hphp/hack/.ocamlformat

Try moving the `.ocamlformat` file to the root of the git repository? Hopefully that will fix it. If it does, you can create a pull request for that too :)

hphp/hack/src/errors/errors.ml
```diff
@@ -1790,17 +1641,25 @@ let typeconst_concrete_concrete_override pos parent_pos =
...
+    | "int" -> ("Add integer type annotation", "int " ^ name)
+    | "float" -> ("Add float type annotation", "float " ^ name)
+    | _ -> ("Add mixed type annotation", "mixed " ^ name)
   in
-  add (Naming.err_code Naming.AddATypehint) pos msg
+  add_list
+    (Naming.err_code Naming.AddATypehint)
+    ~quickfixes:[ Quickfix.make ~title ~new_text pos ]
+    (pos, msg)
+    [ (Pos_or_decl.none, "Declaration is here") ]
```
Wilfred, line 1659: This list is the extra information we show when there's a type error.

```suggestion
    []
```

For example, if I write a bad program like this:

```hack
function takes_int(int $_): void {}

function call_it(): void {
  takes_int("foo");
}
```

Then the type checker complains:

```
$ hh_client
Typing[4110] Invalid argument [1]
-> Expected int [2]
-> But got string [1]

x.php:8:13
    3 | // @format
    4 | 
[2] 5 | function takes_int(int $_): void {}
    6 | 
    7 | function call_it(): void {
[1] 8 |   takes_int("foo");
    9 | }

1 error found.
```

The "expected int", "but got string" are the extra reasons. The main error message here is "invalid argument".

In your case, there's no extra reasons we need to give.

hphp/hack/test/quickfixes/type_annotation_mixed.php
```diff
@@ -0,0 +1,3 @@
+<?hh
+
+const MY_CONST = vec[1,2,3];
```
Wilfred, line 3: Lovely thorough tests :)

Wilfred (changes requested):

hphp/hack/src/errors/errors.ml
vulongphan: I moved the `.ocamlformat` file with `~/hhvm$ mv hphp/hack/.ocamlformat .ocamlformat` and then in `errors.ml` I pressed `option+Shift+F` to reformat but nothing changed. Could you replicate this new formatting on your machine?

Wilfred (approved):

PR #8957, opened 2022-01-04
Ignore `vendor/bin/` in open source builds of Hack

fredemmott:
Previously, `vendor/bin/example.hack` would be a symlink to
`vendor/foocorp/barlib/bin/example.hack`

As of composer 2.2.0, these are shell script wrappers instead of
symlinks, so they now raise type errors.

Always ignore as we aim for Hack to be usable with an empty .hhconfig, and
all projects will need this.

Test Plan:

- cloned github.com/hhvm/hhvm-autoload
- ran `bin/hh-autoload` (not vendor/bin as the top level project is the autoloader)
- ran `hh_client`: no errors. This shows that `vendor/bin/` is ignored, and that `vendor/foo/bar` are not, otherwise `src/` would have undefined class refrences
- moved `vendor/bin/*.hack` to `src/`
- ran `hh_client`; got parse errors, not duplicate definition errors
- moved them to vendor/foo/bar , ditto


fredemmott:
Automated import is having issues, manually imported as FB D33410895

PR #8960, opened 2022-01-20
Add getopt_with_optind native function to support simple subparsers

ianhoffman:
# Rationale

The `getopt()` function as defined in PHP and C returns an `optind` argument by reference. This `optind` indicates the index of the next argument to parse, should `getopt()` be invoked again. This means it's possible to call `getopt()` multiple times, each time passing in the `optind` returned from the previous call, to incrementally parse a set of arguments. Similarly, it's possible to pass a different value for `getopt()` than the value returned from the previous call in order to start parsing at a different offset within the arguments vector.

We can implement simple subcommands on top of this behavior: for example, if we encounter a certain command, we can look for different options in subsequent calls to `getopt()` than we'd look for if we had encountered some other command. Obviously `getopt()` is not the most robust tool atop which to build a CLI framework — I'm aware we have [hh-clilib](https://github.com/hhvm/hh-clilib) for that — but given that `getopt()` is widely used, I think supporting this kind of parsing would be useful. It would be useful where I work, where `getopt()` remains our primary method of parsing command line input across thousands of CLI scripts, and where a migration to a better tool would be non-trivial. https://github.com/facebook/hhvm/pull/7906 also requested this feature.

# Implementation

We'd like to continue using `getopt()` in the usual way, since that suffices for most cases. Furthermore, we want to pass in an `inout int $optind`, and inout variables cannot be optional. Therefore, we introduce a new function, `getopt_with_optind()`. I refactored `ext_std_options.cpp` to share code between `getopt()` and `getopt_with_optind()`. I'm not much of a C++ programmer though, so if I missed something please let me know.

### DIFF
--- a/hphp/hack/hhi/stdlib/builtins_options.hhi
+++ b/hphp/hack/hhi/stdlib/builtins_options.hhi
@@ -53,7 +53,7 @@ function getmyuid();
 function getopt(string $options, $longopts = null);
 <<__PHPStdLib>>
 function getopt_with_optind(
-  string $options, $longopts vec<string>, inout int $optind): dict<string, mixed>;
+  string $options, ?vec<string> $longopts, inout int $optind): dict<string, mixed>;
 <<__PHPStdLib>>
 function getrusage(int $who = 0);
 <<__PHPStdLib>>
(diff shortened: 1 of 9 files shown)


haowanaa:
您好，来信我已收到，我将尽快给您回复。

PR #8961, opened 2022-01-22
Improve error messages for bad method visibility

### LINKED ISSUE #8959: Improve error messages for bad method visibility
## What's the problem?

If the user writes the following code:

```hack
class MyFoo {
  private function bar(): void {}
}

function example(): void {
  $mf = new MyFoo();
  $mf->bar();
}
```

We report the following error:

```
$ hh_client
Typing[4112] You cannot access this member [1]
-> This member is private [2]

x.hack:10:8
     3 | 
     4 | class MyFoo {
[2]  5 |   private function bar(): void {}
     6 | }
     7 | 
     8 | function example(): void {
     9 |   $mf = new MyFoo();
[1] 10 |   $mf->bar();
    11 | }
```

It's confusing to talk about "member", it would be clearer to say "method" here.

## Implementing It

When we check properties or methods on a class, we know if it's a method call:

https://github.com/facebook/hhvm/blob/c49c372928f9b2696586975eb4d85b2661436ec9/hphp/hack/src/typing/typing_object_get.ml#L271-L274

We can pass this boolean into the function that checks visibility:

https://github.com/facebook/hhvm/blob/c49c372928f9b2696586975eb4d85b2661436ec9/hphp/hack/src/typing/typing_object_get.ml#L470

https://github.com/facebook/hhvm/blob/c49c372928f9b2696586975eb4d85b2661436ec9/hphp/hack/src/typing/typing_visibility.ml#L241-L244

You'll eventually want to add an `if is_method then "method" else "property"` here:

https://github.com/facebook/hhvm/blob/c49c372928f9b2696586975eb4d85b2661436ec9/hphp/hack/src/typing/typing_visibility.ml#L162-L173

## Testing it

We have lots of `.exp` files in the repository that test the type checker. Run the tests, then it will suggest a `review.sh` command to help you update the `.exp` files.

ianhoffman:
Closes #8959

The biggest trouble I ran into working on this was getting my dev environment configured. I built HHVM in a docker container using docker-compose and mutagen to handle file-syncing with the container. This mostly worked, except for the files generated by `review.sh`: I had to manually copy them out of the container for some reason. Not sure why they didn't sync to the host automatically.

Just curious - what setup do people at FB use for Hack development? Do y'all build locally? And if you use docker or some other remote setup, do you sync VSCode (or some other IDE?) to an LSP running there?

I should probably take these questions over to Slack.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,8 @@ hphp/hack/cargo_home/
 /hphp/ffi/java/classes
 /hphp/ffi/java/hphp_ffi_java.h
 
+/hphp/ext_hsl_systemlib.hack
+
 /bin
 /output_gd/
 
@@ -57,6 +59,8 @@ TAGS
 /hphp/submodules
 /hphp/third_party
 /third-party/libzip/libzip.dylib
+/third-party/opam/opamDownload-prefix/
+/third-party/*/bundled_*-prefix
 
 # eclipse files
 .project
@@ -82,6 +86,9 @@ install_manifest.txt
 
 # CMake-generated files that need to stay in the source tree because
 # of ocamlbuild
+/hphp/hack/dev_env_common.sh
+/hphp/hack/dev_env_rust_only.sh
+/hphp/hack/hsl/
 /hphp/hack/src/options/buildOptions.ml
 /hphp/hack/opam.stamp
 /hphp/hack/test/.mypy_cache
--- a/hphp/hack/src/server/autocompleteService.ml
+++ b/hphp/hack/src/server/autocompleteService.ml
@@ -143,9 +143,9 @@ let autocomplete_token ac_type x =
 
 let autocomplete_id id = autocomplete_token Acid id
 
-let get_class_elt_types env class_ cid elts =
+let get_class_elt_types ~is_method env class_ cid elts =
   let is_visible (_, elt) =
-    Tast_env.is_visible env (elt.ce_visibility, get_ce_lsb elt) cid class_
+    Tast_env.is_visible ~is_method env (elt.ce_visibility, get_ce_lsb elt) cid class_
   in
   elts
   |> List.filter ~f:is_visible
@@ -294,10 +294,10 @@ let autocomplete_member ~is_static env class_ cid id =
        more stable. *)
     if is_static || match_both_static_and_instance then (
       List.iter
-        (get_class_elt_types env class_ cid (Cls.smethods class_ |> sort))
+        (get_class_elt_types ~is_method:true env class_ cid (Cls.smethods class_ |> sort))
         ~f:(add SearchUtils.SI_ClassMethod);
       List.iter
-        (get_class_elt_types env class_ cid (Cls.sprops class_ |> sort))
+        (get_class_elt_types ~is_method:false env class_ cid (Cls.sprops class_ |> sort))
         ~f:(add SearchUtils.SI_Property);
       List.iter
         (Cls.consts class_ |> sort)
@@ -306,10 +306,10 @@ let autocomplete_member ~is_static env class_ cid id =
     );
     if (not is_static) || match_both_static_and_instance then (
       List.iter
-        (get_class_elt_types env class_ cid (Cls.methods class_ |> sort))
+        (get_class_elt_types ~is_method:true env class_ cid (Cls.methods class_ |> sort))
         ~f:(add SearchUtils.SI_ClassMethod);
       List.iter
-        (get_class_elt_types env class_ cid (Cls.props class_ |> sort))
+        (get_class_elt_types ~is_method:false env class_ cid (Cls.props class_ |> sort))
         ~f:(add SearchUtils.SI_Property)
     )
   )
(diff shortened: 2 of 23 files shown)


hphp/hack/test/typecheck/ctor_privacy.php.exp
```diff
@@ -1,4 +1,4 @@
 File "ctor_privacy.php", line 17, characters 14-16:
-You cannot access this member (Typing[4112])
-File "ctor_privacy.php", line 13, characters 20-30:
-This member is private
+You cannot access this property (Typing[4112])
```
ianhoffman, line 2: It seems wrong to call a constructor a property?

hphp/hack/src/typing/typing_visibility.ml
```diff
@@ -159,16 +159,19 @@ let check_typedef_access ~use_pos ~in_signature env td =
     internal
     td.td_module
 
-let is_visible_for_obj env vis =
+let is_visible_for_obj ?(is_method = false) env vis =
```
Wilfred, line 162: This should probably be a compulsory argument, which will help you spot any other call sites that need to be updated.

```suggestion
let is_visible_for_obj ?(is_method) env vis =
```

hphp/hack/test/typecheck/ctor_privacy.php.exp
Wilfred: Yep, agreed.

.gitignore
```diff
@@ -45,6 +45,8 @@ hphp/hack/cargo_home/
 /hphp/ffi/java/classes
```
ianhoffman, line 45: I had to make all sorts of updates to my `.gitignore`, and I'm not totally sure why. Maybe something about the way I'm building HHVM is different from how y'all do it?

hphp/hack/src/server/autocompleteService.ml
```diff
@@ -143,9 +143,9 @@ let autocomplete_token ac_type x =
 
 let autocomplete_id id = autocomplete_token Acid id
 
-let get_class_elt_types env class_ cid elts =
+let get_class_elt_types ~is_method env class_ cid elts =
   let is_visible (_, elt) =
-    Tast_env.is_visible env (elt.ce_visibility, get_ce_lsb elt) cid class_
+    Tast_env.is_visible ~is_method env (elt.ce_visibility, get_ce_lsb elt) cid class_
```
ianhoffman, line 148: Wondering if there's a better way to do this... given a value of type `class_elt`, can we retrieve the fact that it references a method or property? I wasn't sure how, which is why I'm threading that information down from here.

hphp/hack/src/server/autocompleteService.ml
```diff
@@ -294,10 +294,10 @@ let autocomplete_member ~is_static env class_ cid id =
        more stable. *)
     if is_static || match_both_static_and_instance then (
       List.iter
-        (get_class_elt_types env class_ cid (Cls.smethods class_ |> sort))
+        (get_class_elt_types ~is_method:true env class_ cid (Cls.smethods class_ |> sort))
```
ianhoffman, line 297: Wondering if there's a better way to do this... given a value of type class_elt, can we retrieve the fact that it references a method or property? I wasn't sure how, which is why I'm threading that information in from here.

hphp/hack/src/typing/typing_visibility.ml
```diff
@@ -159,16 +159,19 @@ let check_typedef_access ~use_pos ~in_signature env td =
     internal
```
ianhoffman, line 159: There's actually a lot more we could do to improve the error messages in this file. Functions like `visibility_error`, `is_protected_visible`, and `is_private_visible` could all be updated in the same way we've updated `is_visible_for_obj` and `is_visible_for_class`. I held off on doing that in this PR to make sure it's something we actually want and to keep this diff manageable.

hphp/hack/src/server/autocompleteService.ml
Wilfred: It doesn't look like it. `class_elt` is defined here:

https://github.com/facebook/hhvm/blob/154e3b90833ff874ce8dcac9728754072134766e/hphp/hack/src/typing/typing_defs.ml#L86-L93

and flags only has things like 'is abstract':

https://github.com/facebook/hhvm/blob/154e3b90833ff874ce8dcac9728754072134766e/hphp/hack/src/typing/typing_defs_flags.ml#L34-L47

I suspect this is because we're trying to keep decls ('declarations', i.e. signatures) as small as possible because we have to fetch lots of them when type checking a file that depends on others. I agree the current code is pretty reasonable.

hphp/hack/src/typing/typing_visibility.ml
Wilfred: If this is something you're interested in working on in follow-up PRs, please do :)

I'd like us to avoid using "member" in all our error messages. It's less useful for users.

ianhoffman: Cool. I wonder if there's a way to do this more automatically instead of having to check in every function 😓

PR #8967, opened 2022-01-28
Remove exponential backoff in curl_exec

henryr:
Per [this discussion](https://hacklang.slack.com/archives/CAAGK9GKW/p1642121215011900), remove exponential backoff in the main loop of `curl_exec`, and replace with a fast-slow regime that polls every 10ms, backing off to 100ms if we sleep for 10 loops consecutively. 

The numbers (10ms, 100ms) are opening bids. At Slack we are polling much more frequently (500us) in the fast path without any obvious impact on CPU usage.


fredemmott:
Re-ran tests, still appears to be unrelated failures. Will dig in further if the continue to fail.

PR #8991, opened 2022-02-08
By default open a login shell in VS Code

Atry:
This PR would fix https://github.com/hhvm/packaging/issues/276

Test Plan:
Allocate an EC2 instance from template, and select a Docker image of successful build snapshot, then cherry-pick this PR, then run:

```
cd hphp && test/run test/slow/ext_hsl/ext_os_fork_execve.php
```

It should not core dump any more.


Atry:
Login shell is not necessary any more since #8992 has been fixed.

Issue #8980, opened 2022-02-01
Add quickfix for mixed functions that should return void


Wilfred:
Example bad code:

```hack
function foo(): mixed {
  echo "hello world\n";
}
```

The quickfix should set the return type hint to `void`.

```hack
function foo(): void {
  echo "hello world\n";
}
```

Wilfred:
The way I usually tackle quickfixes:

(1) Search for the error message and find where the error is emitted. In this example:

https://github.com/facebook/hhvm/blob/202003bda2d984cfb2acaf13f4afd7e0a1cd23ae/hphp/hack/src/errors/typing_error.ml#L1169-L1182

The last list should be a quickfix.

(2) Find the position where I want the position. You can make hh_single_type_check dump the full AAST ('annotated abstract syntax tree') available:

```
$ hh_single_type_check --nast ~/scratch/x.hack
[(Fun
    { fd_namespace =
      { Namespace_env.ns_ns_uses = <opaque>; ns_class_uses = <opaque>;
        ns_fun_uses = <opaque>; ns_const_uses = <opaque>; ns_name = None;
        ns_auto_ns_map = []; ns_is_codegen = false; ns_disable_xhp_element_mangling = false };
      fd_file_attributes = []; fd_mode = Mstrict;
      fd_fun =
      { f_span = [1:1-3:2]; f_readonly_this = None; f_annotation = ();
        f_readonly_ret = None; f_ret = ((), (Some ([1:27-32], Hmixed)));
        f_name = ([1:10-13], "\\foo"); f_tparams = []; f_where_constraints = [];
        f_variadic = FVnonVariadic;
        f_params =
        [{ param_annotation = (); param_type_hint = ((), (Some ([1:14-19], Hmixed)));
           param_is_variadic = false; param_pos = [1:20-24]; param_name = "$bar";
           param_expr = None; param_readonly = None; param_callconv = Pnormal;
           param_user_attributes = []; param_visibility = None }
          ];
        f_ctxs = None; f_unsafe_ctxs = None;
        f_body = { fb_ast = [([2:3-18], (Expr ((), [2:3-17], (Call (((), [2:3-11], (Id ([2:3-11], "\\var_dump"))), [], [(Pnormal, ((), [2:12-16], (Lvar ([2:12-16], $bar))))], None)))))] };
        f_fun_kind = FSync; f_user_attributes = []; f_external = false;
        f_doc_comment = None }
      })
  ]
```

[...]

PR #9000, opened 2022-02-10
Add quickfix for function labeled mixed that should return void

mshen63:
#8980


hphp/hack/src/errors/typing_error.ml
```diff
@@ -1166,7 +1167,7 @@ module Primary = struct
     let missing_assign pos =
       (Error_code.MissingAssign, (pos, "Please assign a value"), [], [])
 
-    let non_void_annotation_on_return_void_function is_async pos =
+    let non_void_annotation_on_return_void_function is_async pos hint_pos=
```
Wilfred, line 1170: Don't forget to format your code! You should be able to right click in VS Code and format it, or you can run `ocamlformat` from the command line.

```suggestion
    let non_void_annotation_on_return_void_function is_async pos hint_pos =
```

hphp/hack/src/errors/typing_error.ml
```diff
@@ -1179,7 +1179,19 @@ module Primary = struct
           async_indicator
           return_type
       in
-      (Error_code.NonVoidAnnotationOnReturnVoidFun, (pos, message), [], [])
+      let quickfix =
```
Wilfred, line 1182: I might be tempted to call this `quickfixes` as it's a list of quickfixes, not a single value.

Wilfred:
One tiny comment, but this is looking really good!

It looks like you haven't set your git name and email address yet, so GitHub is just showing 'Ubuntu' for the commits. You should probably fix that so the commits show properly in the web UI. It's not a major problem though.

hphp/hack/test/quickfixes/mixed_annotation_on_return_void_function.php.exp
```diff
@@ -0,0 +1,2 @@
+function foo(): void {}
+async function bar(): Awaitable<void> {}
```
Wilfred, line 2: The test runner also prints the name of the quickfix, so we can check it's sensible. You'll need to do the same here.

I've put up a diff so that `make test` will run this and make it easier for you to spot in future pull requests.

```suggestion
Change to `void`
Change to `Awaitable<void>`


function foo(): void {}
async function bar(): Awaitable<void> {}
```

Issue #8973, opened 2022-01-28
Add quickfix for non-existent type names


Wilfred:
Example bad code:

```hack
class MyClass {
  const type TFooBar = int;
  public function foo(this::TFooBaz $_): MyClass::TFooBaz {
    throw new Exception();
  }
}
```

The quickfix should correct the name (note "TFooBar" with "r").

```hack
class MyClass {
  const type TFooBar = int;
  public function foo(this::TFooBar $_): MyClass::TFooBar {
    throw new Exception();
  }
}
```

PR #9007, opened 2022-02-21
Add quickfix for non existent typeconst names

jewarner57:
Resolves: #8973 Add quickfix for non-existent type names


hphp/hack/test/quickfixes/non-existent-type-const-names.php
```diff
@@ -0,0 +1,9 @@
+<?hh
+
+class MyClass {
+  const type TFooBar = int;
+  const type TFoo = int;
+  public function foo(this::TFooBaz $_): MyClass::TFooBaz {
```
Wilfred, line 6: I suggest you remove the parameter here so it's easier to see what you're testing.

```suggestion
  public function foo(): MyClass::TFooBaz {
```

PR #9012, opened 2022-02-26
Add quickfix for nonexistent type name in params

mshen63:
Second part of #9007, adding quickfix for nonexistent typeconst name in params


hphp/hack/test/quickfixes/non-existent-type-const-name-in-params.php
```diff
@@ -0,0 +1,6 @@
+class MyClass {
+  const type TFooBar = int;
+  public function foo(this::TFooBaz $_) {
+    throw new Exception();
+  }
+}
```
Wilfred, line 6: Reminder about trailing newlines in your files :)

PR #9014, opened 2022-02-28
Migrate `usize` integer to explicit `u64`

muglug:
Once upon a time https://hacklang.org had an in-browser playground where people could experiment with Hack's typechecker. The in-browser playground was based on transpiled-to-JS version of the OCaml source.

As the project migrates to Rust I'm interested in creating something similar with WASM, using the much-more-polished `wasm-pack` utility for Rust. One encumbrance is the platform requirements — Rust `usize` integers are treated as `u32` inside the WASM output, which breaks AAST position information.

This PR attempts to fix that by being explicit about all numerical operations. On 64-bit platforms — all but WASM — these conversions should be no-ops, erased at compile time, since `usize` and `u64` are functionally identical.


fredemmott:
I'm not sure why I'm only seeing these errors on our buck builds, not our cmake-based builds: https://gist.github.com/fredemmott/ee76ad940c3be748a9d641f2d2cc01e1

Issue #8975, opened 2022-01-28
Add quickfix for missing `new`


Wilfred:
Example bad code:

```hack
class Person {}

function foo(): void {
  $p = Person();
}
```

The quickfix should add the `new` keyword when the function doesn't exist and we have a class of that name.

```hack
class Person {}

function foo(): void {
  $p = new Person();
}
```

Wilfred:
Closing as PR is merged :)

PR #9017, opened 2022-03-01
Add quickfix for missing new

jewarner57:
Resolves issue #8975 add quickfix for missing new.

### DIFF
--- a/hphp/hack/src/errors/typing_error.ml
+++ b/hphp/hack/src/errors/typing_error.ml
@@ -4337,7 +4337,7 @@ module Primary = struct
 
   let unbound_name_typing pos name class_exists =
     let (_, start_column) = Pos.line_column pos in
-    
+
     let quickfixes =
       match class_exists with
       | true ->
@@ -4345,7 +4345,7 @@ module Primary = struct
         [
           Quickfix.make
             ~title:("Add " ^ Markdown_lite.md_codify "new")
-            ~new_text:("new ")
+            ~new_text:"new "
             newpos;
         ]
       | false -> []
--- a/hphp/hack/src/errors/typing_error.ml
+++ b/hphp/hack/src/errors/typing_error.ml
@@ -2104,6 +2104,7 @@ module Primary = struct
     | Unbound_name of {
         pos: Pos.t;
         name: string;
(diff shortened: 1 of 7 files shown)


hphp/hack/src/typing/typing.ml
```diff
@@ -247,10 +247,15 @@ let expr_any env p e =
   (env, with_type ty Tast.dummy_saved_env e, ty)
 
 let unbound_name env (pos, name) e =
+  let class_exists =
+    match Env.get_class env name with
```
Wilfred, line 251: This API fetches any 'classish', so `name` might be an interface or trait. We shouldn't offer `new` for things that aren't classes.

hphp/hack/src/errors/typing_error.ml
```diff
@@ -4342,14 +4343,26 @@ module Primary = struct
...
-  let unbound_name_typing pos name =
+  let unbound_name_typing pos name class_exists =
+    let quickfixes =
+      match class_exists with
+      | true ->
+        let subst = String.sub name ~pos:1 ~len:(String.length name - 1) in
+        [
+          Quickfix.make
+            ~title:("Add " ^ Markdown_lite.md_codify "new")
+            ~new_text:("new " ^ subst)
```
Wilfred, line 4354: Rather than re-inserting the name of the class, I suggest you just insert the `"new "`. You can create an empty position where the start and end are the same, so you don't overwrite the existing name.

hphp/hack/src/errors/typing_error.ml
```diff
@@ -4342,14 +4343,26 @@ module Primary = struct
       lazy [(pos2, "Mixed with this kind of field (key => value)")],
       [] )
 
-  let unbound_name_typing pos name =
+  let unbound_name_typing pos name class_exists =
```
Wilfred, line 4346: I think this error occurs for any unknown name, not just function calls. Could you add a test for `$x = Person;` that we don't offer `new`?

Wilfred (changes requested):
Definitely on the right track! A few bits of polish before merging this.

Wilfred (approved):
Looks perfect, nice job :)

Issue #8993, opened 2022-02-09
Add quickfix for void functions with invalid return type


Wilfred:
Example bad code:

```hack
function foo(): int {
  echo "hello world\n";
}
```

The quickfix should set the return type hint to `void`.

```hack
function foo(): void {
  echo "hello world\n";
}
```

Wilfred:
This is essentially the opposite of #8980.

PR #9024, opened 2022-03-08
Add quickfix for void functions with invalid return type

mshen63:
For both sync and async functions where the return type should be set to void but isn't #8993


Wilfred (changes requested):
Making good progress! It looks like one test is now failing, so we'll need to fix that, but otherwise this makes sense.

hphp/hack/test/quickfixes/invalid_annotation_on_sync_return_void_function.php
```diff
@@ -0,0 +1,3 @@
+function foo(): int {
+  echo "hello world\n";
+}
```
Wilfred, line 3: Do please add a trailing newline here too :)

PR #9041, opened 2022-03-22
add quickfix for missing awaitable type

### LINKED ISSUE #8969: Add quickfix for missing Awaitable type
Example bad code:

```hack
async function foo(): void {
}
```

The quickfix should add the `Awaitable` type hint around the existing return type hint.

```hack
async function foo(): Awaitable<void> {
}
```

jewarner57:
Resolves #8969

### DIFF
--- a/hphp/hack/src/parser/aast_check.rs
+++ b/hphp/hack/src/parser/aast_check.rs
@@ -84,7 +84,7 @@ impl Checker {
       end_offset: usize,
       msg: ErrorMsg,
       quickfix_title: &str,
-      edits: Vec<(usize, usize, std::string::String)>,
+      edits: Vec<(usize, usize, String)>,
     ) {
         let quickfixes = vec![SyntaxQuickfix {
             title: quickfix_title.into(),
@@ -109,14 +109,14 @@ impl Checker {
                     }
                 }
                 None => {
-                  let (start_offset, end_offset) = &hint.0.info_raw(); 
+                  let (start_offset, end_offset) = hint.0.info_raw(); 
                   let edits = vec![
-                      (*start_offset, *start_offset, "Awaitable<".into()),
-                      (*end_offset, *end_offset, ">".into())
+                      (start_offset, start_offset, "Awaitable<".into()),
+                      (end_offset, end_offset, ">".into())
                   ];
                   self.add_error_with_quickfix(
-                    *start_offset,
-                    *end_offset,
+                    start_offset,
+                    end_offset,
                     syntax_error::invalid_async_return_hint,
                     "Make return type Awaitable",
                     edits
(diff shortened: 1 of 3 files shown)


hphp/hack/src/parser/aast_check.rs
```diff
@@ -78,6 +78,22 @@ impl Checker {
             .push(SyntaxError::make(start_offset, end_offset, msg, vec![]));
     }
 
+    fn add_error_with_quickfix(
+      &mut self,
+      start_offset: usize,
+      end_offset: usize,
+      msg: ErrorMsg,
+      quickfix_title: &str,
+      edits: Vec<(usize, usize, std::string::String)>,
```
Wilfred, line 87: The `String` type is in the Rust 'prelude', so it's available without needing the full name.

```suggestion
      edits: Vec<(usize, usize, String)>,
```

hphp/hack/src/parser/aast_check.rs
```diff
@@ -92,7 +108,20 @@ impl Checker {
...
-                None => self.add_error(&hint.0, syntax_error::invalid_async_return_hint),
+                None => {
+                  let (start_offset, end_offset) = &hint.0.info_raw(); 
+                  let edits = vec![
+                      (*start_offset, *start_offset, "Awaitable<".into()),
+                      (*end_offset, *end_offset, ">".into())
+                  ];
+                  self.add_error_with_quickfix(
+                    *start_offset,
+                    *end_offset,
```
Wilfred, line 119: You're taking a reference to this tuple, then immediately dereferencing it. You can probably remove the `&` reference, and then the `*` dereferencing shouldn't be necessary.

```suggestion
                  let (start_offset, end_offset) = hint.0.info_raw(); 
                  let edits = vec![
                      (start_offset, start_offset, "Awaitable<".into()),
                      (end_offset, end_offset, ">".into())
                  ];
                  self.add_error_with_quickfix(
                    start_offset,
                    end_offset,
```

Wilfred (changes requested):
Two tiny tweaks, but this is basically spot-on! Nice job :)

hphp/hack/test/quickfixes/add_missing_awaitable_type.php.exp
```diff
@@ -0,0 +1,5 @@
+<?hh
```
Wilfred, line 1: Sorry, missed this in my first review. Don't forget to run the tests!

The .exp files print the quickfix name, as well as the result. That helps us ensure the quickfix name is sensible. Something like this:

```suggestion
Make return type Awaitable
Change to `Awaitable<void>`
<?hh
```

Wilfred (changes requested):

PR #9005, opened 2022-02-17
Spelling

jsoref:
This PR corrects misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

The misspellings have been reported at https://github.com/jsoref/hhvm/commit/e308e3fced995da691531036c61baa6f9cd373a6#commitcomment-66884198

The action reports that the changes in this PR would make it happy: https://github.com/jsoref/hhvm/commit/dc6a358330dd6bdd7f794526304a04f71d1a363b

Note: this PR does not include the action. If you're interested in running a spell check on every PR and push, that can be offered separately.

I normally take efforts to exclude third party directories, but for various reasons I didn't here. I'm quite happy to drop any particular files / directories, just let me know where they live. Also, if there's an upstream for them, I'm generally happy to queue fixes for them.

PR #9044, opened 2022-03-25
Spelling tools

jsoref:
This is a subset of https://github.com/facebook/hhvm/pull/9005, it should be easier to review/test/clear

### DIFF
--- a/hphp/zend/zend-strtod.cpp
+++ b/hphp/zend/zend-strtod.cpp
@@ -1191,7 +1191,7 @@ static CONST double tinytens[] = { 1e-16, 1e-32 };
 #endif
 
 
-static int quorem(Bigint *b, Bigint *S)
+static int quorum(Bigint *b, Bigint *S)
 {
   int n;
   Long borrow, y;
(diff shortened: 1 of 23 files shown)


hphp/zend/zend-strtod.cpp
```diff
@@ -1191,7 +1191,7 @@ static CONST double tinytens[] = { 1e-16, 1e-32 };
 #endif
 
 
-static int quorem(Bigint *b, Bigint *S)
```
Wilfred, line 1194: FWIW this is 'quotient remainder', although it confused me too :)

Wilfred (approved):

hphp/zend/zend-strtod.cpp
jsoref: Doh. I kinda wondered if it might be something like that. But the total lack of a comment before the function left me puzzled (c.f. `Generator::getMarkedCollectable`). I figured a reviewer would set me straight if I was wrong.

Renaming the function to `quoRem` or something would help readers.

jsoref (commented):
@Wilfred: I presume you'll merge this in some magical way that deals w/ conflicts (i.e. that my work on this PR is done). Please let me know if I'm wrong.

Wilfred:
@jsoref yep, no additional work needed on approved PRs :)

Issue #8994, opened 2022-02-09
Add quickfix for misspelled attributes


Wilfred:
Bad code (note spelling of 'override'):

```
class MyParent {
  public function foo(): void {}
}

class MyChild extends MyParent {
  <<__Overide>>
  public function foo(): void {}
}
```

After quickfix:

```
class MyParent {
  public function foo(): void {}
}

class MyChild extends MyParent {
  <<__Override>>
  public function foo(): void {}
}

```

PR #9045, opened 2022-03-29
Add quickfix misspelled attribute

mshen63:
Add quickfix for misspelled attributes #8994


hphp/hack/src/typing/typing_attributes.ml
```diff
@@ -63,8 +63,22 @@ let check_implements
       then
         ()
       else
+        let all_user_attribute =
+          let bindings = SMap.bindings SN.UserAttributes.as_map in
```
Wilfred, line 67: We don't want to offer invalid attributes. If you look at the second value in this map, it says whether an attribute is for classes, for methods, etc.

https://github.com/facebook/hhvm/blob/f3546604cc93d6eea6141346aaf054aa6539dcf7/hphp/hack/src/naming/naming_special_names.ml#L347-L354

You probably want to pass in `attr_interface` and follow the pattern on line 48.

Wilfred (changes requested):
Good start, but we need to make sure we only offer appropriate attributes for the current position.

Wilfred (approved):
Tests are green, looks good :)

PR #9048, opened 2022-04-01
Pass CMAKE_C_COMPILER and CMAKE_CXX_COMPILER to thrid-party libraries

Atry:
As discussed in https://github.com/hhvm/packaging/pull/287 , there is a bug in g++ 11.2 preventing folly and thrift from being compiled. We will compile hhvm on 21.10 with gcc 10 instead. Therefore, this PR is used to pass the compiler to third-party libraries.


fredemmott (approved):
Thanks! Fix looks good, and swapping the compiler to a different one that's already in apt is a really good approach to resolving the compatibility issues

fredemmott:
This also *might* fix the internal Mac flakiness

Atry:
> This also _might_ fix the internal Mac flakiness

Yes. I suspect if #9040 does not fix the Mac flakiness because cmake would inference other settings from `CMAKE_CXX_COMPILER` and `CMAKE_C_COMPILER`.

PR #9049, opened 2022-04-01
Expand the accessibility level of protected to super classes

lexidor:
Super classes can call protected methods of sub classes
if the method is an override of a method in the super class.

```HACK
class A {
  public function example(A $can_be_b): void {
    $can_be_b->protec();
  }
  protected function protec(): void {}
}

class B extends A {
  <<__Override>>
  protected function protec(): void {
    echo 'I was called from ' . static::class;
  }
}
```

### DIFF
--- a/hphp/hack/src/server/serverHover.ml
+++ b/hphp/hack/src/server/serverHover.ml
@@ -353,7 +353,7 @@ let keyword_info (khi : SymbolOccurrence.keyword_with_hover_docs) : string =
     "A `public` method or property has no restrictions on access. It can be accessed from any part of the codebase."
     ^ "\n\nSee also `protected` and `private`."
   | SymbolOccurrence.Protected ->
-    "A `protected` method or property can only be accessed from methods defined on the current class, or methods on subclasses."
+    "A `protected` method or property can only be accessed from methods defined on the current class, or methods on subclasses, or methods on super classes that declare a method with the same name."
     ^ "\n\nIf the current class `use`s a trait, the trait methods can also access `protected` methods and properties."
     ^ "\n\nSee also `public` and `private`."
   | SymbolOccurrence.Private ->
--- a/hphp/hack/src/server/serverHover.ml
+++ b/hphp/hack/src/server/serverHover.ml
@@ -353,7 +353,7 @@ let keyword_info (khi : SymbolOccurrence.keyword_with_hover_docs) : string =
     "A `public` method or property has no restrictions on access. It can be accessed from any part of the codebase."
     ^ "\n\nSee also `protected` and `private`."
   | SymbolOccurrence.Protected ->
-    "A `protected` method or property can only be accessed from methods defined on the current class, or methods on subclasses, or methods on super classes that declare a method with the same name."
+    "A `protected` method or property can only be accessed from methods defined on the current class, or methods on subclasses, or methods on superclasses that declare a method with the same name."
     ^ "\n\nIf the current class `use`s a trait, the trait methods can also access `protected` methods and properties."
     ^ "\n\nSee also `public` and `private`."
   | SymbolOccurrence.Private ->
--- a/hphp/hack/test/hover/visibility_protected.php.exp
+++ b/hphp/hack/test/hover/visibility_protected.php.exp
@@ -1,5 +1,5 @@
 protected
-A `protected` method or property can only be accessed from methods defined on the current class, or methods on subclasses.
+A `protected` method or property can only be accessed from methods defined on the current class, or methods on subclasses, or methods on superclasses that declare a method with the same name.
 
 If the current class `use`s a trait, the trait methods can also access `protected` methods and properties.
 


Atry:
I am not sure about this. Technically the methods in the superclass and the subclass are different symbols. `A::example` is calling the `A::protec`, not `B::protec`, from the type checker's point of view.

lexidor:
> I am not sure about this. Technically the methods, A::protec and B::protec, in the superclass and the subclass are different symbols. A::example is calling the A::protec, not B::protec, from the type checker's point of view, even though A::example is calling B::protec from the runtime's point of view,

```HACK
class A {
  // B::protec()
  public function example(B $must_be_b): void {
    $must_be_b->protec();
  }
  protected function protec(): void {}
}

class B extends A {
  <<__Override>>
  protected function protec(): void {
    echo 'I was called from ' . static::class;
  }
}
```

Now it is calling `B::protec` in type space too.

Atry:
Interesting. Obviously the behavior in the implementation is different from the behavior in the document. 

- The documented behavior is like Scala: https://scastie.scala-lang.org/0dQcY2IWTDSIhnrJJwdylA
- The implemented behavior is like Java: https://replit.com/@Atry/TinyHightechNetworks#Main.java

I am not sure which behavior is desired. 

---
``` scala
class A {
  def example(mustBeB: B): Unit = {
    mustBeB.protec();
  }
  protected def protec(): Unit = {}
}

class B extends A {
  override
  protected def protec(): Unit = {
    println("I was called from " + this.getClass)
  }
}

(new A()).example(new B())
```
Scala Version: `method protec cannot be accessed as a member of (mustBeB : Playground.B) from class A.
 Access to protected method protec not permitted because enclosing class A in object Playground
 is not a subclass of class B in object Playground where target is defined`
---

``` java
class Main {
  static class A {
    // B::protec()
    public void example(B mustBeB) {
      mustBeB.protec();
    }
    protected void protec() {}
  }
  
  static class B extends A {
    @Override
    protected void protec() {
      System.out.println("I was called from " + this.getClass());
    }
  }

  public static void main(String[] args) {
    (new A()).example(new B());
  }
}
```
Java Version: it just compiles

Atry:
@AndrewDiMola what do you think?

AndrewDiMola:
Also looping in @Wilfred, as he might want to update the source of the text.

AndrewDiMola:
This reminds me that these are the only places where we talk about `protected`, and we really need a page that outlines visibility all modifiers and their consequences on scope, like https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html.
* https://docs.hhvm.com/hack/classes/constants (class constants, i.e. always public)
* https://docs.hhvm.com/hack/classes/constructors (constructors must have a modifier, like `protected`)
* https://docs.hhvm.com/hack/functions/function-references (a note on static methods that I don't quite understand)

And just made me realize we have *two* pages on Constants: https://docs.hhvm.com/hack/source-code-fundamentals/constants

Atry:
Calling a subclass method from a superclass is a code smell. I would not allow the access to protected methods like this if I were designing a new language. However, since the behavior is inherited from PHP and we did not submit a HIP to change the decision, I think the desired bahavior would be "allow".

It would be good to have a linter to detect this kind of access though.

lexidor:
> Calling a subclass method from a superclass is a code smell.

I don't agree with that. `abstract protected function doAThing(): void;` depends on calling protected methods from `B` in `A`.

If it would not allow calling `doAThing` on B within the scope of A, but it would allow:
```HACK
((): A ==> $must_be_b)()->doAThing()
```
that would be silly imo.

Every B is an A, therefore if you can do something with A, you must be able to do that same thing with B.

Wilfred:
Yeah, writing good hover docs is hard. We want a short summary, but to still cover all the important points.

We can have a longer, more precise explanation on the docs site, where we can have multiple examples without being overwhelming. Today we don't describe public/protected/private on the docs site though (we should fix that).

> Technically the methods, A::protec and B::protec, in the superclass and the subclass are different symbols.

Hack methods use dynamic dispatch (like virtual in C++). The typechecker knows that if you have a class instance, all the methods have been implemented, so it's happy for you to call an abstract method here.

> Super classes can call protected methods of sub classes if the method is an override of a method in the super class.

It depends what we mean by "a method is defined on". If a superclass has a method definition, I would argue the method is defined there. Any subclass methods need `<<__Override>>` (enforced by lint). If the parent class removes that method, the methods with `<<__Override>>` are type errors.

Hack visibility is weird. I don't want to document all the corner cases in a basic hover summary. For example, you can do funky things like this:

```
class MyParent {
  public function foo(): void {
    if ($this is MyChild) {
      $this->bar();
    }
  }
}

class MyChild extends MyParent {
  protected function bar(): void {
    echo "in MyChild\n";
  }
}

<<__EntryPoint>>
function hello_world(): void {
  (new MyChild())->foo(); // in MyChild
}
```

lexidor:
https://github.com/facebook/hhvm/pull/9049#issuecomment-1093270381

Yes, if this funky thing works, there is no way to describe that rule in a hover doc in a way that makes sense. I can't even figure out why this is allowed, let alone explain it in a Tweet. Thank you for showing me this.

PR #9054, opened 2022-04-04
[security][CVE-2022-27809] Builtins should always take int64_t, not int

Atry:
Builtins are allowed to take 32 or 64-bit integers
(int/int32_t versus int64_t). It's not clear why we allow this, as
integers from Hack are always 64-bit. A builtin taking an int will
just have the upper 32-bits truncated. Depending on the value, this
might change a negative number to a positive one, defeating various
input validity checks.
Remove support for 32-bit integers. Change all of the builtins which
took or returned an int to use int64_t instead. This required a few
fixups here and there, mainly around format strings. If the builtin
uses 32-bits ints internally, I left them. At least the truncation is
now obvious.

Test Plan:
New unit test. Beforehand this large negative input would
be turned into a large positive number, bypassing the "input cannot be
negative" check. Now that it takes int64_t, it is correctly recognized
as a negative number.

---
Initializes m_arr in ArrayInitBase

ArrayInitBase's protected constructor does not initialize
m_arr (presumably because a derived class is meant to do so). However
if a derived class' constructor throws, we'll run the ArrayInitBase
destructor, which will try to release m_arr. If m_arr hasn't been
initialized yet, it can contain garbage and we'll try to release
random memory. The protected constructor should initialize m_arr to
nullptr.

Test Plan:
New unit test which previously failed with ASAN builds

Co-authored-by: Rick Lavoie <rlavoie@fb.com>

### DIFF
--- a/hphp/runtime/base/array-init.h
+++ b/hphp/runtime/base/array-init.h
@@ -146,8 +146,9 @@ protected:
    * Only used by the constructors of derived classes.
    */
   ArrayInitBase(size_t n, CheckAllocation)
+    : m_arr{nullptr}
 #ifndef NDEBUG
-    : m_addCount(0)
+    , m_addCount(0)
     , m_expectedCount(n)
 #endif
   {}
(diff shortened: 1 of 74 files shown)


Atry:
This PR is supposed to manually synchronize the the fix from internal Meta repository to this Github repository. Therefore I will directly merge it instead of letting @facebook-github-bot push the commit.

PR #9055, opened 2022-04-05
Add fix for filtering

mshen63:
add fix for: Add quickfix for misspelled attributes #8994


Wilfred:
Fixed in https://github.com/facebook/hhvm/commit/d5f0428a8838ea4fc1450670f62971df57b60ca7 already :)

Issue #8970, opened 2022-01-28
Add quickfix for missing // FALLTHROUGH


Wilfred:
Example bad code:

```hack
enum MyEnum: int {
  X = 1;
  Y = 2;
  Z = 3;
}

function foo(MyEnum $me): bool {
  switch ($me) {
    case MyEnum::X:
      return true;
    case MyEnum::Y:
      echo "";
    case MyEnum::Z:
      return false;
  }
}
```

The quickfix should add the fallthrough comment, which the type checker looks for.

```hack
enum MyEnum: int {
  X = 1;
  Y = 2;
  Z = 3;
}

function foo(MyEnum $me): bool {
  switch ($me) {
    case MyEnum::X:
      return true;
    case MyEnum::Y:
      echo "";
      // FALLTHROUGH
    case MyEnum::Z:
      return false;
  }
}
```

PR #9056, opened 2022-04-06
Add quickfix switch fallthrough

mshen63:
Add quickfix for missing // FALLTHROUGH #8970


hphp/hack/src/errors/nast_check_error.ml
```diff
@@ -445,7 +446,30 @@ let illegal_context pos name =
...
+      fallthrough_comment
+  in
+
+  let quickfixes =
+    match next_pos with
+    | None -> []
+    | Some next_pos ->
+      let (_, start_col) = Pos.line_column next_pos in
+      let new_pos = Pos.set_col_end (start_col - 5) next_pos in
+      let new_text = addspaces "  // FALLTHROUGH\n" (start_col - 5) in
```
Wilfred, line 463: Why the `- 5` here?

hphp/hack/src/errors/nast_check_error.ml
```diff
@@ -445,7 +446,30 @@ let illegal_context pos name =
...
     []
 
-let case_fallthrough switch_pos case_pos =
+let case_fallthrough switch_pos case_pos next_pos =
+  let rec addspaces fallthrough_comment num_spaces =
+    if num_spaces > 0 then
+      addspaces (fallthrough_comment ^ " ") (num_spaces - 1)
+    else
+      fallthrough_comment
+  in
```
Wilfred, line 455: This works, but appending to string in a loop is O(N^2). Some languages provide a way of doing this for efficiently, e.g. Java has `StringBuilder`.

For OCaml, you can use `String.make num_spaces ' '` ([docs here](https://ocaml.janestreet.com/ocaml-core/v0.12/doc/base/Base/String/index.html#val-make)).

hphp/hack/src/utils/pos.ml
```diff
@@ -390,10 +390,12 @@ let set_line_end line pos =
   | _ -> pos
 
 let set_col_end pos_cnum pos =
+  (* temporarily change this function to change both the start and end's columns *)
```
Wilfred, line 393: Please clean up this temporary code. You could define a separate `set_col_start` function, or even write a helper `empty_pos_from_start` function.

Wilfred (changes requested):
So far so good, but needs a few changes before we merge. Please add a test too.

hphp/hack/src/errors/nast_check_error.ml
mshen63: We only have the position for what the case is matching on, so I moved the start column 5 left to before "case "

hphp/hack/src/errors/nast_check_error.ml
mshen63: Ahh I was thinking there must be an easier way to do it!

hphp/hack/src/errors/nast_check_error.ml
Wilfred: Gotcha. Rather than using a 'magic constant' where the reader has to work that out, this would be clearer as `String.length "case: "`.

hphp/test/quick/switch_case_missing_fallthrough_comment.php
```diff
@@ -0,0 +1,18 @@
+<?hh
```
Wilfred, line 1: This test looks good, but it's in the wrong directory. `hphp/test/quick` is used for runtime tests that run quickly, whereas you want `hphp/hack/test/quickfixes`.

hphp/hack/src/typing/typing.ml
```diff
@@ -2806,12 +2808,35 @@ and case_list parent_locals ty env switch_pos cl dfl =
...
+          env
+          switch_pos
+          pos
+          ~next_pos:(Some next_pos)
+          b
+          ~last
+          ~is_default:false;
+        let (env, tcl) = case_list env (another :: rl) in
+        (env, (te, tb) :: tcl)
       | (((_, pos, _) as e), b) :: rl ->
```
Wilfred, line 2827: Copy-pasting this code makes it harder to see what's going on, and how these cases differ. Instead, you can add an extra pattern match on `rl` and you don't need to repeat yourself in the `case_list` function.

```
let next_pos = match rl with
| ((_, next_pos, _), _) :: _ -> Some next_pos
| [] -> None in
```

Wilfred (changes requested):
Nearly there, but some copy-pasted code and a path to fix.

hphp/hack/src/errors/nast_check_error.ml
```diff
@@ -445,7 +446,29 @@ let illegal_context pos name =
...
+        Pos.set_col_end (start_col - offset)
+        @@ Pos.set_col_start (start_col - offset) next_pos
+      in
+      let new_text =
+        "  // FALLTHROUGH\n" ^ String.make (start_col - offset) ' '
+      in
+
+      [
+        Quickfix.make
+          ~title:"Annotate fallthrough case with // FALLTHROUGH comment"
```
Wilfred, line 466: This message feels a little like it's focusing on the implementation details of adding a comment. How about this?

PR #9057, opened 2022-04-06
Spelling zend

jsoref:

Note: while the top level item is `zend`, if a commit overlaps other directories, I'm lazily including the other changes. If they can safely be dropped from a thing and that makes it easier to merge, I'm happy to do so. By including full spanning commits, I reduce the risk of missing a related change (these tend to result in build/test failures which are painful) at the slight increase in merge conflicts for my general series (these are easy enough for me to clean up)

### DIFF
--- a/hphp/zend/html_tables/generate-html-table.py
+++ b/hphp/zend/html_tables/generate-html-table.py
@@ -190,7 +190,7 @@ output_mapping('mappings/CP1252.TXT', 'entity_table_cs_cp1252')
 output_mapping('mappings/KOI8-R.TXT', 'entity_table_cs_koi8r')
 #output_mapping('mappings/ROMAN.TXT', 'entity_table_cs_macroman')
 
-# table for encoding csjis is esentially empty
+# table for encoding cs-sjis is esentially empty
 write_mapping_table('entity_table_cs_sjis', [])
 
 make('ents_html401.txt', 'entity_table_html401')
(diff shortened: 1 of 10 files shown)


hphp/zend/html_tables/generate-html-table.py
```diff
@@ -190,7 +190,7 @@ def make_multicode_table(file_name, table_name):
 output_mapping('mappings/KOI8-R.TXT', 'entity_table_cs_koi8r')
 #output_mapping('mappings/ROMAN.TXT', 'entity_table_cs_macroman')
 
-# table for encoding csjis is esentially empty
+# table for encoding cs-sjis is essentially empty
```
jsoref, line 193: `cs-sjis` is used elsewhere in the repository, and there didn't seem to be any particular value in this extra spelling.

hphp/zend/zend-html.cpp
```diff
@@ -235,7 +235,7 @@ static entity_table_t ent_uni_8592_9002[] = {
   "ang", "angmsd", "angsph", "mid", "nmid", "par", "npar", "and",
   "or", "cap", "cup", "int", nullptr, nullptr, "conint", nullptr,
   /* 8752 (0x2230) */
-  nullptr, nullptr, nullptr, nullptr, "there4", "becaus", nullptr, nullptr,
+  nullptr, nullptr, nullptr, nullptr, "there4", "because", nullptr, nullptr,
```
jsoref, line 238: This is scary.

PR #9058, opened 2022-04-06
Spelling vixl

jsoref:

Note: while the top level item is `vixl`, if a commit overlaps other directories, I'm lazily including the other changes. If they can safely be dropped from a thing and that makes it easier to merge, I'm happy to do so. By including full spanning commits, I reduce the risk of missing a related change (these tend to result in build/test failures which are painful) at the slight increase in merge conflicts for my general series (these are easy enough for me to clean up)

### DIFF
--- a/hphp/vixl/a64/constants-a64.h
+++ b/hphp/vixl/a64/constants-a64.h
@@ -1465,11 +1465,11 @@ enum LSELoadOp {
   LSE_LDUMIN = 0x00007000
 };
 
-enum LSEAquireRelease {
-  LSEAquireReleaseN  = 0x00000000,
-  LSEAquireReleaseA  = 0x00800000,
-  LSEAquireReleaseL  = 0x00400000,
-  LSEAquireReleaseAL = 0x00C00000
+enum LSEAcquireRelease {
+  LSEAcquireReleaseN  = 0x00000000,
+  LSEAcquireReleaseA  = 0x00800000,
+  LSEAcquireReleaseL  = 0x00400000,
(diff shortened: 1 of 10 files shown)


hphp/vixl/a64/constants-a64.h
```diff
@@ -1465,11 +1465,11 @@ enum LSELoadOp {
...
-enum LSEAquireRelease {
-  LSEAquireReleaseN  = 0x00000000,
-  LSEAquireReleaseA  = 0x00800000,
-  LSEAquireReleaseL  = 0x00400000,
-  LSEAquireReleaseAL = 0x00C00000
+enum LSEAcquireRelease {
+  LSEAcquireReleaseN  = 0x00000000,
+  LSEAcquireReleaseA  = 0x00800000,
+  LSEAcquireReleaseL  = 0x00400000,
+  LSEAcquireReleaseAL = 0x00C00000
```
jsoref, line 1472: The changes here are by far the scariest. I'm happy to drop them if they're a problem.

Atry:
Close and reopen to trigger CI

hphp/vixl/a64/constants-a64.h
Atry: I think they are copied from https://github.com/armvixl/vixl. Maybe we should not fix typo in this library.

jsoref: It looks like that library moved here: https://github.com/Linaro/vixl, I've added it to my list.

PR #9059, opened 2022-04-06
Spelling system

jsoref:

Note: while the top level item is `system`, if a commit overlaps other directories, I'm lazily including the other changes. If they can safely be dropped from a thing and that makes it easier to merge, I'm happy to do so. By including full spanning commits, I reduce the risk of missing a related change (these tend to result in build/test failures which are painful) at the slight increase in merge conflicts for my general series (these are easy enough for me to clean up)


hphp/system/php.txt
```diff
@@ -52,7 +52,7 @@ hphp/system/php/spl/file_handling/SplFileObject.php
 hphp/system/php/spl/file_handling/SplTempFileObject.php
 
 hphp/system/php/lang/ArrayAccess.php
-hphp/system/php/lang/Serializeable.php
+hphp/system/php/lang/Serializable.php
```
jsoref, line 55: I'm pretty sure this change will be dropped (it's a single commit, not a big deal)

PR #9065, opened 2022-04-08
Pass C/C++ compiler to dune

Atry:


Atry:
I am not sure if `dune` support specifying the compiler / linker 

https://github.com/ocaml/dune/issues/3040#issuecomment-599055219

PR #9066, opened 2022-04-09
Pass C compiler to opam_setup.sh

Atry:
As discussed in https://github.com/hhvm/packaging/pull/287 , there is a bug in g++ 11.2 preventing folly and thrift from being compiled. We will compile hhvm on 21.10 with gcc 10 instead. Therefore, this PR is used to pass the compiler to subprojects.

In https://github.com/facebook/hhvm/issues/9048 and https://github.com/facebook/hhvm/pull/9052, we have added the C and C++ compiler to all third-party libraries. This diff in addition passes the compiler setting to opam as well.

`ocamlmklib` internally uses a hard-coded constant to find the C compiler and linker. As a result, the only way to specify the C compiler is to compile `ocamlmklib` with the specific C compiler.

Test Plan:

From [hhvm/packaging](https://github.com/hhvm/packaging) checkout:
```
bin/make-source-tarball && bin/make-package-in-throwaway-container
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,10 @@ hphp.log
 hphp/hack/_build/
 hphp/hack/cargo_home/
 /deps
+/obj-x86_64-linux-gnu
+/debian
+*.buildinfo
+*.changes
 
 /hphp/test/test
 /hphp/test/test_fast.inc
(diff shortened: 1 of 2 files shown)


Atry:
This PR is not necessary because OPAM can find out the C compiler from `CC` environment variable.

PR #9068, opened 2022-04-13
Add quickfix for missing async keywords

### LINKED ISSUE #8963: Add quickfix for missing async keywords
Example bad code:

```hack
function foo(): Awaitable<void> {
  await gen_bar();
}
```

The quickfix should add the `async` keyword directly before the `function` keyword. You'll need to add a new quickfix position type for this.

```hack
async function foo(): Awaitable<void> {
  await gen_bar();
}
```

jewarner57:
Resolves #8963 Add quickfix for missing async keywords.


Wilfred (approved):
Nicely done :)

PR #9072, opened 2022-04-15
Spelling runtime ext

jsoref:

Note: while the top level item is `hphp/runtime/ext`, if a commit overlaps other directories, I'm dropping the commit. By excluding spanning commits, I reduce the risk of missing a related change (these tend to result in build/test failures which are painful) at the slight increase in additional commits for my general series.


Atry:
I cannot merge this PR because it contains conflicting files

PR #9079, opened 2022-05-02
Add a Github Action to build HHVM with Nix

Atry:
This PR add Nix build files and corresponding Github Actions.

- `flake.nix` is the entrypoint of the Nix build files, which references two packages:
  - `hhvm.nix` - This hhvm package
- `.github/workflows/nix.yml` includes the Github Actions to compile and install HHVM. 
  
Other fixes
---
- Added `OPAMASSUMEDEPEXTS=1` and `OPAMNODEPEXTS=1` to `opam`, to prevent `opam` from trying to invoke `brew install` and `apt`, because all the dependencies should have been installed by Nix, Apt or Homebrew.

Test Plan:
---
See Github Actions


Atry:
This PR was based on an HHVM revision about 1 week ago and it builds successfully on my branch.
However, it fails after rebasing onto `master`. It seems related to some recent changes.

Atry:
I rebased this PR onto D36073785, but there is [another error](https://github.com/facebook/hhvm/runs/6263231468?check_suite_focus=true):

```
hhvm> thread 'main' panicked at 'couldn't create output file: Os { code: 2, kind: NotFound, message: "No such file or directory" }', src/hackc/hhbc/cargo/dump-opcodes/../../dump_opcodes.rs:63:38
hhvm> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
hhvm> make[2]: *** [hphp/hack/CMakeFiles/hhbc_ast_cbindgen.dir/build.make:79: hphp/hack/src/hackc/opcodes.rs] Error 101
hhvm> make[1]: *** [CMakeFiles/Makefile2:1923: hphp/hack/CMakeFiles/hhbc_ast_cbindgen.dir/all] Error 2
```

CMake/HPHPSetup.cmake
```diff
@@ -60,7 +60,12 @@ if(HHVM_DYNAMIC_EXTENSION_DIR)
   add_definitions(-DHHVM_DYNAMIC_EXTENSION_DIR="${HHVM_DYNAMIC_EXTENSION_DIR}")
 else()
   if(UNIX)
-    add_definitions(-DHHVM_DYNAMIC_EXTENSION_DIR="${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}/hhvm/extensions")
+    if(IS_ABSOLUTE "${CMAKE_INSTALL_LIBDIR}")
+      set(HHVM_DYNAMIC_EXTENSION_DIR "${CMAKE_INSTALL_LIBDIR}/hhvm/extensions")
```
jtojnar, line 64: Since this is using [`GNUInstallDirs`](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html), you can just switch to `CMAKE_INSTALL_FULL_LIBDIR` to get an absolute path.

Atry: Fixed in #9085

PR #9082, opened 2022-05-05
Add dev container settings

Atry:
This PR is based on #9081, with additional dev container settings.

The Docker image is based on
https://github.com/xtruder/nix-devcontainer


Atry:
By the way, dev container is much slower than a VM. The git would get stuck for every simple command, because this repository contains too many source files.

Issue #257, opened 2010-12-15
Compile Errors: libpam???


gordol:
<pre>
==> Determining latest git revision...
  -> Version found: 20101215
==> Making package: hiphop-git 20101215-1 (Wed Dec 15 13:13:56 CST 2010)
==> Checking Runtime Dependencies...
==> Checking Buildtime Dependencies...
==> Retrieving Sources...
  -> Found hphp.sh
==> Validating source files with md5sums...
    hphp.sh ... Passed
==> Extracting Sources...
==> Removing existing pkg/ directory...
==> Starting build()...
==> Connecting to githup.com GIT server....
Already up-to-date.
==> The local files are updated.
==> GIT checkout done or server timeout
-- CMAKE_PREFIX_PATH was missing, proceeding anyway
-- MySQL Include dir: /usr/include  library dir: /usr/lib
-- MySQL client libraries: mysqlclient_r
-- Found libevent: /opt/hiphop/lib/libevent.so
-- Found GD: /usr/lib/libgd.so
-- Found ICU header files in /usr/include
-- Found ICU libraries: /usr/lib/libicuuc.so
-- Skipping TCmalloc
-- Found Intel TBB
-- Found mcrypt: /usr/lib/libmcrypt.so
-- /usr/lib/libldap.so
-- Found libevent: /opt/hiphop/lib/libevent.so
-- Found libevent 1.4.5+
-- Configuring done
-- Generating done
-- Build files have been written to: /opt/hiphop/hiphop-php
</pre>

<pre>
[ 87%] Building CXX object src/compiler/CMakeFiles/hphp_analysis.dir/parser/hphp.tab.cpp.o
[ 87%] Building CXX object src/compiler/CMakeFiles/hphp_analysis.dir/parser/parser.cpp.o
Linking CXX static library ../../bin/libhphp_analysis.a
[ 87%] Built target hphp_analysis
Scanning dependencies of target hphp
[ 87%] Building CXX object src/hphp/CMakeFiles/hphp.dir/main.cpp.o
[ 88%] Building CXX object src/hphp/CMakeFiles/hphp.dir/externals.cpp.o
Linking CXX executable hphp
/usr/lib/gcc/x86_64-unknown-linux-gnu/4.5.1/../../../../lib/libc-client.a(osdep.o): In function `ssl_onceonlyinit':
(.text+0x7089): warning: the use of `tmpnam' is dangerous, better use `mkstemp'
[...]

PR #9085, opened 2022-05-06
Improve portability in HHVM's CMake build files

Atry:
This PR is a redo of previously reverted #9080. 

This PR includes improvement of portability in HHVM's CMake build files. They are necessary to let HHVM build with Nix, but these changes are not specifically to Nix, because other untested OS might also have similar behavior like Nix:

- `CMAKE_INSTALL_PREFIX` is replaced by `CMAKE_INSTALL_FULL_*`, because manually adding as `${CMAKE_INSTALL_PREFIX}/${xxx}` would point to a wrong path if `xxx` is an absolute path. Note that #9080 incorrectly changes the behavior of the CMake `install` command, because the CMake `install` command only automatically prepends the prefix to directories specified in `DESTINATION`, not in `CODE`. `CMAKE_INSTALL_FULL_*` used by this PR should be fine.
- `PAM_LIB` is added to fix https://github.com/facebook/hhvm/issues/257
- Passing `HAVE_SYSTEM_TZDATA_PREFIX` to `timelib` to customize the `zoneinfo` path instead of hard-coded `/usr/share/zoneinfo`, which does not exist on Nix.
- Passing `CMAKE_OSX_DEPLOYMENT_TARGET` to `thrift` explicitly, otherwise it would be inferred as 10.12 implicitly on Nix, and thrift does not compile when targeting macOS 10.12.
- Skip `zstd` detection by manually setting `CAN_USE_SYSTEM_ZSTD` because the original `zstd` detection script does not work for Nix

Test Plan:
Tested along with the next diff D36068249 on Github Action


third-party/mcrouter/mcrouter-CMakeLists.txt
```diff
@@ -98,5 +98,5 @@ target_include_directories(mcrouter PUBLIC "${CMAKE_CURRENT_BINARY_DIR}" "${JEMA
 
 install(
   TARGETS mcrouter
-  LIBRARY ARCHIVE DESTINATION ${CMAKE_INSTALL_PREFIX}/lib
+  LIBRARY ARCHIVE DESTINATION lib
```
jtojnar, line 101: Why not use the variable here too?

Atry: Updated

Atry: Updated

Atry: I think we should remove the `DESTINATION` because we are installing a library to the default destination.

Atry:
See https://cmake.org/cmake/help/latest/command/install.html and https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html for the installation directory handling.

PR #9100, opened 2022-05-29
Remove disposability from doc comment

lexidor:
This note survived the removal of disposable handles.
The nextConnectionAsync method is the only method (no non-disposable variant exists).


Atry:
Close and reopen this PR to trigger CI

Atry:
Thank you! @lexidor

Issue #8239, opened 2018-06-21
HHI: previous exception in exception's constructor needs to allow \Throwable, not just \Exception


fredemmott:
### HHVM Version

3.27

### Standalone code, or other way to reproduce the problem

```
<?hh

class Foo extends \Exception {
  public function __construct(
    Throwable $previous,
  ) {
    parent::__construct('hello', 0, $previous);
  }
}
```

### Actual result

```
src/__Private/LinterException.php:23:57,65: Invalid argument (Typing[4110])
  /private/tmp/hh_server/hhi_3b2ecd0f/exceptions.hhi:76:16,24: This is an object of type Exception
  src/__Private/LinterException.php:21:6,15: It is incompatible with an object of type Throwable
```

### Expected result

No errors

VitorFalcao:
Hello @fredemmott I am looking to contribute to HHVM, since I program in Hack in a daily basis and I thought this issue could be a good start. Is the solution just a quick change from `?Exception` to `?Throwable` in the following line? https://github.com/facebook/hhvm/blob/25162c745586cd5682db5825ffff0db75c62d6e7/hphp/hack/hhi/exceptions.hhi#L77

fredemmott:
- Add tests to runtime (hphp/test/slow) and typechecker (hphp/hack/test/)
- Make that change to the HHI
- If runtime tests fail, make a similar change to the runtime (probably somewhere in `hphp/system/php` or somewhere in `hphp/runtime/ext/`)

VitorFalcao:
Hi @fredemmott, after writing the runtime test, the error went away. Do I still need to write a test for the typechecker?

VitorFalcao:
Hi @fredemmott sorry for the n00b question, but I am having trouble running the typechecker tests. How do I run them, please?

xixixao:
@fredemmott Couple questions:

1) https://docs.hhvm.com/hack/statements/throw says throw only supports \Exception subclasses, but it supports \Throwable implementations.
2) What's the difference between \Error and \Exception? Should we have just one of them?

fredemmott:
1. can you file an issue against docs.hhvm.com?
[...]

PR #9104, opened 2022-06-09
Update Exception::__construct($previous)

lexidor:
The type should be nullable Throwable since 4.162.0. There may have been good reasons for leaving it though...

### DIFF
--- a/hphp/hack/hhi/exceptions.hhi
+++ b/hphp/hack/hhi/exceptions.hhi
@@ -94,8 +94,8 @@ class Exception implements Throwable {
    * This method isn't pure. Consider using IExceptionWithPureGetMessage
    */
   public function getMessage()[defaults]: string;
-  final public function getPrevious()[]: ?Exception;
-  public final function setPrevious(Exception $previous)[write_props]: void;
+  final public function getPrevious()[]: ?Throwable;
+  public final function setPrevious(Throwable $previous)[write_props]: void;
   public function getCode()[]: int;
   final public function getFile()[]: string;
   final public function getLine()[]: int;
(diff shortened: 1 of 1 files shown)


lexidor:
Fixes #8239

lexidor:
https://github.com/facebook/hhvm/blob/e8ffdc22c2e50a8cbb8be312718116a46dee8476/hphp/hack/hhi/exceptions.hhi#L97 is definitely wrong!

Atry:
Some failure tests:

hphp/hack/test/tast/contravariant_solve.php hphp/hack/test/tast/finally.php hphp/hack/test/tast/reactive.php hphp/hack/test/tast/control_flow/do_throw_while.php hphp/hack/test/tast/control_flow/while_throw.php hphp/hack/test/tast/re_prefixed_string/re_prefixed_string.php

Atry:
@lexidor you can rebase this PR onto master to see the test result from GitHub Actions recently added.

Atry:
Close and reopen this PR to trigger the CI

lexidor:
Ah, the tast contains a a lot of explicit dumps of the signatures. So the test fails because the expect file contains
```hack
function(string=, int=, ?\Exception=)
```

I'll rebase and try to get my local test suite running. That should find all the places where the sig is included in some expect output.

Atry:
The CI on GitHub Actions pass because it runs the quick tests only, not the full set of tests.
https://github.com/facebook/hhvm/blob/e25051f8e9e824966ac7898715411e58a0f21659/flake.nix#L35

lexidor:
This PR is stale and conflicts with HEAD. Closing.

PR #9144, opened 2022-08-11
Use first-party dependencies from third-party/*/src

Atry:


Atry:
Currently we are downloading tarballs from GitHub like this:

https://github.com/facebook/hhvm/blob/03bdbe8329a56e806738eb7e183071c200f5a8ff/third-party/fb-mysql/CMakeLists.txt#L16

However, the tarballs does not contain git submodules, and it will result in errors once we updated it to latest version, according to the build log:

https://github.com/facebook/hhvm/runs/7797790188?check_suite_focus=true

We did not encounter the error previously because fb-mysql was able to build without having the files in the submodules. However, since D32151293, it now requires `rocksdb` submodule to be available in order to build.

Issue #9106, opened 2022-06-30
Automatically update first-party dependencies


Atry:
**Is your feature request related to a problem? Please describe.**
Currently HHVM dependency versions are hard coded in the CMake files like this: https://github.com/facebook/hhvm/blob/3c1feb001d17784ba0e043b3fa59fb9c933b1693/third-party/fb-mysql/CMakeLists.txt#L13-L19

However, HHVM is co-evolved with other OSS projects maintained by Meta Platforms, including folly fbthrift, fb-mysql, etc. When an upstream change is made in a first-party dependency, we will have to wait for the next release of the dependency and then modify the call-site in HHVM. 

**Describe the solution you'd like**
We need a way to automatically update first-party dependencies to make first-party change be available to HHVM as soon as possible.

**Describe alternatives you've considered**
Otherwise we can keep the current approach to manually update dependencies, but it would slow down the HHVM development and is not compatible with the mono repo practice in Meta Platforms.

Atry:
Hi @fredemmott,
Previously the first-party dependency was using submodules, which is easier to update. You changed it to CMake settings last year, which is harder and slower to update. I understand CMake approach makes sense when the dependency is optional, which could be solved from the OS, but I don't understand the purpose of the CMake approach for first-party dependencies, given that we should always want to use the head revision of first-party dependencies.

@fredemmott, do you know why we download first-party dependencies from CMake?

fredemmott:
* Git submodules are much more painful to use; when they were submodule based, a standard part of debugging build failures was `rm -rf third-party; rm -rf build/third-party; git checkout third-party; git submodule update --init --recursive; make`
* it greatly improves build speed and reliability on internal build systems given they can be - and are - cached


> given that we should always want the head revision of first-party dependencies.

Builds need to be reproducible; HHVM today is unlikely to be buildable with folly in 6 months' time

Additionally, head is often not in sync between folly/thrift/... - the tags are.

> and is not compatible with the mono repo practice in Meta Platforms.

No approach is both good externally and good for a mono repo - one is a mono repo with cross-project atomic commits, one isn't, and they have different requirements. Pretending to have atomic commits on github **will** break the ability to use `git bisect` for issues in public builds.

----

It's also important to note: auto-updating is **entirely independent** of submodules vs externalproject_add. They're formulaic, and it would be relatively straightforward to change them to be more formulaic.

fredemmott:
[...]

PR #9150, opened 2022-08-16
Move fb-mysql to a submodule

Atry:
As @fredemmott suggested in https://github.com/facebook/hhvm/issues/9106#issuecomment-1171766088, we can add ShipIt mappings to improve atomicity for first-party dependencies.

This diff is the first step to stop downloading dependencies from tarballs. It temporarily uses git submodule and I will create another diff to remove the git submodule and create ShipIt mappings.

Test Plan:
---
Sandcastle jobs and GitHub Actions should pass


Atry:
This PR is replaced by 3e69567d27f215158afc4e5d219dc32021c10201

PR #9177, opened 2022-09-08
ci: add minimum GitHub token permissions for workflow

boahc077:
### Description
This PR adds minimum token permissions for the GITHUB_TOKEN in GitHub Actions workflows using https://github.com/step-security/secure-workflows.

The GitHub Actions workflow has a GITHUB_TOKEN with write access to multiple scopes.
Here is an example of the permissions in one of the workflow runs:
https://github.com/facebook/hhvm/runs/8220542290?check_suite_focus=true#step:1:19

After this change, the scopes will be reduced to the minimum needed for the following workflow:

- nix-flake-update.yml

The following workflow file already has the least privileged token permission set:

- nix.yml

### Motivation and Context

- This is a security best practice, so if the GITHUB_TOKEN is compromised due to a vulnerability or compromised Action, the damage will be reduced.
- GitHub recommends defining minimum GITHUB_TOKEN permissions.
  https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token
- The Open Source Security Foundation (OpenSSF) [Scorecards](https://github.com/ossf/scorecard) also treats not setting token permissions as a high-risk issue. This change will help increase the Scorecard score for this repository.

Signed-off-by: Ashish Kurmi <akurmi@stepsecurity.io>

### DIFF
--- a/.github/workflows/nix-flake-update.yml
+++ b/.github/workflows/nix-flake-update.yml
@@ -1,8 +1,14 @@
 on:
   schedule:
     - cron: "0 3 * * *"
+permissions:
+  contents: read
+
 jobs:
   nix-flake-update:
+    permissions:
+      contents: write  # for peter-evans/create-pull-request to create branch
+      pull-requests: write  # for peter-evans/create-pull-request to create a PR
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2.4.0


Atry:
Hi @boahc077, could you trigger a CI on your fork, checking if the GitHub Action is able to create a PR?

.github/workflows/nix-flake-update.yml
```diff
@@ -1,8 +1,14 @@
 on:
   schedule:
     - cron: "0 3 * * *"
+permissions:
+  contents: read
+
```
Atry, line 6: Do we need these permissions given that `contents: write` permission has been assigned to the only job of this workflow.

boahc077: Yes, the contents: read is not strictly needed here. Setting that at the workflow level helps if a new job is added to the workflow. In that case, the new job will have _contents: read_, which is a secure-by-default setting.

Atry (changes requested):
Could you trigger the CI from your fork, ensuring the GitHub Action is able to create a PR?

boahc077:
@Atry  It seems to be working fine: https://github.com/boahc077/hhvm/actions

Atry (approved):

Atry:
Thank you! @boahc077

PR #9183, opened 2022-09-10
Edit the installation methods and example OS names

lexidor:
Removing MacOS, because no native MacOS builds are being produced.
Modernize linux versions and add NixOS (not officially supported at time of writing).
Installation methods: Differentiate between deb universal and deb release specific, add nix package.


lexidor:
No new minor packages have been built since hhvm 4.172 was release a year ago. Changing these OS names does not matter until new releases are being made. The hhvm team is not able to do so at this time (see https://hhvm.com/blog).

PR #9194, opened 2022-09-14
Move tests to the checkPhase

Atry:
Previously we put the tests in `checks.quick` so `nix flake check` would run the tests against the `packages.hhvm`. However, since we are going to add another package `hhvm_clang`, we will need tests against every package.
This PR instead move tests to the `checkPhase`, letting the tests be package specific, i.e. the HHVM built packages with gcc and clang could have their own `checkPhase` of tests.

Test Plan:
See GitHub Actions. The test log should be found in the `nix-build` step.

### DIFF
--- a/hhvm.nix
+++ b/hhvm.nix
@@ -231,8 +231,8 @@ stdenv.mkDerivation rec {
   checkPhase =
     ''
       set -ex
-      cd ${./.}
       export HHVM_BIN="$PWD/hphp/hhvm/hhvm"
+      cd ${./.}
       "$HHVM_BIN" hphp/test/run.php quick
     '';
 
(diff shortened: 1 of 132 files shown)


hhvm.nix
```diff
@@ -223,6 +226,17 @@ stdenv.mkDerivation rec {
...
+
+  checkPhase =
+    ''
+      set -ex
+      runHook preCheck
+      export HHVM_BIN="$PWD/hphp/hhvm/hhvm"
+      (cd ${./.} && "$HHVM_BIN" hphp/test/run.php quick)
+      runHook postCheck
+    '';
+
```
Atry, line 239: According to https://nixos.org/manual/nixpkgs/unstable/#ssec-check-phase, `checkPhase` is supposed to used to run test suites of a package.

PR #9208, opened 2022-09-17
Pin OpenSSL version to 1.1

Atry:
#9184 fails because recently nixpkgs unstable changes the default OpenSSL version to 3.0, which is not compatible with Folly.
- See https://github.com/facebook/hhvm/actions/runs/3065306837 for build log.
- See https://github.com/NixOS/nixpkgs/pull/150093 for the nixpkgs commit.

This PR pins OpenSSL version to 1.1 to suppress the error.

Test Plan:
GitHub Actions should pass


henryr:
We're interested in using OpenSSL 3.0 on 22.04 - do you anticipate keeping this pin for a while, or is it a temporary change?

Atry:
1. This PR pins OpenSSL version from nix packages, not the Ubuntu distribution.
2. We HHVM maintainer cannot support OpenSSL right now, because Folly, a HHVM's dependency, has not supported OpenSSL 3.0 yet.

PR #9225, opened 2022-09-23
Remove without-unwind-symbolizer.patch

Atry:


third-party/folly/patches/series
```diff
@@ -2,6 +2,5 @@ disable-liblzma.patch
 folly-cmake-fixup.patch
```
captain-hoity-toity, line 2: For improved code clarity, consider using enumerations instead of raw values.

PR #9249, opened 2022-10-10
Clean foreach helper in thrift

pzhzqt:
Differential Revision: D40240735


Atry:
Thank you for testing the diff with HHVM!

Atry:
The CI failure in https://github.com/facebook/hhvm/actions/runs/3222278940/jobs/5286893057 seems due to flakiness.

PR #9276, opened 2022-10-27
Parse non-fixme /* comments not preceded by newlines

muglug:
This fixes an issue where the contents of `/* ... */` comments are swallowed when they appear on the same line as the previous token.

```hack
function foo(string $a): void {
        // this currently works
	echo(
		/* HH_NOT_A_FIXME[4110] */ $a as nonnull);

	// this currently does not work
	echo(/* HH_NOT_A_FIXME[4110] */ $a as nonnull);
}
```

### DIFF
--- a/hphp/hack/src/parser/core/lexer.rs
+++ b/hphp/hack/src/parser/core/lexer.rs
@@ -1979,7 +1979,7 @@ where
                         acc.push(t);
                         return acc;
                     }
-                    TriviaKind::FixMe | TriviaKind::IgnoreError => {
+                    TriviaKind::FixMe | TriviaKind::IgnoreError | TriviaKind::DelimitedComment => {
                         return acc;
                     }
                     _ => {


viratyosin:
Can you tell me more about the hidden-comment behavior you were trying to fix? What were you doing that resulted in the lost comment?

If I use `hh_parse --full-fidelity-json` on a file containing your code snippet I see `"// this comment is hidden in parser output"` as part of the leading trivia of the `echo` token.

muglug:
@viratyosin yes! But you don't see `HH_NOT_A_FIXME[4110]`

Wilfred:
@muglug I don't think this solves your problem in the general case. You've changed some trailing trivia to be leading trivia, but AFAICS you'll just end up with cases where you can't see the comment associated with a closing bracket rather than an opening one.

Have you considered using HH_FIXME itself instead? We could easily reserve some number prefix (say 8XXXX) for your use case, and then you'd get this for free.

muglug:
> you'll just end up with cases where you can't see the comment associated with a closing bracket rather than an opening one.

I don't see how that could happen in practice — I never see FIXMEs come after the thing they're intending to fix like this

```hack
function foo(string $a): void {
        echo($a as nonnull /* FIX_PREVIOUS_ELEMENT[4110] */);
        // or
        echo(
              $a as nonnull /* FIX_NEXT_AFTER_COMMA[4110] */,
              $b as nonnull,
        );
}
```

Instead those FIXMEs always come directly preceding the element they're intended to cover.

muglug:
As to the `HH_FIXME[8...]` proposal, that's very kind! But I don't think it's a great solution for us — I prefer to use explicit FIXME names rather than force devs to look up what the code means in documentation.

Stuff like

```hack
/* HAKANA_FIXME[UnusedFunction] */
```
is really straightforward, and I prefer it to something like

```hack
/* HH_FIXME[8017] -- Hakana detected an unused function */
```

PR #9282, opened 2022-10-31
Extract public key portion via PEM roundtrip

Atry:
This diff applies the approach similar to https://github.com/php/php-src/commit/26a51e8d7a6026f6bd69813d044785d154a296a3 in order to fix behavior changes in OpenSSL 3

## Test Plan
The existing tests should still pass with OpenSSL 1.1
```
"$HHVM_BIN" hphp/test/run.php hphp/test/slow/ext_openssl/ext_openssl.php
```


Atry:
@Wilfred Why do you close this PR? This PR is necessary to port HHVM to OpenSSL 3, and it is ready for review.

PR #9287, opened 2022-11-01
Remove CMake checks for curl_multi_select/wait

jchaffraix-slack:
libcURL has had `curl_multi_wait` since 7.28.0:
https://curl.se/libcurl/c/curl_multi_wait.html

Note that I couldn't find a reference to
`curl_multi_select` so I don't think cURL
was ever released with it.

All OS should ship with a newer version at this
point so we can remove those unneeded CMake checks.


Wilfred:
Yeah, it looks like an internal patch until curl_multi_wait was available: 941abe0d6200d743cc1682fad179d601b9b2834c

Wilfred (approved):

PR #9289, opened 2022-11-02
Allow optional newline in OpenSSL output

Atry:
Summary: OpenSSL 3 does not include the trailing newline in its output. This diff adjust `expectf` so that it is now compatible with both OpenSSL 1.1 and 3.0.

Differential Revision: D40942191


Sandra638:
bc1qvhragvadmdyw2kdewxetqd8sv9tsam8tjm36lf

PR #9290, opened 2022-11-02
Load the legacy providers from OpenSSL 3

Atry:
Summary:
`openssl_seal` is by default using RC4. However RC4 is only available from the legacy providers in OpenSSL 3, which is not loaded by default.

This diff loads the legacy providers so that `openssl_seal` will not fail with default parameters.

See https://www.openssl.org/docs/man3.0/man7/crypto.html#:~:text=md_whirlpool)%3B%0AEVP_MD_free(md_sha256)%3B-,OPENSSL%20PROVIDERS,-OpenSSL%20comes%20with for providers in OpenSSL 3

## Internal:
We should also consider removing the default value `'RC4'` from `openssl_seal` like [what PHP 8.0 did](https://www.php.net/manual/en/function.openssl-seal.php), because RC4 is vulnerable.

Differential Revision: D40942189

### DIFF
--- a/hphp/runtime/ext/openssl/ext_openssl.cpp
+++ b/hphp/runtime/ext/openssl/ext_openssl.cpp
@@ -28,6 +28,9 @@
 #include <openssl/conf.h>
 #include <openssl/pem.h>
 #include <openssl/pkcs12.h>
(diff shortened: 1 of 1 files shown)


Sandra638:
bc1qvhragvadmdyw2kdewxetqd8sv9tsam8tjm36lf

PR #9306, opened 2022-12-19
[RFC] Add a GitHub Action to build Ubuntu 20.04

jchaffraix-slack:
This PR ports the scripts in
github.com/hhvm/packaging over to
GitHub Actions so we can build the HHVM
package from GitHub.

We run the build script in a container as
the "build-depends` Debian package triggers
some conflict with the packages included
in the default GitHub image.

The only modification to the packaging
scripts are in `make-debianish-package`:
1. The `$IS_NIGHTLY` checks were made consistent as it caused compilation errors.
2. We now fetch the full repository instead of using a source archive.
3. [minor] Simplified the `BUILD_DEPS_{OLD|NEW}` that was unnecessary.
4. [minor] Several variables are derived from the enviroment (`$DISTRO_ROOT`, `$OUT`, ...) instead
of being hardcoded.

This PR doesn't compile as the default GitHub runners only have 7 Gb of memory and thus run out of memory: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources But I have tested the steps in isolation so it should work with larger runners.


ci/skeleton/etc/apache2/mods-available/hhvm_proxy_fcgi.conf
```diff
@@ -0,0 +1 @@
+ProxyPassMatch ^/(.+\.(hh|php|hack)(/.*)?)$ fcgi://127.0.0.1:9000/var/www/$1
```
jchaffraix-slack, line 1: This PR effectively inlines the `packaging` repository in this directory instead of checking it out. Discussing this with Fred, this was discussed in the past and this approach was slightly superior as it ensures the CI and the code are kept in sync.

.github/workflows/ubuntu.yml
```diff
@@ -0,0 +1,47 @@
...
+  build_ubuntu_focal_nightly:
+    runs-on: ubuntu-latest
+    container:
+      image: ubuntu:focal
+      env:
+        DISTRO: ubuntu-20.04-focal
+        IS_NIGHTLY: 1
+    steps:
+    - name: Installing git
+      run: apt update -y && apt install -y git
```
jchaffraix-slack, line 36: We need to fetch git as `actions/checkout@v3` using `submodules: 'recursive'` calls `git submodule`. This should be equivalent to the [instructions for building from source](https://docs.hhvm.com/hhvm/installation/building-from-source).

.github/workflows/ubuntu.yml
```diff
@@ -0,0 +1,47 @@
+name: Ubuntu CI
+
+on:
```
jchaffraix-slack, line 3: This section and the next were copied from the `nix.yml`.

.github/workflows/ubuntu.yml
```diff
@@ -0,0 +1,47 @@
...
+      # should have been covered by GitHub Actions triggered by pull requests.
+      - 'export-D+'
+  pull_request:
+
+concurrency:
+  # If the workflow is triggered by a pull request, then cancel previous runs
+  # for the same pull request, which share the same `github.ref`, otherwise the
+  # run ID is used to identify the concurrency group, which is a no-op because
+  # the run ID is always unique for each trigged event.
+  group: ${{ github.event_name == 'pull_request' && github.ref || github.run_id }}
```
fredemmott, line 16: I think this is leading to the Ubuntu CI job being cancelled - from https://github.com/facebook/hhvm/actions/runs/3732328240 :

> Canceling since a higher priority waiting request for 'refs/pull/9306/merge' exists

Probably want something along the lines of:

```
  group: ubuntu-ci-${{ github.event_name == 'pull_request' && github.ref || github.run_id }}
```

fredemmott: The nix failures seem bogus (and part of a pattern of HHVM's G++ builds flaking but the clang builds being fine), but the Ubuntu CI job isn't being ran - probably because of the above

jchaffraix-slack: Good call on the trigger, I have updated the workflow file as you suggested. It passed on the new run: https://github.com/facebook/hhvm/actions/runs/3838187863/jobs/6534350108

ci/skeleton/usr/share/hhvm/uninstall_fastcgi.sh
```diff
@@ -0,0 +1,219 @@
...
+	then
+		echo "Apache is running, restarting"
+		/etc/init.d/apache2 restart
+		echo "Finished restarting"
+	fi
+	echo "Finished restarting apache"
+	return 0
+}
+
+#!/bin/bash
```
Wilfred, line 104: I know you copied this from the packaging repo, but it's weird. Is this a packaging convention or a typo @fredemmott?

Wilfred (approved):
Wonderful, thanks. I agree that inlining the packaging repository makes sense too.

ci/skeleton/usr/share/hhvm/uninstall_fastcgi.sh
fredemmott: Definitely not a convention; these are just (possibly before my time  😲 ) convenience scripts

I suspect this was just someone combining into one years back and not noticing this.

PR #9318, opened 2023-01-27
ZipArchive::getFromIndex: return a nullable string

jchaffraix-slack:
This updates this function to return null instead of false on error. This allows typing the return as `?string` instead of `mixed`.

Also fixed a return mismatch between builtins_zip.hhi and the extension that was reported by some Slack engineer.


jthemphill:
This is great! We first need to update our internal codebase to be ready for the type errors that this exposed. I have a diff up to do that.

PR #9321, opened 2023-02-04
Delete devcontainer.json

Josephus1123:


Josephus1123:
[my-ublock-backup_2023-03-24_04.03.25.txt](https://github.com/facebook/hhvm/files/11060827/my-ublock-backup_2023-03-24_04.03.25.txt)

PR #9331, opened 2023-02-14
fbshipit-source-id: 6151f651d2d523bfe3121f61a5d75cab7246f3bf

vsiles:
Synchronizing OSS Cargo.toml files


vsiles:
There's an issue with this patch, some dependency on src/parser/ffi was deleted because it is not correctly tracked. Will fix it upstream before submitting a new patch.

PR #9344, opened 2023-03-05
Use third-party/tbb to build HHVM

jchaffraix-slack:
We currently don't use the TBB in third-party/tbb
for building HHVM, which is causing some issues with
the Ubuntu 22.04 update.

This PR updates our CMake files to use the TBB in
third-party. This uses `ExternalProject_add` to build
the existing code in a bundled target. This complemented
with an interface target that exports the static library
to the rest of HHVM.

The approach eschews TBB's CMake recommendation (a
custom function called `tbb_build`) as this seemed
poorly integrated and adds more opacity to our build.
Instead we just call `make` directly to build the code.

This removes the need for calling `find_package(tbb)`
so FindTBB.cmake was removed too.


Wilfred (approved):
Looks great to me, thanks for doing this. I'll confirm that all CI is green before merging this.

third-party/tbb/pic_compile.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/third-party/tbb/src/build/linux.gcc.inc b/third-party/tbb/src/build/linux.gcc.inc
```
Wilfred, line 1: What does this file do?

jchaffraix-slack: This is a patch to enable `-fPIC` when compiling with GCC. Without it, we would get an error when compiling TBB.

All the files in the directory have `PIC_KEY` set and it's the only one that comments it.  TBB uses those `<os>-<compiler>.inc` files to set compilation options.

PR #9404, opened 2023-10-30
Option to set where the perf maps go

jchaffraix-slack:
This matches other options like PerfJitDumpDir
or CoreDumpReportDirectory.

For simplicity, this controls both the perf
map and the perf id map.


jchaffraix-slack:
After spending more time on this code, we realized that only `perf` consumes the perf map and that this PR is actually not needed. Closing...

PR #9446, opened 2024-05-20
Update to grmtools 0.13.

ltratt:
This moves the code away from using the now unmaintained 0.12 branch of grmtools. There is an API incompatibility (see
https://github.com/softdevteam/grmtools/blob/master/CHANGES.md) which requires a minor change.

### DIFF
--- a/hphp/hack/src/Cargo.lock
+++ b/hphp/hack/src/Cargo.lock
@@ -619,11 +619,11 @@ checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
 name = "cfgrammar"
-version = "0.12.0"
+version = "0.13.5"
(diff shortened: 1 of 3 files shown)


mheiber:
Hi @ltratt , thank you for your work on this!
Unfortunately, our Cargo.toml and Cargo.lock files are generated from an internal system where we move at a shared cadence. If we update this file, it'll break the build. 
I'll look into bumping the dep using our internal systems. You had no way of knowing, sorry for the trouble.

mheiber:
Hi @ltratt , thank you for your work on this!
Unfortunately, our Cargo.toml and Cargo.lock files are generated from an internal system where we move at a shared cadence. If we update this file, it'll break the build.
I'll look into bumping the dep using our internal systems. You had no way of knowing, sorry for the trouble.

ltratt:
@mheiber I had guessed there might be an internal versions of this, but I don't see it as wasted effort on my part, so don't worry about that :)

PR #9568, opened 2025-01-07
Update OCaml stubs

mszabo-wikia:
The OSS OCaml stubs under src/stubs have become out of sync with the code referencing them.

Update the stubs to match.


hphp/hack/src/stubs/dune
```diff
@@ -65,16 +65,11 @@
...
+  server_env
   typechecker_options)
  (preprocess
   (pps lwt_ppx ppx_deriving.std)))
 
- ; this stub file is used by both buck2 and dune
-(rule
- (target remote_old_decls_ffi.ml)
- (action
-  (copy# ../facebook/remote_old_decls/stubs/remote_old_decls_ffi.ml remote_old_decls_ffi.ml)))
```
mszabo-wikia, line 76: Alternatively if upgrading to Dune 3.17.0+ is fine, there's a `file-available` macro.

Wilfred:
Could you run ocamlformat and update the PR?

Wilfred:
The sandcastle module looks wrong, I'm seeing a compile error:

```
Error: Library sandcastle is defined twice:
- hack/src/utils/facebook/dune:6
- hack/src/stubs/dune:171
```

mszabo-wikia:
Yeah, looks like the OSS stub will need to be included conditionally in non-fbsource builds only. I'll see what I can do, there are some preexisting examples for this.

mszabo-wikia:
Yeah, the exclusion logic is definitely wrong. I'll look into this, thanks!

mszabo-wikia:
@Wilfred Sorry about the delay, finally had some time to take a look at this.

I think it will be necessary to rename the internal `sandcastle` library to `sandcastle_fb` as part of this patch. This is because `src/stubs/sandcastle/dune`  now defines a shimmed `sandcastle` library that will either point to the internal library in fbsource builds or the stubbed version defined in `src/stubs/dune` in OSS builds, but this requires the internal version to have a different name from the shim, hence the suffix. This seems to be the pattern used for every stubbed internal OCaml library in the project.

Let me know if you have any questions!

hphp/hack/src/stubs/state_loader_lwt.ml
```diff
@@ -6,20 +6,10 @@
...
  *)
 
-let get_project_metadata
-    ~progress_callback:_
-    ~saved_state_type:_
-    ~repo:_
-    ~ignore_hh_version:_
-    ~opts:_ =
+let get_project_metadata ~repo:_ ~ignore_hh_version:_ ~opts:_ =
   failwith "Not implemented"
```
muglug, line 10: This needs to return `Lwt.return (Error ("Not implemented", Telemetry.create ()))` for the IDE integration to work.

Otherwise the `init_with_find` call terminates execution.

hphp/hack/src/stubs/state_loader_lwt.ml
```diff
@@ -6,20 +6,10 @@
...
 
-let load
-    ~ssopt:_
-    ~progress_callback:_
-    ~watchman_opts:_
-    ~ignore_hh_version:_
-    ~saved_state_type:_ =
+let load ~ssopt:_ ~progress_callback:_ ~watchman_opts:_ ~ignore_hh_version:_ =
   failwith "Not implemented"
```
muglug, line 13: This needs to return `Lwt.return (Error "Not implemented")` for the IDE integration to work

hphp/hack/src/stubs/state_loader_lwt.ml
```diff
@@ -30,15 +20,21 @@ let load_internal
...
+
+module FromDisk = struct
+  type load_result = {
+    naming_table_path: Path.t;
+    warning_saved_state_path: Path.t;
+    files_changed: Saved_state_loader.changed_files;
+  }
+
+  let load ~project_metadata:_ ~threshold:_ ~root:_ = failwith "Not implemented"
```
muglug, line 39: This needs to return `Error "Not implemented"` for IDE integration to work — otherwise the `init_with_fetch` call terminates execution.

muglug (commented):
This worked for us, with three lines changed

mszabo-wikia:
> This worked for us, with three lines changed

Thanks! Updated this branch now.

PR #9571, opened 2025-01-07
Update bundled Rust to 1.83.0

mszabo-wikia:
The bundled Rust version is too old for several library dependencies. Update to the nightly build from October 13th, 2024, which corresponds to the 1.83.0 release.

Split from #9564.


Wilfred:
Nit: nightly-2024-11-26 is 1.85.0-nightly. The nearest nightly to 1.83.0 is nightly-2024-10-13.

mszabo-wikia:
Ah, thank you—fixed now.

dtolnay (approved):
Thanks!

PR #9587, opened 2025-01-23
Generate config headers from configs.specification

mszabo-wikia:
Since a41bcb1cebb1a69d423b1f6ad58640dd05b5dd7d, configuration headers and sources are generated using a Rust tool rather than being hardcoded. Implement the CMake logic necessary to compile the tool and generate these sources. Use a Python script to extract config section names from the specification file instead of hardcoding them into CMake build files to reduce the chance of the CMake build getting out of a sync.

A followup PR will handle connecting the hackc_options target to the Rust libraries that depend on it.


Wilfred (changes requested):
Generally looks good to me. Would you mind running `black` to format the Python, so I can merge this?

mszabo-wikia:
Done with `black` 24.10.0. Let me know if I should use a different version!

mszabo-wikia:
@Wilfred It looks like the Cargo.toml manifest got lost after the merge—AFAIK this might be because they're generated automatically from the internal buck2 buildfiles. Would it be possible to do that to allow OSS to compile the config generation tool?

Let me know if this requires any restructuring on the CMake side.

PR #9588, opened 2025-02-01
Use single Rust staticlib for FFI

mszabo-wikia:
Currently, each Rust FFI library is built and linked into the HHVM binaries as separate static libraries. This is [explicitly unsupported](https://cxx.rs/build/other.html#linking-the-c-and-rust-together) by the `cxx` crate, which recommends either using rustc as the final linker or using a single Rust staticlib for the Rust components, and has been the cause of linking errors for the last couple of years (see T146965521).

The former does not seem to be feasible for HHVM OSS, so go with the latter approach and combine Rust FFI libraries into a single staticlib, then link that into HHVM. This will likely need corresponding changes in the internal buck2 build.

### DIFF
--- a/CMake/HPHPFindLibs.cmake
+++ b/CMake/HPHPFindLibs.cmake
@@ -408,11 +408,7 @@ macro(hphp_link target)
   target_link_libraries(${target} ${VISIBILITY} fizz)
   target_link_libraries(${target} ${VISIBILITY} brotli)
   target_link_libraries(${target} ${VISIBILITY} hhbc_ast_header)
-  target_link_libraries(${target} ${VISIBILITY} compiler_ffi)
-  target_link_libraries(${target} ${VISIBILITY} package_ffi)
-  target_link_libraries(${target} ${VISIBILITY} parser_ffi)
-  target_link_libraries(${target} ${VISIBILITY} hhvm_types_ffi)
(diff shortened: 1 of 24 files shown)


mszabo-wikia:
Gentle ping for @dtolnay—I'm curious whether you think this approach might be feasible. I find it a bit curious that the present arrangement worked for as long as it did (until 4b12ed4cbd26cf5a0835c4fdb454bf41aac1dbf8 / 90d3e33675e6a22c9898ca0437a7d26ea415c653) despite multiple staticlibs being linked even before that. Although in fairness the docs do say that "[multiple staticlibs] are likely to conflict" rather than "will conflict", so perhaps it was just the straw that broke the camel's back.

dtolnay (commented):
The internal build uses Buck2 ["advanced unstable linking"](https://github.com/facebook/buck2/blob/35df6149c3d10f5da1ac48e9bcd68540e2902f9c/prelude/rust/rust_toolchain.bzl#L94-L101) which correctly handles arbitrary dependency graphs between Rust and C++.

I don't work on HHVM but the HHVM team will need to decide whether it's time to move the OSS build to Buck2, instead of this PR.

mszabo-wikia:
Thanks for your quick response! This is exactly the kind of valuable context I was looking for.

I'm fine with either a continuation of the preexisting CMake-based approach (#9564 takes care of most build issues) or a community-driven Buck2 build.

PR #9601, opened 2025-04-15
Exclude Meta-specific code in ext/fb

mszabo-wikia:
FBSerialize.h and related files were removed from OSS in 9b12e0bc3bdd8321f33bbc8a4765c6085229c6ae due to them being referenced in many other projects in fbcode, not just HHVM.

As the HHVM functions exposing this serialization are unlikely to have been of use to non-Meta end users to begin with, exclude code referencing FBSerialize in the OSS build.


Wilfred:
Looks good to me, thanks. Would you mind rebasing? That would make it easier for me to merge this.

(No merge conflicts, I just want to re-trigger our internal CI.)

PR #9619, opened 2025-06-06
Update systemlib generation to support precompiled systemlibs

mszabo-wikia:
Since D46606416, we include precompiled unitemitters rather than raw Hack source files for systemlibs into the HHVM binary. This requires corresponding updates to the OSS build system.

* Update embed_all_systemlibs() and embed_systemlibs_byname() to invoke HHVM to compile systemlibs, and update section name generation to match.
* Update systemlib generation to correctly look for systemlib sources under hphp/runtime/ext/core instead of hphp/system after they were moved in D46164829.
* Make tc-print depend on HHVM, since the hhvm binary is required to compile and embed systemlibs into tc-print.
* Fix apparently erroneous contains() check (added in D65835121) in compiler-systemlib.cpp that checks for `file` but then tries to access `"ext_" + file`. This is erroring in OSS, since all extension systemlibs are named ext_something.php, not something.php.


mszabo-wikia:
Resolved elsewhere

PR #9632, opened 2025-09-24
Fix OSS Hack build on Arm

mszabo-wikia:
The `ocaml-option-fp` flag is not supported on aarch64, and the pinned versions of `dune` and `landmarks-ppx` need updating to work on aarch64.


Wilfred:
Could you split the `ocaml-option-fp` change as a separate PR? The dune upgrade is failing our internal CI for some reason.

mszabo-wikia:
@Wilfred Done, split into https://github.com/facebook/hhvm/pull/9644. I'll rebase this once that PR lands and update the description.

PR #9633, opened 2025-09-24
Add various missing includes

mszabo-wikia:
These were originally detected when I was compiling HHVM with clang and libstdc++, but it's generally good to include what you use without relying on another standard library header pulling them in.


Wilfred:
LGTM, could you rebase? Our CI doesn't like merging commits that are more than ~2 days old.

mszabo-wikia:
Done, thanks!

PR #9656, opened 2025-11-04
AArch64: Improve code quality during type checking on Arm targets

david-arm:
When checking a type matches we often need to compare against an arbitrary bit mask (usually fits into a single byte). However, the tst instruction on AArch64 only supports masks of contiguous bits so usually the mask cannot be encoded into the tst instruction. Instead, when compiling for Arm we should prefer using the cmp instruction as this can take any 8-bit immediate.

### DIFF
--- a/hphp/runtime/vm/jit/irlower-internal-inl.h
+++ b/hphp/runtime/vm/jit/irlower-internal-inl.h
@@ -256,6 +256,10 @@ void emitTypeTest(Vout& v, IRLS& env, Type type,
 
     auto const base = type.unspecialize();
     if (type == TStaticStr)     return test(KindOfString);
+    // On Arm targets comparisons with arbitrary bitmasks are more efficient
+    // with cmp than test.
+    if (arch() == Arch::ARM && !type.isUnion())
+      return cmp(type.toDataType(), CC_E);
     if (type.isKnownDataType()) return test(type.toDataType());
     if (base == TArrLike)       return cmp(KindOfKeyset, CC_BE);
     if (type == (TVec|TDict))   return cmp(KindOfVec, CC_BE);


meteorfox:
@david-arm Thanks for the PRs!

This one is failing with this assertion failure

```
-------------------------------Assertion Failure--------------------------------
hphp/runtime/vm/jit/translator-runtime.cpp:235: void HPHP::jit::VerifyParamTypeFail(TypedValue, const Class *, const Func *, int32_t, const TypeConstraint *): assertion `!tc->check(&value, ctx)' failed.
```

It's triggered by 400k tests :)
 
These are some  you can try
* hphp/test/quick/fb_dbt.php
* hphp/test/slow/attributes/class_attribute.php
* hphp/test/slow/attributes/empty_attribute_default_arg_vs_missing.php
* hphp/test/slow/attributes/reflection_attribute.php
* hphp/test/slow/attributes/simplihack.php

david-arm:
> @david-arm Thanks for the PRs!
> 
> This one is failing with this assertion failure
> 
> ```
> -------------------------------Assertion Failure--------------------------------
> hphp/runtime/vm/jit/translator-runtime.cpp:235: void HPHP::jit::VerifyParamTypeFail(TypedValue, const Class *, const Func *, int32_t, const TypeConstraint *): assertion `!tc->check(&value, ctx)' failed.
> ```
> 
> It's triggered by 400k tests :)
> 
> These are some you can try
> 
> * hphp/test/quick/fb_dbt.php
> * hphp/test/slow/attributes/class_attribute.php
> * hphp/test/slow/attributes/empty_attribute_default_arg_vs_missing.php
> * hphp/test/slow/attributes/reflection_attribute.php
> * hphp/test/slow/attributes/simplihack.php

OK thanks I'll take a look!

david-arm:
Hmm, clearly I've not been building with asserts enabled. :)

david-arm:
When I build with asserts even without this patch I see lots of assertion failures in tests, e.g.

```
-------------------------------Assertion Failure--------------------------------
hhvm/hphp/runtime/vm/jit/code-cache.h:98: uint32_t HPHP::jit::CodeCache::toOffset(ConstCodeAddress) const: assertion `addr >= m_base && addr < m_base + kNullptrOffset' failed.
```

and

```
-------------------------------Assertion Failure--------------------------------
hhvm/hphp/runtime/vm/jit/fixup.cpp:132: bool HPHP::jit::FixupMap::(anonymous namespace)::getFrameRegs(VMFrame, VMRegs *): assertion `fixup && "Missing fixup for indirect fixup"' failed.
```

so I must be building and/or testing hhvm incorrectly.

david-arm:
Hi @meteorfox, I've tracked down the source of the assertion in VerifyParamTypeFail and I've discovered that it's actually caused by a pre-existing bug in hhvm. The function `emitCmpTVType` looks like this

```
inline void emitCmpTVType(Vout& v, Vreg sf, DataType s0, Vreg s1) {
  v << cmpbi{static_cast<data_type_t>(s0), s1, sf};
}
```

but unfortunately the lowering for this is wrong for AArch64 targets:

```
vasm-arm.cpp:  void emit(const cmpbi& i) { a->Cmp(W(i.s1), i.s0.ub()); }
```

since we should first zero-extend (or sign-extend depending upon the condition code) the lowest byte of the input register. In my case the 32-bit input register was 0x7a7a7a46 where 46 is the actual data type encoding for Int64. To be honest, I don't actually know how this code in emitTypeTest even works:

```
    if (base == TArrLike)       return cmp(KindOfKeyset, CC_BE);
    if (type == (TVec|TDict))   return cmp(KindOfVec, CC_BE);
    if (type == TNull)          return cmp(KindOfUninit, CC_AE);
```

perhaps in some cases where it's important we just get lucky and bits 8-31 of the register input are 0?

I'll create a separate PR to fix this, but it also means this PR is dead because it would require an extra `and` instruction for the zero-extend which is no better than what we had before. :(

david-arm:
Sorry, I just saw the comments above cmpbi just now so perhaps I'm misunderstanding something:

```
  // note: cmp{bw}[i] are emitted only for narrow comparisons and _do not_ sign
  // extend their arguments--these instructions are lowered to cmp{lq}[i] if
  // the comparison is not narrow or not equality/inequality
```

I'll have a think about it!

david-arm:
On second thoughts, I don't think I can pursue a fix right now. It looks very awkward fixing the cmpbi/cmpb/cmpwi/cmpw bug and will require an audit of all places where those operations are created. Essentially I think it requires new cmpub and cmpsb operations to let the backend know whether the input regs need zero- or sign-extending to get the correct behaviour, but at least you're aware of the bug!

Closing the PR as there is no longer any value in it.

PR #9665, opened 2025-11-12
Fix PIE builds

mszabo-wikia:
When built with jemalloc, HHVM defines several custom arenas that allocate from well-known address ranges. This historically was gated behind the Meta-specific `USE_JEMALLOC_EXTENT_HOOKS` define, which was removed in early September by upstream. PIE builds of HHVM are broken as a result since the entire arrangement appears to rely on at least the translation cache being in a predefined memory location, which PIE ipso facto precludes.

So, have the `ENABLE_PIE` build flag define an `ENABLE_PIE` define and a `use_position_dependent_jemalloc_arenas` constexpr to disable the problematic behavior. Reinstate and use the previous limited arena setup for PIE builds where only the low arena exists and is backed by sbrk(2) rather than mmap, as it was prior to D78567435.

PIE also means we can't use 32-bit SmallPtrs for things like function pointers, so define `FULLPTR_FOR_BUILTINS` in PIE builds and use `emitLdPackedPtr` when referring to them in the JIT rather than 32-bit-only instructions.

### DIFF
--- a/hphp/runtime/base/program-functions.cpp
+++ b/hphp/runtime/base/program-functions.cpp
@@ -2055,13 +2055,15 @@ static int execute_program_impl(int argc, char** argv) {
   }
 
 #if USE_JEMALLOC
-  if (Cfg::Server::Mode) {
-    purge_all();
-    setup_auto_arenas({Cfg::Eval::Num1GPagesForA0, Cfg::Eval::Num2MPagesForA0});
-  }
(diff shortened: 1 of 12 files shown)


mszabo-wikia:
Per our discussion last week, this isn't maintainable for the HHVM team long-term, so closing this out.

PR #9666, opened 2025-11-12
aarch64: Fix JIT for PIE builds

mszabo-wikia:
On aarch64, jitted code ends up segfaulting on PIE HHVM builds. Running a debug build pointed to `makeTarget32` receiving a 64-bit pointer instead of the expected 32-bit pointer, so rework affected usages to operate on 64-bit pointers and registers instead.

x86 is not affected by this issue because it presumably makes no equivalent assumption about pointer sizing.


mszabo-wikia:
Per our discussion last week, this isn't maintainable for the HHVM team long-term, so closing this out.

PR #9667, opened 2025-11-13
Automatically symlink compile_commands.json

mszabo-wikia:
compile_commands.json allows IDEs to provide go-to definition etc. support, but it gets generated in CMAKE_BINARY_DIR, which may be out of tree when doing an out-of-source build. So, symlink it automatically to allow IDEs to pick it up.


Wilfred:
LGTM, would you mind rebasing?

mszabo-wikia:
Done, thanks!

PR #9668, opened 2025-11-13
AArch64: Add folds to use new cbz/cbnz block terminators

david-arm:
We can fold sequences like this:

testl w0, w0
jcc CC_E, label

into a single, new opcode

cbz w0, label

if the flags set by testl are only used by the jcc operation. Not only does this reduce the number of instructions, but it also has the benefit of not clobbering the flags. On AArch64 the branch offset range for cbz/cbnz is identical to b.cc so there is no penalty for using them.

### DIFF
--- a/hphp/runtime/vm/jit/relocation-arm.cpp
+++ b/hphp/runtime/vm/jit/relocation-arm.cpp
@@ -290,13 +290,13 @@ InstrSet findLiterals(Instruction* start, Instruction* end) {
  */
 constexpr auto kFarJccLen = 3 * kInstructionSize;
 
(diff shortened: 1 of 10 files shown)


charles-typ:
Hi @david-arm, this PR has a merge conflict in vasm-simplify-arm.cpp, can you rebase and submit again? Thanks a lot!

david-arm:
> Hi @david-arm, this PR has a merge conflict in vasm-simplify-arm.cpp, can you rebase and submit again? Thanks a lot!

OK, hopefully it's done now!

PR #9672, opened 2025-11-18
Consolidate security-related compiler flags

mszabo-wikia:
Update and extend the list of hardening-related compiler flags used by HHVM to better represent modern distro defaults.

* Convert the existing `ENABLE_SSP` build option into a new `ENABLE_HARDENING` option and put an updated list of security flags behind it. Both clang and GCC have been supporting these options for a while now, so we can set them irrespective of the compiler.
* Put PIE-related options behind a separate `ENABLE_PIE` build option so that we can produce and compare non-PIE and PIE builds once we fix compatibility with PIE.
* Forward `CMAKE_BUILD_TYPE` to vendored subprojects. Lack of this was causing the projects to be built without compiler optimizations, which doesn't play well with `FORTIFY_SOURCE`.

On systems with glibc >= 2.40, https://github.com/facebook/folly/pull/2519 is needed for this option to work.

The overhead from these flags is likely to be limited, as many of them have been set by default for distribution packages for several years now.[1]

[1] https://github.com/jvoisin/compiler-flags-distro


Wilfred:
LGTM, usual rebase request please :)

mszabo-wikia:
Done, thanks!

PR #9676, opened 2025-11-27
AArch64: Use storepair more when storing out constant values

david-arm:
We sometimes see sequences like this:

store %123(...b), [addr + 8]
store %124(...q), [addr]

The existing storepair simplify code requires that the stored registers be physical GP registers, presumably because the lowering for storepair/storepairl cannot handle FP/SIMD regs. However, during register allocation we materialise these constants and end up with the sequence:

ldimmb ... => x0
store x0, [addr + 8]
ldimmq ... => x0
store x0, [addr]

which then makes it very difficult to combine these into storepairs in the post-regalloc simplify pass. This PR permits combining pairs of stores prior to regalloc, provided we can show they are either:

1. A GP physical register, or
2. An integer constant that will be materialised into a GP reg.


ottoni (commented):
looks good, except for small coding style nit below

PR #9678, opened 2025-11-28
AArch64: Allow rebasing pointers to make more use of stp/ldp

david-arm:
If the disp (offset) in a Vptr is negative and outside the range
of stp/ldp, then it's also outside the range of
str/stur/ldr/ldur. This will force the offset to be materialised
in a register and to then use the reg+reg addressing mode for
the loads and stores. We'll end up with assembly like this:

  mov x2, -1024
  str x0, [x29, x2]
  mov x2, -1032
  str x1, [x29, x2]

If the immediate can easily be encoded into an add or sub
instruction, then we can rebase the pointer and it will still be
worth it. We'll then end up with assembly like this:

  add x2, x29, -1024
  stp x0, x1, [x2]

which still saves two instructions.


david-arm:
NOTE: This PR includes https://github.com/facebook/hhvm/pull/9676, so should probably wait for that to land first.

PR #9686, opened 2025-12-05
vasm-graph-color: split penalty vectors when consensus conflicts

igogo-x86:
color_block_initialize() biases each live-in toward the phys reg that its processed predecessors already use by adding weight to every *other* phys reg in the shared penalty vector. When multiple live-ins reuse that vector but their predecessors chose different phys regs, those penalties contradict each other. The resulting vector no longer corresponds to any of the prior assignments, so the allocator picks a register based on whatever mix of penalties remains, and we emit edge copies to fix the mismatches (while also pushing the wrong hint backward).
Detect conflicting consensus assignments and clone the penalty vector on the spot so each live-in keeps its own bias/undo bookkeeping.

### DIFF
--- a/hphp/runtime/vm/jit/vasm-graph-color.cpp
+++ b/hphp/runtime/vm/jit/vasm-graph-color.cpp
@@ -10775,6 +10775,62 @@ size_t color_block_initialize(
   };
   jit::vector<Adjustment> penaltyAdjustments;
 
+  auto const clonePenaltyVectorFor = [&] (Vreg r) {
+    auto& info = reg_info(state, r);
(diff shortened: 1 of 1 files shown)


meteorfox:
Thanks again for the PR!

This one triggers an assertion

```
Assertion failure: fbcode/hphp/runtime/vm/jit/vasm-graph-color.cpp:11212: void HPHP::jit::(anonymous namespace)::assign_colors(State &): assertion `v[r] == 0' failed.
```

We also get SIGABRT crashes (e.g `Process terminated by the signal: 6`)

igogo-x86:
That assert shows some problems with the patch that are quite difficult to fix. Let's try a different approach - https://github.com/facebook/hhvm/pull/9696

PR #9687, opened 2025-12-05
Enable stack trace symbolization for OSS builds

mszabo-wikia:
Historically, HHVM statically linked the system libbfd to power stack trace symbolization in its crash reports and in perf map files. This was disabled for OSS in D2137703[1] because of the wildly differing libbfd ABI across target systems. D3742004[2] and D3855027[3] then brought in folly::Symbolizer for both use cases, but this remained gated behind a Meta-only define.

folly::Symbolizer has been stable in upstream folly for a while now, so let's enable it unconditionally and deshim the experimental/ header. This also allows removing the dead libbfd integration. For this to work, the folly build must be built with and able to find libunwind, which isn't a given when building against libc++ and LLVM libunwind on Ubuntu as it installs includes into a subdirectory, so provide an additional hint for this case.

This allows us to finally have a proper stacktrace rather than raw memory addresses in crash reports, such as:
```
Core dumped: Segmentation fault
Stack trace in /tmp/stacktrace.902095.log
Host: hhvm-jammy-vm
ProcessID: 902095
ThreadID: 281473658339392
ThreadPID: 902095
Name: /usr/local/bin/hhvm
CmdLine: hhvm ../hhvm/hphp/test/quick/dv.php
Type: Segmentation fault
Runtime: hhvm
Version: heads/7.70.0-slack-0-g7be53eff8d22faa0383a128696fae83b224aedd5
DebuggerCount: 0

Arguments: ../hhvm/hphp/test/quick/dv.php
ThreadType: CLI

-------------------------------Treadmill Information----------------------------
Now: 235093382994348
OldestStartTime: 235093373066177
InflightRequestsSize: 1

Active Requests:
  281473658339392 1 235093373066177 (age 9ms) (timeout 0s) CLISession OLDEST

------------------------------------------------------------------------------

CPP Stacktrace:

 # 0  HPHP::jit::FixupMap::processFixupForVMFrame(HPHP::jit::VMFrame)
 # 1  HPHP::jit::FixupMap::fixupWork(HPHP::ActRec*, bool)
 # 2  HPHP::jit::detail::syncVMRegsWork(bool)
 # 3  HPHP::createBacktrace(HPHP::BacktraceArgs const&)
 # 4  HPHP::f_debug_backtrace(long, long)
 # 5  HPHP::throwable_init(HPHP::ObjectData*)
 # 6  HPHP::ObjectData* HPHP::ObjectData::newInstance<false>(HPHP::Class*)
 # 7  HPHP::SystemLib::AllocRuntimeExceptionObject(HPHP::Variant const&)
 # 8  HPHP::SystemLib::throwRuntimeExceptionObject(HPHP::Variant const&)
 # 9  HPHP::throwMissingArgument(HPHP::Func const*, int)

PHP Stacktrace:

 #0  main() called at [/home/mszabo/hhvm/hphp/test/quick/dv.php:28]
 #1  main_entry()

```

[1] https://github.com/facebook/hhvm/commit/a66dd1316e4770f97467ea247add2b9d6d704ffd
[2] https://github.com/facebook/hhvm/commit/c11ad4ebdf17a5d3321533c9603277abf4644453
[3] https://github.com/facebook/hhvm/commit/0e082c4a248494046c61d91456236cac222defbe


juice928:
👋 Hi, I'm an automated AI code review bot. I ran some checks on this PR and found **2 points that might be worth attention** (could be false positives, please use your judgment):

1. <strong>Potential portability issue with ELF-specific structures in debug.cpp</strong>
   - [Location](https://github.com/facebook/hhvm/blob/29831e4fb54d1590e41993240e1788e8bc564e15/hphp/runtime/vm/debug/debug.cpp#L111-L112): `hphp/runtime/vm/debug/debug.cpp:L111-L112`
   - Impact: The unconditional use of `ElfFile` may cause build failures on non-Linux platforms like macOS or Windows.
   - Suggestion: Consider wrapping the `ElfFile` logic in platform guards (e.g., `#ifdef __linux__`) to maintain cross-platform support.

2. <strong>Missing fallback mechanism for stack traces in stack-trace.cpp</strong>
   - [Location](https://github.com/facebook/hhvm/blob/29831e4fb54d1590e41993240e1788e8bc564e15/hphp/util/stack-trace.cpp#L186-L190): `hphp/util/stack-trace.cpp:L186-L190`
   - Impact: Removing the previous fallback could break stack trace functionality on systems where `folly::symbolizer` is not supported.
   - Suggestion: It might be safer to restore the `backtrace()` fallback path to handle cases where the primary symbolizer is unavailable.

If you find these suggestions disruptive, you can reply "stop" , and I'll automatically skip this repository in the future.

PR #9747, opened 2026-03-30
Use monolithic mvfst library

mszabo-wikia:
mvfst uses automatically derived CMake targets since D91009061, so the list of libraries HHVM needs to add to the link line must be updated. Since mvfst is a transitive dependency, let's link in the monolithic library rather than figuring out the exact set of granular targets our direct dependencies consume, which would be brittle.


mszabo-wikia:

PR #9751, opened 2026-03-30
Add magic_enum dependency for HHVM OSS

mszabo-wikia:
blob_writer uses the magic_enum library since D97512521. This isn't packaged until Debian Trixie / Ubuntu Noble, so add corresponding vendoring logic for HHVM OSS.


mszabo-wikia:

PR #9752, opened 2026-03-30
Build all of folly for HHVM OSS rather than patching out bits and pieces

mszabo-wikia:
Currently we patch the folly build system in OSS HHVM to only build a subset. This is generally brittle since the patch needs updating if the upstream listfile changes in an incompatible way. Let's simplify things and build all of folly instead.

Add newly required deps and remove dead patches.


mszabo-wikia:

PR #9753, opened 2026-03-30
Update vendored OSS HHVM fmt to 12.1.0 to fix libc++ 21 compat

mszabo-wikia:
This matches what was done for fbcode_builder in D91067518.


mszabo-wikia:

PR #9754, opened 2026-03-30
Fix spammy -Winconsistent-override in HHVM

mszabo-wikia:
Newer LLVM introduces `-Winconsistent-override` which is quite noisy for a few frequently-included headers in HHVM. Fix occurrences.

### DIFF
--- a/hphp/runtime/ext/reflection/ext_reflection.cpp
+++ b/hphp/runtime/ext/reflection/ext_reflection.cpp
@@ -2592,7 +2592,7 @@ struct ReflectionExtension final : Extension {
       ReflectionExtensionLoader::className());
   }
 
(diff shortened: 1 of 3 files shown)


Wilfred:
FWIW this fails to build:

```
fbcode/hphp/runtime/ext/asio/ext_asio.cpp:174:59: error: 'override' specifier is not allowed outside a class definition
  174 | std::vector<std::string> AsioExtension::hackFiles() const override {
      |                                                           ^~~~~~~~
1 error generated.
```

mszabo-wikia:
Right, should be definitions only - updated, thanks!

Wilfred:
Sorry, I was out in May. Could you rebase?

mszabo-wikia:
Done, thanks!

PR #9773, opened 2026-04-28
Wire up autocomplete tests for OSS

mszabo-wikia:
Build `hh_single_complete` via Dune, and use an explicit enum format method in lieu of an FB-specific migration helper for parse_errors.py. Add a helper script that takes care of invoking Dune with the correct environment variables setup so that e.g. `_build/hphp/hack/hack_oss_dune_invoke.sh build @autocomplete` can be used to run autocomplete tests.


Wilfred:
Seems reasonable, could you rebase this one too?

mszabo-wikia:
I might open a new one for this - I have a fairly extensive build system patch pending following the refactoring in https://github.com/facebook/hhvm/commit/c5f58c11d1707016fbcfa59abd0bc930e22503f9

PR #9788, opened 2026-06-05
Fix enum formatting in ext_watchman.cpp

mszabo-wikia:
Meta's internal tooling does this automatically
but OSS needs some help when formatting types
that aren't formattable by default.

### DIFF
--- a/hphp/runtime/ext/watchman/ext_watchman.cpp
+++ b/hphp/runtime/ext/watchman/ext_watchman.cpp
@@ -245,7 +245,7 @@ folly::dynamic makeDynamic(const HPHP::TypedValue& data) {
     default:
       SystemLib::throwInvalidOperationExceptionObject(fmt::format(
         "HPHP::Variant to folly::dynamic conversion failed! Got an "
-        "unconvertible Variant of type {}...", data.type()));
+        "unconvertible Variant of type {}...", fmt::underlying(data.type())));
   }
   not_reached();
 }


Wilfred:
What happens on the OSS path today?

mszabo-wikia:
This fails to build otherwise:
```
/home/vagrant/hhvm-deps/installed/fmt-yLpppK9TnVNoXWaiZx6gkR07AbmDlYGYil1ryL6NEP8/include/fmt/base.h:2310:45: error: implicit instantiation of undefined template 'fmt::detail::type_is_unformattable_for<HPHP::DataType, char>'
 2310 |     type_is_unformattable_for<T, char_type> _;
      |                                             ^
/home/vagrant/hhvm-deps/installed/fmt-yLpppK9TnVNoXWaiZx6gkR07AbmDlYGYil1ryL6NEP8/include/fmt/base.h:2283:44: note: in instantiation of function template specialization 'fmt::detail::value<fmt::context>::value<HPHP::DataType, 0>' requested here
 2283 |   FMT_CONSTEXPR20 FMT_INLINE value(T& x) : value(x, custom_tag()) {}
      |                                            ^
/home/vagrant/hhvm-deps/installed/fmt-yLpppK9TnVNoXWaiZx6gkR07AbmDlYGYil1ryL6NEP8/include/fmt/format.h:4349:40: note: in instantiation of function template specialization 'fmt::detail::value<fmt::context>::value<HPHP::DataType, 0>' requested here
 4349 |   return vformat(fmt.str, vargs<T...>{{args...}});
      |                                        ^
/home/vagrant/hhvm/hphp/runtime/ext/watchman/ext_watchman.cpp:246:60: note: in instantiation of function template specialization 'fmt::format<HPHP::DataType>' requested here
  246 |       SystemLib::throwInvalidOperationExceptionObject(fmt::format(
      |                                                            ^
/home/vagrant/hhvm-deps/installed/fmt-yLpppK9TnVNoXWaiZx6gkR07AbmDlYGYil1ryL6NEP8/include/fmt/base.h:2150:45: note: template is declared here
 2150 | template <typename T, typename Char> struct type_is_unformattable_for;
      |                                             ^
1 error generated.
```

Actually the proper fix is probably or me to implement an explicit formatter - `fmt::underlying` would just put the underlying `int` values of the enum in the output which isn't probably what we want.

PR #9792, opened 2026-06-08
Fix the build after upstream rework

mszabo-wikia:
Reviewer note: This is a scarily large diff but it only touches OSS build infra, hence it should be safe to land!

D105463266 reworked the OSS build extensively and got rid of building Meta first-party deps via `ExternalProject` in favor of discovering them via `find_package`. Unfortunately, HHVM doesn't build using CMake as-is even with this patch. So fix build issues and take this opportunity to do a wholesale cleanup:

* Remove vendoring for everything that we can build via `getdeps` or trivially install as a system package. This includes all Meta first-party deps, zstd, jemalloc, boost, glog, gflags, fmt, lz4, magic_enum, brotli, double-conversion, libsodium, libzip and re2.
* Require `cargo` and `rustc` to be findable by `find_program` rather than falling back to a bundled default. The Rust version used by HHVM/Hack is updated often so it's easier for people to supply a working version in the PATH via rustup than updating the bundled version in the listfile periodically.
* Move most `find_package` calls to `HPHPFindLibs.cmake` where most of them were living already, so that it's easier to see them at a glance. Generally avoid creating `INTERFACE` libraries for dependencies that already expose CMake targets (notably all Meta first-party deps).
* Vendor the single-header `usdt` library under `third-party/usdt`. D105463266 was trying to load it from a Meta internal directory. The official [usdt docs](https://github.com/libbpf/usdt/blob/b0144b02dd885941cf60b2359365c768ee602b27/README.md)
  explicitly encourage consuming the library in this manner.
* Make `HPHPCompiler.cmake` into a base toolchain file and introduce `HPHPClangToolchain.cmake` so that we can use the same compiler and hardening flags for both getdeps-built dependencies and HHVM proper via a simple `CMAKE_TOOLCHAIN_FILE` stanza. This is especially handy since we (Slack) are using a full LLVM toolchain so we want to make sure every C++ dependency is built with libc++.
* Bring back the explicit Cargo.toml for `generate_configs`. D105463266 tried to replace it with a synthetic Cargo.toml generated in the configure phase, but this isn't very workable because the `options` crate in `hack` also has a dependency on this library. We'll have to ask someone to fix the autocargo exports.
* Partially revert the Rust FFI changes from D105463266 because they broke linking the FFI staticlib into the final HHVM executable. Return to the previous working arrangement instead where we create a Rust staticlib and a C++ staticlib and declare an explicit circular dependency between the two so that the linker can work out the details. Symlink cxxbridge headers from the synthetic Rust staticlib to their original package paths under the build directory as some `#include`s expect to find them there.
* Get rid of `hphpize` and supporting infrastructure, as it's been without use since HHVM dropped support for developing and loading shared library extensions.


Wilfred:
Seems reasonable, but internal CI flaked. Could you rebase?

mszabo-wikia:
Done, thanks!

PR #9793, opened 2026-06-08
Define `MYSQL_CLIENT_CR_COMPRESSION_NOT_SUPPORTED` conditionally in HHVM

mszabo-wikia:
D57873158 added `MYSQL_CLIENT_CR_COMPRESSION_NOT_SUPPORTED` to HHVM's ext_mysql, but the underlying constant is only defined in fb-mysql >= 8.0.32. third-party/fb-mysql however mirrors fb-mysql 8.0.20. So make the constant definition conditional on the source constant's existence until the sync logic is updated.


Wilfred:
LGTM, the usual rebase request please :)

PR #9835, opened 2026-08-17
Fix OSS HHVM build x64 and Arm64

mcfi:
As titled, fix OSS HHVM builds for x64 and Arm64.

Debug/release OSS builds succeeded on Ubuntu 24.04/26.04 and CentOS 9/10. Quick and slow tests passed.


build/oss_build.sh
```diff
@@ -0,0 +1,2095 @@
...
+LIBURING_SHA256="8d052f2622dcb3678cbaee5ff582a87572672a6c0a56533cdda5b65cb636120a"
+LIBURING_SRC_DIR="$HHVM_OSS_WORK_ROOT/sources/liburing"
+LIBURING_PREFIX="$HHVM_OSS_WORK_ROOT/installed/liburing"
+LIBURING_REVISION_STAMP="$LIBURING_PREFIX/.hhvm-revision"
+LIBMCRYPT_VERSION="2.5.8"
+LIBMCRYPT_URL="https://downloads.sourceforge.net/project/mcrypt/Libmcrypt/${LIBMCRYPT_VERSION}/libmcrypt-${LIBMCRYPT_VERSION}.tar.gz"
+LIBMCRYPT_ARCHIVE="libmcrypt-${LIBMCRYPT_VERSION}.tar.gz"
+LIBMCRYPT_SHA256="e4eb6c074bbab168ac47b947c195ff8cef9d51a211cdd18ca9c9ef34d27a373e"
+LIBMCRYPT_SRC_DIR="$HHVM_OSS_WORK_ROOT/sources/libmcrypt"
+LIBMCRYPT_PREFIX="$HHVM_OSS_WORK_ROOT/installed/libmcrypt"
```
mszabo-wikia: if y'all don't use mcrypt internally, maybe this could be removed altogether? we (slack) don't use it and PHP removed the whole ext back in 7.1 due to it being abandonware -  https://wiki.php.net/rfc/mcrypt-viking-funeral

build/oss_build.sh
```diff
@@ -0,0 +1,2095 @@
...
+RUN_LOG_FILE="${HHVM_OSS_LOG_FILE:-$HHVM_OSS_WORK_ROOT/oss_build.log}"
+GETDEPS_REAL_ROOT="$SRC_DIR/build/fbcode_builder"
+GETDEPS="$GETDEPS_REAL_ROOT/getdeps.py"
+SCRATCH_DIR="${HHVM_OSS_SCRATCH_DIR:-$HHVM_OSS_WORK_ROOT/getdeps}"
+GETDEPS_SHADOW_ROOT="$SCRATCH_DIR/fbcode_builder-release-train"
+GETDEPS_SHADOW_MANIFESTS_DIR="$GETDEPS_SHADOW_ROOT/manifests"
+GETDEPS_DIR="$SCRATCH_DIR/installed"
+GETDEPS_PUBLIC_CWD="${HHVM_OSS_GETDEPS_CWD:-${XDG_CACHE_HOME:-$HOME/.cache}/hhvm-getdeps-cwd}"
+GETDEPS_SOURCE_MODE_FILE="$SCRATCH_DIR/.hhvm_getdeps_source_mode"
+RELEASE_TRAIN_TAG="v2026.05.11.00"
```
mszabo-wikia: could this be made to work via the shipit integration that updates commit manifests under `build/`?

mszabo-wikia (commented):
thanks, great stuff! i'm curious if this could integrate more with the existing shipit primitives under fbcode_builder but this is looking very nice

hphp/tools/generate-ptr-def.sh
mszabo-wikia, line 1: not needed possibly since D108765883? possibly just needs the invocation removed from the listfile

build/oss_build.sh
```diff
@@ -0,0 +1,2095 @@
...
+      fi
+      ;;
+    mcrouter)
+      local cmake_file="$src_dir/CMakeLists.txt"
+      local compression_header="$src_dir/mcrouter/lib/Compression.h"
+      if [ -f "$cmake_file" ]; then
+        patch_result="$(replace_text_once \
+          "$cmake_file" \
+          $'  mcrouter/StandaloneUtils.cpp\n' \
+          $'  mcrouter/StandaloneUtils.cpp\n  mcrouter/Server.cpp\n')"
```
mszabo-wikia: I think this was fixed in https://github.com/facebook/mcrouter/commit/7e66d4e8a8a47cdcde60bdc776ff65888cf6f5d4 née D108769791?

build/oss_build.sh
```diff
@@ -0,0 +1,2095 @@
...
+          echo "    mcrouter: patched public CMake source list for Server.cpp" >&2
+        fi
+      fi
+      if [ -f "$compression_header" ]; then
+        patch_result="$(replace_text_once \
+          "$compression_header" \
+          $'#include <sys/uio.h>\n#include <limits>' \
+          $'#include <sys/uio.h>\n#include <cstdint>\n#include <limits>')"
+        if [ "$patch_result" = "patched" ]; then
+          echo "    mcrouter: added the missing cstdint include" >&2
```
mszabo-wikia: I think it's better to fix on the mcrouter side and wait for the mcrouter patch to land rather than have this complexity live here

build/oss_build.sh
```diff
@@ -0,0 +1,2095 @@
...
+
+  case "$dep" in
+    folly)
+      patch_result="$(replace_text_once \
+        "$src_dir/folly/net/NetOps.cpp" \
+        $'static_assert(folly::to_bool(::recvmmsg));\nstatic_assert(folly::to_bool(::sendmmsg));' \
+        $'static_assert(sizeof(&::recvmmsg) > 0);\nstatic_assert(sizeof(&::sendmmsg) > 0);')"
+      if [ "$patch_result" = "patched" ]; then
+        echo "    folly: patched GCC 14 socket declaration checks" >&2
+      fi
```
mszabo-wikia: as below

build/oss_build.sh
```diff
@@ -0,0 +1,2095 @@
...
+    -DHHVM_OSS_TIMELIB_CACHE="$TIMELIB_CACHE_DIR/$TIMELIB_ARCHIVE" \
+    -DLIBMAGICKWAND_INCLUDE_DIRS="$IMAGEMAGICK6_INCLUDE_DIRS" \
+    -DLIBMAGICKWAND_LIBRARIES="$IMAGEMAGICK6_WAND_LIBRARY" \
+    -DLIBMAGICKCORE_LIBRARIES="$IMAGEMAGICK6_CORE_LIBRARY" \
+    -DMYSQL_UNIX_SOCK_ADDR=/dev/null \
+    -DFORCE_BUNDLED_LZ4=OFF \
+    "${JEMALLOC_CMAKE_FLAGS[@]}" \
+    -DRUSTC_EXE="$RUST_PREFIX/bin/rustc" \
+    -DCARGO_EXE="$RUST_PREFIX/bin/cargo" \
+    "${XED_CMAKE_FLAGS[@]}" \
```
mszabo-wikia, line 1849: some of these could be a good candidate to be encoded in a preset

build/oss_build.sh
```diff
@@ -0,0 +1,2095 @@
...
+
+run_hhvm_build() {
+  local build_dir="$1"
+  local libmbfl_eaw_table="$SRC_DIR/third-party/forks/libmbfl/mbfl/eaw_table.h"
+
+  if [ -f "$libmbfl_eaw_table" ] && \
+     ! grep -qF 'mbfl_eaw_table' "$libmbfl_eaw_table"; then
+    echo ">>> Removing an invalid generated libmbfl East Asian width table."
+    rm -f "$libmbfl_eaw_table"
+  fi
```
mszabo-wikia, line 707: ?

hphp/runtime/vm/jit/fixup.cpp
```diff
@@ -72,6 +77,59 @@ std::string Fixup::show() const {
...
+    reinterpret_cast<ActRec*>(_Unwind_GetGR(context, dw_reg::FP));
+  state->found = frame == state->frame;
+  return _URC_NO_REASON;
+}
+#endif
+
+uintptr_t nativeFrameCFA(const ActRec* frame) {
+#if defined(__aarch64__)
+  CFAState state{frame, false, 0};
+  _Unwind_Backtrace(findCFA, &state);
```
mszabo-wikia: We tested this in production and the performance impact was prohibitive. See [D108886444](https://www.internalfb.com/diff/D108886444) for an alternative approach

build/oss_build.sh
```diff
@@ -0,0 +1,2105 @@
...
+  if [ -z "$MAGIC_ENUM_INSTALL" ]; then
+    echo "ERROR: magic_enum build finished but the getdeps install was not found"
+    exit 1
+  fi
+fi
+
+
+# ---------------------------------------------------------------------------
+# Phase 3c: Fetch RocksDB source for fb-mysql warm-storage support
+# ---------------------------------------------------------------------------
```
mszabo-wikia, line 1396: Possibly not needed? AFAICT this is for the mysql server side and not the client

build/oss_build.sh
```diff
@@ -0,0 +1,2105 @@
...
+FBMYSQL_OSS_URL="https://github.com/facebook/mysql-5.6/archive/refs/heads/${FBMYSQL_OSS_BRANCH}.tar.gz"
+FBMYSQL_OSS_ARCHIVE="facebook-mysql-5.6-${FBMYSQL_OSS_BRANCH}.tar.gz"
+FBMYSQL_OSS_SHA256="98d6114b36ce0d080040c94534926600224937e617b516922ede6c4e9fcb8051"
+FBMYSQL_OSS_CACHE_DIR="$DOWNLOAD_CACHE_DIR"
+FBMYSQL_OSS_SRC_DIR="$HHVM_OSS_WORK_ROOT/sources/fb-mysql"
+FBMYSQL_OSS_BOOST_PACKAGE="boost_1_77_0"
+FBMYSQL_OSS_BOOST_URL="https://archives.boost.io/release/1.77.0/source/${FBMYSQL_OSS_BOOST_PACKAGE}.tar.bz2"
+FBMYSQL_OSS_BOOST_ARCHIVE="${FBMYSQL_OSS_BOOST_PACKAGE}.tar.bz2"
+FBMYSQL_OSS_BOOST_SHA256="fc9f85fc030e233142908241af7a846e60630aa7388de9a5fafb1f3a26840854"
+LIBDWARF_OSS_VERSION="20180809"
```
mszabo-wikia: I think this is very old. If the intent is to ship a version that doesn't have the producer ABI break, then `20210528` may be a better target. We use this internally

build/oss_build.sh
```diff
@@ -0,0 +1,2105 @@
...
+      best_prefix="$candidate"
+      best_mtime="$mtime"
+    fi
+  done
+
+  [ -n "$best_prefix" ] || return 1
+  printf '%s\n' "$best_prefix"
+}
+
+find_public_fbcode_builder_cmake() {
```
mszabo-wikia: I think the missing piece here is to configure the same shipit-sync for the mcrouter repo that these other repos use to get `build/fbcode_builder` to appear there. this would also have the very welcome side effect of being able to use getdeps-generated CI over there and then we could sunset the legacy autocrap build it's currently using

build/oss_build.sh
```diff
@@ -0,0 +1,2105 @@
...
+  "$PCRE1_PREFIX" \
+  "lib/libpcre.a" \
+  "include/pcre.h" \
+  "-fPIC" \
+  "--enable-jit --enable-utf --enable-unicode-properties"
+touch "$PCRE1_JIT_STAMP"
+
+# ---------------------------------------------------------------------------
+# Phase 4f: Build ImageMagick 6 for the legacy Imagick extension API
+# ---------------------------------------------------------------------------
```
mszabo-wikia, line 1539: we (slack) don't use the `imagick` extension at all, so this could also be axed wholesale if y'all don't use it either!

build/oss_build.sh
```diff
@@ -0,0 +1,2105 @@
...
+
+mkdir -p "$BUILD_DIR"
+
+DEBUGOPT_FLAGS="-O2 -g"
+if [ "$HHVM_CMAKE_BUILD_TYPE" = "DebugOpt" ]; then
+  CMAKE_BUILD_TYPE_FLAGS=(
+    "-DCMAKE_C_FLAGS_DEBUGOPT=$DEBUGOPT_FLAGS"
+    "-DCMAKE_CXX_FLAGS_DEBUGOPT=$DEBUGOPT_FLAGS"
+  )
+fi
```
mszabo-wikia: feels like this belongs more in HPHPCompiler

build/oss_build.sh
```diff
@@ -0,0 +1,2105 @@
...
+   ! cmake_cache_matches CMAKE_DISABLE_FIND_PACKAGE_libWatchmanClient ON; then
+  echo ">>> Phase 8: Existing cmake cache can still use a system libWatchmanClient. Reconfiguring."
+  RECONFIGURE_CMAKE=true
+fi
+if [ -f "$BUILD_DIR/CMakeCache.txt" ] && \
+   { ! cmake_cache_matches RUSTC_EXE "$RUST_PREFIX/bin/rustc" ||
+     ! cmake_cache_matches CARGO_EXE "$RUST_PREFIX/bin/cargo"; }; then
+  echo ">>> Phase 8: Existing cmake cache does not point at the OSS Rust toolchain. Reconfiguring."
+  RECONFIGURE_CMAKE=true
+fi
```
mszabo-wikia, line 1789: this feels very brittle and overly defensive to check for individual cache variables that may be subject to change. possibly just having the force rebuild option below is enough?

hphp/hack/src/hackc/compile/cargo/options/Cargo.toml
```diff
@@ -19,5 +19,6 @@ oxidized = { path = "../../../../oxidized" }
 serde = { version = "1.0.229", features = ["derive", "rc"] }
 
 [build-dependencies]
+generate_configs_lib = { path = "../../../../../../tools/configs" }
```
mszabo-wikia, line 22: question: would this change persist as-is or would it get overwritten by the next `autocargo` export?

build/oss_build.sh
```diff
@@ -1481,40 +1280,24 @@ else
   echo ">>> Phase 3: Building Meta dependencies with getdeps..."
   echo "    Using $CMAKE_COMPILER_DESCRIPTION"
 
-  # Keep the main Meta C++ stack on one coherent public weekly release train.
+  # Use the ShipIt-synchronized sources bundled with HHVM. Dependencies which
+  # are external cells, such as mvfst, use the exported ShipIt revision files.
```
mszabo-wikia, line 1284: So this seems to somewhat contradict the direction in D105463266 where the intent AFAICT was to stop using the stuff mirrored under `third-party/` in favor of `getdeps`. The latter approach seems to be what most of fboss does

build/oss_build.sh
```diff
@@ -108,28 +106,26 @@ CMAKE_CXX_COMPILER_PATH="${HHVM_OSS_CXX_COMPILER:-}"
...
+  "wangle:$SRC_DIR/third-party/wangle/src"
+  "fbthrift:$SRC_DIR/third-party/thrift/src"
+  "proxygen:$SRC_DIR/third-party/proxygen/src"
+  "mcrouter:$SRC_DIR/third-party/mcrouter/src"
+  "fatal:$SRC_DIR/third-party/fatal"
+)
+SHIPIT_SOURCE_ARGS=()
+for source_mapping in "${SHIPIT_SOURCE_MAPPINGS[@]}"; do
+  SHIPIT_SOURCE_ARGS+=(--src-dir "$source_mapping")
+done
```
mszabo-wikia, line 127: <3

hphp/hack/src/hackc/compile/cargo/options/Cargo.toml
mcfi: I used Codex to prepare this draft so there could be rough edges. :-) Your review comments are highly appreciated!

mszabo-wikia: Yea I'm not familiar with autocargo other than it exists and produces these manifests from BUCK files through some mechanism. If it can't express this interdependency then it may be necessary to write some gnarly CMake to do this as a patch or other configure-time transformation. It's not impossible since currently a similar approach is used to link against a synthetic Rust staticlib but it's staggeringly ugly.

build/oss_build.sh
```diff
@@ -0,0 +1,1871 @@
...
+
+# ---------------------------------------------------------------------------
+# Phase 3: Build Meta dependencies with getdeps
+# ---------------------------------------------------------------------------
+choose_cmake_compilers
+set_getdeps_source_mode
+ensure_public_getdeps_source_mode
+rm -rf "$GETDEPS_MANIFEST_OVERRIDE_DIR"
+prepare_getdeps_boost_override >/dev/null
+prepare_getdeps_download_manifest_override magic_enum "$MAGIC_ENUM_DOWNLOAD_URL" "$MAGIC_ENUM_DOWNLOAD_SHA256" >/dev/null
```
mszabo-wikia, line 1261: this override might not be needed, AFAICT these values match what's currently in https://github.com/facebook/folly/blob/main/build/fbcode_builder/manifests/magic_enum

build/oss_build.sh
```diff
@@ -0,0 +1,1871 @@
...
+fi
+if [ -f "$BUILD_DIR/CMakeCache.txt" ] && \
+   ! cmake_cache_matches \
+     HHVM_THIRD_PARTY_SOURCE_CACHE_PREFIX "$THIRD_PARTY_SOURCE_CACHE_PREFIX"; then
+  echo ">>> Phase 8: Existing cmake cache does not point at the local third-party source cache. Reconfiguring."
+  RECONFIGURE_CMAKE=true
+fi
+if [ -f "$BUILD_DIR/CMakeCache.txt" ] && \
+   ! cmake_cache_matches CMAKE_DISABLE_FIND_PACKAGE_libWatchmanClient ON; then
+  echo ">>> Phase 8: Existing cmake cache can still use a system libWatchmanClient. Reconfiguring."
```
mszabo-wikia, line 1781: I am somewhat shocked that we even had a find module for this.
Submitted [D116532790](https://www.internalfb.com/diff/D116532790) to get rid of that.

build/oss_build.sh
```diff
@@ -0,0 +1,1871 @@
...
+  RECONFIGURE_CMAKE=true
+fi
+if [ -f "$BUILD_DIR/CMakeCache.txt" ] && \
+   ! cmake_cache_matches AWK_EXECUTABLE "$GAWK_EXECUTABLE"; then
+  echo ">>> Phase 8: Existing cmake cache does not point at GNU awk. Reconfiguring."
+  RECONFIGURE_CMAKE=true
+fi
+if [ -f "$BUILD_DIR/CMakeCache.txt" ] && \
+   ! cmake_cache_matches ENABLE_LD_GOLD OFF; then
+  echo ">>> Phase 8: Existing cmake cache still enables the gold linker. Reconfiguring."
```
mszabo-wikia, line 1764: gold itself and this entire logic is pretty much dead AFAICT. cleaning this up in [D116534773](https://www.internalfb.com/diff/D116534773)

build/oss_build.sh
```diff
@@ -0,0 +1,1871 @@
...
+        "$imagemagick6_format format support."
+      exit 1
+    fi
+  done
+  touch "$IMAGEMAGICK6_FORMATS_STAMP"
+fi
+
+# ---------------------------------------------------------------------------
+# Phase 4g: Pin the public opam package index used by Hack
+# ---------------------------------------------------------------------------
```
mszabo-wikia, line 1584: For our build we ended up bumping some of the deps needed so that we fetch versions that exist in the current OPAM registry and/or have aarch64 versions: https://github.com/mszabo-wikia/hhvm/commit/e639644c71c4cc6ba94aa949389077e09d61f1e4, https://github.com/mszabo-wikia/hhvm/commit/1a2eba368d577f96596946d6dab35a88bdfdf542

It could be a potentially simpler option to have an OSS-specific dependency list in `hphp/hack/ocaml_deps_data.sh` that uses these newer versions. The drawback of course is that it'd need updating if the internal deps change, but so would a pinned registry in that case & these seem to change rather rarely. 

The pinned registry approach works too though!

PR #9839, opened 2026-08-25
Fix OSS HHVM build x64 and Arm64

mcfi:
Summary:
Fix OSS HHVM builds on x64 and arm64.

- Pin release-train, Rust, fb-mysql, and other external dependency inputs.
- Fix compiler, linker, dependency, and architecture portability issues, including XED support for tc-print and disabling ImageMagick OpenMP to match the internal build and avoid shutdown crashes.
- Preserve native quick and slow test coverage on both architectures.

Reviewed By: binliu19

Differential Revision: D116318957


mszabo-wikia:
thanks again for your work on this! the facility to designate some tests as internal-only is especially helpful since lack thereof was the largest impediment for me when I was looking into running OSS tests

mcfi:
> thanks again for your work on this! the facility to designate some tests as internal-only is especially helpful since lack thereof was the largest impediment for me when I was looking into running OSS tests

Thanks for your helpful comments! Would you mind pulling this PR and test on your servers? The PR worked well in my local tests, but more test data points are more than welcome.

mszabo-wikia:
> Thanks for your helpful comments! Would you mind pulling this PR and test on your servers? The PR worked well in my local tests, but more test data points are more than welcome.

Sure thing, I'll try to take a look this week!

mszabo-wikia:
I remembered I had some CMake lying around for SBCC that I forgot to upstream, so I did so in [D117586888](https://www.internalfb.com/diff/D117586888). This could be useful to not have to skip the SBCC tests.

mcfi:
Closing this to unblock landing of the internal diff.

PR #9840, opened 2026-08-26
Add CMake for sbcc-compiler

mszabo-wikia:
Add appropriate CMake for `sbcc-compiler` so that it can be built in OSS. For this to work, split out a new `hphp_process_init` target shared between `hhvm` and `sbcc-compiler` and likewise move sbcc format files into their own library.

Happy to rework this to use a target arrangement closer to the internal BUCK build too - if you want me to do so, remember I can't see Phab comments! :)

### DIFF
--- a/CMake/HPHPSetup.cmake
+++ b/CMake/HPHPSetup.cmake
@@ -36,6 +36,7 @@ set(HHVM_WRAP_SYMS)
 set(HHVM_LINK_LIBRARIES
   $<TARGET_PROPERTY:onig,INTERFACE_LINK_LIBRARIES>
   ${HHVM_WRAP_SYMS}
+  hphp_process_init
   hphp_analysis
   hphp_system
   hphp_zend
--- a/hphp/hhvm/CMakeLists.txt
+++ b/hphp/hhvm/CMakeLists.txt
@@ -3,12 +3,14 @@ include(HHVMExtensionConfig)
 set(CXX_SOURCES)
 auto_sources(files "*.cpp" "")
 list(APPEND CXX_SOURCES ${files})
+list(REMOVE_ITEM CXX_SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/process-init.cpp)
+
+add_library(hphp_process_init STATIC process-init.cpp)
+hphp_link(hphp_process_init)
 
 get_object_libraries_objects(additionalObjects ${HHVM_WHOLE_ARCHIVE_LIBRARIES})
 add_executable(hhvm ${CXX_SOURCES} ${additionalObjects})
-target_link_libraries(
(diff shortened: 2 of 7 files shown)


mcfi:
Thanks! I pointed AI to your PR and got the following comments:

  - The CMake binary is named sbcc-compiler and emitted under hphp/runtime/ext/, but the test runner expects hphp/runtime/ext/sbcc/sbcc-build (hphp/test/run.php:920). CMake tests won’t find it.
  - hphp_sbcc_format combines reader, cache, and writer, while Buck deliberately separates reader and writer (hphp/runtime/ext/sbcc/BUCK:19). The compiler only needs the writer.
  - The format library is unconditionally linked into hphp_runtime_ext, so disabling the SBCC extension still builds it.
  - hphp_link(hphp_process_init) gives a small helper library a huge public transitive link interface. Final executables should own that linking.
  - The extraction is incomplete: both test binaries still compile process-init.cpp directly (hphp/test/CMakeLists.txt:11, hphp/runtime/test/CMakeLists.txt:6).

  For hphp/hhvm/CMakeLists.txt, I’d use explicit ownership:

  add_library(hphp_process_init STATIC process-init.cpp)

  get_object_libraries_objects(additionalObjects ${HHVM_WHOLE_ARCHIVE_LIBRARIES})
  add_executable(hhvm main.cpp ${additionalObjects})
  target_link_libraries(
    hhvm
    hphp_process_init
    ${HHVM_LINK_LIBRARIES}
    ${EZC_LINK_LIBRARIES}
    ${HRE_LINK_LIBRARIES}
  )

  Then:

  - Link hphp_process_init into both test executables instead of recompiling its source.
  - Keep sbcc-cache.cpp and sbcc-reader.cpp in the extension.
  - Extract only sbcc-writer.cpp for sharing with the compiler.
  - Put the compiler target in hphp/runtime/ext/sbcc/CMakeLists.txt.
  - Name its output sbcc-build.
  - Configure its HHVM linking explicitly rather than inheriting it through hphp_process_init.
  - Add a CMake build/test for the target.

  Adjacent cleanup: hphp/test/CMakeLists.txt:10 appends ${files} twice and should be removed.

mszabo-wikia:
@mcfi Thanks! Fixed most of those, just a few notes:

> hphp_link(hphp_process_init) gives a small helper library a huge public transitive link interface. Final executables should own that linking.

I agree that it's not idiomatic CMake, however since CMake isn't an officially maintained build system for HHVM, linking in the entire kitchen sink helps reduce maintenance churn for changing library dependencies. Otherwise the CMake build needs updating every time one of the headers included by process-init.cpp ends up pulling in a new library dependency.

I also ended up moving `hphp_process_init` to `HHVM_LINK_LIBRARIES` since its usage neatly lines up with the existing targets that were rebuilding `process-init.cpp`.

PR #9844, opened 2026-08-30
Trigger OSS workflow on pull requests

mcfi:
Add a non-functional comment to the OSS workflow so this pull request exercises the configured pull_request CI trigger.


mcfi:
@paulbiss , would you mind approving this PR so that I can check if the OSS workflow on github works as expected? It seems like the OSS workflow is only triggered if this PR gets approved. This PR is going to be discarded eventually tho.

PR #9853, opened 2026-09-08
JIT: keep generated code RX except during scoped RWX writes

mcfi:
Keep JIT code-cache sections read-execute (RX) by default instead of leaving them writable. Use CodeWriteScope to temporarily change affected pages to read-write-execute (RWX) for code emission and modification, then restore RX when the scope exits.

Apply these scopes to code emission, instruction smashing, relocation, recycling, and stub generation. Serialize protection changes and track nested scopes so overlapping writes restore permissions correctly.

Coordinate writes with pthread_atfork and retain forkable huge-page code mappings so child writes preserve copy-on-write behavior.

### DIFF
--- a/hphp/runtime/vm/jit/code-cache.cpp
+++ b/hphp/runtime/vm/jit/code-cache.cpp
@@ -28,6 +28,18 @@
 #include "hphp/util/numa.h"
 #include "hphp/util/trace.h"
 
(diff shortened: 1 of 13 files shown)


mcfi:
@ottoni, would you guys mind trying this PR on Grace and see if there's any perf impact?

